PR #49, opened 2025-05-16
Use tmux to run server in make dev

phinze:
Fixes the mixing of server logs with shell output and gives us a
platform to run multiple shells inside the dev container

## Summary by CodeRabbit

- **New Features**
  - Added tmux support to the development script, enabling simultaneous viewing of the runtime server output and an interactive shell in split panes.

- **Chores**
  - Updated the development environment to include tmux by default.

### DIFF
--- a/dagger/main.go
+++ b/dagger/main.go
@@ -112,7 +112,13 @@ func (m *Runtime) BuildEnv(dir *dagger.Directory) *dagger.Container {
 		WithExec([]string{"apt-get", "update"}).
 		WithExec([]string{"go", "install", "gotest.tools/gotestsum@latest"}).
 		WithExec([]string{"sh", "-c", "cd /usr/bin && curl https://clickhouse.com/ | sh"}, dagger.ContainerWithExecOpts{}).
-		WithExec([]string{"apt-get", "install", "-y", "iptables", "bash", "iproute2", "inetutils-ping"}).
+		WithExec([]string{"apt-get", "install", "-y",
+			"bash",
+			"inetutils-ping",
+			"iproute2",
+			"iptables",
+			"tmux",
+		}).
 		WithFile("/upstream/containerd.tar.gz", dag.HTTP(containerd)).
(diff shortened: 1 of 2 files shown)


evanphx:
Since I'm gonna run this inside an existing tmux (but I set my default leader to Control-A) this works, but I wonder if we should set the leader to something other than Control-B or Control-A for nested users?

phinze:
I originally had it changed to Control-M for Miren but that messes up a whole host of things because it overlaps with the Enter key's control code 🤦🏻 😂 , so I backed out to the default.

Picking a non-conflicting key combo is [not the easiest](https://superuser.com/questions/74492/whats-the-least-conflicting-prefix-escape-sequence-for-screen-or-tmux)! Perhaps Ctrl-S?

evanphx:
Yeah, Control-S is fine! What about some alias in the shell to open and close splits too, since the context of the shell session is narrow? Maybe like "window" opens a new tmux window or something like that?

phinze:
Also seeing a fun little panic on exit with this, if you resize the window or pane while running it, reported here https://github.com/dagger/dagger/issues/9904#issuecomment-2887162683

I don't necessarily think it's a dealbreaker for this strategy for shells in `make dev` for now... but it makes me wonder about hanging too much on the [dagger terminal](https://docs.dagger.io/api/terminal) in the medium term, as it seems like it's intended for temp debugging.

evanphx:
Yeah, definitely wondering if we're asking too much of dagger terminal.

phinze:
> Yeah Control-S is fine!

Sounds good! Commit incoming.

> What about some alias in the shell to open and close splits too, since the context of the shell session is narrow?

I'm hesitant to build another layer of translation over tmux; IMHO it's simpler to use `tmux new-window` or the default keybinds vs having another set of commands to remember. But if you want to experiment with helpers for your workflow I won't stand in your way!

evanphx:
Totes fair. Let's live with it a bit before we abstract.

evanphx (approved):

PR #55, opened 2025-05-21
Switch mTLS back on

evanphx:
* Add "auth generate" command to create new client configs to connect
  back to the server.

## Summary by CodeRabbit

- **New Features**
  - Added a new CLI command to generate authentication configuration files.
  - Introduced support for specifying additional DNS names and IP addresses for server certificates.
  - Added a Makefile target to launch the development environment in a tmux session.
  - Improved client configuration management with new methods for generating and using service/client configs.

- **Improvements**
  - Enhanced certificate validation and regeneration logic for better security and reliability.
  - Streamlined the development script to support both tmux and non-tmux workflows.
  - Improved logging defaults and error handling in various components.
  - Refined test utilities for more robust namespace cleanup.
  - Updated RPC client creation to use coordinator-provided configurations with TLS support.
  - Added validation for certificate common names to prevent use of reserved or invalid names.

- **Bug Fixes**
  - Fixed issues related to certificate expiration and configuration matching.


hack/dev.sh
```diff
@@ -75,16 +69,25 @@ export HISTFILE=/data/.bash_history
...
+  tmux select-pane -t dev:0.1
+  tmux attach-session -t dev
+else
+  # Start the server in the background
+  ./bin/runtime dev -vv > /tmp/server.log 2>&1 &
+  echo "Server started, logs are in /tmp/server.log"
+
+  # Start a shell
+  bash
+fi
```
phinze, line 93: ❔ Tmux setup not working for ya? Will be good to talk through the experience as we iterate on the ergonomics of the dev setup!

pkg/caauth/validate_test.go
```diff
@@ -0,0 +1,191 @@
...
+			name:        "invalid name with spaces",
+			commonName:  "my service",
+			shouldError: true,
+			errorMsg:    "invalid name format",
+		},
+		{
+			name:        "invalid name with uppercase",
+			commonName:  "MyService",
+			shouldError: false, // Uppercase is allowed in DNS-1123
+		},
```
phinze: 🧹 The rabbit nitpick was right that this test name should be "valid" now, and might make sense to move up to ~L52 to keep the valid / invalid cases batched.

pkg/caauth/caauth.go
```diff
@@ -138,8 +154,52 @@ type ClientCertificate struct {
...
 }
 
+var (
+	dnsName    = regexp.MustCompile(`^([a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?$`)
+	simpleName = regexp.MustCompile(`^[a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?$`)
+)
+
+// ValidateCommonName validates that the provided name is a valid DNS-1123 name
+// or a valid domain name suitable for use as a certificate CommonName.
+func ValidateCommonName(name string) error {
```
phinze: 🧹💭 Nit musing: Should this be private to minimize the API surface area of the package?

pkg/testutils/containerd.go
```diff
@@ -29,19 +29,20 @@ func NukeNamespace(cl *containerd.Client, ns string) {
...
-			fmt.Printf("error clearing namespace: %s\n", err)
 		} else {
 			break
 		}
 
-		time.Sleep(time.Second)
+		time.Sleep(time.Second / 2)
 	}
+
+	fmt.Println("Timed out waiting for namespace to be deleted")
```
phinze, line 45: 💭 I've been noticing this consistently leaving dangling ns/containers too. Will keep an eye on if this longer looping cleans things up consistently.

components/coordinate/coordinate.go
```diff
@@ -142,20 +159,44 @@ func (c *Coordinator) LoadAPICert(ctx context.Context) error {
...
-		c.apiCert = data
-		c.apiKey = key
-		return nil
+		x509Cert, err := caauth.LoadCertificate(data)
+		if err == nil {
+			horizon := time.Now().Add(-48 * time.Hour)
+			if x509Cert.NotAfter.Before(horizon) {
+				c.Log.Info("API cert is expired", "cert", x509Cert.NotAfter, "horizon", horizon)
+				goto regen
+			}
```
phinze: ⚠️ I think either I'm misreading or the logic is off here.

This logic reads like we're giving a grace period _after_ expiry for regen, which implies we're okay loading a cert that expired up to 47h ago.

I wonder if the intent is to preemptively regen any cert that we expect to expire in the _next_ 48h? If so, that would look more like:

```go
horizon := time.Now().Add(48 * time.Hour)   // 48 hours from NOW
if x509Cert.NotAfter.Before(horizon) {      // If cert expires before 48h from now
    c.Log.Info("API cert expired or expires soon", "cert", x509Cert.NotAfter, "horizon", horizon)
    goto regen
}
```

components/coordinate/coordinate.go
```diff
@@ -131,6 +135,19 @@ func (c *Coordinator) LoadCA(ctx context.Context) error {
...
+	names := []string{
+		"localhost",
+	}
+
+	names = append(names, c.AdditionalNames...)
+
+	ips := []net.IP{
+		net.ParseIP("127.0.0.1"),
+		net.ParseIP("::1"),
+	}
```
phinze, line 166: ❓ Will this eventually be remote IPs/hostnames or will our mTLS verification skate over the host bits?

phinze (changes requested):
Looking good overall! One potentially-blocking issue w/ cert expiry ⚠️, a few optional nits and misc Qs

hack/dev.sh
evanphx: Oops! Meant to put this in a different PR. I'll follow up with you on this (dagger crashes on every exit for me).

pkg/testutils/containerd.go
evanphx: 👍🏻

components/coordinate/coordinate.go
evanphx: Hilariously this logic was confusing me too! Here is how I puzzled it out.

```
[past] -------------- [ current time: t1 ] ------- [ t1 + 48 hours: h] ------- [future]
                                             ^: t3                            ^: t2

```

we want to know if expiration is t2 or t3 and if it's t3, we want to regenerate it to push the expiration back out.

so what that would mean is that t3.Before(h) == true, right?

components/coordinate/coordinate.go
evanphx: Correct! In mTLS mode, we're just validating the certs without bothering with the host/ip validation logic.

phinze (approved):

PR #62, opened 2025-05-23
Add config merging functionality and improve config management

evanphx:
## Summary
- Add `Merge` method to `Config` struct with `updateActiveCluster` and `force` flags for safe configuration merging
- Implement new `config load` CLI command for merging configurations from files or stdin
- Enhance `auth_generate` command with configurable cluster names and target hostnames

## Test plan
- [x] Test `config load` command with existing and new configurations
- [x] Verify force flag properly handles duplicate cluster names
- [x] Test `updateActiveCluster` flag behavior
- [x] Validate `auth_generate` generates correct cluster configurations
- [x] Test stdin input support with `-` path

## Summary by CodeRabbit

- **New Features**
  - Added a new CLI command to load and merge configuration files, with options for force overwrite and setting the active cluster.
  - Enhanced configuration management: supports reading configs from standard input, writing to standard output, and merging multiple configs.
  - Added options to specify target hostname and cluster name when generating authentication configs.
- **Improvements**
  - Client certificates now include the CA certificate for easier configuration.
  - Config files now track their source path for improved management.
  - Saving and loading of config files is more flexible and robust.
  - RPC client connections support specifying default endpoints and obtaining clients via config.
  - Authentication config generation now issues fresh client certificates and embeds cluster info explicitly.
  - Improved RPC client initialization in various commands and runner to support config-centric connections with error handling.

### DIFF
--- a/clientconfig/clientconfig.go
+++ b/clientconfig/clientconfig.go
@@ -207,7 +207,7 @@ func (c *Config) Merge(other *Config, updateActiveCluster, force bool) error {
 	// Merge clusters from other config
 	for name, cluster := range other.Clusters {
 		if _, exists := c.Clusters[name]; exists && !force {
-			return fmt.Errorf("cluster %q already exists in current config, use force to overwrite", name)
+			return fmt.Errorf("cluster %q already exists in current config, use --force to overwrite", name)
 		}
 
 		c.Clusters[name] = cluster
--- a/clientconfig/clientconfig.go
+++ b/clientconfig/clientconfig.go
@@ -2,6 +2,7 @@ package clientconfig
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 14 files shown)


clientconfig/clientconfig.go
```diff
@@ -162,6 +196,31 @@ func (c *Config) SaveToHome() error {
...
+	for name, cluster := range other.Clusters {
+		if _, exists := c.Clusters[name]; exists && !force {
+			return fmt.Errorf("cluster %q already exists in current config, use force to overwrite", name)
+		}
+
+		c.Clusters[name] = cluster
+	}
+
+	// Update active cluster if requested and other config has one
+	if c.ActiveCluster == "" || (updateActiveCluster && other.ActiveCluster != "") {
```
teresamychu, line 217: ⚠️  Hmmm. This might be an edge case, but this could result in a state where the `Config.ActiveCluster` would be the `old` cluster (before the merge) but the exact configuration is different if you looked the `Config.ActiveCluster` up by name because it was overwritten in the merge.

I could be misunderstanding the use case so im wondering if that a valid state for ClusterConfig to be in?

clientconfig/clientconfig.go
```diff
@@ -162,6 +196,31 @@ func (c *Config) SaveToHome() error {
...
+// Cluster configurations from the other config will be merged into the current config's clusters
+func (c *Config) Merge(other *Config, updateActiveCluster, force bool) error {
+	if c.Clusters == nil {
+		c.Clusters = make(map[string]*ClusterConfig)
+	}
+
+	// Merge clusters from other config
+	for name, cluster := range other.Clusters {
+		if _, exists := c.Clusters[name]; exists && !force {
+			return fmt.Errorf("cluster %q already exists in current config, use force to overwrite", name)
```
teresamychu:
```suggestion
			return fmt.Errorf("cluster %q already exists in current config, use --force to overwrite", name)
```

teresamychu (changes requested):

clientconfig/clientconfig.go
evanphx: That's a good call. If a user specifies --force and we overwrite the cluster configuration of the cluster referenced by ActiveCluster, then the user will use this new version by default.

I think that's ok because they had to specify --force to get into this situation though.

phinze (approved):
LGTM! Alas no methods named `Moosh` but I'll allow it this time

PR #81, opened 2025-06-02
Improve CLI command routing and entity operations

evanphx:
## Summary
- Re-enable the 'set' command for environment variable management with enhanced functionality
- Add ConfigCentric support to entity get/put operations for improved configuration handling
- Update entity commands to use proper RPC client configuration patterns
- Enhance set command with file-based sensitive variable support using @ prefix syntax

## Test plan
- [ ] Test the re-enabled 'set' command functionality
- [ ] Verify entity get/put operations work with both direct addressing and config-based routing
- [ ] Test file-based sensitive variable loading with @ prefix
- [ ] Confirm proper RPC client initialization in all scenarios

## Summary by CodeRabbit

- **New Features**
	- Enabled the "set" command in the CLI.
	- The "set" command now supports reading sensitive environment variable values from a file if specified.

- **Improvements**
	- Enhanced "entity get" and "entity put" commands to support both configuration-based and address-based RPC client creation, improving flexibility when connecting to the coordinator.
	- Updated the "set" command to use an improved API client and updated environment variable handling.
	- Improved internal logic for establishing RPC connections, providing better error handling and fallback mechanisms.
	- Prevented duplicate entries in app configuration commands and variables.
	- Automatically sets the active app version when creating a new app version.

- **User Experience**
	- Changed the short flag for the "Delete" option in the "set" command from -d to -D.

### DIFF
--- a/cli/commands/entity_get.go
+++ b/cli/commands/entity_get.go
@@ -10,18 +10,47 @@ import (
 func EntityGet(ctx *Context, opts struct {
 	Id      string `short:"i" long:"id" description:"Entity ID" required:"true"`
 	Address string `short:"a" long:"address" description:"Address to listen on" default:"localhost:8443"`
-}) error {
-	// Create RPC client to interact with coordinator
-	rs, err := rpc.NewState(ctx, rpc.WithSkipVerify)
-	if err != nil {
-		ctx.Log.Error("failed to create RPC client", "error", err)
-		return err
-	}
 
-	client, err := rs.Connect(opts.Address, "entities")
+	ConfigCentric
+}) error {
+	var (
+		rs     *rpc.State
+		client *rpc.NetworkClient
+	)
+
+	cc, err := opts.LoadConfig()
 	if err != nil {
-		ctx.Log.Error("failed to connect to RPC server", "error", err)
-		return err
+		addr := opts.Address
+		if addr == "" {
+			addr = "localhost:8443"
+		}
+
+		rs, err = rpc.NewState(ctx, rpc.WithSkipVerify)
+		if err != nil {
+			ctx.Log.Error("failed to create RPC client", "error", err)
+			return err
+		}
+
+		// Create RPC client to interact with coordinator
+		client, err = rs.Connect(addr, "entities")
+		if err != nil {
+			ctx.Log.Error("failed to connect to RPC server", "error", err)
+			return err
+		}
+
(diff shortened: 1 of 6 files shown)


cli/commands/entity_put.go
```diff
@@ -16,23 +16,52 @@ func EntityPut(ctx *Context, opts struct {
...
+			return err
+		}
+
+		// Create RPC client to interact with coordinator
+		client, err = rs.Client("entities")
+		if err != nil {
+			ctx.Log.Error("failed to connect to RPC server", "error", err)
+			return err
+		}
 	}
```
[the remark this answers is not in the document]
evanphx: Yeah, this needs to be cleaned up but I'm going to wait for a future date to do that.

cli/commands/entity_get.go
```diff
@@ -10,18 +10,47 @@ import (
...
+			return err
+		}
+
+		// Create RPC client to interact with coordinator
+		client, err = rs.Client("entities")
+		if err != nil {
+			ctx.Log.Error("failed to connect to RPC server", "error", err)
+			return err
+		}
 	}
```
phinze, line 54: ❓ Hm I don't get why are there two big conditional blocks here. Why are there different clients inits depending on the output of LoadConfig erroring?

evanphx: It was a migration consideration because I didn't want to break the old way. But happy to clean this up and give it a clean break too!

phinze: Ah got it! IMHO we're still well into territory for clean breaks, but I think it's AOK if you want to separate out a "clean up RPC client init" as a separate task since I think we could prolly use a full horizontal sweep.

phinze (approved):
That rabbit error handling thing seems legit, but overall the change LGTM! 🚀

servers/app/app.go
```diff
@@ -194,7 +203,7 @@ func (r *AppInfo) SetConfiguration(ctx context.Context, state *app_v1alpha.CrudS
 
 	appVer.Version = name + "-" + idgen.Gen("v")
 
-	_, err = r.EC.Create(ctx, appVer.Version, &appVer)
+	avid, err := r.EC.Create(ctx, appVer.Version, &appVer)
```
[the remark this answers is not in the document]
evanphx: Done!

PR #83, opened 2025-06-02
Add OCI artifact entity type and improve entity operations

evanphx:
## Summary
- Introduce new Artifact entity type to separate OCI manifest storage from AppVersion
- Improve CLI entity operations to properly use client configuration
- Deduplicate configuration values and update active version references

## Test plan
- [ ] Test that new artifacts are created when pushing to registry
- [ ] Verify entity get/put commands work with client config
- [ ] Test set command for managing environment variables
- [ ] Verify app active version is updated after configuration changes
- [ ] Check that commands and variables are properly deduplicated

## Summary by CodeRabbit

- **New Features**
  - Introduced a new artifact entity for applications, enabling more detailed tracking and management of application artifacts.
  - Application versions now reference associated artifacts, providing clearer linkage between versions and their deployable artifacts.

- **Refactor**
  - Updated internal logic to handle manifest storage and retrieval using artifact entities instead of application versions.
  - Adjusted build process to explicitly associate artifacts with application versions.

### DIFF
--- a/components/ocireg/registry.go
+++ b/components/ocireg/registry.go
@@ -5,7 +5,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -18,7 +17,6 @@ import (
 
 	"miren.dev/runtime/api/core/core_v1alpha"
 	"miren.dev/runtime/api/entityserver"
-	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/entity"
 	"miren.dev/runtime/pkg/idgen"
 )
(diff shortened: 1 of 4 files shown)


components/ocireg/registry.go
```diff
@@ -268,35 +266,30 @@ func (h *RegistryHandler) putManifest(w http.ResponseWriter, r *http.Request, na
...
-				h.log.Error("Error creating app version", "name", name, "error", err)
-				w.WriteHeader(http.StatusInternalServerError)
-				return
-			}
-		} else {
-			h.log.Error("Error getting app version", "name", name, "error", err)
-			w.WriteHeader(http.StatusNotFound)
-			return
-		}
+		h.log.Error("Error getting app during artifact creation, will create orphan artifact", "name", name, "error", err)
```
phinze, line 279: 💭 Wondering in the general case how much we can push down transactional operations to protect against race conditions on "get, then put" type situations. (Somewhat different from this scenario, where this is a reasonable fallback behavior to keep going on the create.) Nothing actionable just noodling. 🍜

phinze (approved):
Nice! Looks good

components/ocireg/registry.go
evanphx: Yeah, it's a real good question. In this case, it's a funny dance because we ask buildkitd to do a push and then want to observe that the code that buildkitd talked to created the item.

PR #89, opened 2025-06-04
Fix flaky integration test in runner component

phinze:
## Summary
This PR fixes a flaky integration test in the runner component that was failing intermittently due to race conditions.

Fixes MIR-192

## Problem
The test `TestRunnerCoordinatorIntegration` had two issues causing intermittent failures:

1. **Timing Issues**: The test was using fixed `time.Sleep()` calls and expecting containers to be created or deleted within those time windows. This caused failures when the system was under load or operations took longer than expected.

2. **Race Condition in Deletion**: The test was failing with "entity file not found" errors during sandbox deletion. This occurred when:
   - The Delete method would be called on an already-deleted sandbox
   - It would try to read the entity file which no longer existed  
   - The error would prevent proper cleanup and cause test timeouts

### 1. Replaced Fixed Sleeps with Polling (Original Fix)
- Container creation: Poll every 100ms for up to 10 seconds until container exists
- Container deletion: Poll every 100ms for up to 10 seconds until container is cleaned up

### 2. Made Sandbox Delete Operation Idempotent (Additional Fix)
- Check if the container exists before attempting deletion
- Handle missing entity files gracefully when container exists
- Make `stopSandbox` resilient to missing containers
- Properly handle containerd `ErrNotFound` errors

### 3. Improved Logging and Cleanup
- Added debug logging to ReconcileController to help trace issues
- Tagged reconcile controller logs with the controller name for better debugging
- Removed stray `spew.Dump` debug statement in runner.go

## Test Results
The test now passes consistently without timeout failures:
```
--- PASS: TestRunnerCoordinatorIntegration (7.70s)
```

## Test Plan
- [x] Modified test to use polling instead of fixed sleeps
- [x] Verified the polling logic works correctly
- [x] Fixed race condition in sandbox deletion
- [x] Run test multiple times to ensure flakiness is resolved

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of sandbox and container deletion, ensuring cleanup proceeds smoothly even if containers or metadata are missing, and preventing unnecessary errors when resources are already deleted.

- **Chores**
  - Enhanced logging with additional context for easier troubleshooting and log identification.
  - Set environment variables to disable usage tracking and notifications in integration scripts.
  - Improved integration test reliability and responsiveness by actively waiting for container events instead of using fixed sleep durations.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -511,7 +511,7 @@ func (c *SandboxController) updateSandbox(ctx context.Context, sb *compute.Sandb
 }
 
 func (c *SandboxController) Create(ctx context.Context, co *compute.Sandbox, meta *entity.Meta) error {
-	c.Log.Info("considering sandbox", "id", co.ID, "status", co.Status)
+	c.Log.Info("considering sandbox create or update", "id", co.ID, "status", co.Status)
 	switch co.Status {
 	case compute.DEAD:
 		return nil
@@ -623,7 +623,7 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 		return err
 	}
 
-	c.Log.Info("sanbox started", "id", co.ID, "namespace", c.Namespace)
+	c.Log.Info("sandbox started", "id", co.ID, "namespace", c.Namespace)
 
 	for _, wp := range waitPorts {
 		c.Log.Info("waiting for ports to be bound", "id", cid, "port", wp.port)
@@ -632,7 +632,7 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 
 	co.Status = compute.RUNNING
 
-	// The contrtoller will detect the updates and sync them back
+	// The controller will detect the updates and sync them back
 	meta.Entity.Update(co.Encode())
 
 	err = c.updateServices(ctx, co, meta, ep)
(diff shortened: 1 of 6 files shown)


phinze:
@coderabbitai review

controllers/sandbox/sandbox.go
```diff
@@ -1427,8 +1428,28 @@ loop:
...
 	if err != nil {
+		// If entity file is missing but container exists, try to reconstruct minimal sandbox info
+		if strings.Contains(err.Error(), "entity file not found") {
+			c.Log.Warn("entity file missing but container exists, attempting cleanup", "id", id)
+			// Create a minimal sandbox just with the ID to attempt cleanup
+			sb := &compute.Sandbox{
+				ID: id,
+			}
+			return c.stopSandbox(ctx, sb)
+		}
```
phinze, line 1448: I don't love this clause but it seems to be the scenario we're hitting in the test. @evanphx I do see I kind of raced you a bit with d2f6f67. I still see the failure w/o these changes, but LMK if any of this smells off.

evanphx: This is fine, mostly the code needs to just try and try again until it's clear it's not going to work out.

controllers/sandbox/sandbox.go
```diff
@@ -1448,52 +1469,59 @@ func (c *SandboxController) stopSandbox(ctx context.Context, sb *compute.Sandbox
 	}
 
 	container, err := c.CC.LoadContainer(ctx, c.containerId(sb.ID))
-	if err == nil {
-		labels, err := container.Labels(ctx)
-		if err != nil {
-			return err
-		}
+	if err != nil {
```
evanphx: Yeah, we raced on this code so you'll pick up my changes on a rebase.

evanphx (approved):
Just need to pickup my changes in sandbox, otherwise good!

controllers/sandbox/sandbox.go
phinze: This is actually the result of a Claude-managed conflict resolution w/ your changes... will take a fresh look with these ole human eyes 👀

phinze: Ok done in [843651e](https://github.com/mirendev/runtime/commit/843651ebe71c174ee01ab208dafd1a3d412a7897)

PR #90, opened 2025-06-05
Overwrite meta-attributes when applying schema

evanphx:
We don't have a way to reconcile things yet and we're still a bit
unstable in terms of schema, so we're going to just jam them in for now.

Using WithOverwrite now only applies the overwrite if the entity is
different than the one stored also, reducing the write load.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved entity creation to allow overwriting existing entities, preventing errors when applying schema changes that involve already existing singletons or attributes.


phinze (approved):
Evergreen schema 🌲

PR #94, opened 2025-06-05
Fix 'set host' CLI command

phinze:
## Summary
- Re-enables the `set host` CLI command that was previously commented out
- Adds a proper app entity client wrapper to handle app-specific operations
- Fixes the implementation to properly create/update http_route entities

## Implementation Details
- Created `api/app/client.go` with an App-specific client that wraps entityserver.Client
- Updated `cli/commands/set_host.go` to use the new app client instead of raw entity access
- The command now properly verifies the app exists before setting the host

Fixes MIR-201

## Summary by CodeRabbit

- **New Features**
  - Introduced a new client for managing applications, enabling retrieval of apps by name and setting the hostname for an application.

- **Refactor**
  - Simplified the process for setting an application's hostname in the CLI by using the new app client, making the command more streamlined and user-friendly.

- **Chores**
  - Re-enabled the "set host" command in the CLI, making it available for use again.

### DIFF
--- a/api/app/client.go
+++ b/api/app/client.go
@@ -1,0 +1,73 @@
+package app
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/api/entityserver"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/api/ingress/ingress_v1alpha"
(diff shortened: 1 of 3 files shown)


phinze:
Ok after many battles with a hanging buildkit in CI, I'm backing out of the added e2e tests and getting those in a separate draft PR. This will be just the code changes now so we can review it.

@coderabbitai review

evanphx (approved):

PR #102, opened 2025-06-06
Fix schemagen bytes encoder emptiness condition

phinze:
## Summary
Fixes the bytes single-value encoder in schemagen which had inverted emptiness logic.

## Problem
The encoder was appending bytes attributes when the slice was empty (`len == 0`) instead of when it contained data. This meant:
- Empty byte slices were being encoded and stored unnecessarily 
- Non-empty byte slices were NOT being encoded (data loss\!)

## Solution
- Changed condition from `len(...) == 0` to `len(...) > 0` for encoding
- Updated to use `j.Len()` instead of `j.Id("len").Call()` for consistency with many-value branch
- Fixed the Empty() method to correctly identify non-empty byte slices

## Test plan
- [x] Added comprehensive test coverage in `bytes_test.go`
- [x] Tests verify single-value bytes only encode when non-empty
- [x] Tests verify Empty() method correctly identifies empty entities
- [x] Tests pass with fix, fail with buggy code

Fixes MIR-191

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive unit tests for schema code generation involving fields of type `bytes`, covering single, multiple, required fields, and mixed-type scenarios.

- **Bug Fixes**
  - Corrected encoding and emptiness logic for single bytes fields to trigger only when the byte slice is non-empty.

### DIFF
--- a/pkg/entity/cmd/schemagen/bytes_test.go
+++ b/pkg/entity/cmd/schemagen/bytes_test.go
@@ -1,0 +1,195 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestBytesFieldEncoding(t *testing.T) {
+	// Test schema with both single and many bytes fields
+	sf := &schemaFile{
+		Domain:  "test",
+		Version: "v1",
+		Kinds: map[string]schemaAttrs{
+			"document": {
+				"content": &schemaAttr{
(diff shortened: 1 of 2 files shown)


evanphx (approved):
✨ Good find!

PR #106, opened 2025-06-11
Allow set-host command to use default app when no app is specified

teresamychu:
## Summary
- Modified the `set host` command to make the `--app` flag optional
- When no app is specified, the command now finds and uses the default app
- Added DNS lookup validation for the provided host

## Test plan
- [x] Test `r set host -h example.com` without specifying an app (should use default app)
- [x] Test `r set host -a myapp -h example.com` with explicit app (should work as before)  
- [x] Test with invalid host (should show DNS lookup warning)
- [x] Test when no default app is set (should show helpful error message)

## Summary by CodeRabbit

- **New Features**
	- The host setting command now automatically uses the default app if no app is specified by the user.
	- Added DNS lookup validation for the provided host, with error and warning messages if the lookup fails.

- **Chores**
	- Updated repository settings to ignore the `.run/` directory.


cli/commands/set_host.go
```diff
@@ -40,14 +50,39 @@ func SetHost(ctx *Context, opts struct {
...
+		var app core_v1alpha.App
+		eac := entityserver_v1alpha.NewEntityAccessClient(client)
+		// Find app marked as default
+		resp, err := eac.List(ctx, entity.Bool(core_v1alpha.AppDefaultId, true))
+		if err != nil {
+			return fmt.Errorf("failed to list default app: %w", err)
+		}
+
+		if len(resp.Values()) == 0 {
+			return fmt.Errorf("No default app is currently set. Use app default set <app>.")
```
evanphx, line 80: 💡  I'd probably say something like "Specify an app with -a or configure a default app."

cli/commands/set_host.go
```diff
@@ -40,14 +50,39 @@ func SetHost(ctx *Context, opts struct {
...
+			return nil
+		}
```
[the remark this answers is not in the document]
evanphx: ⚠️ Yup!

cli/commands/set_host.go
```diff
@@ -1,19 +1,29 @@
...
 }) error {
 	var (
 		rs     *rpc.State
 		client *rpc.NetworkClient
 	)
+	_, err := net.LookupIP(opts.Host)
+	if err != nil {
+		ctx.Printf("Warning: DNS lookup failed for host %q: %v\n", opts.Host, err)
+		return err
+	}
```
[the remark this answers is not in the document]
evanphx: 💡 Agree with coderabbit here, if it's a warning, then should we continue anyway?

evanphx (commented):
A few various bits

cli/commands/set_host.go
```diff
@@ -1,19 +1,44 @@
...
+	var confirm = true
+	if err != nil {
+		err = huh.NewConfirm().
+			Title(fmt.Sprintf("Warning: DNS lookup failed for host %q \n"+
+				"Are you sure you want to continue?", opts.Host)).
+			Affirmative("Yes!").
+			Negative("No.").
+			Value(&confirm).
+			WithTheme(huh.ThemeBase()).
+			Run()
```
teresamychu, line 34: I like this confirm prompt thing but now im thinking maybe it should be in its own function to keep the logic clean.

cli/commands/set_host.go
```diff
@@ -40,14 +65,38 @@ func SetHost(ctx *Context, opts struct {
 		}
 	}
 
-	// Create app client
+	appName := opts.App
+
+	if appName == "" {
+		var app core_v1alpha.App
+		eac := entityserver_v1alpha.NewEntityAccessClient(client)
```
phinze, line 72: 💡 The `appClient` initialized on line 94 is an experiment in wrapping an app-centric API around the raw entity server, and this seems like it could be a good candidate to push down to that layer. The signature could be something like:

```go
// in api/app/client.go
func (c *Client) GetDefault(ctx context.Context) (*core_v1alpha.App, error) {
  // ...
}

phinze (approved):
Optional improvement idea in-line but looks like a good idea overall!

PR #109, opened 2025-06-18
Improve profiler test

phinze:
- Reduce sleep and timeout - save 25s of build time!
- Remove stdout prints and add a few more assertions

This doesn't directly fix the flake we observed but should make it
easier to track down if/when it occurs.

Refs MIR-213

## Summary by CodeRabbit

- **Tests**
  - Reduced test output noise by silencing command execution logs.
  - Shortened test duration for faster execution.
  - Improved test assertions for clearer results.

### DIFF
--- a/observability/profile/profile_test.go
+++ b/observability/profile/profile_test.go
@@ -2,13 +2,13 @@ package profile
 
 import (
 	"context"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"testing"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


observability/profile/profile_test.go
```diff
@@ -55,22 +55,22 @@ func TestProfile(t *testing.T) {
...
 
-		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer cancel()
 
 		err = profiler.Start(ctx)
 		r.NoError(err)
 
-		time.Sleep(30 * time.Second)
+		time.Sleep(5 * time.Second)
```
teresamychu, line 64: NICE!!! 🚀

teresamychu (approved):

PR #111, opened 2025-06-20
Rework default app concept to default route

phinze:
## Summary
This PR reworks the "default app" concept to "default route", improving both naming and implementation. The changes are split into two commits:

1. **Rename "default app" to "default route"** - Pure naming changes throughout the codebase
2. **Refactor implementation** - Moves the default functionality from app entities to http_route entities

Completes MIR-222
Fixes MIR-217

### First Commit: Naming Changes
- Renamed all references from "default app" to "default route"
- Changed CLI commands from `app default` to `default-route`
- Updated controller names, log messages, and documentation

#### Schema Changes
- Moved `default` boolean field from app to http_route entity
- Default routes are now a property of the ingress layer, not the app layer

#### Controller Architecture
- Split into two specialized controllers:
  - **DefaultRouteAppController**: Watches app lifecycle to manage default routes automatically
  - **DefaultRouteController**: Ensures only one http_route can be marked as default

#### New Ingress Client
- Created ingress client library with methods for route management:
  - `Lookup()`: Find routes by hostname
  - `LookupDefault()`: Find the current default route
  - `SetDefault()`: Set a route as default
  - `UnsetDefault()`: Remove default status
  - `EnsureSingleDefault()`: Maintain single default invariant

#### CLI Improvements
- Commands now work with http_route entities directly
- Removed `--host` argument from default-route commands (host is irrelevant for defaults)
- Cleaner implementation using the new ingress client

## Behavioral Changes
- When the first app is created, a default route is automatically created
- When the last app is deleted, all default routes are removed
- Only one route can be marked as default at a time
- Default routes no longer have a host field set

## Test Plan
- [ ] Create an app and verify default route is automatically created
- [ ] Create multiple apps and verify only one default route exists
- [ ] Use `default-route set` to change the default route
- [ ] Use `default-route show` to verify the current default
- [ ] Use `default-route unset` to remove the default route
- [ ] Delete all apps and verify default routes are cleaned up
- [ ] Test HTTP routing falls back to default route when no specific route matches

## Summary by CodeRabbit

- **New Features**
  - Introduced management of default HTTP routes, ensuring only one route is marked as default.
  - Added CLI commands to set, unset, and show the default route for applications.
  - Added ability to retrieve applications by ID and track entity list counts.

- **Changes**
  - Default designation shifted from applications to HTTP routes.
  - Updated schemas and user messages to reference "default route" instead of "default app".
  - Simplified client creation by removing error returns and context parameters in app client constructors.
  - Updated HTTP ingress server to use ingress client for route lookups.

- **Bug Fixes**
  - Enforced uniqueness of the default HTTP route across the system.

- **Refactor**
  - Replaced default app controllers with default route controllers.
  - Updated server and CLI components to utilize new default route client abstractions.
  - Modified controllers and runner setup to manage default routes instead of default apps.

- **Documentation**
  - Enhanced CLI flag descriptions and logs to reflect default route terminology.


controllers/ingress/defaultrouteapp.go
```diff
@@ -0,0 +1,91 @@
...
+func (c *DefaultRouteAppController) Create(ctx context.Context, app *core_v1alpha.App, meta *entity.Meta) error {
+	c.Log.Info("App created", "app", app.ID)
+
+	// Check if this is the first app
+	appList, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, core_v1alpha.KindApp))
+	if err != nil {
+		return fmt.Errorf("failed to list apps: %w", err)
+	}
+
+	// If this is the only app (count == 1), create a default route for it
```
teresamychu, line 50: ❓ So when a user is using runtime for the first time and they initialize their first app, 
this will trigger a default route to be created. 

So that means when they create a second app, there will already be a default route set, so unless they specifically unset the default, there will always be a default route as long as you have initialized an app.

controllers/ingress/defaultrouteapp.go
```diff
@@ -0,0 +1,91 @@
...
+	c.Log.Info("App deleted", "app", id)
+
+	// Check if there are any remaining apps
+	appList, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, core_v1alpha.KindApp))
+	if err != nil {
+		return fmt.Errorf("failed to list apps: %w", err)
+	}
+
+	// If no apps remain, delete all default routes
+	if len(appList.Values()) == 0 {
```
teresamychu, line 96: ❓ what happens if they delete the app itself that the default route was created from? should the route remain? or should it also be deleted?

teresamychu (approved):
Some minor questions but LGTM!

controllers/ingress/defaultrouteapp.go
phinze: That's correct! The first app gets the default route, the second app will need to have either a service or a route added for it. I'm thinking we can add some nice UI messages to make this clear to folks.

controllers/ingress/defaultrouteapp.go
phinze: Ooo good call we should clean up the route if the app is removed. Will add that! 🛠️

phinze: Done in c532f8ff9f5ca598cf7c097c5fe80422bed88434

PR #112, opened 2025-06-23
MIR-73: Add init command to initialize new applications

teresamychu:
## Summary
- Add `runtime init` command to initialize new runtime applications
- Add `GetApp` RPC method for proper app existence checking
- Fix issue where init would fail even when no apps existed

## Details
This PR introduces the `runtime init` command which:
- Creates a new app entity in the runtime
- Generates a `.runtime/app.toml` configuration file
- Defaults to using the current directory name as the app name (can be overridden with `-n/--name`)
- Includes proper rollback on failure

The implementation also adds a `GetApp` method to the app CRUD RPC interface to properly check if an app exists before creating it.

## Test plan
- [ ] Build runtime with `make bin/runtime`
- [ ] In a test directory, run `runtime init`
- [ ] Verify app is created with `runtime apps`
- [ ] Verify `.runtime/app.toml` is created
- [ ] Try running `runtime init` again and verify it fails with appropriate error
- [ ] Test with custom name: `runtime init -n my-app`

## Summary by CodeRabbit

- **New Features**
  - Introduced an "init" command in the CLI to initialize a new runtime application environment.
  - The "init" command checks for existing initialization, creates a new application remotely, and sets up local configuration files.
  - Added support for creating and deleting applications by name through the client interface.
- **Bug Fixes**
  - Rollback mechanisms are in place to clean up resources if initialization fails.

### DIFF
--- a/cli/commands/init.go
+++ b/cli/commands/init.go
@@ -1,0 +1,109 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"miren.dev/runtime/api/app/app_v1alpha"
+)
+
+func Init(ctx *Context, opts struct {
+	Name string `short:"n" long:"name" description:"Application name (defaults to current directory name)"`
+}) error {
+	// Determine app name
+	appName := opts.Name
+	if appName == "" {
+		// Use current directory name as default
+		wd, err := os.Getwd()
+		if err != nil {
+			return fmt.Errorf("failed to get current directory: %w", err)
+		}
+		appName = filepath.Base(wd)
+
+		// Sanitize the app name - replace spaces and special chars with hyphens
+		appName = strings.ToLower(appName)
+		appName = strings.ReplaceAll(appName, " ", "-")
+		appName = strings.ReplaceAll(appName, "_", "-")
+	}
+
+	// Check if already initialized BEFORE creating the app
+	runtimeDir := ".runtime"
+	appTomlPath := filepath.Join(runtimeDir, "app.toml")
+	if _, err := os.Stat(appTomlPath); err == nil {
+		return fmt.Errorf("app.toml already exists in .runtime directory - app already initialized")
+	}
+
+	// Create the app entity in the runtime
+	crudcl, err := ctx.RPCClient("dev.miren.runtime/app")
+	if err != nil {
+		return err
+	}
+
+	crudClient := app_v1alpha.NewCrudClient(crudcl)
+
+	// Check if app already exists
+	_, err = crudClient.GetApp(ctx, appName)
+	if err == nil {
+		// App exists if GetApp succeeds
+		return fmt.Errorf("app '%s' already exists", appName)
+	}
+
+	// Create the app
+	results, err := crudClient.New(ctx, appName)
+	if err != nil {
+		return fmt.Errorf("could not create app '%s': %w", appName, err)
+	}
+
+	// From this point on, if we fail we need to rollback
+	var initErr error
+	defer func() {
+		if initErr != nil {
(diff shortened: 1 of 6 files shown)


cli/commands/init.go
```diff
@@ -0,0 +1,109 @@
...
+	}()
+
+	// Create .runtime directory
+	if err := os.MkdirAll(runtimeDir, 0755); err != nil {
+		initErr = fmt.Errorf("failed to create .runtime directory: %w", err)
+		return initErr
+	}
+
+	// Write app.toml with just the name
+	content := fmt.Sprintf("name = \"%s\"\n", appName)
```
phinze: 💡 Raw string printing seems okay for a one line file, but just noting that we have a [`toml` package hanging out in the imports](https://github.com/mirendev/runtime/blob/adff007be7900ce4d6c70762aa4658d50128f1a9/go.mod#L74) for if/when this gets more complicated and you want to use that to generate the config.

cli/commands/init.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"miren.dev/runtime/api/app/app_v1alpha"
+)
+
+func Init(ctx *Context, opts struct {
+	Name string `short:"n" long:"name" description:"Application name (defaults to current directory name)"`
```
phinze: 💡 Might want to take a `dir` flag like [AppCentric commands do](https://github.com/mirendev/runtime/blob/main/cli/commands/app.go#L25) so users aren't required to run this from inside the project dir. I use `runtime deploy -d testdata/bun` from the root of the project a lot! 😄

cli/commands/init.go
```diff
@@ -0,0 +1,109 @@
...
+		return fmt.Errorf("app.toml already exists in .runtime directory - app already initialized")
+	}
+
+	// Create the app entity in the runtime
+	crudcl, err := ctx.RPCClient("dev.miren.runtime/app")
+	if err != nil {
+		return err
+	}
+
+	crudClient := app_v1alpha.NewCrudClient(crudcl)
```
phinze: ⚠️ I think this Crud RPC stuff is left over from the pre-entity-system prototype and is in the process of getting removed. There's an app-specific client at `api/app/client.go` that has a pretty similar Get method called `GetByName`:

https://github.com/mirendev/runtime/blob/main/api/app/client.go#L37-L44

You'll have to port `New` over from the [old RPC server's implementation](https://github.com/mirendev/runtime/blob/adff007be7900ce4d6c70762aa4658d50128f1a9/servers/app/app.go#L47-L67) (I'd call it `Create` TBH)

(The one thing we _don't_ have figured out yet in the new stuff is whether we want to reproduce the intermediate `AppInfo` struct at that layer where we automatically fetch relations like AppVersion. But I think for this use case you're not relying on that so we don't have to think about it yet.)

phinze (changes requested):
🥳 Nice! One requested change about the RPC client and a few optional ideas, but coming along well! 👯

cli/commands/init.go
teresamychu: OOOO nice catch! Completely forgot we're not supposed to use the RPC server stuff anymore.

phinze (approved):
Discussed live the implications of making an app entity and the config together in the `init` command:

 * We want the CLI to be as easy to use as possible, and we want users to be able to set hosts after initializing an app, so making the app as part of init allows this.
 * Do we want this command to work if you're not connected to a runtime server? Unclear at this time, for now we're saying you must be connected to have a "functioning runtime install"

We can always revisit these UX questions as the app evolves!

LGTM!

PR #118, opened 2025-06-26
Add HTTP metrics tracking with async batching

teresamychu:
## Summary
- Display HTTP metrics in app status command with RPS, hourly stats, and top paths
- Implement async batched recording to handle high load efficiently
- Add comprehensive metrics tracking for request count, latency, status codes, and paths

### User-facing features
- **Current RPS** - Shows requests per second for the last minute
- **Hourly statistics** - Displays request counts, average latency, and error rates
- **Top paths** - Lists the 5 most accessed paths with their metrics

### Technical implementation
- Created `HTTPMetrics` struct in `metrics/http.go` with async buffer and batch inserts to ClickHouse
- Integrated metrics collection into `httpingress` server using responseWriter wrapper
- Added RPC schema types for `RequestStat` and `PathStat`
- Used single flush goroutine pattern to prevent resource exhaustion
- Buffer holds up to 1000 requests, flushes every 5 seconds

### Performance considerations
- Async batching minimizes latency impact on request processing
- Single flush goroutine prevents goroutine explosion under high load
- Acceptable data loss risk: up to 1000 requests or 5 seconds of data in case of unexpected termination
- Load tested with 600 concurrent requests: handled gracefully, completed in 36 seconds

## Test plan
- [x] Test basic metrics collection with curl requests
- [x] Verify RPS calculation accuracy
- [x] Test error rate tracking with error-app endpoints
- [x] Load test with 600 concurrent requests
- [x] Verify metrics display in app status command
- [ ] Test metrics persistence across server restarts
- [ ] Verify TTL cleanup after 7 days

## Summary by CodeRabbit

* **New Features**
  * Added real-time HTTP request metrics, including requests per second, request statistics over the last hour, and top accessed paths, to the application status display.
  * Enhanced the user interface to show HTTP stats alongside CPU and memory usage when available.
  * Introduced detailed HTTP request tracking and persistent storage for comprehensive performance insights.

* **Bug Fixes**
  * Ensured that HTTP stats sections only appear when relevant data is present, preventing empty or misleading displays.

* **Chores**
  * Integrated collection and reporting of HTTP metrics throughout the platform for improved observability.

### DIFF
--- a/metrics/http.go
+++ b/metrics/http.go
@@ -1,0 +1,290 @@
+package metrics
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+
+	"miren.dev/runtime/pkg/asm/autoreg"
+)
+
+// HTTPMetrics tracks HTTP request metrics for applications
+type HTTPMetrics struct {
+	Log *slog.Logger
+	DB  *sql.DB `asm:"clickhouse"`
+
+	// Buffering
+	buffer    []HTTPRequest
+	mu        sync.Mutex
+	flushCtx  context.Context
+	cancel    context.CancelFunc
+	flushChan chan struct{} // Signal channel for flush requests
+	wg        sync.WaitGroup
+}
+
+const defaultBufferSize = 1000
+
+var _ = autoreg.Register[HTTPMetrics]()
+
+func (h *HTTPMetrics) Populated() error {
+	return h.Setup()
+}
+
+func (h *HTTPMetrics) Setup() error {
+	_, err := h.DB.Exec(`
+CREATE TABLE IF NOT EXISTS http_requests (
+    timestamp DateTime64(6) CODEC(Delta(8), ZSTD(1)),
+    app LowCardinality(String) CODEC(ZSTD(1)),
+    method LowCardinality(String) CODEC(ZSTD(1)),
+    path String CODEC(ZSTD(1)),
+    status_code UInt16 CODEC(T64, ZSTD(1)),
+    duration_ms UInt32 CODEC(T64, ZSTD(1)),
+    response_size UInt64 CODEC(T64, ZSTD(1)),
+    INDEX idx_path path TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 1
+) 
+ENGINE = MergeTree
+ORDER BY (app, toUnixTimestamp(timestamp))
+TTL toDateTime(timestamp) + INTERVAL 7 DAY
+SETTINGS ttl_only_drop_parts = 1;
+`)
+	if err != nil {
+		return err
+	}
+
+	// Initialize buffer and channels
+	h.buffer = make([]HTTPRequest, 0, defaultBufferSize)
+	h.flushChan = make(chan struct{}, 1) // Buffered to avoid blocking
+
+	// Start background flush routine
+	h.flushCtx, h.cancel = context.WithCancel(context.Background())
+	h.wg.Add(1)
+	go h.flushLoop()
+
+	return nil
+}
+
+// HTTPRequest represents a single HTTP request for metrics
+type HTTPRequest struct {
+	Timestamp    time.Time
+	App          string
+	Method       string
+	Path         string
+	StatusCode   int
+	DurationMs   int64
+	ResponseSize int64
+}
+
+// RecordRequest adds a request to the buffer for async processing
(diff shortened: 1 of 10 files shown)


servers/httpingress/httpingress.go
```diff
@@ -221,13 +276,27 @@ func (h *Server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-	} else {
-		// Use the http route found
-		targetAppId = route.App
-		routeType = "route"
+		h.Log.Debug("using default route", "host", onlyHost, "app", targetAppId)
 	}
 
-	h.Log.Info("routing request", "host", onlyHost, "app", targetAppId, "type", routeType)
+	// Get app details first to have the name for metrics
+	gr, err := h.eac.Get(ctx, targetAppId.String())
```
phinze, line 283: 💭 I know this is just moving, but looking at the diff has me thinking about these data store queries running on the hot path of every HTTP request. I think this is totally OK for now, but I'm betting we'll want to investigate some performance optimizations (caching, etc) sometime soon-ish. Food for future thought!

metrics/http.go
```diff
@@ -0,0 +1,290 @@
...
+	defer h.wg.Done()
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			h.flush()
+		case <-h.flushChan:
+			h.flush()
```
phinze, line 115: 💭 I don't think these conditions warrant log lines, but these would be good places to tick metrics once we have our _own_ internal metrics setup figured out. It would be helpful to see how often the buffer was flushing for time vs overflow so we can tune the buffer size.

metrics/http.go
```diff
@@ -0,0 +1,290 @@
...
+// HTTPRequest represents a single HTTP request for metrics
+type HTTPRequest struct {
+	Timestamp    time.Time
+	App          string
+	Method       string
+	Path         string
+	StatusCode   int
+	DurationMs   int64
+	ResponseSize int64
+}
```
phinze, line 78: 💭 Asked Copilot to do some back-of-napkin buffer size estimation just for funsies. Here's what it came up with!

----

Given your average values:

- App: <10 characters (let's assume 10 bytes for estimation)
- Method: HTTP verbs, max length is 6 ("DELETE")
- Path: “any HTTP path,” but typically not huge. For estimation, let's assume an average of 100 characters (can adjust if you have real data).

### Memory breakdown per HTTPRequest instance

- **Timestamp:** 16 bytes (time.Time)
- **App:** 16 bytes (string header) + 10 bytes (contents)
- **Method:** 16 bytes (string header) + 6 bytes (contents)
- **Path:** 16 bytes (string header) + 100 bytes (contents)
- **StatusCode:** 8 bytes (`int`)
- **DurationMs:** 8 bytes (`int64`)
- **ResponseSize:** 8 bytes (`int64`)

#### Struct header size (always present in buffer):

- 16 (time.Time)
- 16 (App header)
- 16 (Method header)
- 16 (Path header)
- 8 (StatusCode)
- 8 (DurationMs)
- 8 (ResponseSize)
= **88 bytes**

#### String contents (on heap, per struct):

- App: 10 bytes
- Method: 6 bytes
- Path: 100 bytes
= **116 bytes**

#### Total per struct: 88 + 116 = **204 bytes**

---

### Buffer for 1000 structs

- 204 bytes × 1000 = **204,000 bytes** = **204 KB**

---

**Summary:**

- The buffer of 1000 HTTPRequest structs as described will use about **204 KB of memory** (assuming average path size is 100 bytes).  
- If your paths are typically longer or shorter, adjust the 100 in the estimate accordingly.
- If you want a tighter estimate, sample your actual traffic for path lengths.

phinze (approved):
This is great! Lots to think about in terms of other ways of presenting this info (e.g. graph, slicing info, etc), but these seem like perfect next steps. Nothing but musings inline.

metrics/http.go
teresamychu: for sure! and how that changes with different applications! And this just reminded me that we still have to figure out what kind of internal metrics we want to include in the MVP

servers/httpingress/httpingress.go
teresamychu: Excellent point! I figured this is ok for now cause i didnt want to over-engineer it but caching is definitely something we'll have to figure out before we roll out an external runtime server especially since we dont expect app-name to change that often.

PR #124, opened 2025-07-07
Add comprehensive concurrency controls for apps

phinze:
## Summary
- Adds support for singleton applications with dedicated instance management
- Introduces new concurrency configuration options including `max_instances` and `requests_per_instance` 
- Configures Wordle as the first singleton app with `max_instances: 1`
- Fixes lease acquisition issues for apps with unlimited concurrency
- Adds proactive min instance enforcement

### Concurrency Configuration
- Extended concurrency configuration in core schema to support instance controls
- Renamed `concurrency.fixed` to `concurrency.requests_per_instance` for clarity
- Added `min_instances`, `max_instances`, and `scale_down_delay` fields
- Enhanced activator with singleton app logic and lease management

### Lease Acquisition Fix
- Fixed issue where apps with `requests_per_instance: 0` (unlimited) couldn't reuse instances
- Properly models unlimited concurrency by keeping leaseSlots at 0 instead of forcing to 1
- Changed slot availability check from `<` to `<=` to handle slot boundaries correctly
- Added documentation to clarify the activator/httpingress flow

### Min Instances Controller
- Added controller that watches app entities and ensures min instances are created immediately
- Runs in the coordinator alongside the activator
- Proactively creates required sandboxes when apps are deployed or updated
- Works with activator's background task for ongoing instance maintenance

## Test Plan
- [x] Manual testing of Wordle app as singleton
- [x] Verify lease acquisition and expiry behavior  
- [x] Test traffic routing to singleton instance under load
- [x] Verify min instances are created on app deployment

## Stacked on
- #125 (duration type support for schemagen)

Refs: MIR-227

## Summary by CodeRabbit

* **New Features**
  * Introduced advanced concurrency controls with configurable min/max instances, requests per instance, and scale-down delay.
  * Added a deployment controller to automate application version deployments.
  * Added a sample "wordle" application configuration to test data.

* **Improvements**
  * Enhanced sandbox lifecycle management enforcing min/max instance constraints and improved scaling logic.
  * Improved logging and documentation for activator and HTTP ingress components.
  * Updated configuration schemas and interfaces to support new concurrency options.
  * Integrated controller management into Coordinator for deployment handling.

* **Bug Fixes**
  * Removed old concurrency fields and replaced them with new, granular settings.

* **Chores**
  * Removed obsolete defaults and test concurrency configuration fields.


components/activator/activator.go
```diff
@@ -383,3 +442,59 @@ func (a *localActivator) retireUnusedSandboxes() {
...
+		if minInstances <= 0 {
+			continue
+		}
+
+		maxInstances := int(vs.ver.Config.Concurrency.MaxInstances)
+
+		// Count running sandboxes
+		runningCount := 0
+		for _, sb := range vs.sandboxes {
+			if sb.sandbox.Status == compute_v1alpha.RUNNING {
```
teresamychu: ❓ This might be an edge case? But if you just started up a new sandbox and its status was 'PENDING' or 'NOT READY', do we want to include it in the count? I feel like there is a case here where you just spun up a sandbox but its not good to go yet, but this will try to spin up more instances?

components/activator/activator.go
```diff
@@ -383,3 +442,59 @@ func (a *localActivator) retireUnusedSandboxes() {
...
+				runningCount++
+			}
+		}
+
+		// Create additional sandboxes if needed
+		for i := runningCount; i < minInstances; i++ {
+			// Don't exceed max_instances if set
+			if maxInstances > 0 && len(vs.sandboxes) >= maxInstances {
```
teresamychu: ❓ `len(vs.sandboxes)` would include dead ones i think?

controllers/app/mininstances.go
```diff
@@ -0,0 +1,97 @@
...
+
+	c.Log.Info("ensuring min instances for app",
+		"app", app.ID,
+		"version", appVersion.Version,
+		"min_instances", minInstances)
+
+	// The activator's AcquireLease will create sandboxes up to min_instances
+	// We just need to trigger it for each required instance
+	for i := int64(0); i < minInstances; i++ {
+		lease, err := c.AA.AcquireLease(ctx, &appVersion, "default", "")
```
teresamychu: ⚠️  Forgive me if im misunderstanding something, im still processing....but I think that `AcquireLease` will only create a sandbox if all current sandboxes have no more free slots? It will just return the lease of one of the sandbox that still have available slot capacity.

Which means that calling AcquireLease doesn't guarantee that it will create a new sandbox? It might return the lease of an existing one instead.

teresamychu (changes requested):
Thanks for walking me through this PR today, it was very helpful!

components/activator/activator.go
phinze: Good catch! This actually points at a deeper issue, which is that right now the activator is assuming that nothing changes outside its world view. It builds up the in-memory view in `a.versions` when it creates sandboxes and then... never updates it. 🤦🏻

That means the activator won't catch if a sandbox crashes for any reason... it'll just assume it's running until it retires it.

I'll give this some thought and follow up. 💭

components/activator/activator.go
phinze: It won't because this is just looping over the in-memory view of sandboxes the activator has created and not retired. But that's got other issues, per above comment! 😵‍💫

evanphx:
The sandbox controller should handle crashes, so the activator can count on
it.

controllers/app/mininstances.go
phinze: Yeah you're right. This really only kind of works in the create case, so it passed my manual testing, but it doesn't make much sense as an impl.

Why it works on Create:

 * For apps that have never seen traffic, acquiring a lease will boot the first sandbox
 * That first lease will also cause the activator's background task to trigger, which will make the rest of the sandboxes

But that's sort of a convoluted way to get min_instances booted, and it's definitely going to have issues in the update scenario. Will 💭 and follow up!

phinze:
Good point @evanphx! That means the activator doesn't have to go watching for sandbox status changes... it can just assume that ones it's created will do their best to stay running.

It does look like it'll get out of sync on process restarts though. The sandbox controller will come back up and try and reconcile its view of the world w/ existing sandboxes in etcd and containerd, but the activator will come up w/ new blank in-memory structs and believe it's starting w/ a blank slate.

controllers/app/mininstances.go
phinze: OK! 7bfdc03e4778ce45f527d34d8d408a84cdad7fa4 has an adjusted approach which makes the activator solely responsible for maintaining min_instances and changes the controller's responsibility to just deployment of new app versions when they show up. Take a look and LMK what you think!

components/activator/activator.go
phinze: Swinging back around after my rework in 7bfdc03e4778ce45f527d34d8d408a84cdad7fa4 - we now have the Activator assume that the SandboxController will do everything it can to keep sandboxes alive, so it doesn't carry responsibility for a sandbox's state, merely its existence.

We'll probably need to write some more specific behavior to help people manage crash loops and rollbacks and stuff, but one step at a time! 👣

controllers/app/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Get the app version details
+	vr, err := c.EAC.Get(ctx, app.ActiveVersion.String())
+	if err != nil {
+		return fmt.Errorf("failed to get app version: %w", err)
+	}
+
+	var appVersion core_v1alpha.AppVersion
+	appVersion.Decode(vr.Entity().Entity())
+
```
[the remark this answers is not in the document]
phinze: appVersion.Decode does not return an error

phinze:
Welp that e2e failure is real, there's a deadlock somewhere in all this new code. That's tomorrow's mission! 🌆 🔖

phinze:
@evanphx @teresamychu let's plan to have a team chat about this on Monday. I've got the build passing but I have some Big Questions about various system architecture / MVP scoping pieces around this.

components/activator/activator.go
```diff
@@ -383,3 +489,199 @@ func (a *localActivator) retireUnusedSandboxes() {
...
+
+	// If our version tracking contains this version, we can assume it's been deployed and return early
+	a.mu.Lock()
+	key := verKey{ver.ID.String(), pool}
+	_, exists := a.versions[key]
+	a.mu.Unlock()
+	if exists {
+		a.log.Debug("app version already deployed", "app", app.ID, "version", ver.Version, "pool", pool)
+		return nil
+	}
```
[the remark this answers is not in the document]
phinze: Yeah that's a possibility, I briefly tried an alternate approach where every public method held the mutex for its duration, but that meant that deploys could block incoming traffic for the duration of sandbox boot, which didn't seem desirable.

The reality is there's a lot of contention for the structures here between the lease management and the version tracking.

It might be worth taking a step back and asking some bigger questions about the components involved here. I'll have a discussion with the team next week and decide how to approach.

phinze:
Still planning to swing around to this! Will pull back to draft and re-mark ready for review when I do.

phinze:
Reworked into #162

PR #126, opened 2025-07-08
Adding different apps to testdata.

teresamychu:
Needed more apps for testing when I was adding metrics and I thought 

"this seems useful, maybe i should share!"

## Summary by CodeRabbit

* **New Features**
  * Added sample applications for Node.js, Go, Python (Flask), and Ruby (Sinatra), each with basic endpoints including a root greeting and a health check.
  * Introduced additional endpoints in the Node.js error app to simulate various HTTP error responses and delayed responses.
  * Included configuration files (Procfile, package metadata, and dependency files) for each sample app to support deployment and execution.

### DIFF
--- a/testdata/error-app/Procfile
+++ b/testdata/error-app/Procfile
@@ -1,0 +1,1 @@
+web: node index.js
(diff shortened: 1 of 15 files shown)


phinze (approved):
Ooo nice thanks these will definitely come in handy for future testing!

PR #128, opened 2025-07-09
Add runtime config management commands

teresamychu:
## Summary
- Adds `runtime config set-active` command to switch between configured clusters
- Adds `runtime config remove` command to remove clusters from configuration
- Both commands support interactive selection menus when no cluster name is provided

### runtime config set-active
Allows users to switch the active cluster either by specifying a cluster name or using an interactive menu.

```bash
# Direct selection
runtime config set-active <cluster-name>

# Interactive menu
runtime config set-active
```

### runtime config remove
Allows users to remove clusters from their configuration with safety checks.

```bash
# Remove with confirmation
runtime config remove <cluster-name>
runtime config remove --force <cluster-name>

# Interactive selection
runtime config remove
```

## Safety Features
- Cannot remove the active cluster (must switch first)
- Cannot remove the last cluster
- Requires `--force` flag for confirmation (follows existing CLI patterns)
- Interactive menus show current active cluster

## Test plan
- [x] Test `config set-active` with direct cluster name
- [x] Test `config set-active` with interactive menu
- [x] Test `config remove` without force flag
- [x] Test `config remove` with force flag
- [x] Test removing active cluster (should fail)
- [x] Test removing last cluster (should fail)
- [x] Test interactive selection in both commands

## Summary by CodeRabbit

* **New Features**
  * Added a command to set the active cluster, with interactive selection if no cluster is specified.
  * Added a command to remove a cluster, featuring an interactive UI for selection and safeguards against removing the last or active cluster.
* **User Experience**
  * Interactive terminal menus provide guided selection and clear feedback during cluster management tasks.

### DIFF
--- a/cli/commands/config_remove.go
+++ b/cli/commands/config_remove.go
@@ -3,10 +3,6 @@ package commands
 import (
 	"fmt"
 	"sort"
-	"strings"
-
-	tea "github.com/charmbracelet/bubbletea"
-	"github.com/charmbracelet/lipgloss"
 )
 
 func ConfigRemove(ctx *Context, opts struct {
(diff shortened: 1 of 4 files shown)


cli/commands/config_remove.go
```diff
@@ -0,0 +1,180 @@
...
+			if m.cursor > 0 {
+				m.cursor--
+			}
+
+		case "down", "j":
+			if m.cursor < len(m.clusters)-1 {
+				m.cursor++
+			}
+		}
+	}
```
phinze: 💡 There's some code shared by the two commands here that we'll probably want to consolidate eventually to make sure we don't e.g. accidentally have different keyboard shortcuts on different select boxes. But def not a blocker for the first two!

phinze (approved):
Nice improvement! Muuuuch better than manually messing around with YAML files. 📈

cli/commands/config_remove.go
teresamychu: YUP good catch! I think im going to clean this up now and put all this selection model code into a utils file.

PR #130, opened 2025-07-10
Fix bug preventing restarts after project/default changes

phinze:
The Coordinator would fail to boot after combo.yaml was submitted to a
server, because that file includes a field update for project/default.

We were not catching the Update path on entityserver.Put because the
request did not include an Id, even though db/ident is set in the attrs.
Instead we were relying on an exact-match clause in the response
handling from the Create path.

Implementing a "fixup" at the RPC layer addresses allows the
project/default Put operation to become an update.

There's more work we can do to clarify the semantics here, but this
fixes the active issue for now.

Fixes MIR-231

Co-authored-by: Teresa Chu <teresa@miren.dev>

## Summary by CodeRabbit

* **New Features**
  * Improved logging with additional context in both the entity store and entity server modules.
  * Enhanced error reporting and debug logs for entity creation and update operations.
  * Added a default project owner constant and refined default project creation with a typed project structure.

* **Bug Fixes**
  * Corrected typos in documentation comments related to entity fields.

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -52,6 +52,8 @@ type CoordinatorConfig struct {
 	Logs *observability.LogReader
 }
 
+const DefaultProjectOwner = "miren.system@miren.dev"
+
 func NewCoordinator(log *slog.Logger, cfg CoordinatorConfig) *Coordinator {
 	return &Coordinator{
 		CoordinatorConfig: cfg,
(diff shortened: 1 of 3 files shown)


phinze:
Looks like a couple of real test failures. Will take a look tomorrow morning! (Unless @teresamychu gets to it first 😄 )

teresamychu:
> Looks like a couple of real test failures. Will take a look tomorrow morning! (Unless @teresamychu gets to it first 😄 )

HAHA I did NOT in fact get to it first.

phinze:
OK I think there are more implicit assumptions about the semantics of create vs update spread across the layers than we thought. 

c03b033b7246a80455da4ec77fb13d5f1582b6bd is a rework with smaller bandaid that doesn't do the Fixup globally, which is switching the boot process to use entityserver.Client's CreateOrUpdate instead of EntityAccessClient's Put

teresamychu (commented):
LGTM! Thanks for fixing this Paul!

teresamychu (approved):

components/coordinate/coordinate.go
```diff
@@ -52,6 +52,8 @@ type CoordinatorConfig struct {
 	Logs *observability.LogReader
 }
 
+const DefaultProjectOwner = "miren.system@miren.dev"
```
phinze, line 55: cc @evanphx I'm assuming you have a catchall set up but just in case you don't and want this to be something different lmk

phinze:
> LGTM! Thanks for fixing this Paul!

Thanks for discovering the issue and co-debugging it yesterday! It was a fun bug hunt 🙌🏻 🐞 🔍

components/coordinate/coordinate.go
evanphx: Yeah, I don't even think it works. I think I just typed that in for fun.

PR #135, opened 2025-07-17
Implement automatic cleanup of dead sandboxes

evanphx:
## Summary

This PR implements automatic cleanup functionality that removes dead sandboxes older than 1 hour from the entity store, preventing accumulation of stale records.

### UpdatedAt System Attribute
- Added `UpdatedAt` as a proper system attribute that gets automatically set on entity updates
- Enhanced `EtcdStore.UpdateEntity` to detect and promote `UpdatedAt` values to the entity struct field
- Allows explicit setting of `UpdatedAt` for testing while auto-updating for normal operations

### Periodic Controller Framework
- Extended `ReconcileController` with optional periodic callback support (runs every 10 minutes)
- Added `PeriodicController` interface for controllers that need maintenance tasks
- Modified runner to detect and wire up periodic callbacks automatically

### Dead Sandbox Cleanup
- Implemented `Periodic` method in `SandboxController` that:
  - Lists all sandboxes and checks their status and last update time
  - Deletes sandboxes that have been `DEAD` for more than 1 hour
  - Logs cleanup actions for monitoring
- Runs automatically every 10 minutes via the periodic callback framework

### Supporting Changes
- Fixed entity conversion in `entityserver` to properly populate `CreatedAt`/`UpdatedAt` fields
- Added comprehensive test with proper `Ident` attributes to verify cleanup behavior
- Updated runner to integrate periodic callbacks with sandbox controller

## Test Plan

- [x] Added test that creates sandboxes, marks one as dead with old timestamp, and verifies cleanup
- [x] Test uses proper `Ident` attributes to ensure predictable entity names
- [x] Verified periodic cleanup removes only old dead sandboxes while preserving active ones
- [x] Confirmed `UpdatedAt` attribute promotion and field population works correctly

## Impact

This prevents the entity store from accumulating dead sandbox records indefinitely, improving system cleanliness and potentially reducing storage/query overhead over time. The 1-hour grace period ensures sandboxes aren't cleaned up too aggressively while still maintaining reasonable cleanup intervals.

## Summary by CodeRabbit

* **New Features**
  * Added periodic cleanup to automatically remove dead sandboxes older than one hour.
  * Enhanced sandbox lifecycle management with improved cleanup, retry logic, and robust endpoint deletion.
  * Service controller now has comprehensive tests covering service and endpoint lifecycle, updates, and IP allocation.
  * Introduced periodic task support for controllers, enabling scheduled background operations.
  * Extended entity data to include creation and update timestamps for better tracking.

* **Bug Fixes**
  * Improved handling of entity update timestamps to ensure accurate tracking of changes.

* **Documentation**
  * Expanded test coverage for sandbox and service controllers, improving reliability and maintainability.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1635,3 +1635,56 @@ func (c *SandboxController) stopSandbox(ctx context.Context, sb *compute.Sandbox
 
 	return nil
 }
+
+// Periodic implements the PeriodicController interface
+// It runs every 10 minutes to clean up dead sandboxes that are older than 1 hour
+func (c *SandboxController) Periodic(ctx context.Context) error {
+	c.Log.Info("running periodic cleanup of dead sandboxes")
+
+	// List all sandboxes
+	resp, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, compute.KindSandbox))
+	if err != nil {
+		return fmt.Errorf("failed to list sandboxes: %w", err)
+	}
+
+	now := time.Now()
+	oneHourAgo := now.Add(-1 * time.Hour)
+
+	var deleted int
+	for _, e := range resp.Values() {
+		var sb compute.Sandbox
+		sb.Decode(e.Entity())
+
+		// Check if sandbox is DEAD and UpdatedAt is more than 1 hour ago
+		if sb.Status == compute.DEAD {
+			// Convert UpdatedAt from milliseconds to time.Time
+			updatedAt := time.Unix(0, e.Entity().UpdatedAt*int64(time.Millisecond))
+
+			c.Log.Debug("checking sandbox for cleanup",
+				"id", sb.ID,
+				"status", sb.Status,
+				"updated_at", updatedAt.Format(time.RFC3339),
+				"age", now.Sub(updatedAt).String())
+
+			if updatedAt.Before(oneHourAgo) {
+				c.Log.Info("deleting old dead sandbox",
+					"id", sb.ID,
+					"updated_at", updatedAt.Format(time.RFC3339),
+					"age", now.Sub(updatedAt).String())
+
+				_, err := c.EAC.Delete(ctx, sb.ID.String())
+				if err != nil {
+					c.Log.Error("failed to delete dead sandbox", "id", sb.ID, "error", err)
+					continue
+				}
+				deleted++
+			}
+		}
+	}
+
+	if deleted > 0 {
+		c.Log.Info("periodic cleanup completed", "deleted_sandboxes", deleted)
+	}
+
+	return nil
+}
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -385,7 +385,7 @@ func (c *SandboxController) containerPrefix(id entity.Id) string {
 	return "sandbox." + cid
 }
 
-func (c *SandboxController) containerId(id entity.Id) string {
+func (c *SandboxController) pauseContainerId(id entity.Id) string {
 	return c.containerPrefix(id) + "_pause"
 }
 
@@ -394,7 +394,7 @@ func (c *SandboxController) checkSandbox(ctx context.Context, co *compute.Sandbo
 
 	ctx = namespaces.WithNamespace(ctx, c.Namespace)
 
-	cont, err := c.CC.LoadContainer(ctx, c.containerId(co.ID))
+	cont, err := c.CC.LoadContainer(ctx, c.pauseContainerId(co.ID))
 	if err != nil {
 		return notFound, nil
 	}
@@ -471,7 +471,7 @@ func (c *SandboxController) updateSandbox(ctx context.Context, sb *compute.Sandb
 		c.Log.Error("failed to check if sandbox can be updated in place", "err", err)
 	} else if canUpdate {
 
-		cont, err := c.CC.LoadContainer(ctx, c.containerId(sb.ID))
+		cont, err := c.CC.LoadContainer(ctx, c.pauseContainerId(sb.ID))
 		if err != nil {
 			return fmt.Errorf("failed to load existing sandbox: %w", err)
 		}
@@ -564,7 +564,7 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 		return fmt.Errorf("failed to configure volumes: %w", err)
 	}
 
-	cid := c.containerId(co.ID)
+	cid := c.pauseContainerId(co.ID)
 
 	container, err := c.CC.NewContainer(ctx, cid, opts...)
 	if err != nil {
@@ -991,6 +991,7 @@ func (c *SandboxController) buildSpec(
 	// Make sure the parent directory for the cgroup exists so that gvisor doesn't
 	// try to remove it. This works around a bug in gvisor where it will attempt
 	// to remove the any cgroup directory it creates.
+	// PR with gvisor fix: https://github.com/google/gvisor/pull/11933
 	err = os.MkdirAll(filepath.Join("/sys/fs/cgroup", filepath.Dir(cgroupPath)), 0755)
 	if err != nil {
 		c.Log.Error("failed to create cgroup directory", "path", cgroupPath, "err", err)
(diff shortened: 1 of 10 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -723,6 +729,47 @@ func (c *SandboxController) addEndpoint(
 	return nil
 }
 
+func (c *SandboxController) deleteEndpoints(ctx context.Context, sb *compute.Sandbox, sandboxIPs map[string]bool) error {
```
phinze, line 737: 🙌🏻 Yessss I hit IP address exhaustion yesterday in my long debugging session so this is much appreciated! 🙇🏻

controllers/sandbox/sandbox.go
```diff
@@ -929,6 +985,17 @@ func (c *SandboxController) buildSpec(
 		},
 	}
 
+	// Create unique cgroup path for this sandbox
+	cgroupPath := fmt.Sprintf("/miren/sandbox-%s", sb.ID.PathSafe())
+
+	// Make sure the parent directory for the cgroup exists so that gvisor doesn't
+	// try to remove it. This works around a bug in gvisor where it will attempt
+	// to remove the any cgroup directory it creates.
```
phinze, line 998: 💡 Link to PR in comment so some far future person can figure out when workaround is moot?

controllers/sandbox/sandbox.go
```diff
@@ -929,6 +985,17 @@ func (c *SandboxController) buildSpec(
 		},
 	}
 
+	// Create unique cgroup path for this sandbox
+	cgroupPath := fmt.Sprintf("/miren/sandbox-%s", sb.ID.PathSafe())
```
phinze, line 994: ❓ Just curious what was the old behavior was - does this match it or introduce a new cgroup layout?

controllers/sandbox/sandbox.go
```diff
@@ -379,12 +379,16 @@ func (c *SandboxController) canUpdateInPlace(ctx context.Context, sb *compute.Sa
...
+func (c *SandboxController) containerPrefix(id entity.Id) string {
 	cid := id.String()
 	cid = strings.TrimPrefix(cid, "sandbox/")
 	return "sandbox." + cid
 }
 
+func (c *SandboxController) containerId(id entity.Id) string {
+	return c.containerPrefix(id) + "_pause"
+}
+
```
phinze, line 392: 💡 This is sort of a confusing function w/ the _pause suffix. Consider a rename or, if you want to keep the notion of the pause container being the "id", an explanatory comment?

controllers/sandbox/sandbox.go
```diff
@@ -1331,100 +1402,109 @@ func (c *SandboxController) buildSubContainerSpec(
...
-	// Now, we can delete the subcontainers.
-	for _, container := range sb.Container {
-		id := fmt.Sprintf("%s-%s", c.containerId(sb.ID), container.Name)
+		// Check if any containers still exist
+		remainingContainers := make([]string, 0)
+		for _, id := range containerIds {
+			cont, err := c.CC.LoadContainer(ctx, id)
+			if err != nil {
+				// Container doesn't exist, consider it deleted
+				c.Log.Debug("container not found, considering deleted", "id", id)
```
phinze: 💡 Any way to scope this to a not found error? I'm considering the scenario where containerd gets borked somehow. We wouldn't want this to cruise through a pretend things were cleaned up, only for them to leak when containerd comes back to life. 🧟 

Ideally (provided etcd state is durable) this process should reliably converge to a state that doesn't leak anything miren created, even accounting for restarts of both runtime and containerd.

But I could see calling for a punt here if you don't want to get into these weeds just yet. 🏈

pkg/controller/controller.go
```diff
@@ -284,6 +298,34 @@ func (c *ReconcileController) periodicResync(ctx context.Context) {
...
+// runPeriodic runs the periodic callback every 10 minutes
+func (c *ReconcileController) runPeriodic(ctx context.Context) {
+	c.Log.Info("Starting periodic callback")
+	defer c.Log.Info("Stopping periodic callback")
+
+	defer c.wg.Done()
+
+	// Run every 10 minutes
+	ticker := time.NewTicker(10 * time.Minute)
+	defer ticker.Stop()
```
phinze, line 317: 💡 Consider adding this as an argument to SetPeriodic rather than hard coding the timing.

pkg/controller/controller.go
```diff
@@ -295,6 +337,12 @@ type GenericController[P ControllerEntity] interface {
 	Delete(ctx context.Context, e entity.Id) error
 }
 
+// PeriodicController is an optional interface that controllers can implement
+// to handle periodic callbacks
+type PeriodicController interface {
+	Periodic(ctx context.Context) error
+}
+
```
phinze: 💡 FWIW This isn't used anywhere, since we're using `SetPeriodic` w/ the function explicitly in `runner.go`. If you wanted to implement as automatic interface detection, you'd need some other type assertion dance in the Adapt/Reconcile/constructor flow.

IMHO you can just nix the interface for now and let it be a feature that gets wired explicitly.

pkg/entity/store.go
```diff
@@ -652,6 +652,16 @@ func (s *EtcdStore) UpdateEntity(
 
 	entity.Attrs = primary
 
+	// Set UpdatedAt based on whether it was explicitly provided or already exists
+	// Check if UpdatedAt attribute exists in the entity after update
```
phinze, line 656: 💡 There are a few TODOs relating to this floating around you may want to have a peek at. Some of them might be complete now w/ this addition?

```
❯ git grep 'TODO.*updated'
servers/entityserver/entityserver.go:           // TODO: handle updated_at
servers/entityserver/entityserver.go:   // TODO: handle created_at and updated_at fields
servers/entityserver/entityserver.go:           // TODO: handle updated_at
servers/entityserver/entityserver.go:           // TODO: handle created_at and updated_at fields
```

phinze (approved):
Heck yeah! 💃🏻 A few recommendations and questions, but nothing blocking.

controllers/sandbox/sandbox.go
evanphx: MMm, sure!

controllers/sandbox/sandbox.go
evanphx: it was the countainerd namespace then the container id. This is just to make it extra explicit so we know we're working around the bug.

controllers/sandbox/sandbox.go
evanphx: So the reason I did that is the non-pause containers would use this as a suffix and then ctr would balk at accessing the pause container because it would say "prefix matches multiple containers!" I'm happy to call the pause container something else, but it has to have a unique prefix compared to the sub containers. The `_pause` was because we use `-` + name for the subcontainers.

controllers/sandbox/sandbox.go
evanphx: Sure, that's a fine assumption. I guess what about the scenario where the container was deleted behind the scenes but is still marked DEAD?

pkg/controller/controller.go
evanphx: good call.

pkg/controller/controller.go
evanphx: Oh, yeah, that was a previous bit that I just eliminated by using a function.

pkg/entity/store.go
evanphx: Another good call!

controllers/sandbox/sandbox.go
phinze: Oh yeah the behavior change I'm all for - I've seen that prefix match complaint in action. I'm just nitpicking the naming. Like maybe call the function `pauseContainerId` or `parentContainerId`?

evanphx: ah! sure.

controllers/sandbox/sandbox.go
phinze: >  I guess what about the scenario where the container was deleted behind the scenes but is still marked DEAD?

I was thinking that hopefully that emits some sort of detectable NotFoundError that we could recognize to do the `continue` only in that case instead of on any error.

evanphx: Ahhh! Yes, I can do that.

PR #137, opened 2025-07-17
Improved shutdown logging

phinze:
- Add logging when stopping embedded components (containerd, etcd, clickhouse)
- Use timeout context (30s) for component Stop operations
- Log errors if component cleanup fails
- Add shutdown message to track cleanup progress

This helped diagnose the issue where etcd and clickhouse containers
are orphaned when the runtime exits in standalone mode. It turned out
to be an issue with systemd sending process signals to all members of
the control group, so the fix was to set KillMode on the unit file.

refs MIR-240

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown process for embedded components with explicit timeout handling and enhanced error logging.
  * Enhanced error reporting during shutdown to provide clearer feedback in case of issues.

### DIFF
--- a/cli/commands/dev.go
+++ b/cli/commands/dev.go
@@ -182,7 +182,14 @@ func Dev(ctx *Context, opts struct {
 		ctx.Log.Info("embedded containerd started", "socket", containerdComponent.SocketPath())
 
 		// Ensure cleanup on exit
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Excellent.

PR #140, opened 2025-07-22
MIR-252: Fix buildkit hang when deploying unsupported stacks

teresamychu:
## Summary
- Add early stack detection before launching buildkit sandbox to prevent hangs
- Fix context usage in defer rbk.Close() to respect cancellation
- System now fails fast with clear error message for unsupported stacks

## Problem
When deploying apps with unsupported stacks (like HTML-only apps), buildkit would hang indefinitely, leaving orphaned sandbox entities that continuously failed to reconcile.

## Solution
Check if the stack is supported BEFORE launching the buildkit sandbox. This prevents:
1. Unnecessary buildkit sandbox creation
2. Hanging buildkit processes
3. Orphaned sandbox entities in etcd

## Test plan
- [x] Deploy an HTML app (unsupported stack) and verify it fails quickly with error message
- [x] Verify no buildkit sandboxes are created for unsupported stacks
- [x] Deploy a supported app (Go/Node/Python) and verify it still works correctly
- [x] Check that no orphaned sandbox entities remain after failed deployments

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection and validation of the build stack when the stack type is set to "auto". Builds will now abort with an error if stack detection fails.

* **Improvements**
  * Enhanced logging during the build process to provide clearer status updates.

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -237,7 +237,11 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
 	}
 	b.Log.Info("buildkit launch completed successfully")
 
-	defer rbk.Close(ctx)
+	defer func() {
+		if err := rbk.Close(ctx); err != nil {
+			b.Log.Error("failed to close buildkit", "error", err)
+		}
+	}()
 
 	b.Log.Debug("attempting to get buildkit client")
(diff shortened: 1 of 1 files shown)


phinze (approved):
LGTM!

PR #141, opened 2025-07-22
Fix QUIC stream exhaustion and optimize large file transfers

phinze:
## Summary

  This PR fixes critical performance issues with large file uploads in the RPC system, specifically addressing QUIC stream exhaustion and slow transfer speeds when deploying
  applications with large assets.

  Fixes MIR-247

  ## Problem

  When deploying applications with large files (10MB+), the system would:
  1. Hit QUIC's 1000 bidirectional stream limit with error: "blocked at 1000 bidirectional streams"
  2. Transfer files extremely slowly (~110-185 KB/s instead of expected 5-10 MB/s)
  3. Keep RPC streams open unnecessarily during long build processes

  ## Root Causes

  1. **Stream Exhaustion**: Each inline RPC call created a new QUIC stream, quickly exhausting the 1000 stream limit
  2. **Small Chunk Size**: Using 4KB chunks meant a 10MB file required ~2,500 RPC round-trips
  3. **Stream Lifecycle**: Streams weren't closed promptly after data transfer completed

  ## Solution

  ### 1. QLOG Debugging Support (a3ac7be)
  - Added QLOG tracing to diagnose QUIC performance issues
  - Enabled by setting QLOGDIR environment variable
  - Helped identify the root causes below

  ### 2. Stream Pooling (f18e9e3)
  - Implemented stream pool (size 10) for inline clients
  - Streams are reused instead of being closed after each call
  - Includes proper concurrency control with activeCount tracking to prevent races
  - Testing shows: reduced from 280 unique streams to only 6 streams with massive reuse

  ### 3. Chunk Size Optimization (0c32d5b)
  - Increased chunk size from 4KB to 1MB with buffering
  - Reduces 10MB transfer from ~2,500 to ~10 RPC calls
  - Expected 100-250x throughput improvement

  ### 4. Context Cancellation Check (f5eabe9)
  - Added context check before decode operations
  - Handles client-side cancellations more responsively

  ## Testing

  - Manual testing with large file deployments
  - QLOG analysis confirmed stream reuse working correctly:
    - Before: 280 unique streams created → exhaustion
    - After: Only 6 unique streams, main stream reused 7,444 times
  - Observed transfer rates improved from 110-185 KB/s to expected speeds

  ## Review Notes

  The history has been rebased to simplify the implementation details. The final solution uses stream pooling rather than semaphore-based limiting, as it provides better performance
   and resource management.

## Summary by CodeRabbit

* **New Features**
  * Introduced connection tracing for QUIC connections to enhance observability.
  * Added chunked reading and writing for streaming, optimizing performance and memory usage.

* **Improvements**
  * Implemented a concurrency-limited stream pooling mechanism to reduce overhead and improve efficiency for stream-based operations.
  * Enhanced handling of multiple requests per stream for improved responsiveness and throughput.

* **Bug Fixes**
  * Improved resource cleanup and error handling when closing network clients and streams.

### DIFF
--- a/pkg/rpc/inline.go
+++ b/pkg/rpc/inline.go
@@ -150,6 +150,10 @@ func (c *inlineClient) Call(ctx context.Context, method string, args any, ret an
 	var rr refResponse
 
 	// Read response without timeout loop - let QUIC handle flow control
+	if err := ctx.Err(); err != nil {
+		shouldReturn = false
+		return err
+	}
 	err = conn.dec.Decode(&rr)
 	if err != nil {
 		shouldReturn = false
--- a/pkg/rpc/inline.go
+++ b/pkg/rpc/inline.go
@@ -4,61 +4,155 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"time"
+	"sync"
 
 	"github.com/fxamacker/cbor/v2"
 	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/webtransport"
 )
 
+const inlineStreamPoolSize = 10
+
+type streamConn struct {
+	stream webtransport.Stream
+	enc    *cbor.Encoder
+	dec    *cbor.Decoder
+}
+
 type inlineClient struct {
 	log     *slog.Logger
 	oid     OID
 	ctrl    *controlStream
 	session *webtransport.Session
+
+	// Stream pool
+	poolMu      sync.Mutex
+	pool        chan *streamConn
+	activeCount int // tracks total active streams (in use + in pool)
+	closed      bool
+}
+
+// initPool initializes the stream pool
+func (c *inlineClient) initPool() {
+	c.poolMu.Lock()
+	defer c.poolMu.Unlock()
+
+	if c.pool != nil {
+		return
+	}
+
+	c.pool = make(chan *streamConn, inlineStreamPoolSize)
+}
+
+// getStream gets a stream from the pool or creates a new one if pool is not full
+func (c *inlineClient) getStream(ctx context.Context) (*streamConn, error) {
+	// Ensure pool is initialized
+	c.initPool()
+
+	// Try to get an existing stream from the pool
+	select {
+	case conn := <-c.pool:
+		return conn, nil
+	default:
+		// Pool is empty, check if we can create a new stream
+		c.poolMu.Lock()
+		canCreate := c.activeCount < inlineStreamPoolSize
+		if canCreate {
+			c.activeCount++
+		}
+		c.poolMu.Unlock()
+
+		if canCreate {
+			// We can create a new stream
+			str, err := c.session.OpenStreamSync(ctx)
+			if err != nil {
+				// Decrement counter on error
+				c.poolMu.Lock()
+				c.activeCount--
+				c.poolMu.Unlock()
+				return nil, err
+			}
+
+			return &streamConn{
+				stream: str,
+				enc:    cbor.NewEncoder(str),
+				dec:    cbor.NewDecoder(str),
+			}, nil
+		}
+
+		// Pool is at capacity, wait for a stream to become available
+		select {
+		case conn := <-c.pool:
+			return conn, nil
+		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


phinze:
Here's how I've been testing this, in three copy-pasteable steps.

Get big image:

```bash
curl --output-dir ./testdata/bun -O  https://upload.wikimedia.org/wikipedia/commons/f/ff/Pizigani_1367_Chart_10MB.jpg
```

Modify bun to serve it:

```bash
cat << 'EOF' | patch -p1
diff --git a/testdata/bun/index.js b/testdata/bun/index.js
index 7a14d55..afeff82 100755
--- a/testdata/bun/index.js
+++ b/testdata/bun/index.js
@@ -1,7 +1,51 @@
 const server = Bun.serve({
   port: 3000,
-  fetch(request) {
-    return new Response("Welcome to Bun!");
+  async fetch(request) {
+    const url = new URL(request.url);
+
+    if (url.pathname === "/image") {
+      const file = Bun.file("./Pizigani_1367_Chart_10MB.jpg");
+      return new Response(file, {
+        headers: {
+          "Content-Type": "image/jpeg",
+        },
+      });
+    }
+
+    return new Response(
+      `
+      <!DOCTYPE html>
+      <html>
+      <head>
+        <title>Pizigani 1367 Chart</title>
+        <style>
+          body {
+            margin: 0;
+            padding: 20px;
+            font-family: Arial, sans-serif;
+            text-align: center;
+          }
+          img {
+            max-width: 100%;
+            height: auto;
+            border: 2px solid #333;
+            margin-top: 20px;
+          }
+        </style>
+      </head>
+      <body>
+        <h1>Pizigani 1367 Chart</h1>
+        <p>A medieval portolan chart created by the Pizigani brothers in 1367.</p>
+        <img src="/image" alt="Pizigani 1367 Chart">
+      </body>
+      </html>
+    `,
+      {
+        headers: {
+          "Content-Type": "text/html",
+        },
+      },
+    );
   },
 });
EOF
```

Deploy it (with verbose, explain output):

```bash
bin/runtime deploy -vv -x -d testdata/bun
```

pkg/rpc/inline.go
```diff
@@ -4,61 +4,72 @@ import (
...
+	// Acquire semaphore slot (blocks if at limit)
+	select {
+	case c.streamSem <- struct{}{}:
+		defer func() { <-c.streamSem }() // Release slot when done
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	// Create a new stream for this call
 	str, err := c.session.OpenStreamSync(ctx)
```
evanphx: If I read this right, we're still opening a new stream for each Call, we just now only allow for a max of 50 simultaneous calls. I'm unsure that's what the intention of this fix is, because the change on the stream receiver side is now looping, so that an existing stream can be reused, but this side isn't doing any reuse. Should we be storing a previously opened stream inside the streamSem to use?

evanphx (commented):
Some confusion about the semantics of stream reuse

pkg/rpc/inline.go
phinze: Ah yes good call out. That's a result of the fact that my first stab at this was with a pool, but that resulted in head-of-line blocking. The server side code was modified to support stream reuse but not reverted when we went back to the semaphore strategy. Now that we have bigger chunks I'll try going back to a stream pool to see if the perf is reasonable. 🔍

phinze: (Or at least Claude thought it was HOL blocking, that was before I was verifying stream perf with qvis... it may have been wrong!)

phinze: @evanphx Switched back to the pool approach - went with size 10 just to see.

Results are great:
- Before: 280 streams used
- After: Only 6 streams total!

The main stream got reused 7,444 times. Had to deploy both client and server to see the full benefit - the server needs to reuse inline clients too (this client/server confusion may have been the cause of the mistaken HOL theory earlier).

Also cleaned up that aggressive close in BuildFromTar since it wasn't doing anything useful anyway.

evanphx (approved):
Now we're talking! LFG!

PR #144, opened 2025-07-23
Add include directive support for deploy command

phinze:
## Summary

This PR adds support for "include" directives in AppConfig that allow files to be included in deployment tars even if they're listed in .gitignore. This is part of the work to support private Go modules.

Refs MIR-250

## Changes

- Added `Include` field to AppConfig struct to store include patterns
- Modified `deploy` command to read include patterns from AppConfig and pass them to MakeTar
- Updated `tarx.MakeTar` to accept include patterns that override gitignore rules
- Consolidated duplicate tar utilities by removing `pkg/buildkit/tar.go` (only used in tests)

## Usage

In your `.runtime/app.toml`:
```toml
include = ["dist/*", "*.generated", "vendor/*"]
```

Files matching these patterns will be included in the deployment tar even if they would normally be excluded by `.gitignore`.

## Test plan

- [x] Added comprehensive test `TestMakeTarWithIncludePatterns` to verify include functionality
- [x] Updated all existing tests to work with new MakeTar signature
- [x] Manually tested with example app config
- [ ] Test with actual private Go modules vendored dependencies

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying include patterns in application configuration to control which files are included during deployment, allowing selective overrides of .gitignore exclusions.

* **Bug Fixes**
  * Improved error handling when reading tar archives to ensure immediate reporting of non-EOF errors.

* **Tests**
  * Updated existing tests to use the new include patterns parameter.
  * Added new tests to verify correct behavior when include patterns are specified.

* **Refactor**
  * Updated codebase to use a unified tar handling implementation with enhanced include pattern support.
  * Removed legacy tar utility code in favor of the new implementation.


pkg/tarx/tarx.go
```diff
@@ -58,8 +58,37 @@ func MakeTar(dir string) (io.Reader, error) {
...
+					// Check if the path matches the include pattern
+					match, err := filepath.Match(pattern, rp)
+					if err == nil && match {
+						isIncluded = true
+						break
+					}
+					// Also check if pattern matches any parent directory
+					if info.IsDir() {
+						match, err = filepath.Match(pattern, rp+"/")
+						if err == nil && match {
```
teresamychu: ⚠️ probably want to surface the filepath.Match error when there is one?

pkg/tarx/tarx.go
```diff
@@ -58,8 +58,37 @@ func MakeTar(dir string) (io.Reader, error) {
...
+				for _, pattern := range includePatterns {
+					// Check if the path matches the include pattern
+					match, err := filepath.Match(pattern, rp)
+					if err == nil && match {
+						isIncluded = true
+						break
+					}
+					// Also check if pattern matches any parent directory
+					if info.IsDir() {
+						match, err = filepath.Match(pattern, rp+"/")
```
teresamychu: ⚠️ I dont think that filepath.Match uses the same patterns for filepath matching as .gitignore which might be confusing for users if they cant use the same patterns for .gitignore here?

like.... `**/*.generated` would match `src/foo/bar/file.generated` in .gitignore
but it wont work with `filepath.Match` 

im not sure if this matters, but just wanted to document it.

cli/commands/deploy.go
```diff
@@ -42,7 +43,17 @@ func Deploy(ctx *Context, opts struct {
...
 
-	r, err := tarx.MakeTar(dir)
+	// Load AppConfig to get include patterns
+	var includePatterns []string
+	ac, err := appconfig.LoadAppConfigUnder(dir)
+	if err != nil {
+		return err
+	}
+	if ac != nil && ac.Include != nil {
+		includePatterns = ac.Include
```
teresamychu, line 59: ❓ is there a way to validate the patterns before we load them?

pkg/tarx/tarx.go
```diff
@@ -58,8 +58,37 @@ func MakeTar(dir string) (io.Reader, error) {
...
+			if len(includePatterns) > 0 {
+				for _, pattern := range includePatterns {
+					// Check if the path matches the include pattern
+					match, err := filepath.Match(pattern, rp)
+					if err == nil && match {
+						isIncluded = true
+						break
+					}
+					// Also check if pattern matches any parent directory
+					if info.IsDir() {
```
teresamychu: ❓ Do we want to guard against the user doing pattern matching on files that are outside the project directory? 

right now i think if the user put an include pattern like `../../../etc/*` , it would actually work?

teresamychu (changes requested):
❓ Maybe not specifically for this PR, but do extracted files maintain the same permissions ? Or do we have to explicitly copy the file mode over to preserve the file permissions?

Also this looks great, i just had some minor comments!

pkg/tarx/tarx.go
phinze: Good point! In 1d42c2b3239c5f4aaac233c893465b87ce11c02a I switched to using the same gitignore pattern matching - no reason to have it be different.

pkg/tarx/tarx.go
phinze: Addressed in 1d42c2b3239c5f4aaac233c893465b87ce11c02a

cli/commands/deploy.go
phinze: Added validation in 1d42c2b3239c5f4aaac233c893465b87ce11c02a

pkg/tarx/tarx.go
phinze: Good to ask! But shouldn't be an issue here, since we're only dealing in modifications to the files being tarred up in the current directory. Patterns with `../` will just never match.

phinze:
> ❓ Maybe not specifically for this PR, but do extracted files maintain the same permissions ? Or do we have to explicitly copy the file mode over to preserve the file permissions? 

They are preserved! ✅ 

1. In `MakeTar`: We use `tar.FileInfoHeader(info, "")` which preserves the file permissions from the source file
2. In `TarFS`: We call `f.Chmod(os.FileMode(th.FileInfo().Mode()) & os.ModePerm)` which restores the file permissions when extracting

teresamychu (approved):
WOOT WOOT! Thanks for the quick turnaround! LGTM!

PR #145, opened 2025-07-23
Add vendor directory support to GoStack

phinze:
## Summary

- Add support for vendored Go dependencies in GoStack
- Skip `go mod download` when vendor directory is detected
- Use `go build -mod=vendor` for builds with vendored dependencies

## Benefits

- Enables support for private Go modules without authentication complexity
- Faster builds (no module downloads)
- Deterministic builds with pre-vendored dependencies

## Implementation

Added `hasVendor()` method to check for vendor directory and conditional logic in the build command. When a vendor directory exists, the build uses `-mod=vendor` flag instead of downloading modules.

Also fixed stackbuild tests to use the proper buildkit image reference from imagerefs package.

## Test plan

- [x] Added TestGoWithVendor test case
- [x] All tests pass
- [x] Lint checks pass

Refs MIR-250

## Summary by CodeRabbit

* **New Features**
  * Improved Go build process to automatically detect and use a vendor directory, skipping unnecessary module downloads when vendor is present.

* **Tests**
  * Added a new test to verify correct build behavior when a vendor directory exists.

### DIFF
--- a/pkg/stackbuild/stackbuild.go
+++ b/pkg/stackbuild/stackbuild.go
@@ -677,6 +677,10 @@ func (s *GoStack) commandDir(opts BuildOptions) string {
 	return ""
 }
 
+func (s *GoStack) hasVendor() bool {
+	return s.hasDir("vendor")
+}
+
 func (s *GoStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, error) {
 	// Set up local context with the directory
 	localCtx := llb.Local("context",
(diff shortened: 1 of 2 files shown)


pkg/stackbuild/stackbuild_test.go
```diff
@@ -462,3 +472,55 @@ func TestGo(t *testing.T) {
...
+	// Create a simple Go project without external dependencies for vendor test
+	files := map[string]string{
+		"go.mod": "module test-app\n\ngo 1.23\n",
+		"go.sum": "",
+		"main.go": `package main
+
+import "fmt"
+
+func main() {
+	fmt.Println("Hello, World!")
```
teresamychu, line 493: leftover from testing?

pkg/stackbuild/stackbuild.go
```diff
@@ -707,9 +711,17 @@ func (s *GoStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, error)
 
 	buildDir := s.commandDir(opts)
 
+	// Build command - skip go mod download if vendor directory exists
+	var buildCmd string
+	if s.hasVendor() {
+		buildCmd = fmt.Sprintf("go build -mod=vendor -o /bin/app ./%s", buildDir)
+	} else {
+		buildCmd = fmt.Sprintf("sh -c 'go mod download -json && go build -o /bin/app ./%s'", buildDir)
```
teresamychu, line 719: 💭 Not for this PR but I wish these commands werent just random strings and there was a better place for all of our shell commands that we could reference and validate.

teresamychu (approved):
nit comment about hello world code being left in the test but otherwise LGTM!

pkg/stackbuild/stackbuild.go
phinze: Agreed!

pkg/stackbuild/stackbuild_test.go
phinze: This is part of the test data - making a semi-realistic go project in a temp dir so that `go build` will work.

phinze:
Gonna merge as-is for now but happy to tweak testing strategy post-hoc!

PR #148, opened 2025-07-24
Fix sandbox recycling for slow request patterns

phinze:
When httpingress handles slow traffic (e.g., 1 req/5s), it regularly
expires and reacquires leases from the activator. Previously, when
reusing an existing sandbox for a new lease in AcquireLease, the
lastRenewal timestamp was not updated.

This caused the background retireUnusedSandboxes process to incorrectly
identify these actively-used sandboxes as stale (>2 minutes without
renewal) and retire them, leading to unnecessary sandbox recycling.

Update lastRenewal when reusing a sandbox to properly track its
activity and prevent premature retirement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox management by updating the renewal timestamp when a sandbox is reused, ensuring more accurate tracking of activity.


evanphx (approved):
Easy enough!

PR #153, opened 2025-07-25
Surface build failure logs to users during runtime deploy

teresamychu:
## Summary
- Fixes the issue where build failures only show generic "api version not found" error
- Adds proper error propagation from build server to CLI
- Users now see actual build errors (compilation errors, missing packages, etc.) **without requiring the -x flag**

## Problem
Previously, when `runtime deploy` failed during the build phase, users would only see a generic "api version not found" error. The actual build error messages were only available in server logs or when using the `-x` flag, making it difficult to debug deployment failures.

## Solution
1. Added an 'error' field to the build RPC status message type to propagate server-side errors
2. Modified the build server to send error messages through the status stream using a new `sendErrorStatus` helper
3. Updated the deploy command to:
   - Collect error messages from BuildKit vertex errors
   - Parse BuildKit logs for compilation errors when only generic "exit code" errors are present
   - Display all collected errors to users in both explain and non-explain modes

## Implementation Details
- **RPC Changes**: Added 'error' field to Status union type in `api/build/rpc.yml`
- **Server Changes**: Added `sendErrorStatus` helper in `servers/build/build.go` to consistently send error messages
- **Client Changes**: Enhanced `cli/commands/deploy.go` to extract errors from both vertex errors and logs
- **Optimization**: Only parse logs when necessary (when we have no errors or only generic "exit code" errors)

## Test Results
Tested with multiple failing Go apps:

**Simple import error:**
```
Error detected in building failing-app. No version returned.

Build errors:
  - main.go:5:2: package nonexistent/package is not in std (/usr/local/go/src/nonexistent/package)
  - process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
  - Error building image: failed to solve: process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
```

**Multiple compilation errors:**
```
Error detected in building type-errors-app. No version returned.

Build errors:
  - ./main.go:19:12: cannot use "should be int not string" (untyped string constant) as int value in assignment
  - ./main.go:20:14: cannot use 123 (untyped int constant) as string value in assignment
  - ./main.go:21:13: cannot use "twenty" (untyped string constant) as int value in assignment
  - ./main.go:24:34: too many arguments in call to strconv.Atoi
  - ./main.go:25:2: not enough arguments in call to http.HandleFunc
  - ./main.go:28:14: undefined: undefinedVariable
  - ./main.go:29:12: undefined: someUndefinedFunction
  - ./main.go:32:7: user.DoesNotExist undefined (type User has no field or method DoesNotExist)
  - ./main.go:35:13: assignment mismatch: 3 variables but returnsOnlyTwo returns 2 values
  - ./main.go:38:10: assignment mismatch: 2 variables but 1 value
  - ./main.go:38:10: too many errors
  - process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
  - Error building image: failed to solve: process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
```

## Additional Improvements
- Added service connectivity checks in dev.sh for better debugging in development environment

## Summary by CodeRabbit

* **New Features**
  * Enhanced error reporting during the build and deploy processes, aggregating error messages from multiple sources for consistent user feedback.
  * Introduced a new error status type in build status updates for clearer error communication.

* **Bug Fixes**
  * Improved development environment setup reliability by verifying connectivity to required services (etcd, clickhouse, minio) before proceeding.

* **Chores**
  * Simplified dependency management by removing the list of required modules from the Go module file.


evanphx (commented):
2 things:

1. I think we want to try to expose the output that -x has, which I believe is part of the shell/terminal output that is in the data we get back from buildkit on the channel.
2. Is the dagger stuff necessary? Curious what it's solving for.

teresamychu:
> 2 things:
> 
> 1. I think we want to try to expose the output that -x has, which I believe is part of the shell/terminal output that is in the data we get back from buildkit on the channel.
> 2. Is the dagger stuff necessary? Curious what it's solving for.

1.  OH you're right. That was silly of me, BLARGH this is why we should not be making PRs at 6pm on a Friday haha. I'll fix it. 
 
2. You are correct sir! The dagger stuff is not part of this PR. I just have a pre-commit hook that runs `go mod tidy` and it cleaned up the unused dependencies. Not related to this PR.

hack/dev.sh
```diff
@@ -26,6 +26,38 @@ sleep 1
...
+      echo "ERROR: Could not reach $service after 30 seconds"
+      echo "Checking DNS resolution:"
+      getent hosts $service || echo "DNS lookup failed for $service"
+      break
+    fi
+  done
+  if [ "$count" -lt 30 ]; then
+    echo " OK"
+  fi
+done
```
phinze: 💡 There's a [helper function for waiting for service readiness](https://github.com/mirendev/runtime/blob/1f7b48bbc4d09cb5cca7b7c02d891d2d0f89c6b9/hack/common-setup.sh#L100-L118) you can use instead of rolling your own loop. [Example usage in `test.sh`](https://github.com/mirendev/runtime/blob/cca0a8c3c0fce53ac69aedf201469b53e6850c76/hack/test.sh#L25-L26)

cli/commands/deploy.go
```diff
@@ -159,26 +170,72 @@ func Deploy(ctx *Context, opts struct {
...
+					}
+				}
+
+				// Also check logs for compilation errors if we haven't found specific errors yet
+				if len(buildErrors) == 0 || (len(buildErrors) > 0 && strings.Contains(buildErrors[0], "exit code:")) {
+					for _, log := range status.Logs {
+						if log.Data != nil {
+							logStr := string(log.Data)
+							// Look for Go compilation errors
+							if strings.Contains(logStr, ".go:") && strings.Contains(logStr, ":") {
```
phinze: ⚠️ I don't think we want any go-specific logic at this layer. The build could be any one of our supported stacks or even a custom Dockerfile.

phinze (changes requested):
Took it for a spin w/ a couple of test apps I pushed to [this branch](https://github.com/mirendev/runtime/tree/phinze/bad-deploy-testapps).

The Go one works nicely:

```
root@s8su6es68g462:/src/testdata/compile-error-app# r deploy -a compile-error-app
  ▰▰▰ Building image...
      Fetching 6 items: ███████████████ 100%
 • docker-image://docker.io/library/golang:1.24-alpine
 • docker-image://docker.io/library/busybox:1.37-musl
 • application code
 • [phase] Installing OS packages
 • [phase] Copying application code
 • ERROR [phase] Building Go application



Error detected in building compile-error-app. No version returned.

Build errors:
  - ./main.go:10:41: undefined: port
  - ./main.go:13:32: too many arguments in call to http.HandleFunc
  - ./main.go:23:18: cannot use "not a number" (untyped string constant) as int value in variable declaration
  - ./main.go:31:1: missing return
  - process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
  - Error building image: failed to solve: process "sh -c go mod download -json && go build -o /bin/app ./" did not complete successfully: exit code: 1
```

The node one exposes less:

```
root@tmoldn472m7s4:/src# r deploy -a npm-bad -d testdata/npm-install-error-app/
  ▰▰▰ Building image...
      Fetching 6 items: ███████████████ 100%


Error detected in building npm-bad. No version returned.

Build errors:
  - process "npm install" did not complete successfully: exit code: 1
  - Error building image: failed to solve: process "npm install" did not complete successfully: exit code: 1
```

But per my inline comment I'm not sure we should be putting log filters w/ knowledge of the stacks out at the CLI layer.

I wonder if we should instead dump all the logs from the build on failure? 🤔 Not sure open to thoughts!

dagger/go.mod
```diff
@@ -2,45 +2,6 @@ module dagger/runtime
...
-	github.com/sosodev/duration v1.3.1 // indirect
-	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.27.0 // indirect
-	golang.org/x/net v0.33.0 // indirect
-	golang.org/x/sys v0.28.0 // indirect
-	golang.org/x/text v0.21.0 // indirect
-	google.golang.org/genproto/googleapis/api v0.0.0-20240903143218-8af14fe29dc1 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 // indirect
-	google.golang.org/protobuf v1.35.2 // indirect
-)
-
```
phinze: ⚠️ I think per your earlier back and forth with Evan these dagger go.mod diffs are unrelated and can be removed.

teresamychu:
OK! I think i addressed all the feedback here. 

This PR now prints the build output in its entirety if the build fails and is language agnostic. Also updated dev.sh to use the wait for service helper function!

phinze (approved):
Still some extraneous dagger diffs in the mix but otherwise the code LGTM!

evanphx (approved):

PR #156, opened 2025-07-28
Add cross-architecture builds to PR checks

teresamychu:
## Summary
- Added `build-binaries` job to test.yml to validate cross-platform builds during PR checks
- Tests compilation across linux/darwin and amd64/arm64 architectures
- Helps catch platform-specific build issues before they reach main

## Test plan
- [x] Verify the workflow syntax is correct
- [ ] PR checks should now include build-binaries job for all 4 platforms
- [ ] Builds should complete successfully without uploading artifacts

## Summary by CodeRabbit

* **Chores**
  * Added automated workflow to build and package binaries for multiple operating systems and architectures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,7 @@ jobs:
       run: dagger --progress=plain call test --dir=. --tests="./e2e" --tags="e2e"
 
   build-binaries:
+    needs: lint
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


phinze (approved):
The build failure was transient pretty sure so I gave it a kick. That first 🐰 suggestion is worth the build cycle savings when lint fails IMHO. But overall looks good!

PR #159, opened 2025-07-29
Fix sandbox reconcile error due to CIDR/IP mismatch

phinze:
## Summary
- Fixed MIR-268: Sandbox reconciliation failing with "invalid address" error
- Changed address storage format from plain IP to CIDR notation during sandbox creation
- Updated tests to handle CIDR format correctly

## Problem
When a sandbox is created with automatic network allocation, the address was stored as a plain IP address (e.g., "10.8.0.5") but the reconciliation process expected a CIDR format (e.g., "10.8.0.5/32"). This caused the `netip.ParsePrefix()` call to fail during reconciliation.

## Solution
Modified `controllers/sandbox/sandbox.go` to store the full CIDR format by using `prefix.String()` instead of `prefix.Addr().String()` when saving the network address.

## Test plan
- [x] Code compiles successfully with `make bin/runtime`
- [x] Linting passes with `make lint-changed`
- [x] Manual testing: Create a sandbox and verify reconciliation works
- [x] Unit tests pass in CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and assignment of network addresses in sandbox network configuration to ensure accurate representation and backward compatibility.
  * Updated tests to correctly parse and compare network address values, enhancing test reliability.
  * Ensured sandbox network addresses are consistently validated and formatted as plain IP addresses across URLs and returned values.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log/slog"
 	"math/rand/v2"
+	"net/netip"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


phinze:
E2E failure is real! Need to fix httpingress to handle the sandbox's having CIDR notation too. On it. 🛠️

evanphx (approved):

PR #162, opened 2025-08-05
feat: implement per-service concurrency controls (MIR-227)

phinze:
## Summary

This PR implements RFD 0022 for improved concurrency controls in the Miren runtime, enabling per-service configuration with both auto-scaling and fixed instance modes.

### Key Changes

- **Schema Updates**: Added `services` configuration to `app_version` with per-service concurrency settings
- **Dual Modes**: Support for both `auto` (request-based scaling) and `fixed` (steady-state instances) modes  
- **Service-Aware Activator**: Refactored to maintain separate sandbox pools per service
- **Configuration Validation**: Added comprehensive validation with clear error messages
- **Fixed Instance Management**: Background process ensures target instance counts are maintained
- **Backward Compatibility**: Existing global concurrency settings continue to work

### Default Behaviors

- Web services: Auto mode with 10 requests/instance, 15m scale down delay
- Other services: Fixed mode with 1 instance

### Configuration Example

```toml
# app.toml
name = "my-app"

[services.web.concurrency]
mode = "auto"
requests_per_instance = 80
scale_down_delay = "2m"

[services.worker.concurrency]
mode = "fixed"
num_instances = 3
```

## Test Plan

- [x] Unit tests for configuration parsing and validation
- [x] Activator tests updated and passing
- [ ] Integration tests for new concurrency modes
- [x] Manual testing of auto-scaling behavior
- [x] Manual testing of fixed instance management

## Related

- Implements RFD 0022
- Fixes MIR-227

## Summary by CodeRabbit

* **New Features**
  * Introduced per-service concurrency configuration, allowing each service to define its own scaling mode, instance count, and request handling.
  * Added support for both "auto" and "fixed" concurrency modes on a per-service basis, improving flexibility and scalability.

* **Improvements**
  * Enhanced activator to manage concurrency and scaling independently for each service, including sandbox lifecycle and retention policies.
  * Updated configuration validation to ensure correct concurrency settings for each service.
  * Deprecated global concurrency settings in favor of new per-service configuration.
  * App config now supports detailed service commands and concurrency settings, which take precedence over Procfile commands.

* **Bug Fixes**
  * Improved handling and validation of invalid concurrency configuration values.

* **Tests**
  * Added comprehensive tests for service concurrency configuration, validation, and activator behavior in both fixed and auto modes.
  * Updated activator tests to incorporate service-specific concurrency and leasing logic.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -73,8 +73,9 @@ type verKey struct {
 }
 
 type localActivator struct {
-	mu       sync.Mutex
-	versions map[verKey]*verSandboxes
+	mu               sync.Mutex
+	versions         map[verKey]*verSandboxes
+	pendingCreations map[verKey]int // Track pending sandbox creations per service
 
 	log *slog.Logger
 	eac *entityserver_v1alpha.EntityAccessClient
@@ -84,9 +85,10 @@ var _ AppActivator = (*localActivator)(nil)
 
 func NewLocalActivator(ctx context.Context, log *slog.Logger, eac *entityserver_v1alpha.EntityAccessClient) AppActivator {
 	la := &localActivator{
-		log:      log.With("module", "activator"),
-		eac:      eac,
-		versions: make(map[verKey]*verSandboxes),
+		log:              log.With("module", "activator"),
+		eac:              eac,
+		versions:         make(map[verKey]*verSandboxes),
+		pendingCreations: make(map[verKey]int),
 	}
 
 	// Recover existing sandboxes on startup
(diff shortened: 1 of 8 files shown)


components/activator/activator.go
```diff
@@ -373,6 +440,45 @@
...
+			ScaleDownDelay:      "15m",
+		}
+	}
+
+	// Default for all other services is fixed with 1 instance
+	return &core_v1alpha.ServiceConcurrency{
+		Mode:         "fixed",
+		NumInstances: 1,
+	}
+}
```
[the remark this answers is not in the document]
phinze: You'd think so, but this is actually the effective meaning of the old value in the existing implementation! It may have been a mistake, but it's how it works right now so we should leave it.

components/activator/activator.go
```diff
@@ -350,13 +454,18 @@ func (a *localActivator) RenewLease(ctx context.Context, lease *Lease) (*Lease,
...
+	fixedTicker := time.NewTicker(30 * time.Second)
+	defer fixedTicker.Stop()
 
 	for {
 		select {
-		case <-ticker.C:
+		case <-retireTicker.C:
 			a.retireUnusedSandboxes()
+		case <-fixedTicker.C:
+			a.ensureFixedInstances(ctx)
```
evanphx, line 468: What are instances where the number of fixed instances would be incorrect and need fixing? Sandbox crashing? Can/should we respond to that more directly than detecting it every 30 seconds?

phinze:
Note to reviewers: This gets the basic functionality from RFD 22 down, but we're stopping short of tackling a few issues to avoid a snowball.

1. Both activator.go and build.go have several mixed concerns and are due for cleanup. MIR-314 covers the activator, and I think will end up touching build.go as a part of it.
2. Config parsing and interpretation need some consolidation. It's weird to have defaults assigned in callers and not by some unified representation. I think MIR-314 will touch on this too (or if not it can be extracted when that gets picked up).
3. We're _not_ solving for min instances maintenance on deploy here. Min instances will be maintained after the first request for a new app comes in. Instead of filing this as an issue now, I think we should see how the deployment work pans out.
4. We're also not trying to consolidate "deployment" into a controller like #124 did - leaving the current behavior as-is and leaving MIR-274 as follow up.

evanphx (approved):
looks great! curious about the "oops, missing sandbox" problem. Doesn't need to be solved here, but worth some consideration.

components/activator/activator.go
phinze: Yeah agreed this needs more thought. Right now _effectively_ this only really does its thing when the first request comes in for an app version and theoretically doesn't need to tick assuming the sandbox controller does its job.

But [per my comment above](https://github.com/mirendev/runtime/pull/162#issuecomment-3160964958) the responsibilities need to be reworked more holistically. Activator needs untangling and I think a few more controllers / concepts need to be introduced. MIR-314 and MIR-274 will help guide this.

For now I'm going to leave the tick, but I think it'll be relatively short lived.

PR #165, opened 2025-08-05
Fix activator sandbox recovery with CIDR notation in network addresses

phinze:
## Summary
- Fixes MIR-313: Sandbox network recovery fails after runtime restart
- Strips CIDR notation from network addresses when building URLs during sandbox recovery
- Prevents invalid URLs like `http://10.8.24.21/24:3000` that cause DNS lookup failures

## Problem
After runtime restart, the activator's sandbox recovery was building invalid URLs by including CIDR notation from stored network addresses. This caused proxy errors:
```
http: proxy error: dial tcp: lookup 10.8.24.21/24: no such host
```

## Solution
Updated the `recoverSandboxes` function to parse network addresses and extract just the IP, stripping any CIDR notation. Uses the same `netip.ParsePrefix` approach as the existing `activateApp` function for consistency.

## Test plan
- [x] Added new test `TestActivatorRecoverSandboxesWithCIDR` that verifies CIDR notation is correctly stripped
- [x] All existing activator tests pass
- [x] Manually tested with script demonstrating CIDR stripping logic

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of network addresses with CIDR notation during sandbox recovery, ensuring correct parsing and URL construction for sandbox endpoints.

* **Tests**
  * Added a new test to verify proper recovery of sandboxes with CIDR-formatted network addresses.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -425,7 +425,18 @@ func (a *localActivator) recoverSandboxes(ctx context.Context) error {
 			continue
 		}
 
-		addr := fmt.Sprintf("http://%s:%d", sb.Network[0].Address, port)
+		// Parse the address to extract just the IP from potential CIDR notation
+		ipAddr := sb.Network[0].Address
+		if prefix, err := netip.ParsePrefix(ipAddr); err == nil {
+			// New format: extract IP from CIDR
+			ipAddr = prefix.Addr().String()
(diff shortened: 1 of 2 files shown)


phinze:
Noting that activator could use a larger refactor to clean up its mixed concerns, filed MIR-314 to track that.

evanphx (approved):

PR #167, opened 2025-08-06
Add sandbox stop and delete CLI commands

phinze:
## Summary
- Adds `sandbox stop` command to update sandbox status to stopped
- Adds `sandbox delete` command to remove dead sandboxes with confirmation prompt
- Introduces domain-specific client for compute API following the pattern from ingress

## Details
This PR adds CLI commands for managing sandbox lifecycle:

### `runtime sandbox stop <sandbox-id>`
Updates the sandbox status to "stopped" and lets the sandbox controller handle cleanup.

### `runtime sandbox delete <sandbox-id>`
Deletes a sandbox entity, but only if it's in "dead" status. Includes a confirmation prompt unless `-f/--force` is used.

### Implementation
- Created `api/compute/client.go` following the pattern from `api/ingress/client.go`
- Encapsulates all entity manipulation logic in the compute client
- Simplified CLI commands by using the domain-specific client

## Test plan
- [ ] Test `sandbox stop` on a running sandbox
- [ ] Test `sandbox stop` on an already stopped sandbox
- [ ] Test `sandbox delete` on a dead sandbox with confirmation
- [ ] Test `sandbox delete` with `-f` flag
- [ ] Test `sandbox delete` on non-dead sandbox (should fail)

## Summary by CodeRabbit

* **New Features**
  * Added a command to stop a sandbox from the CLI.
  * Added a command to delete a sandbox from the CLI, with an option to force deletion without confirmation.

* **Improvements**
  * Enhanced sandbox management capabilities via new CLI commands for stopping and deleting sandboxes.

### DIFF
--- a/api/compute/client.go
+++ b/api/compute/client.go
@@ -1,0 +1,83 @@
+package compute
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"miren.dev/runtime/api/compute/compute_v1alpha"
+	"miren.dev/runtime/api/entityserver"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/rpc"
+)
+
+// Client provides a domain-specific client for Sandbox entities
+type Client struct {
+	log *slog.Logger
+	ec  *entityserver.Client
+}
+
+// NewClient creates a new Compute client from an RPC client
+func NewClient(log *slog.Logger, client rpc.Client) *Client {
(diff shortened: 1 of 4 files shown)


phinze:
I've had sandboxes get into a weird state and it'd be helpful to be able to stop them for debug purposes. I think eventually we may tuck all this CLI under a `debug` prefix but for now just following the pattern.

evanphx (approved):

PR #170, opened 2025-08-08
Fix: Fail build on compilation errors in deploy command

teresamychu:
## Summary
- Fixed the deploy command to properly exit with an error when build errors are detected
- Added check for buildErrors collection to fail the build when compilation errors occur
- Previously, the command would continue even if compilation failed

## Changes
1. Check if `buildErrors` array has any errors after collecting them from vertex.Error
2. Return an error with count of build errors when detected
3. Also return error (instead of nil) when no version is returned

## Test plan
- [ ] Deploy an app with compilation errors
- [ ] Verify the deploy command exits with non-zero status
- [ ] Verify error message shows "build failed with N error(s)"
- [ ] Verify build error details are displayed to the user

## Summary by CodeRabbit

* **Bug Fixes**
  * Build failures are now reported immediately instead of continuing silently after build errors are detected.
  * Deployments that produce no version now return an explicit error so failed builds are clearly surfaced.

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -188,6 +188,11 @@ func Deploy(ctx *Context, opts struct {
 					}
 				}
 
+				// Fail the build if we detected any errors
+				if len(buildErrors) > 0 {
+					return fmt.Errorf("build failed with %d error(s)", len(buildErrors))
+				}
+
 				return nil
 			case "message":
 				msg := update.Message()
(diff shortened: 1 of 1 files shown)


phinze (approved):
Yay no more sneaky deploy failures!

phinze:
Gonna pull this in as I'm testing a bunch of stuff in the area and it'll be nice to have it in place.

PR #176, opened 2025-08-15
MIR-338: Schedule ready nodes

evanphx:
This change does two things:

1. It calls Update on the node entry on runner boot so that we're sure the session attribute is updated to Ready (this won't be an issue when we fix the Put() ambiguity issue)
2. Rather than blindly picking the first node we see in a `range` over a map, we pick a random node that has it's status set to ready.

## Summary by CodeRabbit

* New Features
  * Randomly distributes sandbox assignments across READY nodes for better load balancing.

* Bug Fixes
  * Uses only READY nodes for scheduling to avoid assigning work to unavailable nodes.
  * Improves node registration reliability and status handling to reduce transient failures.
  * Clarifies error messages when no nodes are available for mass scheduling.

* Refactor
  * Streamlines node creation/update and sandbox update flows for more consistent state management.

### DIFF
--- a/components/runner/runner.go
+++ b/components/runner/runner.go
@@ -177,12 +177,18 @@ func (r *Runner) setupEntity(ctx context.Context, ec *entityserver.Client) error
 	r.se = sess
 
 	node := compute_v1alpha.Node{
-		Status:      compute_v1alpha.READY,
 		Constraints: types.LabelSet("compute", "generic"),
 		ApiAddress:  r.ListenAddress,
 	}
 
-	res, err := ec.Create(ctx, r.Id, &node)
+	res, err := ec.CreateOrUpdate(ctx, r.Id, &node)
+	if err != nil {
+		return err
+	}
+
+	err = ec.UpdateAttrs(ctx, res, (&compute_v1alpha.Node{
+		Status: compute_v1alpha.READY,
+	}).Encode)
 	if err != nil {
 		return err
 	}
--- a/components/scheduler/scheduler.go
+++ b/components/scheduler/scheduler.go
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"context"
 	"log/slog"
+	"math/rand"
 	"sync"
 
 	"miren.dev/runtime/api/compute/compute_v1alpha"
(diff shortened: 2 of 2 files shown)


components/runner/runner.go
```diff
@@ -177,12 +177,18 @@ func (r *Runner) setupEntity(ctx context.Context, ec *entityserver.Client) error
...
 
-	res, err := ec.Create(ctx, r.Id, &node)
+	res, err := ec.CreateOrUpdate(ctx, r.Id, &node)
+	if err != nil {
+		return err
+	}
+
+	err = ec.UpdateAttrs(ctx, res, (&compute_v1alpha.Node{
+		Status: compute_v1alpha.READY,
+	}).Encode)
```
phinze, line 191: ❓ Is the idea here that the `session: true` property of the `node.status` attribute was not being honored because it was coming in as a part of a larger bag of attributes?

Just trying to figure out if there's an underlying bug we should file.

components/scheduler/scheduler.go
```diff
@@ -167,36 +170,37 @@ func (s *Scheduler) assignSandbox(ctx context.Context, ent *entity.Entity, eac *
 	s.assigning.Lock()
```
phinze, line 170: ❓ I was also squinting at the "Skip if already scheduled" guard above this, which was preventing sandboxes from ever being rescheduled even if their node was gone.

phinze (approved):
Dropped some open questions inline, but this seems AOK as an incremental improvement.

phinze:
Per our discussion in [today's call](https://docs.google.com/document/d/1FDm1-N__BuctG9uPjzx6Nd9ZOPsRLHhXzQe1S84wMUA/edit#bookmark=id.9lq8o4yfobde):

1. **Session status bug**: The explicit UpdateAttrs call to set status to READY is a good workaround for the bug where the create flow fails to store session info when attributes are unchanged. We agreed not to fix the underlying bug separately since the upcoming API rework will address this comprehensively with either client-side query logic or a smarter upsert.

2. **Rescheduling**: Good call on preventing rescheduling by checking for the schedule key. As we discussed [here](https://docs.google.com/document/d/1FDm1-N__BuctG9uPjzx6Nd9ZOPsRLHhXzQe1S84wMUA/edit#bookmark=id.vuvzejm1ymxu), rescheduling needs to be tackled holistically with:
   - Atomic assignment and follow-up verification by the scheduler
   - Proper encapsulation for runner nodes (treating sandboxes as immutable/stateless like K8s pods)
   - Clear separation between runner nodes ("keep running until told to stop") and the central scheduler

This incremental approach is exactly what we need for now. The comprehensive rescheduling solution can come later.

PR #177, opened 2025-08-15
Work around bug in yaml parser

evanphx:
Seems that the yaml parser (that is invoked when calling Encode) really dislikes having a leading newline character! We've openned a bug about it: https://github.com/yaml/go-yaml/issues/65

This is a fix work around for now, but we're considering moving the config file body to something opaque rather than a string.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trim extraneous leading/trailing whitespace from generated build configuration, preventing parsing issues and intermittent build failures.
  * No impact on public APIs; only whitespace handling behavior is adjusted.

### DIFF
--- a/servers/build/launch.go
+++ b/servers/build/launch.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"maps"
+	"strings"
 	"time"
 
 	"github.com/containerd/containerd/v2/client"
@@ -141,7 +142,7 @@ func (l *LaunchBuildkit) Launch(ctx context.Context, addr string, lo ...LaunchOp
 			{
 				Path: "/etc/buildkit/buildkitd.toml",
 				Mode: "0644",
-				Data: config,
+				Data: strings.TrimSpace(config),
 			},
 		},
 	})


phinze (approved):
I thought for a second if we should try and do this more aggressively down in the entityserver, but that seems like it's likely to cause unintended side effects and hopefully the upstream bug gets fixed and we don't need to worry about it soon anyways. 👍

phinze:
As discussed in [our call](https://docs.google.com/document/d/1FDm1-N__BuctG9uPjzx6Nd9ZOPsRLHhXzQe1S84wMUA/edit#bookmark=id.jghgvfj8z6n5) - we agreed this bug theoretically applies to any value in the parser, but keeping the fix narrow for now is the right approach since we haven't seen issues elsewhere.

If the upstream bug (https://github.com/yaml/go-yaml/issues/65) doesn't get fixed, we can consider more aggressive approaches like:
- Trimming every string in the entity server
- Converting config files to base64 blobs (though as you discovered, bytes type → YAML gives us that hilarious literal array of numbers 😄)

Let's monitor how this targeted fix performs and expand if needed.

phinze:
Working on deployment speed stuff and seeing a buildkit hang. Sucking in this in case it solves everything.

PR #184, opened 2025-08-19
Fix Docker-based macOS installation process

teresamychu:
## Summary
This PR fixes the Docker-based macOS installation process by addressing auth generation and config copying issues.

## Changes
- **install.sh**: Updated for reliable auth generation and config copying
- **docker-compose-production.yml**: Added production-ready Docker container configuration  
- **release.yml**: Updated workflow to automatically upload docker-compose.yml and install.sh to assets service

## Test Plan
- [x] Clean environment testing from scratch
- [x] Auth generation and config copying verification
- [x] End-to-end installation flow validation
- [x] Runtime client-server connectivity testing

## Installation
Once merged, users can install via:
```bash
curl -sSL https://api.miren.cloud/assets/install.sh | bash
```

## Result
The complete Docker-based macOS installation now works reliably with proper automation and error handling.

## Summary by CodeRabbit

- New Features
  - Official multi-arch Docker image and docker-compose setup for Miren with etcd and ClickHouse.
  - Containerized startup with automatic client config generation and readiness checks.
  - ClickHouse endpoint now configurable via environment variables.

- Documentation
  - Added Docker deployment guide with quick start, manual setup, configuration, and troubleshooting.
  - Updated README to reflect containerd runtime; added Hot Reload and CLI Tool features.

- Chores
  - Release workflow now builds/publishes Docker images and uploads docker-compose assets with checksums.


docker-compose-production.yml
```diff
@@ -0,0 +1,68 @@
...
+      - "8123:8123"
+    environment:
+      - CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1
+      - CLICKHOUSE_PASSWORD=default
+    volumes:
+      - clickhouse-data:/var/lib/clickhouse
+
+  minio:
+    image: oci.miren.cloud/minio:v1
+    container_name: miren-minio
```
evanphx: So I'm realizing that we're not even using this yet because it was just there for the LSVD stuff, which we don't have wired up yet. So we can remove it for now.

docker/Dockerfile.miren
```diff
@@ -0,0 +1,187 @@
...
+        NERDCTL_URL="https://github.com/containerd/nerdctl/releases/download/v2.0.5/nerdctl-2.0.5-linux-arm64.tar.gz"; \
+    else \
+        NERDCTL_URL="https://github.com/containerd/nerdctl/releases/download/v2.0.5/nerdctl-2.0.5-linux-amd64.tar.gz"; \
+    fi && \
+    curl -L "$NERDCTL_URL" | tar -C /usr/local/bin -xz
+
+# Copy configuration files
+COPY hack/containerd-config.toml /etc/containerd/config.toml
+COPY hack/runsc-ignore /usr/local/bin/runsc-ignore
+RUN chmod +x /usr/local/bin/runsc-ignore
```
evanphx: We've removed gvisor for now, so you can remove these.

docker/Dockerfile.miren
```diff
@@ -0,0 +1,187 @@
...
+    else \
+        RUNSC_URL="https://storage.googleapis.com/gvisor/releases/release/latest/x86_64/runsc"; \
+        RUNSC_SHIM_URL="https://storage.googleapis.com/gvisor/releases/release/latest/x86_64/containerd-shim-runsc-v1"; \
+    fi && \
+    curl -L "$RUNSC_URL" -o /usr/local/bin/runsc && \
+    curl -L "$RUNSC_SHIM_URL" -o /usr/local/bin/containerd-shim-runsc-v1 && \
+    chmod +x /usr/local/bin/runsc /usr/local/bin/containerd-shim-runsc-v1
+
+# Install runsc
+RUN /usr/local/bin/runsc install
```
evanphx: We've removed this for now, so you can pull this out.

docker/Dockerfile.miren
```diff
@@ -0,0 +1,187 @@
...
+    fi && \
+    curl -L "$CONTAINERD_URL" | tar -C /usr/local -xz
+
+# Download and install buildkit
+RUN if [ "$TARGETARCH" = "arm64" ]; then \
+        BUILDKIT_URL="https://github.com/moby/buildkit/releases/download/v0.17.1/buildkit-v0.17.1.linux-arm64.tar.gz"; \
+    else \
+        BUILDKIT_URL="https://github.com/moby/buildkit/releases/download/v0.17.1/buildkit-v0.17.1.linux-amd64.tar.gz"; \
+    fi && \
+    curl -L "$BUILDKIT_URL" | tar -C /usr/local -xz
```
evanphx: When running, we download buildkit as a OCI image, we only used buildkit, the binary, in development for testing. So you can remove this.

docker/Dockerfile.miren
```diff
@@ -0,0 +1,187 @@
...
+RUN chmod +x /usr/local/bin/runsc-ignore
+
+# Copy miren binary from builder
+COPY --from=builder /build/bin/miren /usr/local/bin/miren
+
+# Create necessary directories
+RUN mkdir -p /data/containerd /data/containerd/state /data/buildkit /var/log /root/.config/miren /var/lib/miren/containerd
+
+# Create entrypoint script
+RUN echo '#!/bin/bash' > /entrypoint.sh && \
```
evanphx: Can we just have entrypoint.sh in the docker/ dir and copy it over? This looks like creating the file one line at a time, is there any substitutions happening?

docker/docker-compose.yml
```diff
@@ -0,0 +1,74 @@
...
+      - "8123:8123"
+    environment:
+      - CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1
+      - CLICKHOUSE_PASSWORD=default
+    volumes:
+      - clickhouse-data:/var/lib/clickhouse
+
+  minio:
+    image: oci.miren.cloud/minio:v1
+    container_name: miren-minio
```
evanphx: Like the other one, no need for minio right now.

install.sh
```diff
@@ -0,0 +1,329 @@
+#!/bin/bash
```
evanphx: This whole file should over in @mirendev/cloud, updating the one that is served out by the installer url.

install.sh
```diff
@@ -0,0 +1,329 @@
...
+    if command -v docker &>/dev/null && docker compose version &>/dev/null; then
+        docker compose -f /tmp/docker-compose.yml up -d
+    else
+        docker-compose -f /tmp/docker-compose.yml up -d
+    fi || {
+        print_error "Failed to start miren server containers"
+        exit 1
+    }
+    
+    rm /tmp/docker-compose.yml
```
evanphx: We must put the docker compose file somewhere that the user can get to it, not just delete it afterward or they can't debug or stop it.

.github/workflows/release.yml
```diff
@@ -131,3 +169,27 @@ jobs:
...
+            -H "Authorization: Bearer ${{ steps.get-oidc-token.outputs.token }}" \
+            -F "file=@docker-compose.yml.sha256" \
+            https://api.miren.cloud/assets/docker-compose.yml.sha256
+
+          # Upload install script
+          echo "Uploading install.sh"
+          curl --fail-with-body --retry 3 -X PUT \
+            -H "Authorization: Bearer ${{ steps.get-oidc-token.outputs.token }}" \
+            -F "file=@install.sh" \
+            https://api.miren.cloud/assets/install.sh
```
evanphx: The install script lives here: https://github.com/mirendev/cloud/blob/main/services/installer/install.sh, not in the assets store.

docker-compose-production.yml
```diff
@@ -0,0 +1,68 @@
+version: '3.8'
```
evanphx: What's the difference between this docker compose file and the one in docker/?

evanphx (changes requested):
Some stuff needs to get moved around and a bit of cleanup.

docker-compose-production.yml
teresamychu: OOO this is leftover. Was going to have one for local building for local testing and the other is the one that uses the pre-built image (which would be the production one). 

But since i decided not to commit a "local-testing" version, this shouldnt be here anymore. I'll fix it!

.runtime/app.toml
```diff
@@ -0,0 +1,4 @@
+name = 'workspace'
+post_import = ''
+env = []
+include = []
```
evanphx: This is probably extra.

.github/workflows/release.yml
```diff
@@ -131,3 +169,21 @@ jobs:
...
+          # Upload docker-compose configuration
+          echo "Uploading docker-compose.yml"
+          # Generate SHA256 checksum for docker-compose file
+          sha256sum docker/docker-compose.yml | cut -d' ' -f1 | tee docker-compose.yml.sha256
+          
+          # Upload docker-compose file
+          curl --fail-with-body --retry 3 -X PUT \
+            -H "Authorization: Bearer ${{ steps.get-oidc-token.outputs.token }}" \
+            -F "file=@docker/docker-compose.yml" \
+            https://api.miren.cloud/assets/docker-compose.yml
```
evanphx:
```suggestion
            https://api.miren.cloud/assets/release/miren/${{ github.ref_name }}/docker-compose.yml
```

docker/entrypoint.sh
```diff
@@ -0,0 +1,83 @@
...
+# Wait for containerd to be ready
+sleep 2
+
+# Start buildkitd  
+echo "Starting buildkitd..."
+buildkitd --root /data/buildkit &
+BUILDKITD_PID=$!
+
+# Wait for buildkitd to be ready
+sleep 2
```
evanphx: We don't need buildkit running here.

docker/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+    image: oci.miren.cloud/miren:latest
+    container_name: miren
+    privileged: true
+    ports:
+      - "8080:8080"
+      - "8443:8443/udp"
+    volumes:
+      - miren-data:/data
+      - ${PWD}:/workspace
+    working_dir: /workspace
```
evanphx: Drop this, so the host file system isn't directly leaked into the container.

docker/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+services:
+  miren:
+    image: oci.miren.cloud/miren:latest
+    container_name: miren
+    privileged: true
+    ports:
+      - "8080:8080"
+      - "8443:8443/udp"
+    volumes:
+      - miren-data:/data
```
evanphx: If this is going to be /data, then you need to update the call to `miren server` to tell it to store in /data.

Otherwise you can use `miren-data:/var/lib/miren` which will make Docker create it's volume at the path that server uses by default.

docker/docker-compose-test.yml
```diff
@@ -0,0 +1,48 @@
+version: '3.8'
```
evanphx: Looks like leftovers?

.runtime/app.toml
evanphx: Also left overs.

docker/Dockerfile.miren
evanphx: Still this.

evanphx (changes requested):
Still some leftover bits

evanphx (approved):

PR #185, opened 2025-08-19
fix: add retry logic and prevent queue blockage in controller watch (MIR-346)

phinze:
## Summary
- Adds automatic reconnection logic when entity watch connections are lost
- Prevents controller queue blockage during resync operations
- Prevents indefinite blocking when waiting for ports to be bound
- Ensures controllers remain operational even after entity server restarts

## Problem (MIR-346)
The sandbox controller's entity watch mechanism would permanently disconnect after losing connection to the entity server, causing the controller to stop processing new events. While periodic tasks continued running, the main reconciliation loop became permanently disconnected, requiring a service restart to restore functionality.

Additionally, the controller could block indefinitely waiting for ports to be bound during sandbox startup, causing the entire reconciliation to stall.

### Impact
- New sandboxes were not created/updated/deleted
- Existing sandboxes were not reconciled with desired state
- Only periodic cleanup continued to function
- Controller could become completely blocked waiting for port binding

### Commit 1: Add retry logic to controller watch mechanism
- Implemented exponential backoff retry logic in `WatchIndex` to automatically reconnect
- Fixed entityserver to properly check and propagate watch errors
- Added comprehensive test to verify watch reconnection behavior
- Controllers now automatically reconnect when the entity server restarts or network issues occur

### Commit 2: Prevent controller queue blockage during resync
- Made queue sends non-blocking with default case to prevent deadlocks
- Added warnings when events are dropped due to full queue
- Ensured periodicResync continues running even under queue pressure
- Added test to verify resync behavior (currently skipped due to timing sensitivity)

### Commit 3: Add timeout to waitForPort to prevent indefinite blocking
- Added context and timeout parameters to waitForPort function (30 second default)
- Controller continues with sandbox startup even if port binding times out
- Prevents entire controller from stalling when individual sandboxes have port binding issues
- Added comprehensive test coverage for timeout behavior

### Commit 4: Additional activator logging
- The sandbox server got into a state where it was leaking rfd app sandboxes (running sandboxes grew to ~40 in 24h), but it disappeared after I deployed this branch to it.
- This additional logging was meant to diagnose the issue; leaving it in so we have it if the issue recurs.

## Testing
- Added unit tests for watch reconnection logic
- Added test for queue overflow handling during resync
- Added tests for port wait timeout and context cancellation
- Manually tested controller behavior during entity server restarts

## Files Changed
- `pkg/controller/controller.go` - Core retry logic and queue handling
- `pkg/controller/controller_test.go` - Comprehensive test coverage
- `servers/entityserver/entityserver.go` - Error propagation fixes
- `pkg/entity/mock.go` - Test infrastructure improvements
- `controllers/sandbox/sandbox.go` - Port wait timeout implementation
- `controllers/sandbox/sandbox_test.go` - Port timeout test coverage

Fixes MIR-346

## Summary by CodeRabbit

* **New Features**
  * Sandbox startup no longer blocks on port binding; per-port timeouts and cancellation allow startup to continue.
  * Mock store gains a hook to customize index listing for tests.

* **Bug Fixes**
  * Index watch is now resilient: automatic reconnect with exponential backoff, non-blocking event queuing (drops logged when full), explicit cancellation/error handling, and improved logging.
  * Resyncs now warn and drop events when the resync queue is full.

* **Tests**
  * Added tests for watch reconnection, queue overflow, and port-wait success/timeout/cancellation.

### DIFF
--- a/controllers/sandbox/sandbox_test.go
+++ b/controllers/sandbox/sandbox_test.go
@@ -13,6 +13,9 @@ import (
 	"testing"
 	"time"
 
+	"log/slog"
+	"sync"
+
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/pkg/cio"
 	"github.com/containerd/containerd/v2/pkg/namespaces"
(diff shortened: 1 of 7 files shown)


phinze:
@coderabbitai full review

controllers/sandbox/sandbox_test.go
```diff
@@ -1329,4 +1332,65 @@ func TestSandbox(t *testing.T) {
...
+		}()
+
+		start = time.Now()
+		err = c.waitForPort(ctx, "test-id", 6666, 5*time.Second)
+		elapsed = time.Since(start)
+
+		r.Error(err, "should error on context cancellation")
+		r.Contains(err.Error(), "context cancelled")
+		r.True(elapsed >= 50*time.Millisecond && elapsed < 150*time.Millisecond,
+			"should detect context cancellation quickly, got %v", elapsed)
```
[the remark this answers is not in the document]
phinze: Will worry about this down the road.

pkg/controller/controller.go
```diff
@@ -288,14 +343,17 @@ func (c *ReconcileController) periodicResync(ctx context.Context) {
 				return
 			case c.workQueue <- ev:
 				// Event added to queue
+			default:
+				// Queue is full, log and skip this event
+				c.Log.Warn("Work queue full during resync, dropping event", "entity", ev.Id, "queueSize", len(c.workQueue))
```
evanphx, line 348: I worry a little about dropping stuff but on the other hand, a healthy controller is the more important thing. Let's double check the size of workQueue's buffer to be sure we can absorb a situation where, say, 20 sandboxes are all scheduled at the same time.

evanphx (approved):
Approve with the minor concern about the size of the workQueue influencing dropped events so quickly.

pkg/controller/controller.go
phinze: workQueue is 1000 deep right now. Claude wanted to up it to 10000, but I felt like at this stage masking issues is worse than dropping events.

phinze: Whaddya think should we hop to 10k or keep it as is for now? I welcome your instincts

evanphx: 1k seems totally fine for now. I just wanted to be sure it was, like, 10.

PR #186, opened 2025-08-21
MIR-245: Wire in the ability to connect to miren.cloud and run rbac policy

evanphx:
This is the first phase of adding miren.cloud auth support. Here we've wired in the ability to register a cluster with miren.cloud and that server can then validate JWTs generated by miren.cloud and execute rbac policy that is created there.

## Summary by CodeRabbit

* **New Features**
  * Cluster registration (register / register status) with on-disk persistence and polling.
  * RBAC debug tools to fetch policies and run local access tests.
  * Cloud-based RPC authentication (JWT validation, service auth flow, token caching) with policy fetching and evaluator integration; server auto-enables cloud auth when registered.
  * ED25519 keypair utilities and signing helpers.

* **RBAC**
  * New policy model and evaluator with tag/group-based rule evaluation and caching.

* **Tests**
  * Extensive tests for JWT, policy fetching, RBAC evaluator, and RPC auth.

* **Chores**
  * Add lint-pr Makefile target and linter config; add JWT dependency.


go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/fxamacker/cbor/v2 v2.7.0
 	github.com/go-logr/logr v1.4.2
 	github.com/go-ozzo/ozzo-validation/v4 v4.3.0
+	github.com/golang-jwt/jwt/v5 v5.3.0
```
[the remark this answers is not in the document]
evanphx: I'd rather use our simple version.

pkg/auth/jwt_test.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	// Create and sign token with Ed25519
+	token := jwt.NewWithClaims(&jwt.SigningMethodEd25519{}, claims)
+	token.Header["kid"] = "test-key-1"
+	
+	tokenString, err := token.SignedString(privateKey)
+	if err != nil {
+		t.Fatalf("failed to sign token: %v", err)
+	}
```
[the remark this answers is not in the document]
evanphx: You're incorrect.

cli/commands/register.go
```diff
@@ -0,0 +1,244 @@
...
+func Register(ctx *Context, opts struct {
+	ClusterName string            `short:"n" long:"name" description:"Cluster name" required:"true"`
+	CloudURL    string            `short:"u" long:"url" description:"Cloud URL" default:"https://api.miren.cloud"`
+	Tags        map[string]string `short:"t" long:"tag" description:"Tags for the cluster (key:value)"`
+	OutputDir   string            `short:"o" long:"output" description:"Output directory for registration"`
+}) error {
+	if opts.OutputDir == "" {
+		// Default to /var/lib/miren/server
+		opts.OutputDir = "/var/lib/miren/server"
+	}
```
phinze: 💡 Can just default this in the tags like `CloudURL`

cli/commands/register.go
```diff
@@ -0,0 +1,244 @@
...
+	opts.Tags = clean
+
+	// Check if already registered
+	existing, err := registration.LoadRegistration(opts.OutputDir)
+	if err != nil {
+		return fmt.Errorf("failed to check existing registration: %w", err)
+	}
+	if existing != nil {
+		// Check if it's a pending registration that hasn't expired
+		if existing.Status == "pending" && existing.ExpiresAt.After(time.Now()) {
```
phinze: 💡 Probably want a little buffer here so we don't try and resume a registration that's destined to expire in 5s.

cli/commands/register.go
```diff
@@ -0,0 +1,244 @@
...
+				fmt.Print(".")
+			})
+			if err != nil {
+				ctx.Warn(" Failed!")
+				return fmt.Errorf("registration failed: %w", err)
+			}
+			ctx.Info(" Approved!")
+
+			// Update and save the approved registration
+			existing.Status = "approved"
```
phinze, line 73: 💡 We know the status is "approved" because of the API of `PollForApprovalWithCallback`, but this still feels a little awkward to be just falling back to a random string here. Maybe just pass status.Status here for consistency?

phinze: Ah now that I see we set "initializing" and "pending" below I'm realizing we're already sort of the authority on the state machine strings in this file, so this is probably fine for now.

If we end up using registration concepts elsewhere someday we can get all enum-y

pkg/auth/jwt.go
```diff
@@ -0,0 +1,299 @@
...
+// It fetches public keys from the JWKS endpoint at {cloudURL}/.well-known/jwks.json
+// and caches them for efficient validation. Only Ed25519 keys are supported.
+type JWTValidator struct {
+	cloudURL   string
+	httpClient *http.Client
+
+	mu        sync.RWMutex
+	keys      map[string]crypto.PublicKey // Map of kid -> public key
+	keyExpiry time.Time
+}
```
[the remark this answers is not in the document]
evanphx: Not needed atm.

pkg/cloudauth/rpc_adapter.go
```diff
@@ -0,0 +1,200 @@
...
+	// Set default CloudURL if not provided
+	if config.CloudURL == "" {
+		config.CloudURL = DefaultCloudURL
+	}
+
+	// Validate configuration
+	if err := config.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid configuration: %w", err)
+	}
+
```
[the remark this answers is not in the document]
evanphx: The context cancelation handles that.

pkg/registration/registration.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if err := os.WriteFile(path, data, 0600); err != nil {
+		return fmt.Errorf("failed to write registration file: %w", err)
+	}
+
+	// Also save the private key separately for easier access
+	keyPath := filepath.Join(dir, "service-account.key")
+	if err := os.WriteFile(keyPath, []byte(reg.PrivateKey), 0600); err != nil {
+		return fmt.Errorf("failed to write private key: %w", err)
+	}
```
phinze, line 253: 💡 Don't think this is used anywhere. Remove for simplicity?

pkg/auth/jwt.go
```diff
@@ -0,0 +1,299 @@
...
+		},
+	}
+}
+
+// Claims represents the JWT claims from miren.cloud
+type Claims struct {
+	jwt.RegisteredClaims
+	OrganizationID string   `json:"organization_id,omitempty"`
+	GroupIDs       []string `json:"group_ids,omitempty"`
+}
```
phinze, line 47: 💭 This is the first "maybe a monorepo is worth considering" thought that's popped into my head (cause it might be slightly nice to share representations of things like auth claims). It's not strong. But it was a twinge so I say it out loud. 😄

pkg/auth/jwt.go
```diff
@@ -0,0 +1,299 @@
...
+
+		// Only support Ed25519 keys
+		if jwk.Kty != "OKP" || jwk.Crv != "Ed25519" {
+			continue // Skip non-Ed25519 keys
+		}
+
+		publicKey, err := parseEd25519Key(jwk)
+
+		if err != nil {
+			// Log error but continue with other keys
```
phinze, line 165: 💡 lol when the comment says something but then the code does not do it

pkg/cloudauth/policy_fetcher.go
```diff
@@ -0,0 +1,231 @@
...
+	"log/slog"
+	"net/http"
+	"sync"
+	"time"
+
+	"miren.dev/runtime/pkg/rbac"
+)
+
+// DefaultRefreshInterval is the default interval for refreshing policies
+const DefaultRefreshInterval = 60 * time.Second
```
phinze, line 17: 💭 This feels a little aggressive... but I also get that you don't want people waiting around for sync. I say leave it for now and let's keep an eye on server load from this. I have a feeling we're going to want to move from a polling lifestyle to a persistent connection we can use for bidirectional triggered events, but we can see how things evolve!

pkg/cloudauth/rpc_adapter.go
```diff
@@ -0,0 +1,200 @@
...
+	"miren.dev/runtime/pkg/auth"
+	"miren.dev/runtime/pkg/rbac"
+)
+
+// DefaultCloudURL is the default URL for miren.cloud
+const DefaultCloudURL = "https://api.miren.cloud"
+
+// RPCAuthenticator adapts cloud authentication for RPC usage
+type RPCAuthenticator struct {
```
phinze: 💡 Match filename concept name just for search-ability?

pkg/registration/registration.go
```diff
@@ -0,0 +1,280 @@
...
+	return &Client{
+		managementURL: managementURL,
+		config:        config,
+		httpClient:    &http.Client{Timeout: 30 * time.Second},
+	}
+}
+
+// StartRegistration initiates the registration process
+func (c *Client) StartRegistration(ctx context.Context) (*Result, error) {
+	url := fmt.Sprintf("%s/api/v1/clusters/register/initiate", c.managementURL)
```
phinze, line 117: 💭 Not an immediate thing, but noting there are a lot of URL paths scattered around here. At some point before wide release might want to implement some sort of discovery/indirection here to give us flexibility to move services around.

pkg/registration/registration.go
```diff
@@ -0,0 +1,280 @@
...
+		return nil, fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return &result, nil
+}
+
+// PollForApproval polls the registration status until approved or timeout
+func (c *Client) PollForApproval(ctx context.Context, pollURL string, pollInterval time.Duration) (*Status, error) {
+	return c.PollForApprovalWithCallback(ctx, pollURL, pollInterval, nil)
+}
```
phinze: 💡 Never used, just remove and rename the other one to remove `WithCallback` and let the API be that it's optional?

pkg/rpc/server.go
```diff
@@ -305,7 +305,49 @@ func (s *Server) setupMux() {
...
+		allowed, identity, err := s.state.authenticator.NoAuthorization(r.Context(), r)
+		if err != nil {
+			s.state.log.Warn("authentication check failed", "error", err, "path", r.URL.Path)
+			http.Error(w, "authentication failed", http.StatusUnauthorized)
+			return
+		}
+		if !allowed {
+			// NOTE: This is a change from the old behavior. Previously during early development,
+			// the authentication was "best attempt" and we didn't reject requests that didn't
+			// present creds. We've since changed that to require	authentication.
```
phinze, line 340: 🙌🏻 Woohoo we're locking the doors now! I could see us wanting some backdoor for localhost requests... but ideally we can just make sure to set up our servers so that when you SSH in you have a client with auth material available.

phinze (approved):
Looking good! All nonblocking nits and musings inline.

pkg/auth/jwt.go
evanphx: Yeah, I had the exact same thought.

pkg/auth/jwt.go
evanphx: hahah

cli/commands/register.go
evanphx: Yeah, this is strictly so that we can reuse the keys and other registration data across attempts so that we don't end up creating a bunch of random cluster records because the this side was retried.

pkg/registration/registration.go
evanphx: We do use it a few places.

pkg/cloudauth/policy_fetcher.go
evanphx: Totes fair.

pkg/registration/registration.go
evanphx: I had the same thought. I'm not sure what the right pattern is yet.

pkg/rpc/server.go
evanphx: Yeah, I think we need to grow more auth mechanisms rather than leaving the side door open.

pkg/registration/registration.go
phinze: We got a lot of value out of Terraform's simple mechanism for service discovery: https://github.com/hashicorp/terraform/blob/e6dfb537e9f7f8419866873297186b0526fa50e9/internal/backend/remote/testing.go#L209-L215

Some bits different about our situation such that I wouldn't want exactly that, but I liked the simplicity of the approach.

PR #190, opened 2025-08-25
fix: prevent duplicate services in app version config (MIR-350)

phinze:
## Summary
Fixes MIR-350: Prevent duplicate services when building new app versions.

## Problem
When `nextVersion()` copies configuration from the previous app version, it includes the entire Services list. Later in `BuildFromTar()`, when we iterate through services with commands and apply concurrency configs (lines 412-441), we were appending new service configurations without checking for duplicates. This caused services to appear multiple times in the configuration.

## Solution
Replace the append logic with a proper deduplication approach:
1. Build a service map to track unique services by name
2. Rebuild the Services list from the deduplicated map

This ensures each service appears only once in the final configuration with the most up-to-date settings.

## Test plan
- [x] Code compiles successfully
- [x] Linting passes
- [ ] Manual testing with an app that has service concurrency configs
- [ ] Verify no duplicate services in resulting app version

## Summary by CodeRabbit

* **Bug Fixes**
  * Eliminates duplicate services in the final build output.
  * Ensures services defined only in configuration are included even without associated commands.
  * Preserves per-service concurrency settings from app config.

* **Refactor**
  * Consolidates configuration-defined and command-defined services into a single unified list.
  * Final service ordering may differ from previous versions (non-deterministic).

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -407,8 +407,9 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
 	}
 
 	// Apply service-level concurrency from AppConfig if present
-	// We iterate through all services that have commands (from srvMap)
-	// and apply concurrency config if available in app config
+	// Build a new services list to replace the old one, ensuring no duplicates
+	serviceMap := make(map[string]core_v1alpha.Services)
+
 	for serviceName := range srvMap {
 		svc := core_v1alpha.Services{
 			Name: serviceName,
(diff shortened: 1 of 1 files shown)


evanphx (approved):
I saw this another place in this file too, but I think we fixed it?

phinze:
> I saw this another place in this file too, but I think we fixed it?

If you're thinking of `serviceCmds` further up yeah we redid those when the per-service concurrency came in. Overall this thing is due for a big ole refactor. Saga RFD incoming! ✍🏻 😄 

Merging this as a quick fix in the meantime.

PR #193, opened 2025-08-27
Fix HTTP ingress context handling for lease acquisition

evanphx:
## Summary
- Fixed a critical issue where HTTP request contexts were being used for lease acquisition
- Request contexts are automatically cancelled when clients disconnect, which can interrupt important operations
- Now using a separate context for lease acquisition to ensure it completes regardless of client connection status

## Problem
When performing sequential operations in HTTP handlers, the request context cannot be trusted to remain valid throughout. If a client disconnects mid-request, the context is cancelled, which was causing lease acquisition operations to fail unexpectedly.

## Solution
Created a separate context specifically for the lease acquisition operation that isn't tied to the client connection lifetime. This ensures that once we start acquiring a lease, the operation will complete even if the client goes away.

## Test plan
- [x] Verify that lease acquisition completes even when clients disconnect early
- [x] Ensure no resource leaks with the new context management
- [x] Test with slow/flaky client connections

## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Improved resilience of request handling to client disconnects, reducing intermittent failures during lease acquisition and downstream processing.

- Documentation
  - Clarified behavior of request context cancellation and provided guidance for managing critical work when clients disconnect.

### DIFF
--- a/servers/httpingress/httpingress.go
+++ b/servers/httpingress/httpingress.go
@@ -240,6 +240,11 @@ func (h *Server) serveHTTPWithMetrics(w http.ResponseWriter, req *http.Request, 
 
 	ctx := req.Context()
 
+	// CRITICAL TO KNOW
+	// The context on requset is closed automaticaly when the client on the over side closes!
+	// So if you're doing critical work, don't use this context! Use a separate context and ping
+	// this one to figure out if you should continue with your critical work or clean up.
+
 	// Use ingress client to lookup route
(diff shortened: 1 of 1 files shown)


phinze (approved):
THE MYSTERY IS SOLVED THANK HEAVENS

evanphx:
https://www.youtube.com/watch?v=guHXLWmJRks

PR #194, opened 2025-08-27
feat: implement local shared storage for apps (RFD 26)

phinze:
## Summary
Implements [RFD 26: Local Shared Storage](https://github.com/mirendev/rfd/pull/36) to add persistent, host-local storage to Miren applications.

This PR adds a bind mount at `/miren/data/local` inside containers that persists across container restarts and is shared between all containers within an app.

## Changes
- Modified `controllers/sandbox/sandbox.go` to add local storage mount in `buildSubContainerSpec`
- When creating application containers, the controller:
  1. Fetches the AppVersion entity to get the App ID
  2. Creates a directory at `/var/lib/miren/data/local/{app_id}/` on the host
  3. Bind mounts this directory to `/miren/data/local` inside the container

## Storage Properties
- **Persistent**: Data survives container restarts and redeployments
- **Shared**: All containers within an app (across all versions) share the same storage
- **Isolated**: Each app has its own isolated storage directory
- **Well-organized**: Apps store data at `/var/lib/miren/data/local/app/{app-name}` on the host

## Use Cases
This enables applications to use:
- SQLite databases that persist across deployments
- File-based session storage
- User upload directories
- Any file-based data that needs to survive container lifecycle

## Testing
- [x] Code compiles successfully
- [x] Passes lint checks (`make lint-changed`)
- [x] Manual testing with a sample app using SQLite
- [x] Verify storage persists across container restarts

## Notes
- Only application containers get the mount (not the pause container) to keep it simple
- The implementation handles app IDs with prefixes (e.g., `app/myapp`) correctly
- Errors in fetching AppVersion are logged at ERROR level since this impacts storage availability

## Summary by CodeRabbit

* **New Features**
  * Adds per-app local storage inside sandboxed containers, providing a writable, persistent data area for each app when an app version and data path are present.
  * Request logging: sandboxed apps now append request logs to a local server.log in the app data area.

* **Bug Fixes**
  * More resilient startup: if app metadata or storage setup fails, the container proceeds without the local storage mount and logs the issue.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1198,6 +1198,41 @@ func (c *SandboxController) buildSubContainerSpec(
 		},
 	}
 
+	// Add local storage mount if this sandbox has an associated app
+	if sb.Version != "" && c.DataPath != "" {
+		// Fetch the AppVersion to get the App ID
+		res, err := c.EAC.Get(ctx, sb.Version.String())
+		if err == nil {
+			var appVer core_v1alpha.AppVersion
+			appVer.Decode(res.Entity().Entity())
+
+			if appVer.App != "" {
+				// Create the local storage path for this app
+				// Using 0777 for simplicity - containers run as non-root app user and need write access
+				// TODO: Consider more restrictive permissions with proper UID/GID mapping or ACLs for production
+				localStoragePath := filepath.Join(c.DataPath, "data", "local", appVer.App.String())
+				if err = os.MkdirAll(localStoragePath, 0777); err != nil {
+					c.Log.Warn("failed to create local storage directory", "path", localStoragePath, "error", err)
+				} else {
+					// Explicitly chmod to 0777 since MkdirAll respects umask
+					if err = os.Chmod(localStoragePath, 0777); err != nil {
+						c.Log.Warn("failed to set permissions on local storage directory", "path", localStoragePath, "error", err)
+					}
+					// Add the bind mount for local storage
+					mounts = append(mounts, specs.Mount{
+						Destination: "/miren/data/local",
+						Type:        "bind",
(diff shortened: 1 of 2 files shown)


phinze:
## Addressed Code Review Feedback

Thanks for the review @coderabbitai\! I've addressed the feedback in commit 1b8fd79:

### ✅ Implemented:
1. **Fixed error variable shadowing** - Changed from `err :=` to `err =` 
2. **Updated directory permissions** - Changed from 0755 to 0770 to support non-root containers
3. **Added security mount options** - Added `rprivate`, `nosuid`, `nodev` for defense in depth

### ⏭️ Skipped (with rationale):
- **Duplicate mount check**: We're in full control of the container lifecycle through the entity system. The `co.Mount` field comes from user-configured Container entities, not from Docker images. A duplicate would only occur if someone explicitly configured a mount to `/miren/data/local` in their Container entity, which would be an intentional override. The check seems overly defensive for our current use case.
- **chmod after MkdirAll**: Since we fully control the `/miren/data` path and nothing else creates directories there, the chmod is unnecessary. The `os.MkdirAll` with 0770 will set the correct permissions on creation.

The implementation is now more secure and compatible while keeping it simple and focused on the core functionality.

evanphx (approved):
The description on the PR is wrong, which is no biggy. This is storing in /var/lib/miren/local/{app}. Should it be /var/lib/miren/data/local/{app}?

phinze:
> The description on the PR is wrong, which is no biggy. This is storing in /var/lib/miren/local/{app}. Should it be /var/lib/miren/data/local/{app}?

Whoops! This is what I get for squeezing to get something done at the end of the day. 🤦🏻 You're correct it's just the description that's wrong. Fixing in descrip and commit message before so it's not confusing to some future human.

PR #199, opened 2025-08-29
Handle wrong servers better in 'config bind'

evanphx:
1. Don't try localhost at the start even if the server has it in it's list
2. If we can contact a server and it has the wrong fingerprint, keep trying
3. Always try localhost if all the other servers fail

## Summary by CodeRabbit

- New Features
  - Per-address TLS certificate fingerprint verification during cluster address probing when a CA fingerprint is provided.
  - Localhost fallback attempts (127.0.0.1 and [::1]:8443) if advertised cluster addresses fail.
  - Smarter address normalization for IPv4/IPv6, ports, and URL prefixes to improve probing accuracy.

- Bug Fixes
  - Avoids probing bind/local addresses (0.0.0.0, 127.0.0.1, ::1, localhost variants) during cluster discovery.
  - More reliable selection of a working cluster address, reducing connection and binding failures.

### DIFF
--- a/cli/commands/config_bind.go
+++ b/cli/commands/config_bind.go
@@ -131,6 +131,47 @@ func ConfigBind(ctx *Context, opts struct {
 			break
 		}
 
+		// If all normal addresses failed, try localhost as a fallback
+		if workingAddress == "" {
+			ctx.Info("All cluster addresses failed, trying localhost as fallback...")
+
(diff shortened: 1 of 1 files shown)


phinze (approved):
IMHO ultimately we should filter the noise addresses from the server at registration/discovery time instead of ignoring them down on the client, but this works as an incremental fix!

PR #200, opened 2025-09-02
Fix ClientConfig Save() to exclude clientconfig.d entries

evanphx:
## Summary
- Fixed a bug where `Save()` was incorrectly writing out data merged from `clientconfig.d` files
- `clientconfig.d` entries should remain in their separate files and not be saved to the main config

## Changes
- Added tracking fields to the `Config` struct to distinguish main config entries from merged ones
- Modified loading logic to mark clusters/identities from the main config file
- Updated `SaveTo()` to only write clusters/identities from the main config file
- Added comprehensive test coverage for various scenarios

## Test Plan
- [x] Added new test `TestSaveDoesNotIncludeConfigDirEntries` to verify merged entries are excluded
- [x] Added test `TestSavePreservesNewlyAddedClusters` for programmatically added clusters
- [x] Added test `TestSaveEmptyMainConfigWithConfigD` for edge cases
- [x] All existing tests pass

## Summary by CodeRabbit

- New Features
  - Per-file "leaf" configs supported under clientconfig.d; new leaves can be staged and persisted separately; saving writes main config and new leaves independently.
  - Unified config API: accessors for clusters and identities that honor main + leaf overrides and provide counts, names, iteration, and safer setters/getters.

- CLI Changes
  - Config-related commands now use the unified API and respect leaf overrides; safer add/remove/set-active flows and clearer leaf storage messages.

- Bug Fixes
  - Improved save/load semantics, stdout save error handling, and hostname/active-cluster validation.

### DIFF
--- a/cli/commands/auth_generate.go
+++ b/cli/commands/auth_generate.go
@@ -61,17 +61,15 @@ func AuthGenerate(ctx *Context, opts struct {
 		return fmt.Errorf("target hostname is empty, use --target to specify a hostname")
 	}
 
-	lcfg := &clientconfig.Config{
-		ActiveCluster: opts.ClusterName,
-
-		Clusters: map[string]*clientconfig.ClusterConfig{
-			opts.ClusterName: {
-				Hostname:   tgt,
-				CACert:     string(cc.CACert),
-				ClientCert: string(cc.CertPEM),
-				ClientKey:  string(cc.KeyPEM),
-			},
-		},
+	lcfg := clientconfig.NewConfig()
+	lcfg.SetCluster(opts.ClusterName, &clientconfig.ClusterConfig{
+		Hostname:   tgt,
+		CACert:     string(cc.CACert),
+		ClientCert: string(cc.CertPEM),
+		ClientKey:  string(cc.KeyPEM),
+	})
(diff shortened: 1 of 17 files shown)


phinze (approved):

evanphx:
@phinze Re-requested review because I ended up refactoring the ClientConfig to remove the issue all together.

phinze (approved):
Good refactoring! The encapsulation with private fields + proper getters/setters is much cleaner, and the leaf config system in clientconfig.d/ is a nice way to handle modular
configs.

One thing that might be confusing: when you run miren server, it drops a `clientconfig.d/50-local.yaml` file that sticks around in your user config even after the server is gone. So you end up with a "local" cluster showing up in miren config info that points to nothing. I get why it's there (runner needs to auth with the server), but mixing server-runtime configs with user CLI configs feels a bit odd. Maybe worth revisiting later - could use a different location for server-internal configs or clean them up on shutdown.

Not blocking though - this is a good improvement overall!

evanphx:
@phinze  Agree about the '50-local.yaml' entry. We might want to find a more elegant way to wire the client into local access.

PR #201, opened 2025-09-02
Add comprehensive tests for clientconfig.d loading scenarios

evanphx:
## Summary
- Fixed `IsEmpty()` to properly check both clusters and identities
- Added comprehensive tests for various clientconfig.d loading scenarios

## Changes
- Updated `IsEmpty()` method to return `false` when identities exist, even without clusters
- Added tests verifying config loads correctly when only clientconfig.d files exist (no main config)
- Added tests for configs with only identities (no clusters)
- Added tests for mixed config.d files containing different types of configurations
- Added real-world scenario tests simulating actual user environments

## Test Coverage
The new tests ensure:
- ✅ Configs load successfully from clientconfig.d when no main config exists
- ✅ Configs with only identities are recognized as non-empty
- ✅ Mixed config.d files (some with clusters, some with identities) merge correctly
- ✅ Empty config.d directories properly return `ErrNoConfig`
- ✅ Environment variable paths work correctly with clientconfig.d

## Test Plan
- [x] All new tests pass
- [x] All existing tests continue to pass
- [x] Verified `IsEmpty()` behavior with various configurations

## Summary by CodeRabbit

- Bug Fixes
  - Updated empty-configuration detection: configurations are only treated as empty when both clusters and identities are missing.
  - Improved robustness when the primary config file is absent by correctly loading from the config.d directory and setting the active cluster. Returns a clear error when config.d is empty.

- Tests
  - Added extensive unit and real-world scenario tests covering config loading from config.d, environment variable overrides, identity/cluster merging, active cluster selection, and error handling.

### DIFF
--- a/clientconfig/clientconfig.go
+++ b/clientconfig/clientconfig.go
@@ -59,7 +59,7 @@ func NewConfig() *Config {
 
 // IsEmpty checks if the configuration has no clusters defined
 func (c *Config) IsEmpty() bool {
-	return len(c.Clusters) == 0
+	return len(c.Clusters) == 0 && len(c.Identities) == 0
 }
 
 // Validate checks if the configuration is valid
--- a/clientconfig/clientconfig_only_dir_test.go
+++ b/clientconfig/clientconfig_only_dir_test.go
@@ -1,0 +1,167 @@
+package clientconfig
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


phinze (approved):

evanphx:
#200 Is going to rework the config pretty dramatically so I'm just going to fix the issue there.

PR #205, opened 2025-09-03
Improve sandbox cleanup to prevent resource leaks

evanphx:
## Summary
- Ensures containerd snapshots are always deleted when container boot fails
- Prevents IP address and port monitoring leaks during error scenarios
- Centralizes cleanup logic for consistency and maintainability

## Changes
- **Explicit snapshot deletion**: Added fallback using `snapshotter.Remove()` that runs even if `container.Delete()` fails
- **Network cleanup**: Added `deallocateNetwork()` function that releases IP addresses back to the subnet on all error paths
- **Port monitoring cleanup**: Integrated `StopMonitoring()` calls into the container cleanup flow
- **Refactored cleanup logic**: Created `cleanupContainer()` function to centralize all cleanup operations

## Test plan
- [x] Code compiles successfully (`make bin/miren`)
- [x] Code formatted with `go fmt`
- [x] Test sandbox creation failures verify resources are cleaned up
- [x] Verify no containerd snapshot leaks after failures
- [x] Verify IP addresses are released back to subnet pool

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents IP and network resource leaks when sandbox setup or boot fails.
  * Ensures pause and subcontainers, tasks and snapshots are consistently stopped and removed on errors or panics.
  * Marks failed sandboxes as DEAD and cleans up partial boot progress.

* **Behavior Change**
  * Worker processing now logs item errors but continues to allow subsequent update handling.

* **Refactor**
  * Consolidated cleanup into unified helpers for consistent error handling across sandbox creation paths.

* **Notes**
  * No changes to public APIs.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -513,6 +513,11 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 			// Clean up the pause container using the common cleanup function
 			pauseID := c.pauseContainerId(co.ID)
 			c.cleanupContainer(ctx, container, pauseID)
+
+			// Update sandbox status to DEAD in entity store
+			co.Status = compute.DEAD
+			meta.Attrs = co.Encode()
+			c.Log.Info("marked sandbox as DEAD due to boot failure", "id", co.ID)
 		}
 	}()
 
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -515,8 +515,7 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 			c.destroySubContainers(ctx, co)
 
 			// Clean up the pause container using the common cleanup function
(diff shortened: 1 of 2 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -1053,6 +1080,71 @@ type waitPort struct {
...
 
+// cleanupContainer removes a container and its snapshot during failure scenarios
+func (c *SandboxController) cleanupContainer(ctx context.Context, cont containerd.Container, containerID string) {
+	if cont == nil {
+		return
+	}
+
+	if containerID == "" {
+		containerID = cont.ID()
+	}
```
phinze: 💡 Might be simpler to just take the `Container` and remove the extra arg if the id can be set in terms of it?

phinze (approved):

controllers/sandbox/sandbox.go
evanphx: very true!

PR #206, opened 2025-09-03
Detect when a sandbox dies on boot, flow error to ingress

evanphx:
We were not terminating the watch of the sandbox status if it transitioned to other states than running. So when a sandbox died, we watched it literally until the end of time.

Now we stop as soon as the sandbox transitions to running OR a terminal state.

Additionally, the httpingress can now detect when this happens and display a nicery message to the user.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate response when an application fails to boot: returns HTTP 408 (Request Timeout) with a clear message instead of a generic 500 in certain cases.
* **Chores**
  * Improved startup error handling and more granular logging for sandbox status transitions to aid troubleshooting.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -245,6 +245,8 @@ func (a *localActivator) AcquireLease(ctx context.Context, ver *core_v1alpha.App
 	return a.activateApp(ctx, ver, pool, service)
 }
 
+var ErrSandboxDiedEarly = fmt.Errorf("sandbox died while booting")
+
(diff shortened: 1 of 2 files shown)


phinze (approved):
This is a great step in the right direction! There are still some failure modes where the container will keep running but e.g. not listen on the right port or whatever, and we'll have to cover those with a fully fledged health check thingy. But we also need this, so it's very much worth having now.

PR #207, opened 2025-09-03
Propagate data-path from server CLI to components

evanphx:
This uses a little used feature of injecting the CLI flags into the registry. It also makes a small change, allowing the flags from the CLI to override any existing values, so there can be defaults.

This was found because net.db was always being written to /var/lib/miren regardless of the value of -d on the command line.

## Summary by CodeRabbit

- Refactor
  - Introduced override-aware component inference to improve registration handling.
- Chores
  - Added internal annotation to the server data path option; no change to defaults or CLI behavior.
  - General internal updates to align with new inference flow.


phinze (approved):
Nice catch!

PR #209, opened 2025-09-04
Refresh RBAC rules after rejection with 30-second cooldown

evanphx:
## Summary
- Automatically refresh RBAC rules from miren.cloud when access is denied
- Implement 30-second cooldown to prevent excessive API calls
- Clear decision cache after successful refresh to ensure fresh evaluation

## Details
When RBAC denies a request, the system now automatically triggers a refresh of the RBAC rules. This helps ensure the runtime has the latest permissions without waiting for the next periodic refresh cycle (default 60 seconds).

### Implementation
- Added `RefreshIfNeeded()` method to `PolicyFetcher` that:
  - Checks if 30 seconds have passed since last refresh attempt
  - Runs the refresh asynchronously to avoid blocking the rejection response
  - Clears the RBAC evaluator's decision cache on successful refresh
  
- Integrated with `RPCAuthenticator` to call `RefreshIfNeeded()` when RBAC denies access

### Benefits
- Faster permission updates when rules change in miren.cloud
- Reduced latency between permission grant and access
- No excessive API calls due to 30-second cooldown
- Non-blocking implementation maintains response times

## Test Plan
- [x] Unit tests pass
- [x] Code builds successfully
- [x] Linting passes
- [ ] Manual testing with RBAC rejections
- [ ] Verify cooldown prevents excessive refreshes
- [ ] Confirm cache is cleared after refresh

## Summary by CodeRabbit

- **New Features**
  - Automatically refreshes access policies in the background when needed.
  - Triggers a quick policy refresh after an access denial (with cooldown) to keep decisions current.
  - Improved coordination between policy updates and authorization checks.

- **Bug Fixes**
  - Clears cached evaluations after successful refreshes to reduce stale authorization results.
  - Avoids caching denial decisions so access denials are re-evaluated, improving consistency.

### DIFF
--- a/pkg/cloudauth/policy_fetcher.go
+++ b/pkg/cloudauth/policy_fetcher.go
@@ -56,11 +56,13 @@ type PolicyFetcher struct {
 	httpClient      *http.Client
 	logger          *slog.Logger
 	refreshInterval time.Duration
+	evaluator       *rbac.Evaluator
 
 	mu          sync.RWMutex
 	policy      *rbac.Policy
 	lastFetched time.Time
 	lastError   error
+	lastRefresh time.Time
 
 	stop func()
 	wg   sync.WaitGroup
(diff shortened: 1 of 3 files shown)


phinze (approved):
We just hit this while trying to add @teresamychu to all our instances! No more 5m cache timeout waiting is very welcome 😄

PR #212, opened 2025-09-05
Print the name of the cluster being deployed to

evanphx:
Deploy is a big operation, and so we should be clear where the deploying is coming to.

A user mistakenly had the local cluster as the active one and was deploying to it and confused why the deploy wasn't going to the correct cluster. This gives them some feedback to be able to solve that issue.

## Summary by CodeRabbit

* **New Features**
  * Deploy now displays the target cluster (colorized, with a checkmark) before starting, so you see the deployment destination up front.
  * CLI captures and exposes the resolved cluster name in the command context for clearer status and flows.

* **User-facing Behavior**
  * If no cluster is specified and no active cluster is set, the command now reports an error instead of proceeding.

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/progress/progresswriter"
 
@@ -27,6 +28,9 @@ func Deploy(ctx *Context, opts struct {
 	Explain       bool   `short:"x" long:"explain" description:"Explain the build process"`
 	ExplainFormat string `long:"explain-format" description:"Explain format" choice:"auto" choice:"plain" choice:"tty" choice:"rawjson" default:"auto"`
 }) error {
+	greenStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("10"))
+	ctx.Printf("  ✓ %s: %s\n", greenStyle.Render("Deploying to cluster"), ctx.ClusterName)
+
 	cl, err := ctx.RPCClient("dev.miren.runtime/build")
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


cli/commands/deploy.go
```diff
@@ -27,6 +28,21 @@ func Deploy(ctx *Context, opts struct {
 	Explain       bool   `short:"x" long:"explain" description:"Explain the build process"`
 	ExplainFormat string `long:"explain-format" description:"Explain format" choice:"auto" choice:"plain" choice:"tty" choice:"rawjson" default:"auto"`
 }) error {
+	// Display target cluster if multiple clusters are configured
+	cfg, _ := opts.LoadConfig()
+	if cfg != nil && cfg.GetClusterCount() > 1 {
```
phinze: 💡 Should we just always print the cluster name? Seems like a consistent intro line might be nice even for single-cluster users to make it clear the code is going somewhere else

phinze (approved):
🤔 The config should already be loaded if we're in an AppCentric CLI, so it seems like there should be an easier way to get "the configured cluster" than this... but the change looks simple enough that it's nonblocking!

cli/commands/deploy.go
evanphx: Sure, i'll make that change.

evanphx:
Oh der! yes.

cli/commands/global.go
```diff
@@ -118,11 +119,12 @@ func setup(ctx context.Context, flags *GlobalFlags, opts any) *Context {
 	}
 
 	if lc, ok := opts.(interface {
-		LoadCluster() (*clientconfig.ClusterConfig, error)
+		LoadCluster() (*clientconfig.ClusterConfig, string, error)
```
phinze, line 122: 💭 Ah the evergreen conundrum of "when you have a `map[string]Thing` do you nest the `string` as a field of the `Thing` (thereby making it convenient to ship around, but duplicating it) or pass it around separately?" There is no answer to this question, we will be pondering it for eternity. 🙃

phinze (approved):

cli/commands/global.go
evanphx: so much! I was like "should I add Name to ClusterConfig?" and just noped out of deciding.

PR #222, opened 2025-09-20
A number of small QOL fixes found during demo prep

evanphx:
These are all about making the default experience predictable.

## Summary by CodeRabbit

* **New Features**
  * Standalone mode now auto-downloads the required release when missing, choosing a global or user path based on permissions.
  * Coordinator configuration accepts additional hostnames and IPs for cluster coordination.

* **Improvements**
  * Safer cleanup on failed downloads to prevent partial installs.
  * Smarter server address handling: auto-adds default port and normalizes host.

* **Defaults**
  * Server now binds to all interfaces by default (:8443).
  * Standard TLS is enabled by default.

### DIFF
--- a/cli/commands/download_release.go
+++ b/cli/commands/download_release.go
@@ -17,12 +17,16 @@ import (
 	"time"
 )
 
-func DownloadRelease(ctx *Context, opts struct {
-	Branch string `short:"b" long:"branch" description:"Branch name to download" default:"main"`
-	Global bool   `short:"g" long:"global" description:"Install globally to /var/lib/miren/release"`
-	Force  bool   `short:"f" long:"force" description:"Force download even if release directory exists"`
-	Output string `short:"o" long:"output" description:"Custom output directory"`
-}) error {
+// DownloadReleaseOptions contains options for downloading a release
+type DownloadReleaseOptions struct {
+	Branch string
+	Global bool
+	Force  bool
+	Output string
+}
+
+// PerformDownloadRelease performs the actual download with the given options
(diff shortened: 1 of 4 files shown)


cli/commands/download_release.go
```diff
@@ -17,12 +17,16 @@ import (
...
-	Force  bool   `short:"f" long:"force" description:"Force download even if release directory exists"`
-	Output string `short:"o" long:"output" description:"Custom output directory"`
-}) error {
+// DownloadReleaseOptions contains options for downloading a release
+type DownloadReleaseOptions struct {
+	Branch string
+	Global bool
+	Force  bool
+	Output string
+}
```
phinze, line 26: 💡 IMHO you could leave the field tags on this struct and use it for the CLI and pass it on through vs duplicating it.

cli/commands/server.go
```diff
@@ -91,10 +92,52 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
...
+
+					// Download the release
+					if err := PerformDownloadRelease(ctx, DownloadReleaseOptions{
+						Branch: "main",
+						Global: downloadGlobal,
+						Force:  false,
+						Output: downloadPath,
+					}); err != nil {
+						return fmt.Errorf("failed to download release: %w", err)
+					}
```
phinze, line 136: 💡 Feels like we're close to having enough release-oriented code to wrap up in a pkg between download release and this chunk. Not critical though.

phinze (approved):
Good improvements!

cli/commands/download_release.go
evanphx: I considered it but opted to just let it live it's own life.

PR #234, opened 2025-10-08
feat: Add deployment tracking system

teresamychu:
## Summary
- Implements deployment tracking as specified in RFD 32 (Multi-User Workflow)
- Records who deployed what, when, and captures git information at deployment time
- Deployment locking will be implemented as a separate feature after this PR

## Key Features
- **New deployment entity** to track all deployment attempts (successful and failed)
- **Records deployment metadata**: user, timestamp, git SHA/branch/message, deployment status
- **Tracks deployment phases**: preparing, building, pushing, activating
- **Captures build errors and logs** for failed deployments
- **New CLI commands**:
  - `m app status`: Shows current active deployment with git info
  - `m app history`: Shows deployment history (default: compact view with essential columns)
  - `m app history --detailed`: Shows full details including git SHA, branch, and commit message

## Technical Changes
- Added deployment entity schema to core entity system
- New deployment RPC service for managing deployment records
- Updated deploy command to create and update deployment records throughout the deployment lifecycle
- Added git info capture during deployment (SHA, branch, commit message, author)
- Proper error handling with separate context for status updates (ensures deployments are marked as failed even if main context is cancelled)

## Testing
- Successfully deployed apps show as "active" with full git information
- Failed builds are properly marked as "failed" with error messages and build logs
- Deployment history shows all attempts in chronological order
- Timeout scenarios properly mark deployments as failed

## Notes
- The "deployed by" field currently shows a placeholder value. This will be populated with actual user information when the deployment permissions PR is merged.

## Next Steps
After this PR is merged, deployment locking will be implemented as a separate feature to prevent concurrent deployments of the same app.

## Summary by CodeRabbit

- **New Features**
  - Deployment service: create, update, list, fetch deployments and active deployment per app/cluster.
  - Deployment tracking: status, phase, timestamps, error messages, build logs, and rich Git metadata captured at deploy time.
  - New CLI commands: app history (filtered, detailed view) and app status (config, active deployment, recent activity, masked secrets).
  - Git metadata utility to gather repo info for deployments.

- **Tests**
  - Comprehensive server tests covering creation, listing, retrieval, status transitions, and GitInfo handling.


pkg/git/info.go
```diff
@@ -0,0 +1,144 @@
...
+		diffOutput, _ := runGitCommand(absDir, "diff", "HEAD", "--no-ext-diff")
+		
+		// Combine both outputs to create a unique fingerprint of the dirty state
+		combinedState := fmt.Sprintf("status:\n%s\ndiff:\n%s", statusOutput, diffOutput)
+		
+		// Create a hash of the combined state
+		cmd := exec.Command("git", "hash-object", "--stdin")
+		cmd.Dir = absDir
+		cmd.Stdin = strings.NewReader(combinedState)
+		output, err := cmd.Output()
```
evanphx, line 73: This is a really cool concept, hashing the working status. In this case, the body is just bytes that are in a git blob, so maybe we should use sha1.Hash natively in git rather than shell out? Not a big deal.

pkg/git/info.go
```diff
@@ -0,0 +1,144 @@
...
+		
+		if err != nil || len(output) == 0 {
+			// Fallback: use a simple hash of the status
+			info.WorkingTreeHash = "dirty"
+		} else {
+			// Use first 8 characters of the hash
+			info.WorkingTreeHash = strings.TrimSpace(string(output))[:8]
```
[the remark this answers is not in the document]
evanphx: Also, if we use sha1.Hash here, we don't need this extra logic because it will always work, rather than shelling out which might error.

pkg/git/info.go
```diff
@@ -0,0 +1,144 @@
...
+	email, err := runGitCommand(absDir, "log", "-1", "--pretty=%ae")
+	if err == nil {
+		info.CommitEmail = strings.TrimSpace(email)
+	}
+
+	// Get commit timestamp (ISO 8601 format)
+	timestamp, err := runGitCommand(absDir, "log", "-1", "--pretty=%cI")
+	if err == nil {
+		info.CommitTimestamp = strings.TrimSpace(timestamp)
+	}
```
evanphx, line 106: Could we call `log --pretty` once with all these options and split the output?

servers/deployment/server.go
```diff
@@ -0,0 +1,635 @@
...
+	}
+
+	appName := args.AppName()
+	clusterId := args.ClusterId()
+	appVersionId := args.AppVersionId()
+
+	// Get user info from context (will be implemented with auth integration)
+	// For now, use placeholder values
+	userId := "user-" + uuid.New().String()
+	userEmail := "user@example.com"
```
evanphx, line 55: Not critical at this first phase, but I *think* we have the auth info already. I'd have to check the api to access it (something we can retrieve via ctx perhaps?)

evanphx (approved):
Looks great! It's worth a convo after this is merged on if we should do more of the deployment object managed on the server side inside build, but doing it inside the `deploy` command is totally fine too.

api/core/schema.yml
```diff
@@ -126,4 +126,86 @@ kinds:
...
               doc: Whether or not the value is sensitive
 
+  deployment:
+    app_name:
+      type: string
+      doc: The name of the app being deployed
+      indexed: true
+    
+    app_version:
+      type: string
```
phinze, line 136: 💡 This can be `type: ref` so the schema knows we're referencing another entity. It will do a few extra checks to ensure the reference is valid.

api/core/schema.yml
```diff
@@ -126,4 +126,86 @@ kinds:
...
+      doc: The app version ID or temporary value (pending-build, failed-{id})
+    
+    cluster_id:
+      type: string
+      doc: The cluster where the deployment is happening
+      indexed: true
+    
+    status:
+      type: string
```
phinze, line 145: 💡 This can be a`type: enum` which will make the entity system ensure only valid statuses get stored to it.

cli/commands/app_history.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+func AppHistory(ctx *Context, opts struct {
+	AppCentric
+
+	Limit      int32  `short:"n" long:"limit" description:"Maximum number of deployments to show" default:"20"`
+	Status     string `short:"s" long:"status" description:"Filter by status (active, failed, rolled_back)"`
+	All        bool   `long:"all" description:"Show deployments from all clusters"`
+	ShowFailed bool   `long:"show-failed" description:"Include failed deployments (shown by default)"`
+	HideFailed bool   `long:"hide-failed" description:"Hide failed deployments"`
```
phinze, line 19: 💡 I think both `ShowFailed` and `HideFailed` are redundant flags given we already have `Status` filtering if users want to filter the output.

cli/commands/app_history.go
```diff
@@ -0,0 +1,242 @@
...
+			errorStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("9")).Italic(true)
+			ctx.Printf("             %s\n", errorStyle.Render("Error: "+dep.ErrorMessage()))
+		}
+	}
+
+	return nil
+}
+
+// formatRelativeTime formats a time as a relative string (e.g. "2 hours ago")
+func formatRelativeTime(t time.Time) string {
```
phinze, line 223: 💡 There's a helper nearly identical to this in `sandbox_list.go` called `humanFriendlyTimestamp` - it should be in scope to call directly from this file, and ideally we could move it to `pkg/ui` for further sharing.

servers/deployment/server.go
```diff
@@ -0,0 +1,635 @@
...
+// decodeEntity is a helper to decode RPC entity to struct
+func decodeEntity(rpcEntity *entityserver_v1alpha.Entity, target interface{}) {
+	type decoder interface {
+		Decode(entity.AttrGetter)
+	}
+
+	if d, ok := target.(decoder); ok {
+		d.Decode(&rpcEntityWrapper{entity: rpcEntity})
+	}
+}
```
phinze, line 600: 💡 The `decodeEntity` helper and `rpcEntityWrapper` type aren't needed here. The entity response already provides direct access to the underlying `*entity.Entity` (which implements`AttrGetter`) via `.Entity().Entity()`.

So instead of:

```go
var deployment core_v1alpha.Deployment
decodeEntity(deploymentResp.Entity(), &deployment)
```

You can use:

```go
var deployment core_v1alpha.Deployment
deployment.Decode(deploymentResp.Entity().Entity())
```

Here are a couple of examples of this pattern is used elsewhere in the codebase:

- https://github.com/mirendev/runtime/blob/c6a4526475633369682df20261c0ad13407004f1/servers/httpingress/httpingress.go#L348-L352
- https://github.com/mirendev/runtime/blob/c6a4526475633369682df20261c0ad13407004f1/controllers/sandbox/sandbox.go#L194
- https://github.com/mirendev/runtime/blob/c6a4526475633369682df20261c0ad13407004f1/controllers/service/service.go#L54

phinze (approved):
Exciting stuff! All inline comments are nonblocking nits so giving it the big ole ✅

teresamychu:
Going to address these non-blocking comments in my next PR

PR #238, opened 2025-10-09
Add systemd service management with graceful drain and cloud registration

evanphx:
This PR adds comprehensive systemd service management commands for the miren server, including graceful node draining, idempotent installation, and cloud registration integration.

Closes MIR-442

## Summary

- New CLI commands: `miren server install`, `miren server uninstall`, `miren server status`
- Graceful node drain on SIGUSR2 signal before shutdown
- Automatic cloud registration during installation
- Improved RPC error handling with user-friendly messages

## Summary by CodeRabbit

* **New Features**
  * Linux CLI: add server install/uninstall/status commands to manage the server as a system service (install/enable/start, stop/disable, status, log follow, optional data backup/remove).
  * Graceful server drain on signal to safely stop active workloads and drain running sandboxes.

* **Bug Fixes**
  * Clearer "access denied" reporting with a user-facing warning for remote operations.
  * More descriptive, categorized errors when resolving remote capabilities for improved diagnostics.

### DIFF
--- a/cli/commands/server_install.go
+++ b/cli/commands/server_install.go
@@ -243,8 +243,28 @@ func ServerUninstall(ctx *Context, opts struct {
 		return fmt.Errorf("service file not found at %s", servicePath)
 	}
 
+	// Check if service is running
+	isRunningCmd := exec.Command("systemctl", "is-active", "miren.service")
+	output, err := isRunningCmd.CombinedOutput()
+	isRunning := err == nil && strings.TrimSpace(string(output)) == "active"
+
+	// If service is running, drain it first
+	if isRunning {
+		ctx.Info("Service is running, draining before shutdown...")
+		ctx.Info("Sending SIGUSR2 to drain runner...")
+
+		// Send SIGUSR2 to the service using systemctl kill
+		killCmd := exec.Command("systemctl", "kill", "-s", "SIGUSR2", "miren.service")
+		if err := killCmd.Run(); err != nil {
+			ctx.Warn("Failed to send SIGUSR2: %v", err)
+		} else {
+			ctx.Info("Drain signal sent, waiting 5 seconds for graceful shutdown...")
+			time.Sleep(5 * time.Second)
+		}
+	}
+
 	// Stop the service if requested or if it's running
-	if opts.Stop {
+	if opts.Stop || isRunning {
 		ctx.Info("Stopping miren service...")
 		cmd := exec.Command("systemctl", "stop", "miren.service")
 		if output, err := cmd.CombinedOutput(); err != nil {
--- a/cli/commands/server_install.go
+++ b/cli/commands/server_install.go
@@ -1,0 +1,346 @@
+//go:build linux
+// +build linux
+
+package commands
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+// ServerInstall sets up systemd units to run the miren server
+func ServerInstall(ctx *Context, opts struct {
+	Address   string `short:"a" long:"address" description:"Server address to bind to" default:"0.0.0.0:8443"`
+	Verbosity string `short:"v" long:"verbosity" description:"Verbosity level" default:"-vv"`
+	Start     bool   `long:"start" description:"Start the service immediately after installation"`
+	Force     bool   `short:"f" long:"force" description:"Overwrite existing service file"`
+}) error {
+	// Check if running with sufficient privileges
+	if os.Geteuid() != 0 {
+		return fmt.Errorf("server install requires root privileges (use sudo)")
+	}
+
+	// Verify the miren binary exists at the expected location
+	mirenPath := "/var/lib/miren/release/miren"
+	if _, err := os.Stat(mirenPath); err != nil {
+		return fmt.Errorf("miren binary not found at %s. Run 'miren download release --global' first", mirenPath)
+	}
+
+	ctx.Info("Installing miren systemd service...")
+
+	// Check if service file already exists
+	servicePath := "/etc/systemd/system/miren.service"
+	if _, err := os.Stat(servicePath); err == nil && !opts.Force {
+		return fmt.Errorf("service file already exists at %s (use --force to overwrite)", servicePath)
+	}
+
+	// Build ExecStart command
+	var execStartParts []string
+	execStartParts = append(execStartParts, mirenPath, "server")
+
+	if opts.Verbosity != "" {
+		execStartParts = append(execStartParts, opts.Verbosity)
+	}
+
+	if opts.Address != "" {
+		execStartParts = append(execStartParts, fmt.Sprintf("--address=%s", opts.Address))
+	}
+
+	execStartParts = append(execStartParts, "--serve-tls")
+
+	execStart := strings.Join(execStartParts, " ")
+
+	// Create systemd service file content
+	serviceContent := fmt.Sprintf(`[Unit]
+Description=Miren Service
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+Environment="NO_COLOR=1"
+ExecStart=%s
+Restart=always
+RestartSec=10
+StandardOutput=journal
+StandardError=journal
+SyslogIdentifier=miren
+User=root
+WorkingDirectory=/var/lib/miren/release
+KillMode=process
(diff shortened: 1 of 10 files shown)


cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+	// Stop the service if requested or if it's running
+	if opts.Stop || isRunning {
+		ctx.Info("Stopping miren service...")
+		cmd := exec.Command("systemctl", "stop", "miren.service")
+		if output, err := cmd.CombinedOutput(); err != nil {
+			ctx.Warn("Failed to stop service: %v\nOutput: %s", err, output)
+		} else {
+			ctx.Completed("Service stopped")
+		}
+	}
```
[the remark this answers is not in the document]
evanphx: That's fine because the process will still get a chance to continue it's cleanup when systemd sends it a stop signal.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+		if info.Mode().IsRegular() {
+			file, err := os.Open(path)
+			if err != nil {
+				return fmt.Errorf("failed to open file %s: %w", path, err)
+			}
+			defer file.Close()
+
+			if _, err := io.Copy(tarWriter, file); err != nil {
+				return fmt.Errorf("failed to write file %s to tar: %w", path, err)
+			}
```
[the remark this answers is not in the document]
evanphx: We're in a callback function so the defer will run pretty quickly.

components/runner/runner.go
```diff
@@ -79,6 +82,63 @@ func (r *Runner) Close() error {
...
+		}
+	}
+
+	if drainErr != nil {
+		return fmt.Errorf("errors during drain: %w", drainErr)
+	}
+
+	r.Log.Info("runner drained successfully", "id", r.Id, "sandboxes_stopped", stoppedCount)
+	return nil
+}
```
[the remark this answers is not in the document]
evanphx: It won't.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+	CloudURL     string            `short:"u" long:"url" description:"Cloud URL for registration" default:"https://miren.cloud"`
+	Tags         map[string]string `short:"t" long:"tag" description:"Tags for the cluster (key:value)"`
+}) error {
+	// Check if running with sufficient privileges
+	if os.Geteuid() != 0 {
+		return fmt.Errorf("server install requires root privileges (use sudo)")
+	}
+
+	// Check if miren binary exists, download if not
+	mirenPath := "/var/lib/miren/release/miren"
```
phinze, line 38: 🔍 Several paths in here are inline and duplicated - I'd like to consolidate these into the `serverconfig` package, but that can be a separate pass. That package is just a first draft at this point and there's still a lot of dancing that happens in `server.go` to set `ReleasePath`

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+			}
+		}
+	} else {
+		ctx.Info("Skipping cloud registration (--without-cloud specified)")
+	}
+
+	ctx.Info("Installing miren systemd service...")
+
+	// Check if service file already exists
+	servicePath := "/etc/systemd/system/miren.service"
```
phinze, line 119: 🔍 Same with this, which is set inline 3 times. I think endgame here is this command can optionally take the full server config and this becomes:

```go
cfg, err := serverconfig.Load(configFile, &opts, ctx.Log)
servicePath := cfg.GetServicePath()
```

Again I don't think we have to tackle this now... just sketching for directionality.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+StandardOutput=journal
+StandardError=journal
+SyslogIdentifier=miren
+User=root
+WorkingDirectory=/var/lib/miren/release
+KillMode=process
+TimeoutStopSec=90s
+
+[Install]
+WantedBy=multi-user.target
```
phinze, line 169: 💭 Yeah then this would move into `serverconfig` which is where I'd expect to find something like this vs inline.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+		ctx.Info("Service is running, draining before shutdown...")
+		ctx.Info("Sending SIGUSR2 to drain runner...")
+
+		// Send SIGUSR2 to the service using systemctl kill
+		killCmd := exec.Command("systemctl", "kill", "-s", "SIGUSR2", "miren.service")
+		if err := killCmd.Run(); err != nil {
+			ctx.Warn("Failed to send SIGUSR2: %v", err)
+		} else {
+			ctx.Info("Drain signal sent, waiting 5 seconds for graceful shutdown...")
+			time.Sleep(5 * time.Second)
```
phinze, line 261: 💭 I wonder if that healthcheck we just added could grow a status field for "draining" -> "drained" we can poll. Could be a nice future enhancement.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+	fmt.Println()
+	ctx.Info("To view logs:")
+	fmt.Println("  journalctl -u miren -f")
+
+	return nil
+}
+
+// ServerUninstall removes the systemd service and optionally removes /var/lib/miren
+func ServerUninstall(ctx *Context, opts struct {
+	Stop       bool   `long:"stop" description:"Stop the service before uninstalling"`
```
phinze: ❓ Why is this optional? Isn't it required if it's running?

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+					return fmt.Errorf("failed to create backup: %w", err)
+				}
+				ctx.Completed("Backup created at %s", backupPath)
+			} else {
+				ctx.Warn("Skipping backup as requested")
+			}
+
+			// Remove the directory
+			ctx.Info("Removing %s...", mirenDataDir)
+			if err := os.RemoveAll(mirenDataDir); err != nil {
```
phinze, line 325: ❓ This deletes the `miren` command out from under itself. I wonder if we should do an early "are you sure" prompt (skippable with a `--force` flag)? See `EnvDelete` for example of using the `ui.Confirm` control in this pattern.

cli/commands/server_install.go
```diff
@@ -0,0 +1,450 @@
...
+		ctx.Info("Note: The miren data at /var/lib/miren has not been removed.")
+		ctx.Info("To remove it with backup: sudo miren server uninstall --remove-data")
+		ctx.Info("To remove it without backup: sudo rm -rf /var/lib/miren")
+	}
+
+	return nil
+}
+
+// ServerStatus shows the status of the miren systemd service
+func ServerStatus(ctx *Context, opts struct {
```
phinze, line 350: ✨ Convenient!

phinze (approved):
Looking great! All inline musings are non-blocking.

cli/commands/server_install.go
evanphx: I agree, there is a cleanup pass PR I'd like to do.

cli/commands/server_install.go
evanphx: That was my thought too!

cli/commands/server_install.go
evanphx: Oh, yeah, that's a mistake, fixing.

cli/commands/server_install.go
evanphx: It deletes the one in /var/lib/miren only, I presume that's not the one the user is running?

phinze: Hm... In all my server installs I'm installing globally and symlinking `/usr/local/bin/miren` to `/var/lib/miren/release/miren` to avoid confusion with two different binaries floating around.

What's your imagined layout in the ideal case?

phinze: Reference for what our servers are doing now - happy to tweak to align with what we decide:

https://github.com/mirendev/infra/blob/25922a89c67a68077442da82856e8f3c79275356/terraform/miren-development/modules/miren-instance/scripts/50-miren-install.sh#L3-L41

evanphx: Well, when you do the normal install, it either writes to /usr/local/bin or ~/.miren/bin. The user is then running `miren server install` or `miren server uninstall` against the miren in either of those paths.

phinze: Ah right okay. I still would prefer to get away from there being two different copies of `miren` on the happy path, but this makes sense for now.

PR #241, opened 2025-10-10
Add standalone component support to entity schema system

phinze:
This PR adds support for standalone component types in the entity schema system, enabling component definitions to be shared across multiple entity kinds without duplication.

## Context

As part of our activator refactoring (RFD-34, MIR-435/MIR-436), we need to separate immutable build-time configuration from runtime state in the Sandbox entity. The goal is to extract fields like `version`, `container[]`, and `logEntity` into a `SandboxSpec` component that can be:

1. Referenced by `Sandbox` entities (alongside runtime fields like `status` and `network.address`)
2. Stored in `SandboxPool` entities as a template for creating sandboxes
3. Shared without duplication between these two entity kinds

This requires the schema system to support standalone component types that can be referenced by name, rather than only inline component definitions.

## What This Adds

**New Schema Syntax** - Top-level `components:` section in schema.yml:
```yaml
components:
  sandbox_spec:
    version:
      type: ref
      doc: Application version reference
    container:
      type: component
      many: true
      attrs:
        image:
          type: string
          required: true

kinds:
  sandbox:
    status:
      type: enum
      choices: [pending, running, stopped, dead]
    spec:
      type: sandbox_spec  # References component by name
  
  sandbox_pool:
    desired_instances:
      type: int
    sandbox_spec:
      type: sandbox_spec  # Same component referenced here
```

**Generated Code**:
- Standalone component types generate structs with `Decode()`, `Encode()`, `Empty()`, and `InitSchema()` methods
- Component types do **NOT** generate entity-specific methods (`Kind()`, `EntityId()`, `ShortKind()`, `Is()`)
- Entity kinds can reference components by type name, supporting both single values and `many: true`
- Component attributes are namespaced with `component.` prefix (e.g., `dev.miren.compute/component.sandbox_spec.version`)
- Nested structures within components use prefixed names (e.g., `SandboxSpecContainer`) to avoid collisions
- Nested structures within kinds use simple names (e.g., `Container`) for backward compatibility

The extension is fully backwards compatible - existing schemas without `components:` section work exactly as before.

## Next Steps

This enables the next phases of the activator refactoring:
- **MIR-435**: Introduce SandboxPool entity with `sandbox_spec` field
- **MIR-436**: Extract SandboxPoolManager controller to manage pool lifecycle

## Summary by CodeRabbit

- **New Features**
  - Added first-class standalone components to schema generation with component-aware encode/decode, support for single/multiple references, nested inline components, and namespaced identifiers while preserving entity-only helpers for non-component kinds.

- **Tests**
  - Added extensive tests validating standalone components, component references (single/many), nesting and naming isolation, enums namespacing, and encode/decode/empty behaviors.

### DIFF
--- a/pkg/entity/cmd/schemagen/generator.go
+++ b/pkg/entity/cmd/schemagen/generator.go
@@ -19,10 +19,11 @@ const (
 )
 
 type schemaFile struct {
-	Domain  string                 `yaml:"domain"`
-	Version string                 `yaml:"version"`
-	Major   string                 `yaml:"kind-major"`
-	Kinds   map[string]schemaAttrs `yaml:"kinds"`
+	Domain     string                 `yaml:"domain"`
+	Version    string                 `yaml:"version"`
+	Major      string                 `yaml:"kind-major"`
+	Components map[string]schemaAttrs `yaml:"components"`
+	Kinds      map[string]schemaAttrs `yaml:"kinds"`
 }
 
 type schemaAttrs map[string]*schemaAttr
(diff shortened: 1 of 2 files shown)


evanphx (approved):
gif

PR #243, opened 2025-10-10
Optimize Docker build in release workflow (20+ min → 2-5 min)

phinze:
## Summary
Significantly speeds up the `build-and-push-docker` job in the release workflow by eliminating redundant builds and downloads.

### 1. Use pre-built artifacts instead of rebuilding
- The `package` job already builds miren binaries with all runtime dependencies (containerd, runc, nerdctl, etc.)
- Download these pre-built artifacts and extract them before the Docker build
- Copy the extracted binaries directly into the Docker image instead of building from source

### 2. Eliminate redundant downloads
- Removed slow `curl` downloads for containerd, runc, and nerdctl from Dockerfile
- Removed the entire Go build stage from the multi-stage Dockerfile
- All binaries now come from the pre-built packages

## Performance Impact
- **Before:** 20+ minutes for Docker build
- **After:** ~2-5 minutes (just copying binaries + installing apt packages)
- **Savings:** ~15-18 minutes per release

## Testing
- [ ] Verify Docker image builds successfully for both amd64 and arm64
- [ ] Confirm all expected binaries are present in the final image
- [ ] Test that the miren server starts correctly from the built image

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow to download prebuilt amd64/arm64 binaries and stage them for Docker builds, streamlining the build-and-push process.

* **Refactor**
  * Simplified Docker image build to use prebuilt binaries instead of building in-container.
  * Removed runtime downloads and builder stages, improving build speed and reliability across architectures.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,6 +76,24 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Download amd64 package
+        uses: actions/download-artifact@v4
+        with:
+          name: miren-base-linux-amd64
+          path: artifacts/amd64
+
+      - name: Download arm64 package
+        uses: actions/download-artifact@v4
+        with:
+          name: miren-base-linux-arm64
+          path: artifacts/arm64
+
+      - name: Extract pre-built binaries
+        run: |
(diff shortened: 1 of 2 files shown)


evanphx (approved):
gif

PR #246, opened 2025-10-15
Add SandboxPoolManager Controller

phinze:
**Stacked on:** #244 (MIR-435)

### Summary

Adds the SandboxPoolManager controller that reconciles SandboxPool entities by ensuring actual sandbox counts match desired instance counts.

### What This Does

- Watches all SandboxPool entities cluster-wide
- For each pool:
  - Counts actual sandboxes (filtered by service + version)
  - Scales up when `actual < desired` by creating Sandbox entities
  - Updates pool status (`current_instances` and `ready_instances`)
- Runs on Coordinator (global, not per-node)
- Created sandboxes are picked up by existing Scheduler → distributed to nodes

### What This Doesn't Do Yet

- Scale down (logged but not implemented)
- Consume pool sandboxes in activator (still creates directly)
- Provide any user-facing functionality

**User Impact:** Zero. Pools are created but manage nothing yet - all activator behavior unchanged.

### Tests

6 comprehensive tests covering:
- Scale up from zero and partial scaling
- Service isolation between pools
- Version filtering
- Status updates and optimizations

All tests passing using in-memory entity server.

### Next Steps

Phase 4 will make the activator consume sandboxes from pools instead of creating them directly.

## Summary by CodeRabbit

* **New Features**
  * Added a Sandbox Pool Manager that continuously reconciles sandbox pools: creates sandboxes to meet desired counts, handles partial successes, respects service/version isolation, updates pool status (current/ready), and is exposed at server startup for discovery by other components.

* **Tests**
  * Added comprehensive tests covering scale-up, partial creation, service/version isolation, status updates, and no-op scenarios.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -592,6 +592,9 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 
 	aa := co.Activator()
 
+	// Create SandboxPoolManager to reconcile pool entities
+	spm := co.SandboxPoolManager()
+
 	// HttpRequestTimeout is already validated by cfg.Validate() to be >= 1
 	// No need for additional checks here
 
@@ -601,6 +604,7 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 	hs := httpingress.NewServer(ctx, ctx.Log, ingressConfig, client, aa, &httpMetrics)
 
 	reg.Register("app-activator", aa)
+	reg.Register("sandbox-pool-manager", spm)
 	reg.Register("resolver", res)
 
 	rcfg, err := co.NamedConfig("runner")
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -26,6 +26,7 @@ import (
 	"miren.dev/runtime/clientconfig"
 	"miren.dev/runtime/components/activator"
 	"miren.dev/runtime/components/netresolve"
+	"miren.dev/runtime/controllers/sandboxpool"
 	"miren.dev/runtime/metrics"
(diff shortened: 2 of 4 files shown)


evanphx (approved):
Not critical, but curious why this doesn't use the controller abstraction in ./pkg/controller.

phinze:
> Not critical, but curious why this doesn't use the controller abstraction in ./pkg/controller.

Good Q! We initially avoided it due to a semantic mismatch, but on revisiting it we're planning to update the framework to support this better.

## The original issue

The `pkg/controller` abstraction (specifically `GenericController` + `AdaptController`) is designed for **1:1 entity-to-resource mappings**, but SandboxPoolManager has a **1:N relationship**:

**Good fit (SandboxController):**
- 1 Sandbox entity → 1 container
- `Create(sandbox)` → create container
- `Delete(id)` → delete container

**Awkward fit (SandboxPoolManager):**
- 1 SandboxPool entity → **N** Sandbox entities
- We're doing **declarative reconciliation** (ensure desired state), not imperative operations
- Both `Create()` and `Update()` would just call the same `reconcile()` function

## The plan

We're going to add a `ReconcileController` interface to `pkg/controller` that's designed for aggregate state management (1:N relationships). This will give us:

- Work queues + parallel workers (important as we scale to dozens of apps)
- Consistent retry/restart logic
- Standard observability hooks
- A clean `Reconcile(ctx, pool, meta)` method instead of awkward Create/Update split

Once that's in place, we'll refactor SandboxPoolManager to use it. This will help w/ our broader goal of having consistent patterns for reactive management across the entity system.

Filed as MIR-456

PR #249, opened 2025-10-20
fix: verify sandbox network connectivity during recovery

phinze:
## Problem

After miren restarts, recovered sandboxes may have broken network state where they cannot respond to ARP requests, even though the container is still running. This causes traffic failures as the host cannot learn the sandbox's MAC address.

### Observed Failure Pattern

```
01:43:01  Miren restarts, recovers sandbox at 10.8.81.29
01:43:01-03  Initial ARP requests fail → brief connection failures
01:43:05-47:39  Stale ARP cache entry allows traffic to work (~4 minutes)
01:47:40+  Kernel marks ARP entry as FAILED → total traffic lockup
```

### Evidence

From failure inspection:
```
10.8.81.29 FAILED  (recovered sandbox - no MAC address)
10.8.81.3  lladdr ca:26:c9:d1:99:8e REACHABLE  (fresh sandbox - working)
10.8.81.5  lladdr 7a:66:f2:48:9f:e3 REACHABLE  (fresh sandbox - working)
```

All veth interfaces are UP, but the recovered sandbox doesn't respond to ARP.

## Solution (Band-Aid Fix)

This PR adds network verification during sandbox recovery:
1. Ping the sandbox to verify it responds to ARP and ICMP
2. If verification fails, mark the sandbox as DEAD
3. Let the system create a fresh sandbox with working network state

This prevents prolonged outages by ensuring only network-healthy sandboxes are recovered.

## Root Cause Investigation (TODO)

This is a band-aid fix. We still need to understand **why** recovered sandboxes can't respond to ARP:

Possible causes:
- Gvisor network stack not properly restored after containerd/miren restart
- Network namespace disconnected from veth pair
- Sandbox's eth0 interface DOWN or misconfigured
- Something in CNI or bridge state not restored

## Testing

- [ ] Test with apps using auto concurrency mode
- [ ] Test with apps using fixed concurrency mode  
- [ ] Verify fresh sandboxes get created when recovery fails
- [ ] Monitor logs for network verification failures
- [ ] Confirm no prolonged outages after restart

## Notes

- The mirendev app (fixed concurrency) appears unaffected, possibly because it doesn't rely on recovered sandboxes as heavily
- The cloud app (auto concurrency) consistently fails after restart
- This suggests a timing or usage pattern difference that we should investigate

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox recovery process now includes automatic network health verification. During recovery and startup, unreachable sandboxes are identified and skipped automatically, preventing failed operations on unavailable instances and improving overall system reliability.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"log/slog"
 	"math/rand/v2"
+	"net"
+	"os/exec"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


phinze:
This was better handled by subsequent PRs to setup gateway and DNS for recovered sandboxes

PR #256, opened 2025-10-22
Integrate activator with SandboxPoolManager

phinze:
This PR integrates the activator component with SandboxPoolManager, establishing a clean separation of concerns between traffic-driven lease management and declarative pool-based capacity management.

  ## Architecture

  The system now uses SandboxPool entities as the coordination layer between activator and pool manager:

  - **Activator** (app/traffic-aware): Manages leases for specific app versions and services. When capacity is needed, increments `SandboxPool.DesiredInstances` and waits for sandboxes to become available. Tracks lease activity via in-memory `lastRenewal` timestamps and syncs to `Sandbox.LastActivity` every 30s.

  - **SandboxPoolManager** (container lifecycle): Watches SandboxPool entities and reconciles actual sandbox count to match `DesiredInstances`. Monitors `Sandbox.LastActivity` and scales down idle sandboxes based on per-service concurrency configuration.

  - **Concurrency abstraction**: Strategy pattern encapsulates auto-scaling vs fixed-instance behavior, providing unified lease tracking and capacity management regardless of mode.

  ## Key Changes

  ### Pool-Based Capacity Management

  - Activator no longer creates sandboxes directly - instead creates/updates SandboxPool entities with desired capacity
  - SandboxPoolManager provisions sandboxes to match pool's DesiredInstances
  - Background scale-down monitor retires idle sandboxes based on LastActivity age and service-specific ScaleDownDelay
  - Proactive background monitoring prevents relying solely on reconciliation triggers

  ### Concurrency Handling

  - Refactored from global concurrency settings to per-service configuration (`AppVersion.Config.Services[].ServiceConcurrency`)
  - ConcurrencyTracker and ConcurrencyStrategy provide mode-agnostic lease management
  - Handles both auto-scaling (requests-per-instance) and fixed-instance modes with unified interface

  ### Sandbox Data Model Migration

  - Migrated from legacy top-level fields (Container, Version, LogEntity, etc.) to immutable `Sandbox.Spec` field
  - Spec field is now the single source of truth for sandbox configuration
  - Boot-time migration converts existing sandboxes to Spec format
  - Fixed protocol enum mapping to use correct fully-qualified paths

  ### Lease Activity Tracking

  - Level-triggered approach: background task syncs in-memory `lastRenewal` → `Sandbox.LastActivity` every 30s
  - Ensures HTTPIngress lease renewals are visible to scale-down logic
  - Throttled entity updates prevent etcd chatter while maintaining accurate idle detection

  ### Concurrency Improvements

  - RWMutex for activator allows concurrent capacity scanning with exclusive writes
  - Sentinel pattern prevents duplicate pool creation races during concurrent scale-up
  - Fixed lease acquisition race in ConcurrencyTracker via proper synchronization

  ## Testing

  All tests pass with updated coverage for:
  - Activator lease operations, recovery, and concurrent access patterns
  - SandboxPoolManager scale-up/down, pool isolation, and label filtering
  - Protocol enum mapping for legacy sandbox migration
  - Concurrency tracker race conditions

  ## Related

  MIR-436

## Summary by CodeRabbit

* **New Features**
  * Node-level port (node_port) support for container networking.
  * Sandbox pool manager with pool-backed activator, background auto-scaling and idle scale-down monitoring.
  * Per-service concurrency settings replacing the previous global concurrency.

* **Bug Fixes**
  * Legacy-to-spec sandbox migration to preserve existing sandboxes during rollout.

* **Chores**
  * Migration to a nested Spec-based sandbox representation; tests and startup flows updated accordingly.


components/activator/activator.go
```diff
@@ -258,193 +149,168 @@ func (a *localActivator) AcquireLease(ctx context.Context, ver *core_v1alpha.App
...
-	a.log.Debug("watching sandbox", "app", ver.App, "sandbox", pr.Id())
+		_, err := a.eac.Put(ctx, &rpcE)
+		if err != nil {
+			return nil, fmt.Errorf("failed to update pool: %w", err)
+		}
 
-	var localErr error
+		a.log.Info("incremented pool capacity", "pool", existingPool.ID, "desired_instances", existingPool.DesiredInstances)
+		return existingPool, nil
+	}
```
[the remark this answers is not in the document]
phinze: The optimistic concurrency locking system needs a holistic review, tracked in MIR-219. We'll hold off on interacting with that layer in this effort, and cover it in its own sweep.

controllers/sandboxpool/manager.go
```diff
@@ -245,3 +424,140 @@ func (m *Manager) updatePoolStatus(ctx context.Context, pool *compute_v1alpha.Sa
...
+			).Attrs())
+
+			if _, err := m.eac.Put(ctx, &rpcE); err != nil {
+				return fmt.Errorf("failed to update pool desired instances: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
phinze: The optimistic concurrency locking system needs a holistic review, tracked in MIR-219. We'll hold off on interacting with that layer in this effort, and cover it in its own sweep.

components/activator/activator.go
```diff
@@ -110,341 +119,288 @@ func NewLocalActivator(ctx context.Context, log *slog.Logger, eac *entityserver_
...
+		// Check if pool exists or is being created (read lock)
+		a.mu.RLock()
+		state, exists := a.pools[key]
+		a.mu.RUnlock()
+
+		if exists {
+			// If creation is in progress, wait for it to complete
+			if state.inProgress {
+				a.log.Debug("pool creation in progress, waiting", "service", service)
+				<-state.done
```
evanphx: We should select on ctx.Done here too, so we don't hang on a shutdown under unwell conditions.

components/activator/activator.go
```diff
@@ -110,341 +119,288 @@ func NewLocalActivator(ctx context.Context, log *slog.Logger, eac *entityserver_
...
+				// Check if creation succeeded or failed
+				if state.err != nil {
+					return nil, fmt.Errorf("pool creation failed: %w", state.err)
+				}
+				// Creation succeeded, fall through to increment logic below
+				a.mu.RLock()
+				state = a.pools[key]
+				a.mu.RUnlock()
```
evanphx: In this case, should we always be incrementing desired instances? Seems like if we were racing to get a pool started, we should try the whole logic again here rather than assuming we need another instance.

components/activator/activator.go
```diff
@@ -534,195 +454,245 @@ func (a *localActivator) watchSandboxes(ctx context.Context) {
...
-		// Extract service from sandbox labels - default to "web" if not found
-		var metadata core_v1alpha.Metadata
-		metadata.Decode(ent.Entity())
-		service := getLabel(&metadata, "service", "web") // Default to web if not found
+		// Extract service from sandbox labels
+		var md core_v1alpha.Metadata
+		md.Decode(ent.Entity())
+		service, _ := md.Labels.Get("service")
+		if service == "" {
+			a.log.Warn("sandbox missing service label", "sandbox", sb.ID)
```
evanphx: Don't warn here, there could be lots of different kinds of sandboxs in the system other than ones the activator is managing.

evanphx (approved):
Looks good! A couple minor things. lots and lots of mutex juggling which makes me a little itchy, but I think the saga stuff will help there.

phinze:
> lots and lots of mutex juggling which makes me a little itchy, but I think the saga stuff will help there.

I spent a lot of time itching myself, and did quite a bit of analysis around this to prep for talking about it with the team... bear with me for a context dump! 🚚 

On the tail end of this refactor I dug into the mutex usage to understand if this is a code smell or just inherent complexity.

**Short answer: It's inherent to what the activator does, and we're actually in better shape than before the refactor.**

### What's the activator doing?

The activator is fundamentally a coordination point between multiple concurrent goroutines. You've got:
- Many request threads hitting `AcquireLease` simultaneously (the hot path)
- A background `watchSandboxes` thread discovering new sandboxes from etcd
- A background `syncLastActivity` thread updating the entity store every 30s

All of these need to read and update the same shared state (which sandboxes exist, what their capacity is, who's waiting for new ones). This is exactly the kind of problem mutexes solve.

### Did the refactor make it worse?

Lock count went from ~15-20 operations to ~22. The increase comes from three new responsibilities:

1. **Pool management** - The sentinel pattern in `ensureSandboxPool` prevents duplicate pool creation races. It's complex (7 lock/unlock pairs) but correct.

2. **Channel notifications** - We need to track waiting requests and notify them when sandboxes appear. This requires locks when registering/unregistering channels and when sending notifications.

3. **More shared state** - We now track `pools` and `newSandboxChans` in addition to `versions`.

But we also got better in some ways. We moved to an `RWMutex` (allows concurrent reads on the hot path) and kept all critical sections short. The double-check pattern prevents TOCTOU bugs, and the sentinel pattern prevents race conditions in pool creation. Race detector is happy, no deadlocks.

### Could we do it differently?

I looked at a few alternatives:

An **actor model** (single goroutine owns state, everything goes through channels) would actually be slower - the hot path needs many concurrent readers, and an actor would serialize everything.

**Lock-free data structures** (atomics and CAS loops) might help for the capacity counters, but we'd still need locks for the map updates. The added complexity probably isn't worth it.

**Per-key mutexes** (separate lock per app/version) would reduce contention if we had tons of apps, but that's premature optimization without profiling data showing it's actually a bottleneck.

### Where does saga fit in?

The saga pattern could definitely help with `ensureSandboxPool` specifically - that's the most complex locking code (the sentinel pattern state machine). Instead of manually managing locks through the different states (check existence, create sentinel, build spec, create pool, replace sentinel), a saga would make that state machine explicit.

But saga won't eliminate locks from the hot path. `AcquireLease` still needs to quickly scan available sandboxes and grab capacity, and that's always going to need some form of synchronization. The saga would really just clean up the pool creation flow.

### What should we do?

I think the right next step is to add clear comments explaining the three main patterns (double-check, sentinel, channel notification) and why they're necessary. That way when folks read the code, they understand it's intentional complexity rather than accidental.

For the saga evaluation, my take is: wait until the saga framework is ready, then try it specifically for `ensureSandboxPool`. That's where it would help most. But I wouldn't expect it to reduce the overall lock count significantly - just make that one complex piece clearer.

If we're worried about actual contention (not just visual complexity), we should add pprof mutex profiling in production first. That'll tell us if this is a real bottleneck or just looks busy.

Thoughts?

api/core/core_v1alpha/extra.go
```diff
@@ -1,9 +1,25 @@
...
+// Returns an error if the service is not found - all service configs should be
+// hydrated with defaults during app version creation.
+func GetServiceConcurrency(ver *AppVersion, serviceName string) (*ServiceConcurrency, error) {
+	for _, svc := range ver.Config.Services {
+		if svc.Name == serviceName {
+			return &svc.ServiceConcurrency, nil
+		}
+	}
+	return nil, fmt.Errorf("service %q not found in version config (services should be hydrated with defaults)", serviceName)
+}
```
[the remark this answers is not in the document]
phinze: Callers are meant to be read-only... should we just make this return a value instead?

components/activator/activator.go
phinze: Done in b630ea327ea4763eeb4d4ba297452893afade0f4

components/activator/activator.go
phinze: Done in b630ea327ea4763eeb4d4ba297452893afade0f4

components/activator/activator.go
phinze: Done in b630ea327ea4763eeb4d4ba297452893afade0f4

api/core/core_v1alpha/extra.go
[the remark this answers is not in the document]
phinze: Decided to go for that change - done here 71145a2c58f1ab37dcf53553f3ac64460584863e

components/activator/activator.go
```diff
@@ -534,195 +487,245 @@ func (a *localActivator) watchSandboxes(ctx context.Context) {
...
-		metadata.Decode(ent.Entity())
-		service := getLabel(&metadata, "service", "web") // Default to web if not found
+		// Extract service from sandbox labels
+		var md core_v1alpha.Metadata
+		md.Decode(ent.Entity())
+		service, _ := md.Labels.Get("service")
+		if service == "" {
+			// Skip sandboxes without service label (e.g., buildkit, other non-app sandboxes)
+			return nil
+		}
```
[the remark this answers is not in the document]
phinze: Done in d6b7dcb7999d85d818cac30b37a9650cd8bf4907

evanphx:
Thanks for the context dump! Makes total sense. Sometimes the code is inherently locky! Let's keep going down the road and find next place to pause and we can take a look to see if there is a way to improve it, but no need to do anything different today.

PR #257, opened 2025-10-22
nbd: cleanup large request size handling

evanphx:
So, turns out that nbd servers don't have a way to tell the client to clamp the maximum payload size. And the linux nbd ioctl/netlink interfaces don't expose the ability to do that either.

So instead we need to honor the clients size request up to some insane maximum.

Oh, but also, the code was _wrong_ in that the 3rd field of the block size info is supposed to be the maximum payload size. I don't think any nbd clients actually even look at it, but I've corrected it just for completeness.

We don't actually use nbd.Handle in our implementation because we're using the loopback connection to the kernel, so we go straight to nbd.HandleTransport.

## Summary by CodeRabbit

* **Bug Fixes**
  * Read-only mode now accepts incoming write/trim requests without failing, responding successfully while avoiding backend writes — improves client compatibility.
  * Added a sanity check to reject/guard against excessively large requests, preventing resource exhaustion.

* **Improvements**
  * Simplified block-size negotiation and request sizing behavior to reduce mismatches and improve performance and reliability during storage operations.

### DIFF
--- a/lsvd/pkg/nbd/nbd.go
+++ b/lsvd/pkg/nbd/nbd.go
@@ -18,7 +18,8 @@ var (
 )
 
 const (
-	defaultMaximumRequestSize = 32 * 1024 * 1024 // Support for a 32M maximum packet size is expected: https://sourceforge.net/p/nbd/mailman/message/35081223/
+	defaultMaximumRequestSize = 32 * 1024 * 1024  // Support for a 32M maximum packet size is expected: https://sourceforge.net/p/nbd/mailman/message/35081223/
+	MaxInsaneLength           = 512 * 1024 * 1024 // Reject requests larger than 512M as insane
 )
 
 type Export struct {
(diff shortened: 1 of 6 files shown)


lsvd/pkg/nbd/nbd.go
```diff
@@ -491,42 +502,20 @@ func HandleTransport(log *slog.Logger, conn net.Conn, backend Backend, options *
...
 		case TRANSMISSION_TYPE_REQUEST_WRITEZ:
-			if options.ReadOnly {
-				_, err := io.CopyN(io.Discard, conn, int64(length)) // Discard the write command's data
-				if err != nil {
-					return err
-				}
-
-				if err := sendReply(TRANSMISSION_ERROR_EPERM, handle); err != nil {
+			if !options.ReadOnly {
+				if err := backend.ZeroAt(int64(offset), int64(length)); err != nil {
```
phinze, line 506: This is just a wee cleanup I presume

phinze (approved):
Image

lsvd/pkg/nbd/nbd.go
evanphx: Correct, I realized look at it this code it was wrong. We've never hit this because we aren't using readonly disks atm but felt I should just correct it now.

PR #258, opened 2025-10-23
entity: add read-time migration for sandbox entity.cbor files

evanphx:
## Summary
- Add automatic read-time migration for entity.cbor files in sandbox controller
- Migrate old entity format (struct fields) to new format (attributes)
- Add defensive migration in Entity.UnmarshalCBOR as safety net

## Background
The recent entity system overhaul migrated metadata (ID, Revision, CreatedAt, UpdatedAt) from struct fields to attributes in the CBOR encoding. However, the sandbox controller stores entity.Meta structs in entity.cbor files that may still be in the old format.

## Changes
1. **Added `MigrateMetaFromBytes` in `pkg/entity/migrate.go`**:
   - Tries to decode as old format first
   - Detects migration need by checking if `entity.ID` is set
   - Migrates all metadata fields to attributes
   - Preserves absence of timestamps if they weren't set
   - Falls back to new format if no migration needed

2. **Updated `readEntity` in `controllers/sandbox/sandbox.go`**:
   - Uses `MigrateMetaFromBytes` instead of direct `Decode`
   - Automatically migrates old files on read
   - Files are re-saved in new format when `saveEntity` is called

3. **Added defensive migration in `Entity.UnmarshalCBOR`**:
   - Safety net for any other code paths that deserialize entities directly
   - Logs warnings when migration is triggered
   - Can be removed once all data is confirmed migrated

## Test plan
- [x] Added comprehensive tests in `TestMigrateMetaFromBytes`:
  - Migrates old Meta format with all metadata fields
  - Handles new Meta format without migration
  - Preserves missing timestamps correctly
- [x] All tests pass
- [x] Build succeeds

## Summary by CodeRabbit

* **New Features**
  * Automatic read-time migration of legacy entity/meta formats so older persisted data is transparently upgraded.
  * Controller read paths now use the centralized migration routine to handle older formats and missing legacy fields, emitting warnings when fallback occurs.

* **Tests**
  * Added tests validating migration for old/new formats and cases with missing timestamps.

* **Chores**
  * Test harness uses an isolated data path for coordinator artifacts.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -693,14 +693,13 @@ func (c *SandboxController) readEntity(ctx context.Context, id entity.Id) (*enti
 		return nil, fmt.Errorf("failed to open entity file: %w", err)
 	}
 
-	var meta entity.Meta
-
-	err = entity.Decode(data, &meta)
+	// Use MigrateMetaFromBytes for automatic migration from old format
+	meta, err := entity.MigrateMetaFromBytes(data)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode entity file: %w", err)
 	}
 
-	return &meta, nil
+	return meta, nil
 }
 
 func (c *SandboxController) updateSandbox(ctx context.Context, sb *compute.Sandbox, meta *entity.Meta) error {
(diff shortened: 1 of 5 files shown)


phinze (approved):
TIL entity.cbor files! Will look more when back at desk but here’s a check to unblock

PR #268, opened 2025-10-24
entity: Minor revision cleanup

evanphx:
Don't log when the revisions are different in the detection case, the code is working the way it should. But also, be sure to update the sandbox version when you do an update!

Lastly, because we use the etcd revision as the entity revision, be more careful about not storing the Revision attr into etcd to reduce confusion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Revision field is now exclusively managed by the system and will not accept user-provided values during create, update, replace, and patch operations.
  * Sandbox version label is now applied consistently to avoid mismatches.

* **Chores**
  * Removed debug logging related to sandbox version detection.
  * Optimized internal revision update logic.

* **Configuration**
  * Added an extra cache path to the ISO config to include /root/.cache.

### DIFF
--- a/.iso/config.yml
+++ b/.iso/config.yml
@@ -2,4 +2,6 @@ privileged: true
 workdir: /src
 volumes:
   - /data
+cache:
   - /go/pkg
+  - /root/.cache
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -549,12 +549,10 @@ func (c *SandboxController) checkSandbox(ctx context.Context, co *compute.Sandbo
 	}
 
 	if _, ok := labels[sandboxVersionLabel]; !ok {
-		c.Log.Debug("sandbox version label not found, assuming new sandbox")
 		return differentVersion, nil
 	}
 
 	if labels[sandboxVersionLabel] != fmt.Sprint(meta.Revision) {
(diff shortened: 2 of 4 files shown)


phinze (approved):
Seems like a legit test failure, but here's a check on the code to unblock you in case you find time to bang on this:

```
104 : [6m26s] | 22:39:15.384 [ERROR] coordinator.etcdstore failed to update entity in etcd │ error: <nil> id: sandbox/test-1761345555 rev: 0 server-rev: 513
104 : [6m26s] | 22:39:15.384 [ERROR] failed to assign sandbox │ error: "remote error: generic unknown: failed to update entity in put: conflict in entity: sandbox/test-1761345555" sandbox: sandbox/test-1761345555
104 : [6m26s] | 22:39:15.384 [ERROR] failed to assign sandbox │ error: "remote error: generic unknown: failed to update entity in put: conflict in entity: sandbox/test-1761345555" sandbox: sandbox/test-1761345555
104 : [6m26s] |     integration_test.go:153: 
104 : [6m26s] |         	Error Trace:	/src/components/runner/integration_test.go:153
104 : [6m26s] |         	Error:      	Timeout waiting for container creation
104 : [6m26s] |         	Test:       	TestRunnerCoordinatorIntegration
```

PR #278, opened 2025-10-27
MIR-466: Migrate metrics from ClickHouse to VictoriaMetrics

evanphx:
## Issue

Closes MIR-466

## Summary

Migrates metrics storage from ClickHouse to VictoriaMetrics, a high-performance time-series database optimized for metrics workloads.

### VictoriaMetrics Integration
- Added `VictoriaMetricsComponent` for lifecycle management
  - Supports embedded and external deployment modes
  - Configurable HTTP port, data path, and retention period
  - Graceful shutdown with SIGTERM/SIGKILL handling

- Implemented `VictoriaMetricsWriter` for metric ingestion
  - Buffered writes (1000 points, 5s flush interval)
  - Automatic metric name sanitization
  - Concurrent write safety
  - Graceful shutdown flushes remaining metrics

- Implemented `VictoriaMetricsReader` for querying
  - Instant and range query support
  - MetricsQL query language
  - Helper methods for common patterns

### Metrics Updates
- Updated `CPUUsage`, `MemoryUsage`, and `HTTPMetrics` to use VictoriaMetrics
- Made VictoriaMetrics dependencies optional with asm tags
- Preserved existing metrics API for backward compatibility

### Testing
- **47 test cases** covering writer, reader, and integration scenarios
- Mock VictoriaMetrics/VictoriaLogs HTTP servers for test isolation
- Updated sandbox tests to use mock infrastructure
- Removed legacy ClickHouse verification from observability tests

### Configuration
- Added `victoriametrics` configuration to server config schema
- Updated server startup to initialize VictoriaMetrics component
- Updated Dockerfile and config for VictoriaMetrics image

## Test Results

✅ All sandbox tests pass
✅ All observability tests pass
✅ All metrics tests pass (47 test cases)
✅ Linter: 0 issues

## Test Plan

- [x] Unit tests for VictoriaMetrics writer/reader
- [x] Integration tests for component lifecycle
- [x] Sandbox tests with CPU usage and log reattachment
- [x] Observability tests for log write/read
- [x] Linter verification

## Summary by CodeRabbit

* **New Features**
  * VictoriaMetrics added as the primary metrics backend (embedded or external) and VictoriaLogs as the logs backend; configurable address, port, timeout and retention exposed.

* **Chores**
  * Replaced ClickHouse with VictoriaMetrics/VictoriaLogs across runtime, config, defaults, and packaging; updated development/docker/iso scaffolding and docs.

* **Tests**
  * Extensive unit and integration tests for VictoriaMetrics writer, reader, component lifecycle, and end-to-end metrics/logs flows.


phinze (approved):
Haven't had a chance to manually test it yet (per make dev tweaks in slack), but looking good overall!

## Gauges vs Histograms for Latency Metrics

Quick note on the approach here - we're storing latency as raw gauge values and using `quantile_over_time()` to calculate P95/P99. This works and maintains parity with the old ClickHouse implementation (which also used raw values with `quantile()`).

**The traditional Prometheus way** would be histograms with pre-bucketed values:
- Pro: Constant storage (fixed buckets), aggregates correctly across instances, faster queries
- Con: Approximate percentiles, need to pick buckets upfront

**Our current approach** (gauges + quantile_over_time):
- Pro: Exact percentiles, simpler implementation, flexible
- Con: More storage (one sample per request), queries can slow down at high volume (>1000 req/s), doesn't aggregate well across multiple instances

**Bottom line:** This is totally fine for the migration - keeps things simple and maintains behavioral parity. We should just keep an eye on storage growth and query performance. If we start seeing issues at scale, histograms would be the next evolution.

The main thing to watch for is **path cardinality** - if we're getting tons of unique paths like `/user/123`, `/user/456`, etc., we can consider normalizing them to patterns like `/user/:id` to avoid metric explosion.

**References:**
- [VictoriaMetrics: quantile_over_time](https://docs.victoriametrics.com/metricsql/#quantile_over_time) (what we're using)
- [VictoriaMetrics: histogram_quantile](https://docs.victoriametrics.com/metricsql/#histogram_quantile) (alternative approach)
- [Prometheus: Histograms and Summaries](https://prometheus.io/docs/practices/histograms/)

phinze (approved):
We walked through this together live and while there are a few display/windowing hiccups it seems to be recording well enough to merge and we can follow up with improvements from here.

PR #290, opened 2025-11-03
Implement key reuse in login to prevent server-side key proliferation

evanphx:
Add key management system to store and reuse authentication keys across login attempts. Previously, each login created a new keypair and registered it with the server, leading to key pollution when users repeatedly logged in.

Changes:
- Add KeyConfig to clientconfig for storing reusable keys separately
- Add KeyRef to IdentityConfig to reference keys instead of embedding them
- Store keys in clientconfig.d/key-{name}.yaml for reuse across identities
- Add --key-name CLI option (default: "miren-cli") for custom key names
- Implement getOrCreateKey() to check for existing keys before generating
- Update key consumers to resolve keys via GetPrivateKeyPEM() helper
- Maintain backward compatibility with legacy direct PrivateKey fields
- Add comprehensive test suite with 11 tests covering all key operations

Keys are now identified by name (default "miren-cli") and reused on subsequent logins, reducing server-side storage and preventing duplicate registrations.

## Summary by CodeRabbit

* **New Features**
  * Introduced named cryptographic key management allowing keys to be created, reused, and referenced across multiple identities. Keys now include fingerprint tracking and metadata for improved organization and lifecycle management within configurations.

### DIFF
--- a/cli/commands/cluster_add.go
+++ b/cli/commands/cluster_add.go
@@ -65,7 +65,7 @@ func ClusterAdd(ctx *Context, opts struct {
 	if opts.Cluster == "" && opts.Address == "" {
 		ctx.Info("Fetching available clusters from identity server...")
 
-		clusters, err := fetchAvailableClusters(ctx, identity)
+		clusters, err := fetchAvailableClusters(ctx, mainConfig, identity)
 		if err != nil {
 			return fmt.Errorf("failed to fetch available clusters: %w", err)
 		}
--- a/cli/commands/config_bind.go
+++ b/cli/commands/config_bind.go
@@ -74,7 +74,7 @@ func ConfigBind(ctx *Context, opts struct {
 	if opts.Cluster == "" && opts.Address == "" {
 		ctx.Info("Fetching available clusters from identity server...")
 
-		clusters, err := fetchAvailableClusters(ctx, identity)
+		clusters, err := fetchAvailableClusters(ctx, mainConfig, identity)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


phinze (approved):
LGTM!

Few observations as I was walking through this:
- Keys never expire (like SSH keys, users manually revoke)
- Key-per-machine model means users accumulate keys as they login from new machines (capped at 10)
- No CLI commands for key management yet

Things for the backlog:
- Add miren keys list/revoke to CLI and/or Web UI (looks like we have API for revoke but no UI wired yet)

Nice to haves:
- Would be nice to include hostname in default key name: `miren-cli@laptop`
- Maybe support optional key expiry? (`--key-ttl 90d`)

This is a solid next step, good to ship! :shipit:

teresamychu (approved):
Excellent! Thanks for looking into this!

evanphx:
@phinze The initial version had the hostname on the key but I was worried the users would think we're constraining the key? Happy to add it back though.

phinze:
> The initial version had the hostname on the key but I was worried the users would think we're constraining the key?

Ah interesting. Yeah I think in general it's nicer to have more metadata instead of just accumulating a long list of "terraform cli" keys that you're left guessing where they came from.

I guess it comes down to whether we want to guide users towards shipping around private keys or making new ones per machine. I'd lean towards the latter myself, with endgame pie in the sky being (optional) un-exportable secure enclave private keys.

evanphx:
I gotcha. Let me do a follow up that lets us attach some metadata to keys. And agree, eventually we'll want to move towards a secure enclave of keys!

PR #292, opened 2025-11-04
Remove miren.dev email address when getting a cert

evanphx:
Let's Encrypt doesn't even do anything with the email addresses anymore anyway.

## Summary by CodeRabbit

* **Chores**
  * Modified how contact information is handled during automatic TLS certificate provisioning. Email will no longer be automatically supplied during ACME certificate initialization.


phinze (approved):
Ah nice this is good - I was a little concerned about racking up rate limits against this email. We'll probably end up needing to research this a bit more for shared instances... the failure case for cert allocation errors is pretty gross right now.

evanphx:
Now that Let's Encrypt no longer sends expiration notification emails, there is little use in sending them an email anyway it seems.

PR #295, opened 2025-11-04
Add metadata support to key configuration

evanphx:
## Summary

Adds generic metadata support to keys in the clientconfig, enabling better key lifecycle management and tracking.

## Changes

- Added `Metadata map[string]string` field to `KeyConfig` struct
- When generating new keys during login, automatically populate metadata with:
  - `hostname`: The hostname where the key was created
  - `created_at`: ISO 8601 timestamp in UTC

## Benefits

- **Generic design**: The metadata map can be extended with additional fields without schema changes
- **Backward compatible**: Existing keys without metadata continue to work (`omitempty`)
- **Trackable**: Users can now see when and where each key was created
- **Extensible**: Future features can add more metadata (e.g., `last_used`, `description`)

## Test Plan

- [x] All existing tests pass
- [x] Verified metadata is correctly populated on new key generation
- [x] Confirmed backward compatibility with existing keys

## Example

Generated keys are saved to `~/.config/miren/clientconfig.d/key-<name>.yaml`:

```yaml
keys:
  miren-cli:
    name: miren-cli
    type: ed25519
    fingerprint: SHA256:abc123...
    metadata:
      hostname: mycomputer.local
      created_at: "2025-11-04T18:30:00Z"
```

## Release Notes

* **New Features**
  * Authentication keys now automatically capture and store metadata, including hostname and creation timestamp, for improved credential tracking and management.

### DIFF
--- a/cli/commands/login.go
+++ b/cli/commands/login.go
@@ -535,6 +535,18 @@ func saveKeyPairToConfig(identityName, cloudURL string, keyPair *cloudauth.KeyPa
 
 	// Save the key separately (if it doesn't already exist)
 	if !mainConfig.HasKey(keyName) {
+		// Get hostname for metadata
+		hostname, err := os.Hostname()
+		if err != nil {
+			hostname = "unknown"
+		}
+
+		// Create metadata with hostname and creation timestamp
+		metadata := map[string]string{
+			"hostname":   hostname,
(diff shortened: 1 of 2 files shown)


phinze (approved):
Love it!

PR #296, opened 2025-11-04
login: always register a key

evanphx:
Even if we find a key to reuse on disk, we always need to register it against the account. This allows keys to be reused across accounts, which we're considering a feature.

## Summary by CodeRabbit

* **Bug Fixes**
  * Login flow now treats existing and newly created authentication keys identically, preventing erroneous failures.
  * Key registration treats "already-registered" as non-error, reducing duplicate or confusing messaging.

* **Tests**
  * Updated tests to reflect the simplified key creation/registration behavior and removed obsolete checks.

### DIFF
--- a/cli/commands/login.go
+++ b/cli/commands/login.go
@@ -75,39 +75,39 @@ type KeyRegistrationResponse struct {
 
 // getOrCreateKey checks if a key with the given name already exists in the config,
 // and reuses it if found. Otherwise, it generates a new key.
-// Returns the keypair, whether it was already registered, and any error.
-func getOrCreateKey(ctx *Context, keyName string) (*cloudauth.KeyPair, bool, error) {
+// Returns the keypair and any error.
+func getOrCreateKey(ctx *Context, keyName string) (*cloudauth.KeyPair, error) {
 	// Try to load existing config
 	config, err := clientconfig.LoadConfig()
 	if err != nil && err != clientconfig.ErrNoConfig {
-		return nil, false, fmt.Errorf("failed to load config: %w", err)
+		return nil, fmt.Errorf("failed to load config: %w", err)
 	}
 
 	// Check if key already exists
 	if config != nil && config.HasKey(keyName) {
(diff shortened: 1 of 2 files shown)


cli/commands/login.go
```diff
@@ -76,38 +76,38 @@ type KeyRegistrationResponse struct {
 // getOrCreateKey checks if a key with the given name already exists in the config,
 // and reuses it if found. Otherwise, it generates a new key.
 // Returns the keypair, whether it was already registered, and any error.
```
phinze: 💡 Should make this match the new sig

cli/commands/login_keys_test.go
```diff
@@ -288,19 +285,17 @@ func TestKeyManagementIntegration(t *testing.T) {
...
 		// Step 2: Save the key to config
 		err = saveKeyPairToConfig(identityName, cloudURL, keyPair1, keyName)
 		require.NoError(t, err)
 
 		// Step 3: Simulate a new login attempt - should reuse the key
-		keyPair2, alreadyRegistered2, err := getOrCreateKey(ctx, keyName)
+		keyPair2, err := getOrCreateKey(ctx, keyName)
 		require.NoError(t, err)
-		assert.True(t, alreadyRegistered2, "Should reuse existing key")
 		assert.Equal(t, fingerprint1, keyPair2.Fingerprint(), "Should have same fingerprint")
```
phinze, line 299: ❓ I'm sort of confused by the test changes... I'd presume to see a changed assertion or a new test for the behavior change, but I think there's just one removed assertion?

phinze (approved):
Change looks good overall. Two inline nitty nits.

cli/commands/login.go
evanphx: We should!

cli/commands/login_keys_test.go
evanphx: The tests are still valid even though we're not indicating if it's already registered.

PR #320, opened 2025-11-10
Wire LSVD disks to fixed count services in sandboxes

evanphx:
## Summary

This PR enhances the disk lease controller to better handle disk provisioning states and adds event-driven reconciliation. It also consolidates volume providers by removing the old LSVD-based implementation.

### Disk Lease Controller Improvements
- Modified `handlePendingLease()` to wait for disk provisioning instead of failing immediately when disk is in PROVISIONING state
- Added `watchDiskChanges()` to reconcile dependent leases when disk state changes
- Updated `Start()` to use 10-second resync period for faster reconciliation
- Added event-driven reconciliation via disk entity watcher
- Added comprehensive tests for provisioning wait behavior

### Volume Provider Consolidation
- Removed old LSVD-based "miren" volume provider implementation
- Renamed disk-lease-based "disk" provider to "miren"
- Updated error messages and cleaned up unused imports
- Removed obsolete tests for old LSVD provider

### Deployment and Debugging Improvements
- Updated `specsMatch()` in `deployment/launcher.go` to return `(string, bool)` with detailed mismatch reasons
- Improves debugging when sandbox pools aren't being reused
- Added new integration test for disk lease functionality

### Configuration Updates
- Added `stop_sandboxes_on_shutdown` flag to server config schema (implementation in separate commit)
- Regenerated config code from schema changes
- Updated API schemas and regenerated code

## Key Benefits

- ✅ Disk leases now wait for provisioning instead of failing prematurely
- ✅ Event-driven reconciliation reduces latency for lease binding (no need to wait for resync period)
- ✅ Consolidated volume providers simplify codebase maintenance
- ✅ Better debugging with detailed spec mismatch reasons
- ✅ Foundation for graceful sandbox shutdown in development

## Test Plan

- [x] Existing tests pass
- [x] Added new test: `TestDiskLeaseController_WaitsForProvisioning`
- [x] Verified disk lease waits for provisioning
- [x] Verified event-driven reconciliation triggers correctly
- [x] Code compiles and passes linter

## Summary by CodeRabbit

* **New Features**
  * Services can attach persistent disks (name, mount path, size, filesystem, read-only, lease timeout).
  * Per-instance disk provisioning/naming for fixed-concurrency services.
  * New server option to stop sandboxes on shutdown.

* **Improvements**
  * Sandbox pool now manages instance-numbered sandboxes for reuse.
  * Containers receive instance env var; improved volume mounting, leasing, change-watching, and clearer pool-reuse diagnostics.
  * Specs comparison now surfaces detailed mismatch reasons.

* **Tests**
  * Added unit and integration tests for disk provisioning, leasing, and sandbox spec behavior.

* **Bug Fixes**
  * Prevented a potential panic during segment flush.


pkg/serverconfig/schema.yml
```diff
@@ -138,6 +138,15 @@ configs:
...
           min: 1
 
+      stop_sandboxes_on_shutdown:
+        type: bool
+        default: false
+        cli:
+          long: stop-sandboxes-on-shutdown
+          description: Stop all sandboxes when server shuts down (useful in development)
+        env: MIREN_SERVER_STOP_SANDBOXES_ON_SHUTDOWN
+        toml: stop_sandboxes_on_shutdown
```
phinze, line 148: Good idea 🤌

controllers/disk/disk_lease_controller.go
```diff
@@ -753,20 +766,84 @@ func (d *DiskLeaseController) CleanupOldReleasedLeases(ctx context.Context) erro
...
+		// Only care about updates (state changes)
+		if op.OperationType() != entityserver_v1alpha.EntityOperationUpdate {
+			return nil
+		}
+
+		diskId := entity.Id(op.EntityId())
+		d.Log.Debug("Disk state changed, finding dependent leases", "disk", diskId)
+
+		// Find all leases that reference this disk
+		listResp, err := d.EAC.List(ctx, entity.Ref(entity.EntityKind, storage_v1alpha.KindDiskLease))
```
phinze: 💡 Leases are indexed on disk_id, so you don't have to list all of them.

controllers/sandbox/volume.go
```diff
@@ -48,130 +55,308 @@ func (c *SandboxController) configureHostVolume(sb *compute.Sandbox, volume comp
...
-	dataPath := filepath.Join(c.DataPath, "miren-volumes", volume.Name, "data")
-	err := os.MkdirAll(dataPath, 0755)
-	if err != nil {
-		return fmt.Errorf("failed to create volume directory: %w", err)
+	c.Log.Info("configuring miren volume",
+		"sandbox", sb.ID,
+		"disk_name", volume.DiskName,
+		"mount_path", volume.MountPath)
+
+	// Append instance number to disk name to ensure each instance gets its own disk
```
phinze, line 98: ❓ Bookmarking this for further conversation. It's unclear at first blush why we want every instance to get its own disk by default.

controllers/sandbox/volume.go
```diff
@@ -48,130 +55,308 @@ func (c *SandboxController) configureHostVolume(sb *compute.Sandbox, volume comp
...
+		"lease", leaseID,
+		"disk_mount_path", diskMountPath)
+
+	// Return the disk mount path so it can be used directly in the container spec
+	return diskMountPath, nil
+}
+
+func (c *SandboxController) ensureDisk(ctx context.Context, diskName string, sizeGB int64, filesystem string) (entity.Id, error) {
+	// Search for existing disk by name
+	listResp, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, storage.KindDisk))
```
phinze: 💡 Disk name is also indexed

controllers/sandbox/volume.go
```diff
@@ -48,130 +55,308 @@ func (c *SandboxController) configureHostVolume(sb *compute.Sandbox, volume comp
...
 		}
+		leaseTimeout = duration
+	}
 
-		sa, err = lsvd.NewS3Access(c.Log, pu.Host, pu.Path, cfg)
-		if err != nil {
-			return fmt.Errorf("failed to create s3 access: %w", err)
-		}
+	// Look up or create Disk entity using instance-specific name
+	diskID, err := c.ensureDisk(ctx, actualDiskName, sizeGB, filesystem)
```
phinze, line 131: 💭 Thinking through race conditions on disk creation... I don't think we need to solve for everything now, but I'm left with high level questions like "what's the behavior we want if a bunch of sandboxes all simultaneously request the `assets` disk"

appconfig/appconfig.go
```diff
@@ -127,42 +138,73 @@ func Parse(data []byte) (*AppConfig, error) {
...
+
+			for i, disk := range svcConfig.Disks {
+				if disk.Name == "" {
+					return fmt.Errorf("service %s: disk[%d] must have a name", serviceName, i)
+				}
+				if disk.MountPath == "" {
+					return fmt.Errorf("service %s: disk[%d] (%s) must have a mount_path", serviceName, i, disk.Name)
+				}
+				if disk.Filesystem != "" && disk.Filesystem != "ext4" && disk.Filesystem != "xfs" && disk.Filesystem != "btrfs" {
+					return fmt.Errorf("service %s: disk[%d] (%s) has invalid filesystem %q, must be ext4, xfs, or btrfs", serviceName, i, disk.Name, disk.Filesystem)
```
phinze, line 198: 💭 IIRC we were also going to validate on tools being available. But I think this is all client-side right now yeah? So would need an RPC-tinged validation to do that. Just something for the backlog.

phinze (approved):
Image

controllers/sandbox/volume.go
evanphx: if we set num_instances to `2` then we'll get an error without this, because we'll try to create one disk and 2 different leases against the same disk.

controllers/sandbox/volume.go
evanphx: we should move the disk creation to much earlier rather than create them here, but I'm opting to do that later.

evanphx: As for requests, the lease system should prevent multiple leases on the same disk.

appconfig/appconfig.go
evanphx: agree, this is just to prevent easy, early footguns.

controllers/sandbox/volume.go
evanphx: Yes! fixing.

controllers/disk/disk_lease_controller.go
evanphx: Yes! Fixing.

PR #324, opened 2025-11-11
Move to a proper one-cell wide spinner

evanphx:
Turns out the paralellograms are in some really weird fonts and can't really be rendered right! So switch to dots which can be.

## Summary by CodeRabbit

* **Style**
  * Updated deploy UI spinner to a smoother Unicode-based animation for improved readability.
  * Made spinner rendering consistent across the whole status text (no per-character styling).
  * Increased animation frame rate for a quicker, more responsive spinner during deployments.


phinze (approved):
A classic! 

Image

PR #336, opened 2025-11-11
Add DNS support for app.miren internal domain

evanphx:
## Summary

Implements DNS resolution for inter-service communication within sandboxes using the `app.miren` internal domain. Sandboxes can now discover and connect to each other using service names (e.g., `web.app.miren` or just `web`).

## DNS Query Support

- **Query format**: `<service-name>.app.miren` (e.g., `web.app.miren`)
- **Source IP lookup**: Determines which app is making the request
- **Service discovery**: Returns all IPs for running sandboxes with matching service label
- **Response handling**: Empty NOERROR for unknown sources or missing services
- **TTL**: 30 seconds for dynamic service discovery
- **TXT**: Queries to `TXT app.miren.` return the list of services that are known about the current app
- **PTR**: Reverse DNS ptr records are resolved properly

```shell
miren2025:/app$ dig +short web.app.miren A
10.8.71.4
miren2025:/app$ dig +short app.miren TXT
"web"
miren2025:/app$ dig +short -x 10.8.71.4
web.app.miren.
```

### Entity Watching
- Watches Sandbox entities to maintain in-memory DNS mappings
- Tracks three mappings: `ipToApp`, `appServiceToIPs`, and `entityToIP`
- Handles CREATE, UPDATE, and DELETE operations properly
- Supports cleanup via both DELETE events and status transitions
- Recovers existing sandboxes on startup

### Resolver Configuration
- Adds `search app.miren` to sandbox resolv.conf
- Enables short-form queries (e.g., `web` automatically becomes `web.app.miren`)

### Lifecycle Management
- Proper DNS server shutdown with goroutine tracking (WaitGroup + context cancellation)
- Fixed deadlock in `ServiceManager.ShutdownAll()` by extracting unlocked helper method
- Calls `NetServ.ShutdownAll()` in `SandboxController.Close()`
- Ensures DNS servers shut down between tests to prevent resource leaks

## Test plan

- [x] All sandbox controller tests pass (28 tests)
- [x] Proper cleanup between tests verified
- [x] No goroutine leaks or nil pointer dereferences
- [x] Handles entity DELETE operations correctly

## Summary by CodeRabbit

* **New Features**
  * App-scoped DNS resolution for sandbox services via the app.miren domain, including service listings and PTR responses.
  * Live sandbox watching keeps DNS mappings up to date.

* **Improvements**
  * Network services now support graceful shutdown and better context-aware DNS operations.
  * Sandbox configs include app.miren as a DNS search domain.

* **Chores**
  * Example app startup command adjusted; test server performs an external IP lookup at launch.

### DIFF
--- a/network/services.go
+++ b/network/services.go
@@ -58,9 +58,7 @@ func (sm *ServiceManager) SetupDNS(ctx context.Context, bc *BridgeConfig) error 
 	sm.bridges[bridgeName] = bs
 
 	// Store context for DNS watcher
-	if sm.ctx == nil {
-		sm.ctx = ctx
-	}
+	sm.ctx = ctx
 
 	for _, addr := range bs.ips {
 		// Create and start DNS server with entity client and logger
--- a/network/services.go
+++ b/network/services.go
@@ -1,12 +1,14 @@
 package network
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 	"net/netip"
 	"slices"
 	"sync"
 
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
 	"miren.dev/runtime/pkg/asm/autoreg"
 	"miren.dev/runtime/pkg/dns"
 )
@@ -14,9 +16,11 @@ import (
 // ServiceManager handles network services (DNS, etc) for bridges
 type ServiceManager struct {
 	Log *slog.Logger
+	EAC *entityserver_v1alpha.EntityAccessClient
 
 	mu      sync.Mutex
 	bridges map[string]*BridgeServices
+	ctx     context.Context
 }
 
 var _ = autoreg.Register[ServiceManager]()
@@ -33,7 +37,7 @@ type BridgeServices struct {
 }
 
 // SetupDNS ensures a DNS server is running for the given bridge
-func (sm *ServiceManager) SetupDNS(bc *BridgeConfig) error {
+func (sm *ServiceManager) SetupDNS(ctx context.Context, bc *BridgeConfig) error {
 	sm.mu.Lock()
 	defer sm.mu.Unlock()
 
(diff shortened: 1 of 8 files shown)


pkg/controller/interfaces.go
```diff
@@ -10,6 +10,9 @@ import (
 
 // EntityAccessClient defines the interface for interacting with the entity server
 type EntityAccessClient interface {
+	// Get returns a single entity by its ID
+	Get(ctx context.Context, id string) (*entityserver_v1alpha.EntityAccessClientGetResults, error)
```
phinze, line 14: How did we not have this already lol

testdata/go-server/main.go
```diff
@@ -2,11 +2,23 @@ package main
...
+	r, err := http.Get("http://ipconfig.io/json")
+	if err != nil {
+		fmt.Printf("Error fetching IP config: %v\n", err)
+	} else {
+		defer r.Body.Close()
+		fmt.Println("Successfully fetched IP config")
+		data, _ := io.ReadAll(r.Body)
+		fmt.Println(string(data))
+	}
+
```
phinze, line 21: ❓ Was this just to double check DNS config? I suppose nbd to leave it in - IP address detection won't hurt in the test app.

pkg/dns/dns.go
```diff
@@ -70,12 +109,383 @@ func (s *Server) handleRequest(w dns.ResponseWriter, r *dns.Msg) {
...
+	if len(sb.Network) == 0 {
+		s.log.Warn("sandbox has no network addresses", "sandbox", sb.ID)
+		return
+	}
+
+	// Extract IP from address (may be in CIDR format like "10.0.0.5/24")
+	ipAddr := sb.Network[0].Address
+	if strings.Contains(ipAddr, "/") {
+		ipAddr = strings.Split(ipAddr, "/")[0]
+	}
```
phinze, line 497: 💡 `netip.ParsePrefix` can do this with a lil more robustness:

```go
  if prefix, err := netip.ParsePrefix(ipAddr); err == nil {
      ipAddr = prefix.Addr().String()
  }
```

network/services.go
```diff
@@ -53,18 +57,30 @@ func (sm *ServiceManager) SetupDNS(bc *BridgeConfig) error {
 	// Create new services entry if needed
 	sm.bridges[bridgeName] = bs
 
+	// Store context for DNS watcher
+	if sm.ctx == nil {
+		sm.ctx = ctx
+	}
```
phinze: 💡 I don't think the guard is getting us anything here... if we call SetupDNS multiple times (tho unlikely) we'd probably expect the last context to win.

phinze (approved):
Image

pkg/controller/interfaces.go
evanphx: it's a separate one, we need probably to cleanup our interface defs at some point.

testdata/go-server/main.go
evanphx: Yup and yup.

pkg/dns/dns.go
evanphx: ah yeah, let's use netip.

network/services.go
evanphx: Fair!

phinze (approved):

PR #338, opened 2025-11-12
logging: reducing unnecessary logging

evanphx:
## Release Notes

* **Refactor**
  * Improved logging structure with module context labels for enhanced traceability in diagnostic output
  * Reduced verbose debug logging to enhance log clarity and reduce noise during normal operation
  * Streamlined internal helper functions for improved code maintainability

### DIFF
--- a/cli/commands/global.go
+++ b/cli/commands/global.go
@@ -95,7 +95,7 @@ func setup(ctx context.Context, flags *GlobalFlags, opts any) *Context {
 		Level: slog.LevelDebug,
 	})).With("module", "user")
 
-	ctx = slogrus.WithLogger(ctx, s.Log)
+	ctx = slogrus.WithLogger(ctx, s.Log.With("module", "slogrus"))
 	slogrus.OverrideGlobal(s.Log)
 
 	s.Server.Log = s.Log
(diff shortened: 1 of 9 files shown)


cli/commands/global.go
```diff
@@ -95,7 +95,7 @@ func setup(ctx context.Context, flags *GlobalFlags, opts any) *Context {
 		Level: slog.LevelDebug,
 	})).With("module", "user")
 
-	ctx = slogrus.WithLogger(ctx, s.Log)
+	ctx = slogrus.WithLogger(ctx, s.Log.With("module", "slogrus"))
```
phinze, line 98: ❓ Does setting this help us at all? Seems like if it's global it'll just add "slogrus" everywhere?

phinze (approved):
🧹 👏🏻

cli/commands/global.go
evanphx: it helps with all the output from containerd/log that shows up whenever we download images at least.

evanphx: I thought about tagging it "containerd" but it might get picked up elsewhere and that would be extra confusing.

PR #339, opened 2025-11-12
controller: fix concurrency entity processing races

evanphx:
The system needs multilpe workers per controller to function currently (workers occasionally block to wait for a condition). We have always had a situation where 2 workers could concurrently be processing the same entity, which is super bad.

This solves the problem by doing a sort of inverted work stealing. If a worker detects that an event is for an entity that is currently being processed by another worker, it enqueues that event FOR THAT OTHER WORKER. Then when a worker finishes it's work, it checks if anyone dropped additional events for processing and it handles them there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures events for the same entity are processed sequentially (queued when already in-flight), preventing duplicate/concurrent handling and improving reliability.
* **Tests**
  * Adds tests validating per-entity sequencing and concurrency limits across multiple workers and entities, ensuring no cross-entity interference and correct event ordering.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -55,6 +55,12 @@ func WorkerId(ctx context.Context) string {
 // HandlerFunc is a function that processes an entity
 type HandlerFunc func(ctx context.Context, event Event) ([]entity.Attr, error)
 
+// inFlightEntry tracks an entity being processed and queues additional events
+type inFlightEntry struct {
+	revision      int64
+	pendingEvents []Event
+}
+
 // ReconcileController implements the Controller interface
 type ReconcileController struct {
 	Log *slog.Logger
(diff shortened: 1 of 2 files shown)


phinze:
Oooooo nice find! Can you see if you can get Claude to write a test exposing this issue and then covering the regression? It's a lot of behavior to toss into a core cog w/o any explicit test coverage, which makes me itchy.

phinze:
> This solves the problem by doing a sort of inverted work stealing. If a worker detects that an event is for an entity that is currently being processed by another worker, it enqueues that event FOR THAT OTHER WORKER. Then when a worker finishes it's work, it checks if anyone dropped additional events for processing and it handles them there.

Does this imply that we still require min 2 workers per controller for correct behavior? If so I think there are a few controller constructors we'd need to tweak (and we should add a validation to require it).

evanphx:
We're currently setting up a lot of controllers with 3 workers as it's the default.

I'll have claude work up a test for this.

phinze:
> We're currently setting up a lot of controllers with 3 workers as it's the default.

All the [coordinator controllers](https://github.com/mirendev/runtime/blob/f06d9aae13273706102ff1b7a066f388e82571f2/components/coordinate/coordinate.go#L508-L529) and the [default route controller](https://github.com/mirendev/runtime/blob/f06d9aae13273706102ff1b7a066f388e82571f2/components/runner/runner.go#L442-L452) are currently using 1. If it results in incorrect behavior IMHO we should make it a panic.

phinze (approved):
Test looks good! Just that one question about the outstanding controllers with 1 worker configured.

PR #340, opened 2025-11-12
disk: Properly recheck leases on disk state changes

evanphx:
The previous method was never actually called (oops). This cleans up the pattern a little bit by building a simple controller that enqueues events for another controller. A decent pattern to build off as well.

## Summary by CodeRabbit

* **Refactor**
  * Restructured disk reconciliation: a new disk-watch controller now observes disk state changes and triggers lease reconciliations.
  * The disk-lease controller no longer manages its own startup/watch lifecycle; lease reconciliation is driven via the new disk-watch flow.

### DIFF
--- a/components/runner/runner.go
+++ b/components/runner/runner.go
@@ -491,8 +491,8 @@ func (r *Runner) SetupControllers(
 			entity.Ref(entity.EntityKind, storage_v1alpha.KindDisk),
 			eas,
 			controller.AdaptController(diskWatchController),
-			time.Minute, // No periodic resync needed
-			1,           // Single worker sufficient
+			time.Minute,
+			1, // Single worker sufficient
 		),
 	)
 
--- a/components/runner/runner.go
+++ b/components/runner/runner.go
@@ -482,5 +482,19 @@ func (r *Runner) SetupControllers(
 
 	cm.AddController(diskLeaseRC)
 
+	// Add disk watch controller to trigger lease reconciliation on disk changes
+	diskWatchController := disk.NewDiskWatchController(log, eas, diskLeaseRC)
+	cm.AddController(
+		controller.NewReconcileController(
+			"disk-watch",
+			log,
+			entity.Ref(entity.EntityKind, storage_v1alpha.KindDisk),
+			eas,
+			controller.AdaptController(diskWatchController),
+			time.Minute, // No periodic resync needed
+			1,           // Single worker sufficient
+		),
+	)
+
 	return cm, nil
 }
--- a/controllers/disk/disk_watch_controller.go
+++ b/controllers/disk/disk_watch_controller.go
@@ -1,0 +1,94 @@
+package disk
+
+import (
+	"context"
+	"log/slog"
+
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/api/storage/storage_v1alpha"
+	"miren.dev/runtime/pkg/controller"
+	"miren.dev/runtime/pkg/entity"
+)
+
+// DiskWatchController watches for disk state changes and triggers reconciliation of dependent leases
+type DiskWatchController struct {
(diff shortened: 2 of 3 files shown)


components/runner/runner.go
```diff
@@ -482,5 +482,19 @@ func (r *Runner) SetupControllers(
...
+	diskWatchController := disk.NewDiskWatchController(log, eas, diskLeaseRC)
+	cm.AddController(
+		controller.NewReconcileController(
+			"disk-watch",
+			log,
+			entity.Ref(entity.EntityKind, storage_v1alpha.KindDisk),
+			eas,
+			controller.AdaptController(diskWatchController),
+			time.Minute, // No periodic resync needed
+			1,           // Single worker sufficient
```
phinze: ❓ Same Q as #339 about single workers... are they never ok?

controllers/disk/disk_watch_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	// New disk created - reconcile any pending leases waiting for it
+	return d.reconcileDependentLeases(ctx, disk.ID)
+}
+
+// Update handles updates to a disk entity (triggers lease reconciliation)
+func (d *DiskWatchController) Update(ctx context.Context, disk *storage_v1alpha.Disk, meta *entity.Meta) error {
+	// Disk state changed - reconcile dependent leases
+	d.Log.Debug("Disk state changed, finding dependent leases", "disk", disk.ID)
+	return d.reconcileDependentLeases(ctx, disk.ID)
+}
```
phinze, line 47: 💡 I added an interface for this pattern so you just have to implement `Reconcile` instead of wiring all the ~meaningless verbs.

Example usage: https://github.com/mirendev/runtime/blob/f06d9aae13273706102ff1b7a066f388e82571f2/components/coordinate/coordinate.go#L513

components/runner/runner.go
```diff
@@ -482,5 +482,19 @@ func (r *Runner) SetupControllers(
...
+	// Add disk watch controller to trigger lease reconciliation on disk changes
+	diskWatchController := disk.NewDiskWatchController(log, eas, diskLeaseRC)
+	cm.AddController(
+		controller.NewReconcileController(
+			"disk-watch",
+			log,
+			entity.Ref(entity.EntityKind, storage_v1alpha.KindDisk),
+			eas,
+			controller.AdaptController(diskWatchController),
+			time.Minute, // No periodic resync needed
```
phinze: ⚠️ This comment is currently a lie - did you mean to disable periodic?

controllers/disk/disk_watch_controller.go
```diff
@@ -0,0 +1,94 @@
...
+
+		// Trigger reconciliation by creating an event with the full entity
+		ev := controller.Event{
+			Type:   controller.EventUpdated,
+			Id:     lease.ID,
+			Entity: e.Entity(),
+		}
+
+		// Add to the lease reconcile controller's work queue
+		d.LeaseController.Enqueue(ev)
```
phinze, line 83: 🙌🏻  Nice pattern will def be useful elsewhere

phinze (changes requested):
Couple of nonblockers and one blocking Q about the periodic comment mismatch

components/runner/runner.go
phinze: Per convo live -> they are sometimes okay! It is not always incorrect w/ 1 worker, it depends on what a given controller does.

controllers/disk/disk_watch_controller.go
evanphx: yeah, agree. I'll add a ticket to clean up this pattern later.

components/runner/runner.go
evanphx: lies! the comment lies!

phinze (approved):

PR #341, opened 2025-11-12
activator: track ver => pool => sandboxes, not ver => sandboxes

evanphx:
Becuase there is not a 1-to-1 relationship between versions and pools, we need to track the ready sandboxes against the pool, rather than the rev.

Tracking them ver => sandbox causes a strange bug. If a new version is introduced and tacked onto an existing pool, the activator will ignore the existing sandboxes (as it's tracking them per version). But when it goes to try reconsie the pool to get sandboxes, the code gets confused because the pool in question already has the right size. This causes the activator to wait for a condition that never occurs, as the pool is already the correct size.

New versions can be easily created by deploying identical code. The config code currently won't create a version that matches previous versions (as the diff is empty), but pushing the same code can result in the same container image being created, which causes the bug.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized sandbox and pool tracking to a pool-centered model, improving lifecycle consistency, recovery, and concurrency handling.

* **Tests**
  * Updated and expanded tests to validate pool-centric ownership, recovery, migration, and concurrency behaviors.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -96,13 +96,6 @@ type sandbox struct {
 	tracker     *concurrency.ConcurrencyTracker
 }
 
-type verSandboxes struct {
-	ver       *core_v1alpha.AppVersion
-	sandboxes []*sandbox
-
-	strategy concurrency.ConcurrencyStrategy
-}
-
 type verKey struct {
 	ver, service string
 }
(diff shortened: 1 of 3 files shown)


phinze (approved):
Looking good! 👏🏻

PR #342, opened 2025-11-12
build: detect identical artifacts

evanphx:
Buildkit can generate an existing image based on the input if it wishes (it's an optimization!). But when that happened before this, we'd get an error because the ocireg will refuse to create a new artifact entity with the same sha256 as an existing one (this is an important property because we have to lookup artifacts FROM the sha256, so we want there to be a 1-to-1 between them).

This fixes the issue by running buildkit, but then looking up the resulting artifact after we build and using that one. This results in multiple deploys using the same artifact properly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability with manifest validation and clearer error/status reporting.
  * More accurate artifact tracking by switching to digest-based lookups.

* **New Features**
  * Builds now capture manifest identifiers from exporter responses, improving traceability.
  * Image URLs are updated to reflect the actual artifact used (including reused/deduplicated artifacts).


phinze (approved):
Image

evanphx:
@coderabbitai review

PR #343, opened 2025-11-12
pool: Add sandbox-labels and sandbox-prefix to sandbox-pool

evanphx:
When a pool is created by the deployer, it's helpful (see later commits about DNS) to have the sandbox entity tagged with the app it's from. This lets the deployer setup a set of labels that the pool manager will dutifully apply to any sandboxes it creates.

Additionally, finally make the entity ID for sandboxes in a pool meaningfuly, so sandbox_prefix informs the pool manager to name the sandboxes it creates with a particular naming schema for easily understanding.

## Summary by CodeRabbit

* **New Features**
  * Sandbox pools now support custom labels that automatically apply to all sandboxes created from the pool
  * Added configurable name prefix for sandbox entities to improve organization and identification


api/compute/schema.yml
```diff
@@ -346,6 +346,17 @@ kinds:
...
+    # Labels to apply to sandboxes created from this pool
+    sandbox_labels:
+      type: label
+      doc: Labels that will be added to the metadata of sandboxes created from this pool
+      many: true
+
+    # Prefix for sandbox entity names created from this pool
+    sandbox_prefix:
+      type: string
+      doc: Prefix used when generating sandbox entity names (e.g., "myapp-web" produces "myapp-web-abc123")
```
phinze, line 358: ❓ Just want to squint at the semantics of this for a second... why are these attributes of the `pool` and not the `sandbox_spec` (which would imply static immutable things resolved at build-time)?

phinze: Currently all the usage is to set things that are known at build-time... so trying to imagine future usages where it'd be valuable for this to be a runtime concern.

evanphx: Because they aren't really about the spec of the sandbox, they're about how the pool creates sandboxes. In other words, the spec is what goes into the body of the sandbox, but sandbox_labels are what go into the metadata of the entity, and sandbox_prefix are what is used to generate the id of the sandbox. So putting them in the spec felt like the wrong place.

phinze: Seems reasonable... the sandbox_spec should be the things copied down to sandboxes. Makes me think maybe sandbox_spec.Version needs rethinking. Anyways LGTM!

phinze (approved):

PR #345, opened 2025-11-12
dns: Track sandbox's earlier than running

evanphx:
Depends on #343

## Summary by CodeRabbit

* **Improvements**
  * More reliable sandbox-to-service tracking using entity-based guards for DNS mappings
  * Deferred app/service identification until after IP resolution to reduce false positives
  * Skip processing for sandboxes missing service labels, reducing erroneous entries
  * Enhanced diagnostic logging with application and service context and preserved startup recovery for running sandboxes


evanphx:
@coderabbitai review

phinze (approved):

PR #346, opened 2025-11-12
sandbox: patch a sandbox after allocating it's network details

evanphx:
This allows the DNS subsystem to see the address before the sandbox has fully booted. This is critical because it allows dependent sandboxes to see the names of their dependent services when they're finally fully booted.

Without this, service A that needs service B but boots before service B will get a negative DNS result about service B. With the change, service A will get the address of service B even when service B hasn't finished booting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure sandbox network address data is persisted into sandbox metadata immediately after network allocation, improving metadata consistency during creation.

* **Tests**
  * Updated tests to persist and retrieve entities from the store and synchronize revision values before metadata operations, increasing test accuracy and resilience.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -915,7 +915,9 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 
 	patchAttrs := entity.New(networkAttrs...)
 
-	res, err := c.EAC.Patch(ctx, patchAttrs.Attrs(), meta.Revision)
+	// Use 0 as the revision in the case that the sandbox has been updated before we got
+	// here. This update must go through.
+	res, err := c.EAC.Patch(ctx, patchAttrs.Attrs(), 0)
 	if err != nil {
 		return fmt.Errorf("failed to patch sandbox with network address: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -904,6 +904,22 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
...
+		entity.Ref(entity.DBId, co.ID),
+	}
+
+	for _, v := range co.Network {
+		networkAttrs = append(networkAttrs, entity.Component(compute.SandboxNetworkId, v.Encode()))
+	}
+
+	patchAttrs := entity.New(networkAttrs...)
+
+	_, err = c.EAC.Patch(ctx, patchAttrs.Attrs(), meta.Revision)
```
phinze: ❓ Do we need to update `co` and `meta.Revision` with the resulting new values for any subsequent operations in this path?

evanphx: probably a good idea, yeah.

evanphx: co already has it, we're actually pulling the patch material out of it. But i'm updating meta.Revision too.

phinze (approved):
💭 An Update inside the Create is going to fire the Watch... we'll have to keep an eye on the emergent behavior here.

PR #352, opened 2025-11-13
hack: fix ./hack/run by not munging the second argument

evanphx:
## Release Notes

* **Chores**
  * Improved internal build and test script infrastructure for better argument and environment variable handling.

### DIFF
--- a/hack/run
+++ b/hack/run
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-VERBOSE=1 iso run bash hack/test.sh "$1" -run "$2"
+iso run VERBOSE=1 -- bash hack/test.sh "$1" -run "$2"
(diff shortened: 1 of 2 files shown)


hack/run
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-VERBOSE=1 iso run bash hack/test.sh "$1" -run "$2"
+iso run VERBOSE=1 -- bash hack/test.sh "$1" -run "$2"
```
phinze, line 3: ❓ Interesting... does the env var set in this position work? And didn't it pass through before?

phinze (approved):

hack/run
evanphx: It does! Correct, it wasn't being passed through before.

PR #357, opened 2025-11-13
Remove entity.cbor persistence from sandbox controller

phinze:
## Summary

Removes disk-based entity.cbor persistence to eliminate race conditions and simplify the sandbox lifecycle. The entity store and containerd API are now the sole sources of truth for sandbox state.

## Changes

- **Remove disk I/O functions**: `saveEntity()` and `readEntity()` no longer needed
- **Remove in-place update logic**: `canUpdateInPlace()` and `updateSandbox()` deleted (sandboxes are immutable in modern architecture)
- **Improve Delete()**: Reconstructs sandbox state from containerd containers instead of reading stale entity.cbor files
- **Simplify checkSandbox()**: Removes version checking (immutable sandboxes don't need version comparison)
- **Remove differentVersion handling**: Constant and all related code paths removed
- **Clean up tests**: Removed 3 obsolete tests that tested removed functionality

## Benefits

✅ **Eliminates race conditions**: No more "entity file not found" flakes  
✅ **Simpler code**: Removes ~300 lines of legacy update-in-place logic  
✅ **More reliable**: containerd and entity store are authoritative sources  
✅ **Better architecture**: Aligns with modern immutable sandbox model via sandboxpools  

## Test Plan

- All 24 sandbox controller tests pass
- Verified linting passes
- Net code reduction: **-307 lines** (removed 347, added 40)

## Stack

Stacked on: PR #350 (sandbox recreate parameter fix)

When PR #350 merges, this can be rebased to main.

## Summary by CodeRabbit

* **Refactor**
  * Sandboxes are now treated as immutable — updates create new sandboxes instead of in-place mutation.
  * Cleanup and deletion now operate via ID-based discovery and stop/teardown flows; lifecycle logging and metrics aligned with ID-centric operations.
  * Simplified health/version checks: sandboxes validated for health only, version changes handled by replacing sandboxes.

* **Tests**
  * Legacy in-place update/rebuild/migration tests removed, reducing coverage of mutation/update paths.


controllers/sandbox/sandbox.go
```diff
@@ -2156,37 +1990,74 @@ func (c *SandboxController) destroySubContainers(ctx context.Context, sb *comput
...
+			if existing.Name == name {
+				alreadyExists = true
+				break
 			}
-			return c.stopSandbox(ctx, sb)
 		}
-		c.Log.Error("failed to read existing entity", "id", id, "err", err)
-		return fmt.Errorf("failed to read existing entity: %w", err)
+		if !alreadyExists {
+			sb.Spec.Container = append(sb.Spec.Container, compute.SandboxSpecContainer{
```
evanphx: I think this line will panic in the case that the sandbox was removed from the entity store, as Spec will be nil because we were unable to hydrate the spec.

evanphx (changes requested):
Maybe we should figure out how to store all the info we need for this case inside containerd? I think that would mostly mean a stopSandbox not requiring the spec?

controllers/sandbox/sandbox.go
phinze: Ok! Just IDs and Labels now as we discussed. Feeling nicer 🤌

controllers/sandbox/sandbox.go
```diff
@@ -2156,37 +1990,74 @@ func (c *SandboxController) destroySubContainers(ctx context.Context, sb *comput
...
+		}
 	}
 
-	var oldSb compute.Sandbox
-	oldSb.Decode(oldMeta)
+	if !foundAny {
+		c.Log.Info("Delete called but no containers found, continuing with cleanup", "id", id)
+	} else {
+		c.Log.Debug("reconstructed sandbox from containers", "id", id, "containers", len(discoveredContainers))
+	}
```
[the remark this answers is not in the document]
phinze: This concern has been superseded by the latest refactoring. The Delete() function (lines 1998-2003) now simply calls stopSandbox(id) without any spec reconstruction. The stopSandbox() function discovers everything it needs directly from containerd - no more spec hydration or nil pointer concerns.

See commits https://github.com/mirendev/runtime/commit/c1d9dc2b88d34bb94495ed91468b8bcfdcccffd1 and https://github.com/mirendev/runtime/commit/3e3b5affb361cf5d9c5b7b4d1844be6cb69b3349 which completely rewrote this logic.

evanphx (approved):

PR #366, opened 2025-11-14
clientconfig: Fix RemoveCluster operating on leaf configs

evanphx:
Prior to this, RemoveCluster calls ignored clusters defined in leaf configs. Now they're not ignored!

## Summary by CodeRabbit

* **New Features**
  * Cluster removal now supports clusters defined in leaf configurations, in addition to main configuration clusters.

### DIFF
--- a/clientconfig/clientconfig.go
+++ b/clientconfig/clientconfig.go
@@ -712,18 +712,43 @@ func (c *Config) SetCluster(name string, cluster *ClusterConfig) {
 
 // RemoveCluster removes a cluster from the configuration
 func (c *Config) RemoveCluster(name string) error {
-	if !c.HasCluster(name) {
-		return fmt.Errorf("cluster %q not found in configuration", name)
-	}
-
 	// Don't allow removing the active cluster
 	if c.active == name {
 		return fmt.Errorf("cannot remove active cluster %q", name)
 	}
 
-	delete(c.clusters, name)
+	// Check if cluster exists in main config
+	if _, existsInMain := c.clusters[name]; existsInMain {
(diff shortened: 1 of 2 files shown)


clientconfig/clientconfig.go
```diff
@@ -712,18 +712,43 @@ func (c *Config) SetCluster(name string, cluster *ClusterConfig) {
 
 // RemoveCluster removes a cluster from the configuration
 func (c *Config) RemoveCluster(name string) error {
-	if !c.HasCluster(name) {
-		return fmt.Errorf("cluster %q not found in configuration", name)
```
teresamychu, line 716: 💭 Hmmm.. should this one error out or should it just return nil?

clientconfig/clientconfig.go
```diff
@@ -712,18 +712,43 @@ func (c *Config) SetCluster(name string, cluster *ClusterConfig) {
...
+					Identities: leafConfig.identities,
+					Keys:       leafConfig.keys,
+				}
+				c.unsavedLeafConfigs[leafName] = configData
+			}
+			return nil
+		}
+	}
+
+	return fmt.Errorf("cluster %q not found in configuration", name)
```
teresamychu, line 751: 💭 Same thought as above, if the cluster doesnt exist, should it error or should it be considered a "successful deletion" since the config doesnt have this cluster anyways?

teresamychu (approved):

clientconfig/clientconfig.go
evanphx: I opted to not change the behavior for now. A good question though!

PR #368, opened 2025-11-14
Implement soft delete for apps

phinze:
## Summary

Implements first-class app deletion using soft delete with a `deleted_at` timestamp field. Deleted apps are hidden from default listings but can be restored.

- Add `deleted_at` field to app entity (not indexed - negative queries don't benefit)
- Add `miren app delete` command for soft deletion
- Add `miren app undelete` command to restore deleted apps
- Update `miren app list` to filter deleted apps by default
- Add `--deleted` flag to show deleted apps in list
- Update controllers to skip deleted apps (launcher, HTTP ingress, default route)

## Changes

**Entity Schema:**
- Added `deleted_at` timestamp field to app entity (api/core/schema.yml)

**CLI Commands:**
- `miren app delete -a <app>` - Soft deletes an app
- `miren app undelete -a <app>` - Restores a deleted app
- `miren app list --deleted` - Shows deleted apps

**Controller Updates:**
- Launcher controller skips deleted apps in reconciliation
- HTTP ingress returns 404 for deleted apps  
- Default route controller handles soft delete events

**Implementation Notes:**
- The undelete command uses entity list to find deleted apps by name, since Get by name doesn't return soft-deleted entities
- The app client's Destroy method now sets deleted_at instead of calling hard delete

## Test Plan

Tested complete workflow:
- ✅ Deploy app → delete app → verify hidden from list
- ✅ `app list --deleted` shows deleted apps
- ✅ Undelete app → verify appears in default list
- ✅ Deleted apps don't get deployed (launcher skips)
- ✅ Deleted apps return 404 on HTTP requests

## Summary by CodeRabbit

* **New Features**
  * Apps can now be soft-deleted and restored instead of permanent deletion.
  * Added `app delete` and `app undelete` CLI commands for managing deleted applications.
  * Added `--deleted` flag to `app list` command to display deleted apps.
  * Deleted apps return HTTP 404 errors and are excluded from deployment reconciliation and default routing.


phinze:
Closing this for now as we're not pursuing it soon. Just leaving it out there as a reference point.

PR #378, opened 2025-11-17
disk: Support operations when nbd isn't available

evanphx:
The disks won't be persistented by miren.cloud, but they'll at least work.

## Summary by CodeRabbit

* **New Features**
  * System now automatically detects NBD availability and switches to directory-based disk operations when unavailable, improving compatibility across different environments.

* **Tests**
  * Comprehensive test coverage added for directory-based disk and lease management scenarios.

### DIFF
--- a/controllers/disk/disk_controller.go
+++ b/controllers/disk/disk_controller.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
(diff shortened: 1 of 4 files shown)


controllers/disk/disk_controller.go
```diff
@@ -63,7 +88,13 @@ func NewDiskControllerWithMountPath(log *slog.Logger, eac *entityserver_v1alpha.
 
 // Init initializes the disk controller
 func (d *DiskController) Init(ctx context.Context) error {
-	// No special initialization needed
+	// Check if NBD is available
+	if !isNBDAvailable() {
+		d.directoryMode = true
+		d.Log.Warn("NBD kernel module not available - using directory-only mode for disks")
```
phinze, line 94: 💭 We probably want an even noisier version of this at some point. Maybe something shows up in the Miren Cloud cluster entry with a caution about data not being persisted.

phinze (approved):

PR #371, opened 2025-11-14
Add sandbox health checking to prevent zombie sandboxes

phinze:
## Summary

This PR implements comprehensive health checking for sandboxes to detect and replace "zombie" sandboxes - those that appear running but are actually unreachable.

## Problem

On 2025-11-14, rfd.miren.garden experienced 502 errors for over an hour due to a zombie sandbox (`rfd-web-CV86X712QinHpbMs4GraX`):
- Sandbox had `status: running` in entity store
- Containerd task was running
- **But the gvisor network had failed** - `dial tcp 10.8.48.130:3000: connect: no route to host`
- HTTP ingress kept selecting this sandbox, causing all requests to fail with 502

The system had no mechanism to detect this type of failure until manual intervention.

## Solution

This PR adds three layers of health checking:

### 1. Periodic Health Checks (sandbox controller)
- **What**: Background goroutine runs every 60 seconds
- **Checks**:
  - Containerd task status (existing logic)
  - TCP port connectivity (new)
- **Action**: Marks unhealthy sandboxes as DEAD → triggers replacement

Key code: `controllers/sandbox/sandbox.go:625-732`

### 2. Application-Level Port Checks
- **What**: Extends `isContainerHealthy()` with TCP dial test
- **When**: Called during periodic checks and boot reconciliation
- **Timeout**: 2 seconds for quick failure detection

Key code: `controllers/sandbox/sandbox.go:608-623`

### 3. Proxy Error Tracking (HTTP ingress)
- **What**: Tracks consecutive proxy failures per sandbox URL
- **Threshold**: 3 consecutive errors → mark sandbox DEAD
- **Reset**: Successful requests clear the error counter
- **Action**: Proactively removes broken sandboxes from rotation

Key code: `servers/httpingress/httpingress.go:549-653`

## Testing

- ✅ Build: `make bin/miren` - passes
- ✅ Sandbox controller tests: `hack/it controllers/sandbox` - passes
- ✅ HTTP ingress tests: `hack/it servers/httpingress` - passes
- ⏳ Full test suite: Will run in CI

Verified health check starts and stops correctly:
```
21:03:38.653 [INFO] starting periodic sandbox health checks │ interval: 60s
21:03:55.499 [INFO] stopping periodic health checks
```

### Benefits
- **Automatic recovery**: Zombie sandboxes detected within 60s (vs manual intervention)
- **Faster detection**: Proxy error tracking catches failures in ~10s (3 requests)
[...]

PR #386, opened 2025-11-18
Integrate network health checks into sandbox lifecycle

phinze:
## Summary

This PR integrates network connectivity checks into the sandbox lifecycle to ensure `RUNNING` is a strong guarantee that sandboxes are ready to serve traffic. This addresses zombie sandboxes that appear running but are unreachable due to network failures.

## Problem

On 2025-11-14, rfd.miren.garden experienced 502 errors for over an hour due to a zombie sandbox:
- Sandbox had `status: RUNNING` in entity store
- Containerd task was running
- **But the gvisor network had failed**: `dial tcp 10.8.48.130:3000: connect: no route to host`
- HTTP ingress kept routing to this sandbox, causing all requests to fail

The current system only checks containerd task status, not network reachability.

## Solution

This PR takes a **fail-fast approach** that:
1. Makes network connectivity a hard requirement before marking sandboxes RUNNING
2. Continuously validates network health during reconciliation
3. Leverages existing backoff mechanisms at the pool/activator level
4. Avoids complexity of grace periods and separate health check goroutines

### Changes

**1. Make `waitForPort` strict** (`sandbox.go:743-753`)
- **Before**: Timeout → log warning → mark RUNNING anyway
- **After**: Timeout → error → mark DEAD → pool retries with backoff
- Timeout increased from 5s to 10s for slow-starting apps
- Fast-path: Most apps bind in 1-2s and get marked RUNNING immediately

**2. Add `checkNetworkHealth()` function** (`sandbox.go:526-581`)
- Verifies TCP connectivity to sandbox ports using 2-second timeout
- Skips check for sandboxes without ports (background workers)
- Skips check for sandboxes without network allocated
- Returns false if ports are unreachable

**3. Integrate into reconciliation** (`sandbox.go:461-468`)
- Extends `checkSandbox()` to call `checkNetworkHealth()` for RUNNING sandboxes
- If unreachable → return `unhealthy` → existing flow marks DEAD → cleanup → replacement

**4. Comprehensive tests** (`sandbox_test.go:1414-1501`)
- Test network health for sandboxes without ports
- Test network health for sandboxes without network
- Test network health failure for unreachable ports

### Design Decisions

✅ **Fail-fast over patience**: With backoff at pool/activator level, we can fail fast at sandbox level  
✅ **Same logic everywhere**: No separate code paths for boot vs reconciliation  
✅ **No grace period**: Simpler implementation, backoff handles crash loops  
✅ **Port-optional**: Background workers without ports skip network checks  
✅ **Strong guarantee**: `RUNNING` means "ready to serve traffic"

### Rejected Alternatives (PR #371)

❌ Separate periodic health checker goroutine (wrong pattern)  
❌ httpingress marking sandboxes DEAD (layering violation)  
❌ Proxy error tracking (reactive, not proactive)

## Testing

- ✅ Build: `make bin/miren` - passes
- ✅ Sandbox tests: `./hack/it controllers/sandbox` - **32/32 tests pass**
  - Including 3 new network health tests
- ⏳ Full test suite: Will run in CI

### Benefits
- **Prevents zombie sandboxes**: Detects network failures within one reconciliation cycle
- **Faster failure detection**: Port binding issues fail within 10s instead of silently proceeding
- **Better reliability**: Only hands off sandboxes that are actually ready to serve traffic

### Risks
- **Slower startups for apps that bind late**: Now fail after 10s (was: proceed anyway)
  - Mitigation: Backoff at pool level handles this gracefully
- **False positives during network issues**: Transient network problems could mark healthy sandboxes unhealthy
  - Mitigation: 2-second timeout is forgiving for transient issues

## Related

- Issue: [MIR-528](https://linear.app/miren/issue/MIR-528)
- Rejected approach: PR #371
- Incident: rfd.miren.garden outage 2025-11-14

## Test Plan

1. Deploy to staging
2. Verify healthy apps start normally (should be ~1-2s to RUNNING)
3. Simulate network failure (iptables drop) and verify sandbox marked DEAD
4. Verify slow-starting apps (Java, etc.) still succeed within 10s
5. Monitor for any increased failure rates

## Summary by CodeRabbit

* **New Features**
  * Automatic network health verification for sandboxes now runs before marking them RUNNING.

* **Bug Fixes**
  * Increased port wait timeout during startup and treat port failures as hard errors to improve startup reliability.
  * Startup now marks RUNNING only after successful port binding and, if applicable, network verification.
  * Improved network-related logging with sandbox and container identifiers.

* **Tests**
  * Added tests covering network-health scenarios: no network, no exposed ports, and unreachable ports.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"net/netip"
 	"os"
@@ -458,6 +459,15 @@ func (c *SandboxController) checkSandbox(ctx context.Context, co *compute.Sandbo
 		}
 	}
 
+	// Check network connectivity for RUNNING sandboxes
+	// This catches cases where the task is running but the network has failed
+	if co.Status == compute.RUNNING {
+		if !c.checkNetworkHealth(ctx, co) {
+			c.Log.Warn("sandbox network health check failed", "sandbox_id", co.ID)
+			return unhealthy, nil
+		}
+	}
+
 	return same, nil
 }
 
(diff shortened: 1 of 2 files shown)


evanphx (approved):
I like it! I like the idea that our health checks tie directly in sandbox status rather than trying to have status bolted on by another component (like having activator try to keep it's own sandbox status).

PR #387, opened 2025-11-19
logs: log in the app context on building and httpingress

evanphx:
- Add tracking to httputil.ReverseProxy to capture request/response metrics
- Log HTTP requests from router to app logs with Heroku logfmt format (method, path, host, source_ip, body, status, response bytes, duration)
- Log deployments from builder to app logs when new versions are deployed
- Add source field to LogEntry API to distinguish log sources (router, builder, etc)
- Display source in CLI logs with truncation for long names
- Add X-Forwarded-Proto to automatic headers we set
- Bump the API response from 100 to 1000 entries

Users can now see router requests and deployments in `miren logs` alongside application output, providing better visibility into app activity.

Example:

```
❯ miren logs
2025-11-18 21:46:47: [builder] version=demo-vCVGii2ib2F8dZAinQHoCt artifact=demo-aCVGii2ib2LCvfJiNaKTM2 status=deployed
2025-11-18 21:46:50: [dem…rB9D6YwE] Server running at http://localhost:3000
2025-11-18 21:46:54: [router] method=GET path="/" host=localhost source_ip=172.17.0.1 body=0 status=200 response=4447 duration_ms=2
```

NOTE: The reverseproxy.go here is a vendored version of the stdlib one with minor additions to perform tracking of the request/response as it's sent and receieved.

## Summary by CodeRabbit

* **New Features**
  * Log entries now include source information for better tracking and identification.
  * CLI log display shows a source prefix for each log entry when available.
  * HTTP ingress now routes via a full reverse proxy and records richer request metrics (method, path, host, source IP, timing).

* **Improvements**
  * Default log retrieval limit increased from 100 to 1000 entries.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1222,6 +1222,7 @@ func (c *SandboxController) logConsumer(sb *compute.Sandbox, container string) *
 
 	attrs := map[string]string{
 		"sandbox": sb.ID.String(),
+		"source":  strings.TrimPrefix(sb.ID.String(), "sandbox/"),
 	}
 
 	if container != "" {
--- a/observability/logs.go
+++ b/observability/logs.go
@@ -211,6 +211,8 @@ func logsQLQuote(s string) string {
 	return `"` + s + `"`
 }
 
+const DefaultLogReadLimit = 1000
+
 func (l *LogReader) Read(ctx context.Context, id string, opts ...LogReaderOption) ([]LogEntry, error) {
 	var o logReadOpts
 
@@ -220,7 +222,7 @@ func (l *LogReader) Read(ctx context.Context, id string, opts ...LogReaderOption
 
 	limit := o.Limit
 	if limit == 0 {
-		limit = 1000
+		limit = DefaultLogReadLimit
 	}
 
 	// Build LogsQL query - use simple field matching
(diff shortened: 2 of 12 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -1222,6 +1222,7 @@ func (c *SandboxController) logConsumer(sb *compute.Sandbox, container string) *
 
 	attrs := map[string]string{
 		"sandbox": sb.ID.String(),
+		"source":  strings.TrimPrefix(sb.ID.String(), "sandbox/"),
```
phinze, line 1225: 💭 I've been thinking about how we've been relatively inconsistent with how we use the path-style ID prefixes - we generally strip them off for users on both inputs and outputs... does that mean we should ditch them entirely? The information encoded in the path prefix (`sandbox/`) and the hyphened prefix (`sb-`) are sort of redundant. Worth discussing IMO!

observability/logs.go
phinze: 💡 🐰 is right we should match this (and constantize it)

phinze (approved):
Nice! Been wanting the sandbox id prefixes for a while. Just the one limit match thingy.

controllers/sandbox/sandbox.go
evanphx: yeah, been thinking about that A LOT too.

observability/logs.go
evanphx: Should!

PR #388, opened 2025-11-19
Fix activator sandbox tracking memory leak

evanphx:
## Summary

Fixes a bug where the activator continued tracking sandboxes in its internal maps after they were deleted from the entity store, causing memory leaks and stale references.

## Changes

- **Added `removeSandboxFromTracking()` helper function**: Safely removes sandboxes from the `poolSandboxes` map with proper locking
- **Handle entity deletion in `watchSandboxes()`**: When `!op.HasEntity()` (entity deleted), now calls the cleanup helper using `op.EntityId()`
- **Implement cleanup in `syncLastActivityOnce()`**: When a sandbox returns `ErrNotFound` during last activity sync, remove it from tracking (implements the TODO comment)
- **Use `slices.Delete`**: Safe slice element removal instead of manual slice manipulation

## Testing

- ✅ All 26 activator tests pass
- ✅ Code compiles successfully
- ✅ Linter passes with no issues

## Impact

This ensures deleted sandboxes are properly cleaned up from memory, preventing:
- Memory leaks from accumulating deleted sandbox references
- Stale data in the activator's internal state
- Potential errors from referencing deleted entities

The fix maintains thread safety by using the same locking patterns as the rest of the activator code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Remove sandboxes from in-memory tracking when a sandbox entity is deleted, preventing stale entries.
  * On sync errors where a sandbox is not found, log the issue and remove the inaccessible sandbox from tracking, keeping pool counts accurate.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -35,6 +35,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math/rand/v2"
+	"slices"
 	"sync"
 	"time"
 
@@ -886,6 +887,11 @@ func (a *localActivator) watchSandboxes(ctx context.Context) {
 
 		_, err := a.eac.WatchIndex(ctx, entity.Ref(entity.EntityKind, compute_v1alpha.KindSandbox), stream.Callback(func(op *entityserver_v1alpha.EntityOp) error {
 			if !op.HasEntity() {
+				// Entity was deleted - clean up from tracking
+				// The ID should still be available in the operation even without the entity
+				if op.HasEntityId() {
+					a.removeSandboxFromTracking(entity.Id(op.EntityId()))
+				}
 				return nil
 			}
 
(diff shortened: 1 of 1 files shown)


components/activator/activator.go
```diff
@@ -886,6 +887,11 @@ func (a *localActivator) watchSandboxes(ctx context.Context) {
 
 		_, err := a.eac.WatchIndex(ctx, entity.Ref(entity.EntityKind, compute_v1alpha.KindSandbox), stream.Callback(func(op *entityserver_v1alpha.EntityOp) error {
 			if !op.HasEntity() {
+				// Entity was deleted - clean up from tracking
```
phinze, line 890: 💡 Seems better to check the operation type instead of assuming a delete when the entity is gone.

https://github.com/mirendev/runtime/blob/3db72654d7dc5e6108d719f066dfd8e69e22e399/api/entityserver/entityserver_v1alpha/constants.go#L11-L12

phinze (approved):
Nice catch! Just the one optional nit inline

components/activator/activator.go
phinze: Or better yet an `IsDelete()` predicate

https://github.com/mirendev/runtime/blob/3db72654d7dc5e6108d719f066dfd8e69e22e399/api/entityserver/entityserver_v1alpha/constants.go#L30-L33

evanphx: 💯

PR #394, opened 2025-11-21
Fix sandbox controller to update PENDING->RUNNING on reconciliation

phinze:
## Summary

Fixes a defensive backstop issue where sandboxes could get stuck in PENDING state forever when the initial status update to RUNNING failed due to optimistic concurrency conflicts.

## The Bug

During the mirendev production deployment on 2025-11-20, two sandboxes got stuck in PENDING state even though:
- Containers were running and healthy
- Ports were bound successfully  
- The sandbox was fully operational

This happened because:
1. Sandbox controller created sandbox and waited for ports (successful)
2. While starting, activator made multiple rapid updates to the entity
3. Controller tried to update status PENDING → RUNNING but got conflict error (entity revision changed)
4. Status update was silently lost (separate issue to address)
5. Subsequent reconciliations saw "sandbox exists and is healthy" and returned early without fixing the status

## The Fix

Added a check in the `same` case (line 655-672): if the sandbox exists, is healthy, but status is PENDING, we now explicitly update it to RUNNING using `Patch` with OCC (`meta.Revision`).

This provides a backstop to recover from status update failures regardless of why they occurred.

## Test

Added comprehensive test `PENDING sandbox with running containers should transition to RUNNING` that:
1. Creates a sandbox (which starts successfully)
2. Manually resets status to PENDING (simulating failed update)
3. Calls Create again to trigger reconciliation
4. Verifies status is corrected to RUNNING

## Part of MIR-531

This is the first of several PRs to fully address the root cause:
- ✅ This PR: Defensive fix in sandbox controller
- ⏭️ Next: Retry logic in controller framework for conflict errors
- ⏭️ Future: Consider reducing activator update frequency

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a case where a sandbox stuck in PENDING while containers were already healthy now correctly transitions to RUNNING.
  * Increased startup port wait timeout and polling frequency to improve readiness handling.

* **Tests**
  * Added a test ensuring a PENDING sandbox with active containers transitions to RUNNING after reconciliation.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -653,6 +653,41 @@ func (c *SandboxController) Create(ctx context.Context, co *compute.Sandbox, met
 		} else {
 			switch searchRes {
 			case same:
+				// If sandbox exists and is healthy but status is PENDING,
+				// update it to RUNNING. This is a fallback to recover from failed
+				// status updates (e.g. due to OCC conflicts during creation).
+				// Only do this if entity is stale (created > 2 minutes ago)
+				// to avoid conflicting with active goroutines working on booting.
+				if co.Status == compute.PENDING {
+					createdAt := meta.GetCreatedAt()
+					age := time.Since(createdAt)
+					const staleThreshold = 2 * time.Minute
+
+					if age > staleThreshold {
+						c.Log.Info("sandbox exists and is healthy but status is PENDING (stale), updating to RUNNING",
+							"id", co.ID,
+							"createdAt", createdAt,
+							"age", age)
+						patchAttrs := entity.New(
+							entity.Ref(entity.DBId, co.ID),
+							(&compute.Sandbox{
+								Status: compute.RUNNING,
+							}).Encode,
+						)
+						_, err := c.EAC.Patch(ctx, patchAttrs.Attrs(), meta.Revision)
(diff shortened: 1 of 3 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -653,6 +653,24 @@ func (c *SandboxController) Create(ctx context.Context, co *compute.Sandbox, met
...
+							Status: compute.RUNNING,
+						}).Encode,
+					)
+					_, err := c.EAC.Patch(ctx, patchAttrs.Attrs(), meta.Revision)
+					if err != nil {
+						c.Log.Error("failed to update sandbox status to RUNNING", "id", co.ID, "error", err)
+						return fmt.Errorf("failed to update sandbox status to RUNNING: %w", err)
+					}
+					return nil
+				}
```
phinze, line 690: This should be different... we should only progress from PENDING->RUNNING if there hasn't been an update in N minutes w/ a comment. This is the state transition of last resort and we don't want to conflict with legit running goroutines actively working on booting sandboxes, but we _ALSO_ don't want to dig into some funky coordination logic right now.

evanphx (approved):
Pending (har har) the update to check the last update time, looks good!

PR #395, opened 2025-11-21
Add RingSet to skip self-generated watch events in controllers

phinze:
## Summary

Implement RFD-41 to reduce controller reconciliation noise by tracking recently written entity revisions and skipping watch events for writes the controller just made.

## Background

During the MIR-531 production incident, we observed that controllers receive watch events for their own writes, triggering unnecessary reconciliation. While this is eventually consistent and correct, it creates noise and wastes CPU cycles processing events where the controller already knows the state.

## Implementation

**RingSet Data Structure:**
- Thread-safe ring buffer tracking 1000 recent entity revisions (~8KB memory)
- O(1) lookup using both ring buffer and hash map
- Automatic eviction when buffer fills

**Controller Integration:**
- `RecordWrite(revision)`: Records revisions after successful Put operations
- Watch callback: Checks `recentWrites.Contains(revision)` before queueing events
- Skipped events logged at DEBUG level

**Code Cleanup:**
- Standardized WaitGroup usage to `wg.Go()` throughout controller.go
- Removed manual `defer wg.Done()` calls that were causing double-Done() panics

## Testing

**Unit Tests (5):**
- BasicOperations, Wraparound, Concurrent, DuplicateAddsSamePosition, ZeroCapacity

**Integration Tests (4):**
- `TestReconcileController_SkipsSelfGeneratedWatchEvents` - verifies revision 2 is skipped while 1 and 3 are processed
- `TestReconcileController_RingWraparound` - verifies evicted revisions are processed
- `TestReconcileController_PutRecordsRevision` - verifies Put calls record revisions
- `TestReconcileController_FailedWriteDoesNotRecordRevision` - verifies failed writes don't pollute the ring

## Expected Impact

In production during MIR-531, we saw ~13 revisions over 3 seconds (~4 rev/sec). With this change:
- Controllers skip their own writes, reducing reconciliation by 30-50%
- Ring size of 1000 provides ~250 seconds of buffer (very conservative)
- Minimal memory overhead: ~8KB per controller, ~80KB total

## References

Part of MIR-531
Implements RFD-41

## Summary by CodeRabbit

* **New Features**
  * Controller now skips processing of self-generated watch events by tracking recent revisions, reducing redundant reconciliations.

* **Bug Fixes**
  * Prevents enqueuing during shutdown and improves goroutine lifecycle handling to reduce race conditions and noisy processing.

* **Tests**
  * Added tests covering self-event skipping, revision tracking, ring-buffer wraparound, failed-write semantics, concurrency, and related edge cases.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -193,7 +193,6 @@ func (c *ReconcileController) Start(top context.Context) error {
 
 				// Skip watch events for revisions we recently wrote to reduce reconciliation noise
 				if ev.Rev > 0 && c.recentWrites.Contains(ev.Rev) {
-					c.Log.Debug("Skipping watch event for self-generated revision", "entity", ev.Id, "rev", ev.Rev, "eventType", ev.Type)
 					return nil
 				}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -82,6 +82,10 @@ type ReconcileController struct {
 	inFlight   map[entity.Id]*inFlightEntry
 	inFlightMu sync.Mutex
 
+	// Recent writes tracking to skip self-generated watch events
+	// Controllers record revisions from their writes to reduce reconciliation noise
+	recentWrites *RingSet
+
 	// periodic is an optional periodic callback
 	periodic     func(ctx context.Context) error
 	periodicTime time.Duration
@@ -99,6 +103,7 @@ func NewReconcileController(name string, log *slog.Logger, index entity.Attr, es
 		workers:      workers,
 		workQueue:    make(chan Event, 1000),
 		inFlight:     make(map[entity.Id]*inFlightEntry),
+		recentWrites: NewRingSet(1000), // Track last 1000 revisions written by this controller
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/controller/controller.go
```diff
@@ -187,6 +191,12 @@ func (c *ReconcileController) Start(top context.Context) error {
 					ev.Rev = aen.Revision()
 				}
 
+				// Skip watch events for revisions we recently wrote to reduce reconciliation noise
+				if ev.Rev > 0 && c.recentWrites.Contains(ev.Rev) {
+					c.Log.Debug("Skipping watch event for self-generated revision", "entity", ev.Id, "rev", ev.Rev, "eventType", ev.Type)
```
evanphx: I'd argue we don't even log this in Debug because it's going to be pretty noisy and likely not actionable.

evanphx (approved):
I like it! Clean and easy.

pkg/controller/controller.go
phinze: Yeah fair! https://github.com/mirendev/runtime/pull/395/commits/2161cd17c5c6dc65abad5999540440871d29a869

I'm really itching for a metrics subsystem.... will probably wire one up soon. I need graphs to stare at! 😹

phinze: loool rabbit gets me 🐰 🤝 

2025-11-24 at 13 45 37@2x

PR #396, opened 2025-11-21
Switch controller automatic writes from Put to Patch

phinze:
## Summary

Switch the automatic entity write at the end of controller actions from Put to Patch, which is more semantically correct (Patch didn't exist when this was originally written).

Also extracts duplicate update logic into a reusable helper method.

### 1. Switch from Put to Patch

**Before:**
```go
var rpcE entityserver_v1alpha.Entity
rpcE.SetId(string(event.Id))
rpcE.SetAttrs(updates)
result, err := c.esc.Put(ctx, &rpcE)
```

**After:**
```go
attrs := append([]entity.Attr{entity.Ref(entity.DBId, event.Id)}, updates...)
result, err := c.esc.Patch(ctx, attrs, 0)
```

**OCC Decision:** Uses revision 0 (no optimistic concurrency control) to avoid breaking unforeseen code that may depend on this working without conflict detection. This is the conservative approach - we can revisit OCC protection in a more holistic pass.

### 2. Extract Duplicate Code

Lines 342-363 and 390-411 had identical code for applying updates (main event processing vs pending event processing). Extracted into new `applyUpdates` helper method.

**Benefits:**
- Eliminates code duplication
- Centralizes update logic (validation, logging, Patch call, revision recording)
- Makes future changes easier

## Manual Patch Calls in SandboxController

There are 4 manual Patch calls in SandboxController (lines 671, 719, 1545, 2055) that were identified but not changed in this PR. These could benefit from manual RecordWrite calls to skip self-generated watch events, but that would require adding a ReconcileController field to SandboxController. Since eventual consistency is acceptable, this is documented as future work.

This is addressed in the follow-up PR #397 using the WriteTracker pattern.

## Testing

- All controller tests pass (`./hack/it ./pkg/controller`)
- End-to-end tests pass (`make test-e2e`)

## Stack

- Follows: #395 (skip-self-generated-watch-events)
- Followed by: #397 (add WriteTracker pattern)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of update operations with stronger validation, error handling, and concurrency robustness to reduce failed or duplicated updates.
  * Reduced self-generated event noise so the system skips redundant internal events and maintains accurate revision tracking.
  * Enhanced operational logging for clearer success/error visibility during update processing.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/entity"
 	"miren.dev/runtime/pkg/idgen"
 	"miren.dev/runtime/pkg/rpc/stream"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -425,6 +383,35 @@ func (c *ReconcileController) processItem(ctx context.Context, event Event) ([]e
...
+		return
+	}
+
+	c.Log.Info("updating entity with updates produced by controller", "event", event, "updates", len(updates))
+
+	// Add entity ID to attrs for Patch
+	attrs := append([]entity.Attr{entity.Ref(entity.DBId, event.Id)}, updates...)
+
+	// Use Patch with OCC based on the event revision to detect concurrent modifications
+	result, err := c.esc.Patch(ctx, attrs, event.Rev)
```
phinze: Talked about this with @evanphx and this should set Revision to 0 for now to protect against any unforeseen code that is depending on this working. We can revisit with a more holistic pass.

evanphx (approved):

PR #427, opened 2025-12-02
Update developer preview download command.

teresamychu:
## Summary by CodeRabbit

* **Documentation**
  * Quick Start installation command updated to use safer curl flags (`-fsSL`) and to invoke `bash` for a more consistent installation instruction across docs.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -11,7 +11,7 @@ Get up and running with Miren in minutes.
 Install Miren with a single command:
 
 ```bash
-curl https://miren.cloud/install | sudo bash
+curl -fsSL https://miren.cloud/install | sudo bash
 ```
 
 This script will:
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started.md
```diff
@@ -11,7 +11,7 @@ Get up and running with Miren in minutes.
 Install Miren with a single command:
 
 ```bash
-curl https://miren.cloud/install | sudo sh
+curl https://miren.cloud/install | sudo bash
```
phinze: While we're in here making tweaks, can we add `-fsSL` flags to `curl`?

- `-f` fail silently on HTTP errors (don't execute error pages)
- `-s` silent mode
- `-S` show errors even in silent mode
- `-L` follow redirects

This will make it a little more resilient to any network weirdness or changes.

phinze (approved):

docs/docs/getting-started.md
evanphx: Agreed!

teresamychu: Indeed!

PR #433, opened 2025-12-03
disk: add remote volume lease management and server-generated IDs

evanphx:
## Summary

- Add remote volume lease management with `AcquireVolumeLease`/`ReleaseVolumeLease` calls throughout the disk lifecycle
- Change `CreateVolumeInSegmentAccess` to return server-generated volume IDs instead of accepting a requested ID
- Add metadata (node, app) to volume lease requests for tracking
- Handle 409 Conflict as success when volume already exists (idempotent creates)
- Close LSVD disk when unmounting volumes and release leases on close
- Resolve app version to app ID when creating disk leases in sandbox controller
- Restrict disk usage to single instance services
- Remove noisy RPC logs

## Test plan

- [x] Verify volume creation works with server-generated IDs
- [x] Verify lease acquisition before disk initialization
- [x] Verify lease release on unmount/delete/close
- [x] Verify 409 handling for existing volumes
- [x] Run existing disk controller tests

## Summary by CodeRabbit

* **New Features**
  * Introduced volume lease management with acquire/release flow and app-level context for disk leases.

* **Bug Fixes**
  * Disks may now only be attached to services in fixed concurrency mode with num_instances=1.

* **Refactor**
  * Switched to deterministic, org-scoped and server-generated volume IDs and clarified provisioning flows (directory, remote, fallback).

* **Chores**
  * Removed verbose debug logging from authentication paths.


controllers/disk/lsvd_client.go
```diff
@@ -151,125 +165,174 @@ func NewLsvdClientWithReplica(log *slog.Logger, dataPath string, authClient *clo
...
+	}
+
+	volumeInfo := &lsvd.VolumeInfo{
+		Name: diskName,
+		Size: units.GigaBytes(sizeGb).Bytes(),
+		UUID: u.String(),
+		Metadata: map[string]any{
+			"filesystem": filesystem,
+		},
+	}
```
phinze, line 205: 💡 The server expects this in a `data_format` field, so there's a column blank in the UI

Image

lsvd/segment_access.go
```diff
@@ -128,20 +165,11 @@ func (t *replicaWriter) OpenVolume(ctx context.Context, vol string) (Volume, err
...
-	}
-
-	if wInfo.Name != rInfo.Name || wInfo.Size != rInfo.Size || wInfo.Parent != rInfo.Parent || wInfo.UUID != rInfo.UUID {
-		return nil, os.ErrInvalid
+	if err == nil {
+		return wInfo, nil
 	}
 
-	return wInfo, nil
+	return t.replica.GetVolumeInfo(ctx, vol)
```
phinze, line 175: ❓ This is a notable error handling change.

**Before**: Required both primary and replica to have matching volume info.
**After**: Silently falls back to replica if primary fails.

Is this intentional? If so I could use a 'splain. 😄 👂

pkg/cloudauth/rpc_authenticator.go
```diff
@@ -189,9 +189,11 @@ func (a *RPCAuthenticator) NoAuthorization(ctx context.Context, r *http.Request)
...
 
 		// Extract the identity from the certificate
-		a.logger.Debug("request allowed via client certificate",
-			"subject", cert.Subject.String(),
-			"path", r.URL.Path)
+		/*
+			a.logger.Debug("request allowed via client certificate",
+				"subject", cert.Subject.String(),
+				"path", r.URL.Path)
+		*/
```
phinze: 💡 IMHO delete instead of comment here

phinze (approved):
Took this for a spin with a local cloud instance. Tested the remote volume lease flow end-to-end:

1. Created a remote-only disk → got server-generated volume ID ✓
2. Deployed an app with a disk (db-app style config) → lease acquired, disk mounted, sandbox started ✓
3. Deleted the app → lease released, volume unmounted ✓

The lease acquire/release cycle is working nicely!

Initially thought there might be an issue with lease release not cleaning up if the lease wasn't in the in-memory activeLeases map (e.g. after server restart), but `handleBoundLease`  recovers that state from EAS before any release would happen, so it's fine.

Some inline minor things, but overall LGTM! 🚀

lsvd/segment_access.go
```diff
@@ -86,15 +94,44 @@ func (t *replicaWriter) InitContainer(ctx context.Context) error {
...
+		// Use the server-generated ID for the local volume
+		localVol := *vol
+		localVol.Name = volumeID
+
+		if err := t.primary.InitVolume(ctx, &localVol); err != nil {
+			return "", err
+		}
+
+		return volumeID, nil
 	}
```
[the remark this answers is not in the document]
evanphx: We're not using this path for now, so fine.

lsvd/segment_access.go
evanphx: I'll add a comment. Basically I need to rev the LSVD api because FileAccess is using Name as the directory to store the data in, but our remote uses Name as the name that the user gave it. What results in this code failing because the Names are different even though they're the same.

controllers/disk/lsvd_client.go
evanphx: mmm yeah, I think we just leave it for now and improve it later.

PR #434, opened 2025-12-03
Add server readiness check and client config to server install

evanphx:
## Summary
- Add server readiness waiting to `miren server install` (matches Docker install behavior)
- Check systemd service status during health checks to fail fast if server crashes
- Generate and save client configuration for the local cluster after server starts
- Fix ownership and permissions on config files when running under sudo

## Test plan
- [ ] Run `sudo miren server install` on a fresh system
- [ ] Verify server readiness is checked before completing
- [ ] Verify client config is written to `~/.config/miren/clientconfig.d/50-local.yaml`
- [ ] Verify file ownership is correct when running under sudo
- [ ] Test that install fails fast if service crashes during startup

## Release Notes

* **New Features**
  * Server install now waits for and validates server readiness before completing.
  * Local client configuration is automatically generated and activated during post-install to simplify onboarding.
* **Observability**
  * Improved logging and explicit success/failure messages for health checks and client setup to aid troubleshooting.

### DIFF
--- a/cli/commands/server_install.go
+++ b/cli/commands/server_install.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


phinze (approved):
Rabbit seems right, but otherwise looks good

PR #447, opened 2025-12-05
Add basic docs about disks

evanphx:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Disks documentation covering overview, lifecycle, configuration examples, provisioning, leasing, and deletion workflows
  * Added a dedicated CLI Disk reference describing list/status/create/delete/lease/mount commands with examples and troubleshooting
  * Added "Persistent Storage with Disks" to Getting Started and a new "Disks" entry under Key Concepts and Next Steps navigation


docs/docs/cli/disk.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+## miren debug disk lease-release
+
+Release a disk lease (mark it for cleanup).
+
+### Usage
+
+```bash
+miren debug disk lease-release -i <lease-id>
```
phinze, line 182: This combination of words triggers a vinyl skip in my brain. They're the correct words! The bug is in my brain. 😂 

Image

docs/docs/disks.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Disk Sizing
+
+- Disks use a "thin provisioning" technology, enabling it to only allocated storage when needed
+- Choose a size that accommodates growth
+
+### Filesystem Choice
+
+- **ext4**: Best general-purpose choice, widely compatible
+- **xfs**: Better for large files and high-throughput workloads
```
phinze, line 168: Missing `brtfs` here though mentioned above

docs/docs/disks.md
```diff
@@ -0,0 +1,168 @@
...
+
+**NOTE:** Your server must have the mkfs tools to format the disk types.
```
phinze, line 170: Spot checking just to see what has what:

 - Oracle Linux has all three
 - Ubuntu on Oracle Cloud has only ext4
 - Ubuntu on Google Cloud has all three

docs/docs/getting-started.md
```diff
@@ -161,6 +161,22 @@ miren env
...
+
+```toml
+[[services.db.disks]]
+name = "myapp-data"
+mount_path = "/data"
+size_gb = 10
+```
+
+**NOTE:** Disks can only be used with fixed instance scheduled services, which the default
+web service is not.
```
phinze, line 176: 💡 Might want to show a config snippet of what it looks like to make a web service fixed instance.

phinze (approved):
Love it! Just a few inline optional thingies

phinze:
⚠️ Ah something we should mention somewhere: **how much data do I get to store and is this free forever?**

I think we can be honest and say it's un-metered right now but we're planning to design a free-tier + paid service around it. Just would be good to have something we can point to since I'm sure folks will ask.

docs/docs/disks.md
evanphx: Yeah, I removed it on purpose I don't want people going down that route yet.

docs/docs/disks.md
evanphx: perf.

docs/docs/getting-started.md
evanphx: I think I want to steer people away from that for now.

PR #449, opened 2025-12-05
Initial cut at talking about scaling

evanphx:
## Summary by CodeRabbit

* **Documentation**
  * Added a full autoscaling guide covering auto vs fixed modes, scale-to-zero, configuration options, instance-calculation logic, tuning tips, and monitoring commands.
  * Documented persistent-disk constraint (requires fixed single instance).
  * Added comprehensive Services reference (configuration, routing, storage, examples, lifecycle).
  * Updated intro and getting-started with new key concepts, links, and next steps; added "Service" terminology entry.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -81,6 +81,21 @@ That's it! Miren will:
 
 Note: Your first app gets a default route automatically. For subsequent apps, you'll need to configure routes manually. See [Working with Routes](#working-with-routes).
 
+### "Why didn't I have to configure how many copies of my application to run?"
+
+A core philosophy of Miren is that guessing replica/instance/copy counts is the wrong way to manage
+application scaling by default. For that reason, from day 1, Miren has built around autoscaling application
+instances. It does this using the same technique that Google Cloud Run uses, namely that as
+the amount of traffic to the application increases, additional instances are automatically launched.
+
+And as the traffic reduces, the instance counts are automatically reduced.
+
+If an application isn't stateless and needs to only run a certain number of copies, the application
+can be switched to fixed mode where you set the number of instances. This is commonly used for services
+that your application also needs, like databases, background workers, etc.
+
+See [Application Scaling](/scaling) for full documentation on configuring scaling behavior.
+
 ## Check Your Application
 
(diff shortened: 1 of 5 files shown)


docs/docs/scaling.md
```diff
@@ -0,0 +1,227 @@
+---
+sidebar_position: 5
```
phinze, line 2: 💡 We're colliding on a couple PRs here, I think we can just sort them out once it all lands.

docs/docs/scaling.md
```diff
@@ -0,0 +1,227 @@
...
+|------|-------------|----------|
+| `auto` | Scales instances based on traffic | Stateless web services, APIs |
+| `fixed` | Runs a set number of instances | Databases, workers, stateful services |
+
+## Default Behavior
+
+When you deploy without explicit configuration:
+
+- **`web` service**: Auto mode with 10 concurrent requests per instance, 15-minute scale-down delay
+- **All other services**: Fixed mode with 1 instance
```
phinze, line 48: ❓ Let's chat in our sync about how this interacts with MIR-559. Feels to me like we want to talk about services and scaling together, just want to talk through how to organize the pages.

phinze: Actually I think I see how this could work - we just link out to the services page here. Will stack on this for the service stuff if I get to it today.

phinze (approved):
Looking good!

docs/docs/getting-started.md
```diff
@@ -81,6 +81,21 @@ That's it! Miren will:
 
 Note: Your first app gets a default route automatically. For subsequent apps, you'll need to configure routes manually. See [Working with Routes](#working-with-routes).
 
+### "Why didn't I have to configure how many copies of my application to run?"
```
teresamychu: I think "Why didnt i..?" is too theoretical, we probably want the FAQ to be practical questions so they can easily find the question they're looking for. 

Something like: 

**"How do I configure multiple instances of my application?"**

And then the answer goes: 

**You dont! A core philosophy of Miren....**

evanphx: Ah perf!

PR #457, opened 2025-12-08
docker: Let container boot a bit before making UDP requests

evanphx:
The behavior is so bizarre. If we make an http3 request to the forwarded port before the container has had enough time to startup (note: doesn't have to actually wait for the server to listen on the port inside the container, just for the container to actually boot), then docker desktop removes it's listener for the port!

I have to presume it's some pathalogical condition in docker desktop itself, but since it's closed source, we can't pin point it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker server installation reliability by introducing an initialization delay during server startup. This addresses intermittent health check failures experienced in Docker Desktop environments, ensuring more stable and successful server deployments.

### DIFF
--- a/cli/commands/server_install_docker.go
+++ b/cli/commands/server_install_docker.go
@@ -104,6 +104,14 @@ func ServerInstallDocker(ctx *Context, opts struct {
 
 	// Wait for server to be ready
 	ctx.Info("Waiting for miren server to initialize...")
+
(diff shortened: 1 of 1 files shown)


teresamychu (approved):
😴

phinze (approved):
honk shoo 🕯️

PR #468, opened 2025-12-09
CLI error output now redirects the user to check server status and links demo server instructions

teresamychu:
Giving the user some clearer guidance in CLI output when the Miren client cannot find the cluster / has no server installed. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when no clusters are configured with guidance on checking server status
  * Enhanced setup instructions to include information about Miren's hosted demo server as an alternative option

### DIFF
--- a/cli/commands/cluster.go
+++ b/cli/commands/cluster.go
@@ -28,7 +28,7 @@ func ClusterList(ctx *Context, opts struct {
 				}
 				return PrintJSON([]ClusterInfo{})
 			}
-			ctx.Printf("No clusters configured\n")
+			ctx.Printf("No clusters configured. Is your miren server running? Check server status with `miren server status`\n")
 			ctx.Printf("\nUse 'miren cluster add' to add a cluster\n")
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


cli/commands/cluster.go
```diff
@@ -28,7 +28,7 @@ func ClusterList(ctx *Context, opts struct {
 				}
 				return PrintJSON([]ClusterInfo{})
 			}
-			ctx.Printf("No clusters configured\n")
+			ctx.Printf("No clusters configured. Is your miren server running? Check server status with `miren server status`\n")
```
evanphx, line 31: Why would the server running effect this output?

evanphx (approved):

cli/commands/cluster.go
teresamychu: It is not clear through the CLI output that a server needs to be running in order for the cluster list to be populated (on first use with no client configuration)

Users will most likely do a "miren deploy" or a "miren login" and will be presented with "no clusters" error, even though the real issue is that they are not connected to a server.

This message just prompts the user to check their server status so they can differ between cluster issues and server issues.

Based off of feedback from Simon in the discord


image

evanphx: Sure, but a server connection issue wouldn't show up at this level ever. In the case of simon, I feel like the issue is in what `miren apps` says, it should say "no cluster configured, add a cluster using `miren cluster add` or install one locally".

PR #469, opened 2025-12-10
Fix authentication bypass in local/non-cloud mode

evanphx:
## Summary

- Add `LocalOnlyAuthenticator` that requires valid client certificates for all requests
- Use `LocalOnlyAuthenticator` when cloud auth is disabled instead of `NoOpAuthenticator`
- Previously, unauthenticated requests could access the API when cloud auth was not enabled

## Summary by CodeRabbit

* **New Features**
  * Added client-certificate (local-only) authentication when cloud auth is not enabled.
  * Added a runtime option to fully disable authentication (intended for testing); startup logs emit clear warnings when auth is disabled.

* **Tests**
  * Added TLS/X.509 tests covering certificate acceptance and rejection across authentication scenarios.

### DIFF
--- a/pkg/cloudauth/rpc_authenticator.go
+++ b/pkg/cloudauth/rpc_authenticator.go
@@ -114,7 +114,8 @@ func (a *RPCAuthenticator) AuthenticateRequest(ctx context.Context, r *http.Requ
 	// Extract token from Authorization header
 	authHeader := r.Header.Get("Authorization")
 	if authHeader == "" {
-		// No auth header - could be certificate-based auth
+		// No auth header - shouldn't happen because rpc server only calls
+		// this method if an auth header is present
 		return false, "", nil
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/cloudauth/rpc_authenticator.go
```diff
@@ -114,7 +114,8 @@ func (a *RPCAuthenticator) AuthenticateRequest(ctx context.Context, r *http.Requ
 	// Extract token from Authorization header
 	authHeader := r.Header.Get("Authorization")
 	if authHeader == "" {
-		// No auth header - could be certificate-based auth
+		// No auth header - shouldn't happen because rpc server only calls
+		// this method if an auth header is present
```
phinze, line 118: Does that mean this should return an error then?

phinze (approved):
🔒 👍

pkg/cloudauth/rpc_authenticator.go
evanphx: It returns false indicating don't allow authentication.

evanphx: But you make a good point!

PR #470, opened 2025-12-10
entity: Don't enforce a rev check if fromRevision is 0

evanphx:
We have an early check for fromRevision being 0, but we were still using it later on when we actually perform the operation.

## Summary by CodeRabbit

* **Improvements**
  * Entity update/replace/patch operations now proceed when no prior revision is specified, avoiding unnecessary revision checks and allowing successive changes to apply smoothly.

* **Tests**
  * Added tests validating that update, replace, and patch succeed and advance revisions when no fromRevision is provided.

### DIFF
--- a/pkg/entity/store.go
+++ b/pkg/entity/store.go
@@ -626,11 +626,20 @@ func (s *EtcdStore) UpdateEntity(
 
 	txopt = append(txopt, coltxopt...)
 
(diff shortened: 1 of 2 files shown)


phinze (approved):
💭 Worth a stab at a test to cover this?

PR #473, opened 2025-12-10
Improve error message when server is not responding

teresamychu:
Show actionable steps instead of confusing network error when the miren server is unreachable (timeout, connection refused, etc).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling to provide clearer feedback when the server is not responding, improving user experience during network connectivity issues.

### DIFF
--- a/cli/commands/global.go
+++ b/cli/commands/global.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"net"
 	"os"
 	"os/signal"
 	"reflect"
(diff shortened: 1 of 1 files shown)


cli/commands/global.go
```diff
@@ -421,18 +422,36 @@ func (c *Context) RPCClient(name string) (*rpc.NetworkClient, error) {
...
+		// Check for connection/network errors
+		if resolveErr.Kind == rpc.ResolveHTTPError {
+			var netErr net.Error
+			if errors.As(resolveErr.Err, &netErr) {
+				c.Warn("server not responding at %s", c.Config.ServerAddress)
+				c.Info("")
+				c.Info("To fix:")
+				c.Info("  - Check if miren is running: sudo systemctl status miren")
+				c.Info("  - Start it: sudo systemctl start miren")
+				c.Info("  - Or use our demo cluster: https://miren.dev/docs/getting-started#using-our-demo-cluster")
```
evanphx, line 440: Since they've already added a cluster, I think we leave off the demo cluster info here. I'd say we should make this a bit wider, maybe something like:

```
- Check that the server is running
- Check that your client can reach the server at %s
- If the server is local, check that it's started: sudo systemectl status miren
```

evanphx (approved):

teresamychu:
Actually revamped to have more specific guidance CLI output for different errors in a new PR. Closing this one.

PR #477, opened 2025-12-11
lsvd: use Go File.Write instead of raw unix.Write for NBD responses

phinze:
## Summary

- Replace `unix.Write` syscall with Go's `*os.File.Write()` for NBD read responses
- Go's Write handles transient errors (EAGAIN) with internal retry logic, which the raw syscall does not

## Context

NBD transport errors like "resource temporarily unavailable" could cause the disk-lease controller to hang when socket buffer pressure led to EAGAIN errors that weren't being retried.

The sendfile path is unchanged since `unix.Sendfile` is significantly faster and we haven't observed the same issue there yet.

## Test plan

- [x] Deployed to miren-garden and verified disk provisioning/mounting works
- [x] Monitored logs for NBD errors during disk operations - none observed
- [ ] Long-term monitoring to verify fix under load conditions that previously triggered the issue

Part of: MIR-578

### DIFF
--- a/lsvd/nbd.go
+++ b/lsvd/nbd.go
@@ -153,8 +153,6 @@ func (n *nbdWrapper) ReadIntoConn(b []byte, off int64, output *os.File) (bool, e
 		return false, err
 	}
 
-	wfd := output.Fd()
-
 	var written int
 
 	left := len(b)
@@ -164,9 +162,9 @@ func (n *nbdWrapper) ReadIntoConn(b []byte, off int64, output *os.File) (bool, e
 
 		off := 0
 		for left > 0 {
-			written, err = unix.Write(int(wfd), b[off:])
+			written, err = output.Write(b[off:])
 			if err != nil {
-				n.log.Error("error sending data via write(2)", "error", err)
+				n.log.Error("error sending data via Write", "error", err)
 				return true, nil
 			}
 
@@ -184,6 +182,7 @@ func (n *nbdWrapper) ReadIntoConn(b []byte, off int64, output *os.File) (bool, e
 		return true, nil
 	}
 
+	wfd := output.Fd()
 	rfd := cps.fd.Fd()
 	sendfileResponses.Inc()
 


evanphx (approved):

controllers/disk/lsvd_client.go
```diff
@@ -1393,16 +1451,9 @@ func (c *lsvdClientImpl) formatDevice(ctx context.Context, devicePath string, fi
...
 		if _, err := ext4.ReadExt4SuperBlock(devicePath); err == nil {
 			formatted = true
-			// Run fsck to ensure filesystem is clean with timeout to prevent indefinite hangs
-			fsckCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
-			defer cancel()
-
-			cmd := exec.CommandContext(fsckCtx, "e2fsck", "-f", "-y", devicePath)
-			output, err := cmd.CombinedOutput()
+			// Run fsck to ensure filesystem is clean
+			output, err := c.runCommandWithNBDTimeout(ctx, nbdIndex, 2*time.Minute, "e2fsck", "-f", "-y", devicePath)
```
evanphx: We could make the argument that 2 minutes isn't enough for a big e2fsck. Maybe we should make this timeouts bigger and/or if they timeout, see if we can issue a read on the nbd ourselves to check if it's a big check or if the device is broken?

evanphx (approved):

controllers/disk/lsvd_client.go
phinze: Mmm good point 💭 

Will stare at it a bit more and maybe back it out for more noodling

phinze: Yeah ok good catch on this - decided to back out the timeout/disconnect change for now.

Digging into it more, there are a few interrelated issues that need more thought:

1. **Lock contention**: `MountVolume` holds `c.mu.Lock()` for the entire duration including `formatDevice`. A long e2fsck would block other disk operations on the same LSVD client
instance.

2. **Timeout scaling**: Looked at linear scaling by disk size (e.g. 2 min + 1 min per 10gb disk size, but that gets unwieldy for large disks (we're cloud-backed with no max size). Capping at 30min still means 30min of blocked disk ops.

3. **Forced fsck**: We're running `e2fsck -f` which forces a check even on clean filesystems. Without `-f`, checks would only run after crashes and be much faster in the normal case.

Shipping just the `File.Write` fix for now since it's a clear win. Will regroup on the timeout/recovery approach - probably needs a combination of:
- Removing forced fsck (`-f` flag)
- Per-volume locking or releasing lock during long ops
- Then timeout + disconnect as a last-resort recovery

PR #486, opened 2025-12-15
Enable BuildKit layer caching with persistent storage

phinze:
## Summary

BuildKit layer caching should make subsequent deploys fast by reusing cached build layers. However, caching wasn't working at all - every deploy was doing a full rebuild.

## What wasn't working

The previous implementation attempted local filesystem caching but had several issues:

1. **Inaccessible cache directory** - The cache dir was created on the host at `{tempDir}/buildkit-cache`, but BuildKit runs in an ephemeral sandbox that couldn't access this path
2. **Missing cache export mode** - BuildKit logs showed `skipping invalid cache export mode: ""` indicating misconfigured cache options
3. **Ephemeral sandboxes** - Even if the cache worked, BuildKit sandboxes are short-lived and any internal state was lost between builds

## What we changed

**Persistent BuildKit storage:**
- Mount a persistent host volume at `{dataPath}/buildkit` into the BuildKit sandbox at `/var/lib/buildkit`
- This preserves BuildKit's internal layer cache, snapshots, and build history across builds
- This is the primary mechanism for caching - BuildKit's internal layer cache now persists and subsequent builds show `CACHED` for unchanged layers

**Registry-based inline caching:**
- Enable `BUILDKIT_INLINE_CACHE=1` which embeds cache metadata in exported images
- Configure cache imports from the registry (attempts to pull cache from previously built image)
- Note: We generate a unique image tag for every deploy, so registry cache imports will currently always miss. The layer caching from persistent storage is what provides the speedup. Registry caching is configured for potential future use with stable tags.

**Friendlier error handling:**
- Cache import "not found" is expected on every build (since we use unique image tags)
- Changed output from `#1 ERROR: failed to configure registry cache importer...` to `#1 DONE (cache not found)`
- Switched to vendored progresswriter so these UI changes apply consistently

## Test plan
- [x] First deploy shows "cache not found" (expected)
- [x] Second deploy shows CACHED for unchanged build layers
- [x] Build output shows friendly "DONE (cache not found)" instead of ERROR
- [x] No error block printed at end of successful builds
- [x] `make lint` passes

## Release Notes

* **Bug Fixes**
  * Cache import failures no longer incorrectly trigger build errors; these are now handled gracefully during deployment.

* **New Features**
  * Persistent BuildKit state storage via host volumes for more efficient builds.
  * Registry-based inline caching for faster image rebuilds with improved cache efficiency.

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -12,8 +12,9 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/moby/buildkit/client"
-	"github.com/moby/buildkit/util/progress/progresswriter"
 	"golang.org/x/term"
+
+	"miren.dev/runtime/pkg/progress/progresswriter"
 
 	"miren.dev/runtime/api/build/build_v1alpha"
 	"miren.dev/runtime/api/deployment/deployment_v1alpha"
@@ -578,6 +579,10 @@ func createBuildStatusCallback(
 			// Extract error messages from status
 			for _, vertex := range status.Vertexes {
 				if vertex.Error != "" {
+					// Skip cache import errors - these are expected when no cached image exists
+					if isCacheImportError(vertex.Name, vertex.Error) {
+						continue
+					}
 					*buildErrors = append(*buildErrors, vertex.Error)
 				}
 			}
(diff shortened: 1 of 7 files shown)


phinze:
Ah hang on, this does speed things up (via persistent local layer cache) but it breaks concurrency for multiple deploys, since buildkit does not support having multiple copies of itself sharing data dirs. I think the fix here is to rework buildkit into a longer-running service. It'll be a bigger change, but I think it's the right direction. Closing this for now!

PR #487, opened 2025-12-15
Add --service flag to filter logs by service name

phinze:
## Summary
- Adds `--service` flag to `miren logs` command for filtering by service name (e.g., `web`, `worker`)
- Service filter is combined with existing `--grep` filter for server-side LogsQL filtering
- Adds upgrade warning for users on older servers without optimized log streaming

## Usage
```bash
# Filter logs to only show 'web' service
miren logs --service web

# Combine with text filter
miren logs --service worker -g "error"

# With other options
miren logs -a myapp --service web -f --last 1h
```

## Test plan
- [x] Unit tests for `buildFilterWithService` helper
- [x] Unit tests for `normalizeSandboxID` helper
- [x] `go build` passes
- [x] `make lint` passes

## Summary by CodeRabbit

* **New Features**
  * Added `--service` flag to filter logs by service name, enabling targeted log viewing.
  * Support for combining service filtering with existing text filters.

* **Documentation**
  * Updated with dedicated "Filtering by Service" section and usage examples.

* **Improvements**
  * Added server version compatibility check; warns if `--service` is used with older server versions.

### DIFF
--- a/cli/commands/logs.go
+++ b/cli/commands/logs.go
@@ -22,6 +22,19 @@ func normalizeSandboxID(sandboxID string) string {
 	return "sandbox/" + sandboxID
 }
 
+// buildFilterWithService combines a user filter with a service filter for LogsQL.
+// Service filter is added as a field match: service:"value"
+func buildFilterWithService(userFilter, service string) string {
+	if service == "" {
+		return userFilter
+	}
+	serviceFilter := fmt.Sprintf("service:%q", service)
+	if userFilter == "" {
+		return serviceFilter
+	}
+	return serviceFilter + " " + userFilter
+}
+
 func Logs(ctx *Context, opts struct {
 	ConfigCentric
 
@@ -31,6 +44,7 @@ func Logs(ctx *Context, opts struct {
 	Sandbox string         `short:"s" long:"sandbox" description:"Show logs for a specific sandbox ID"`
 	Follow  bool           `short:"f" long:"follow" description:"Follow log output (live tail)"`
 	Filter  string         `short:"g" long:"grep" description:"Filter logs (e.g., 'error', '\"exact phrase\"', 'error -debug', '/regex/')"`
+	Service string         `long:"service" description:"Filter logs by service name (e.g., 'web', 'worker')"`
 }) error {
 	// Check for conflicting options
 	if opts.App != "" && opts.Sandbox != "" {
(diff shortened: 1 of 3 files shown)


evanphx (approved):

teresamychu:
LOL i dont know if you created this PR before or after my comments on your RFD, but I left a comment about a --service flag and a prompt for the user to upgrade their server on your RFD earlier today.

so i apologize if you made this PR first for my duplicate comment.

cli/commands/logs.go
```diff
@@ -75,10 +89,20 @@ func Logs(ctx *Context, opts struct {
...
+	combinedFilter := buildFilterWithService(opts.Filter, opts.Service)
+
 	// Check if server supports streaming (prefer chunked for efficiency)
 	if cl.HasMethod(ctx, "streamLogChunks") {
-		return streamLogChunks(ctx, cl, opts.App, opts.Sandbox, opts.Last, opts.Follow, opts.Filter)
+		return streamLogChunks(ctx, cl, opts.App, opts.Sandbox, opts.Last, opts.Follow, combinedFilter)
+	}
+
+	// Older server - warn about upgrade and limited functionality
+	ctx.Printf("Warning: server does not support optimized log streaming. Upgrade your server for better performance and --service filtering.\n")
```
teresamychu, line 101: 💭 we can probably give the user the explicit command to upgrade the server. 

```suggestion
	ctx.Printf("Warning: server does not support optimized log streaming. Use `miren server upgrade` to upgrade your server for better performance and --service filtering.\n")
```

teresamychu (approved):

cli/commands/logs.go
phinze: I love the thought and in general big +1 to specific guidance! But in this case I'm hesitant to give a specific command if we're not certain the user can run it immediately and get the correct behavior. Since we're on the client side here, we don't know if we're talking to a server the user controls, or if they're potentially using `miren server docker`. If they were to run `miren server upgrade` immediately it could give them some confusing output.

phinze:
> LOL i dont know if you created this PR before or after my comments on your RFD

It was in response to your comments! So you can take credit for all this 😄

cli/commands/logs.go
phinze: Gonna merge as is but happy to tweak in separate PR if you want to talk about this more!

PR #489, opened 2025-12-16
Implement exec command functionality

evanphx:
## Summary
- Add `miren app exec` command to create ephemeral sandboxes for interactive shell access to apps
- Re-enable `miren sandbox exec` command to exec into existing sandboxes by ID
- Implement WatchEntity-based monitoring for efficient sandbox status detection (replacing polling)
- Add comprehensive test coverage for exec proxy including mock sandbox controller

## Changes
- `cli/commands/app_exec.go` - New command for `miren app exec -a <appname>` (renamed from console.go)
- `cli/commands/sandbox_exec.go` - Fixed to use "id" category for sandbox lookup
- `servers/exec_proxy/exec_proxy.go` - Ephemeral sandbox creation with WatchEntity monitoring
- `pkg/entity/mock.go` - Added WatchEntity implementation with watcher notification
- `pkg/entity/testutils/mock_sandbox_controller.go` - Mock controller for testing sandbox lifecycle

## Test plan
- [x] All exec_proxy tests pass
- [x] Manual test: `miren app exec -a <appname>` creates ephemeral sandbox and opens shell
- [x] Manual test: `miren sandbox exec --id <sandbox-id>` opens shell in existing sandbox

## Summary by CodeRabbit

* **New Features**
  * Added interactive shell commands: "sandbox exec" (existing sandbox) and "app run" (new ephemeral sandbox).
  * Ephemeral sandbox creation and automatic cleanup for app run.
  * Enabled TTY and stdin support for container runtimes.
  * Terminal window resize now supported during exec sessions.

* **Tests**
  * Added mock sandbox controller and comprehensive tests for sandbox lifecycle and exec flows.

* **Documentation**
  * CLI docs updated with app run and sandbox exec usage and examples.

* **Bug Fixes**
  * Improved RPC/stream robustness to avoid leftover-argument issues.


servers/exec_proxy/exec_proxy.go
```diff
@@ -143,3 +157,205 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+	appResp, err := s.EAC.Get(ctx, app.ID.String())
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get app metadata: %w", err)
+	}
+
+	var appMD core_v1alpha.Metadata
+	appMD.Decode(appResp.Entity().Entity())
+
+	// Create sandbox entity
+	sbName := idgen.GenNS(fmt.Sprintf("%s-web", appMD.Name))
```
phinze: 💡 Prolly let's call this `%s-exec` yeah?

pkg/entity/testutils/mock_sandbox_controller.go
```diff
@@ -0,0 +1,350 @@
...
+	Log *slog.Logger
+	EAC *entityserver_v1alpha.EntityAccessClient
+
+	// PollInterval controls how often the controller checks for pending sandboxes.
+	// Defaults to 50ms if not set.
+	PollInterval time.Duration
+
+	// StartupDelay is the delay before transitioning a sandbox from PENDING to RUNNING.
+	// Defaults to 0 (immediate transition).
+	StartupDelay time.Duration
```
phinze, line 28: 💡 This has some features (StartupDelay, callbacks, per-sandbox MarkFailed) that aren't used by the exec_proxy tests - they're only exercised by the mock's own unit tests.

Is that intentional? Like, building out the utility for anticipated future test needs? Or could we trim it down to just what's needed now and add features when actual tests need them?

Not a blocker either way, just curious about the thinking here.

phinze (commented):
Looking good so far! A few wee nonblockers and will finish testing tomorrow 👍

pkg/entity/testutils/mock_sandbox_controller.go
evanphx: Pure anticipation as I had Claude flesh it out.

servers/exec_proxy/exec_proxy.go
evanphx: Done!

pkg/entity/testutils/mock_sandbox_controller.go
phinze: Understood! I'm vaguely hesitant to have so much functionality in a mock, directionally I'm hoping we can move towards carving out collaborators so we can use the real SandboxController. But all this will change so much with saga refactoring that it's not worth thinking about too much at this stage.

cli/commands/app_exec.go
```diff
@@ -12,9 +12,8 @@ import (
 	"miren.dev/runtime/pkg/rpc/stream"
 )
 
-func Console(ctx *Context, opts struct {
+func AppExec(ctx *Context, opts struct {
```
phinze: 💭 Coming at this after sleeping on it, I wonder if it would be clearer to call this `app run`.

That would align with [`docker run`](https://docs.docker.com/reference/cli/docker/container/run/) and [`heroku run`](https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-run) both of which launch new one-off containers.

And it would prevent confusion from overloading `exec` as a verb w/ two fairly distinct behaviors.

phinze (approved):
Working well in testing! One outstanding thought on command naming.

PR #490, opened 2025-12-16
Persistent BuildKit daemon for faster builds (RFD-43)

phinze:
## Summary

Implements RFD-43: Replace ephemeral BuildKit sandboxes with a persistent daemon for faster container builds.

- **Persistent daemon**: BuildKit runs as a long-lived containerd container, eliminating ~10s startup overhead per build
- **Layer caching**: Daemon maintains internal cache across builds - no more cold starts
- **External mode**: Distributed deployments can point to external BuildKit via `--buildkit-socket`
- **Clean design**: Removed 300+ lines of ephemeral sandbox code

## Key Changes

| File | Change |
|------|--------|
| `components/buildkit/` | New component managing BuildKit lifecycle |
| `pkg/serverconfig/schema.yml` | Added `[buildkit]` config section |
| `cli/commands/server.go` | Start embedded or connect to external BuildKit |
| `servers/build/build.go` | Use component's `Client()` instead of launching sandboxes |
| `servers/build/launch.go` | **Deleted** - ephemeral sandbox code removed |

## Config Options

```toml
[buildkit]
start_embedded = true           # default in standalone mode
socket_path = ""                # for external BuildKit (distributed mode)
socket_dir = "/run/miren/buildkit"
gc_keep_storage = "10GB"
gc_keep_duration = "7d"
```

## Test plan

- [ ] `make test` passes
- [ ] Deploy in standalone mode, verify builds work with persistent daemon
- [ ] Verify layer cache persists across builds (second build faster)
- [ ] Test external BuildKit mode with `--buildkit-socket`

## Summary by CodeRabbit

* **New Features**
  * BuildKit integration (embedded or external) with configurable socket, GC storage/duration, and registry host handling; BuildKit lifecycle and metrics exposed to server and build flows.

* **Chores / Refactor**
  * Replaced ephemeral launches with a persistent BuildKit daemon; simplified wiring so build processes use a shared BuildKit component.
  * Added configuration block and defaults for BuildKit settings and CLI/env mappings.

* **Tests**
  * Added BuildKit lifecycle tests; removed one obsolete build test case.


cli/commands/server.go
```diff
@@ -1020,3 +1082,78 @@ func stopAllSandboxContainers(ctx context.Context, log *slog.Logger, cc *contain
 	log.Info("stopped sandbox containers", "count", stoppedCount)
 	return nil
 }
+
+// parseStorageSize parses a human-readable storage size (e.g., "10GB") to bytes
+func parseStorageSize(s string) int64 {
```
evanphx: We have https://github.com/mirendev/runtime/blob/main/pkg/units/data.go#L174 to do this instead

components/buildkit/buildkit.go
```diff
@@ -0,0 +1,484 @@
...
+	image, err := c.CC.Pull(ctx, buildkitImage, containerd.WithPullUnpack)
+	if err != nil {
+		return fmt.Errorf("failed to pull buildkit image: %w", err)
+	}
+
+	// Set up paths
+	dataPath := filepath.Join(c.DataPath, "buildkit")
+	socketDir := config.SocketDir
+	if socketDir == "" {
+		socketDir = "/run/miren/buildkit"
```
evanphx, line 123: Should we put this along side the containerd socket instead, which defaults to inside the datadir (defaulting to /var/lib/miren).

components/buildkit/buildkit.go
```diff
@@ -0,0 +1,484 @@
...
+
+	if !running {
+		return nil, fmt.Errorf("buildkit component not running")
+	}
+
+	return buildkitclient.New(ctx, "unix://"+socketPath)
+}
+
+func (c *Component) generateConfig(gcKeepStorage, gcKeepDuration int64, registryHost string) string {
+	if registryHost == "" {
```
evanphx, line 270: I'm a bit surprised that we aren't having to inject on /etc/hosts entry for the registry host. How is buildkit resolving the host address?

components/buildkit/buildkit.go
```diff
@@ -0,0 +1,484 @@
...
+		oci.WithImageConfig(image),
+		// No host networking - BuildKit uses its own network namespace
+		// This is more secure than host networking since BuildKit doesn't need
+		// to accept external TCP connections (we use Unix sockets)
+		oci.WithPrivileged, // Required for BuildKit
+		oci.WithProcessArgs(
+			"/usr/bin/buildkitd",
+			"--config=/etc/buildkit/buildkitd.toml",
+		),
+		oci.WithHostHostsFile,
```
evanphx: This means that we're using /etc/hosts and /etc/resolv.conf from the host, but they won't have the cluster.local entry...

evanphx (changes requested):
The only thing is that I'm concerned about how buildkit will resolve cluster.local.

components/buildkit/buildkit.go
```diff
@@ -0,0 +1,539 @@
...
+debug = true
+root = "/var/lib/buildkit"
+insecure-entitlements = [ "network.host", "security.insecure" ]
+
+[log]
+  format = "text"
+
+[dns]
+  nameservers=["1.1.1.1","8.8.8.8"]
+  options=["edns0"]
```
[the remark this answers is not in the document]
phinze: Gonna skip this for now, the old code had these hardcoded too. It's something we can address separately down the line.

cli/commands/server.go
phinze: Nice, flopped out.

components/buildkit/buildkit.go
phinze: Good call, moved in https://github.com/mirendev/runtime/pull/490/commits/e874e1c0b258e9f28e95fabfc54b7ac923941643

components/buildkit/buildkit.go
phinze: Yeah this was my bad, the code as I was testing it had left in the "fallback" of the old ephemeral container and I noticed it right at the end and ripped it out, not realizing I was only testing the fallback. Now I've actually tested it!  😅

components/buildkit/buildkit.go
phinze: Now we're injecting the cluster.local entry properly

evanphx (approved):
Looks good!

PR #491, opened 2025-12-17
Miren doctor improvements

teresamychu:
Miren Doctor now recoomends commands for users to run instead of running commands for them. 

`miren doctor`

image

-------------------------------

`miren doctor config`
image
image

-------------------------------

`miren doctor auth`
image
image

-------------------------------

`miren doctor server`
image

image

## Summary by CodeRabbit

* **New Features**
  * Clickable routes in CLI app listings and richer Ready/Desired/Health/Routes output.
  * Interactive guided diagnostics (config, server, auth) with picker-driven help.
  * Streamlined login flow with sensible defaults.

* **Improvements**
  * Better connectivity checks (HTTP/HTTPS) and tailored local vs remote guidance.
  * Config view separates local/remote clusters and shortens displayed paths.
  * Table/JSON status semantics refined for clearer health reporting.

* **Removals**
  * Removed legacy diagnostic command aliases and aggregated apps/routes diagnostics.


cli/commands/doctor_auth.go
```diff
@@ -74,5 +75,49 @@ func DoctorAuth(ctx *Context, opts struct {
...
+			return nil
+		}
+
+		switch selected.ID() {
+		case "How do I login with a different account?":
+			showLoginDifferentAccountHelp(ctx)
+		case "How do I add authentication to this server?":
+			showAddAuthToServerHelp(ctx)
+		}
+	}
```
evanphx: I still think that we want to do this as more of a guide setup later. So we should just display the question and then the output, no prompt.

cli/commands/doctor_config.go
```diff
@@ -2,40 +2,180 @@ package commands
...
+		}
+
+		switch selected.ID() {
+		case "How do I add an existing cluster?":
+			showAddClusterHelp(ctx, cfg)
+		case "How do I register a new cluster?":
+			showRegisterClusterHelp(ctx, existing)
+		}
+	}
+}
```
evanphx: Same as the other, let's just display the questions and the answers rather than prompting them.

cli/commands/doctor_server.go
```diff
@@ -48,55 +54,211 @@ func DoctorServer(ctx *Context, opts struct {
...
+
+		switch selected.ID() {
+		case "How do I start a local server?":
+			showStartLocalServerHelp(ctx)
+		case "How do I connect to a known remote server?":
+			showConnectRemoteServerHelp(ctx)
+		case "How do I fix https/http connectivity?":
+			showFixConnectivityHelp(ctx)
+		}
+	}
```
evanphx: Ditto again.

evanphx (changes requested):
Let's not prompt them, I think we want to think through a proper guided setup rather than adding them into doctor.

teresamychu:
> Let's not prompt them, I think we want to think through a proper guided setup rather than adding them into doctor.

Got it! Updated to just print the question and the commands without prompting the user.

image
image
image

evanphx (approved):
Looks good!

PR #493, opened 2025-12-17
cli: Migrate from mitchellh/cli to miren.dev/mflags

evanphx:
Replace the mitchellh/cli command-line framework with mflags, a custom flag parsing library. This provides more control over command dispatch, flag handling, and help output generation.

The big change with this vs our old approach is the ability to have interspersed arguments. Ie `miren -a blah app run` works now and didn't before because mitchellh/cli did command dispatching before flag resolution.

Key changes:
- Refactor command registration to use mflags Dispatcher
- Update Infer() to return mflags.Executor instead of cli.Command
- Add ErrExitCode for explicit exit code control
- Enhance admin command with unknown flag passthrough for method params
- Add method-specific help output for admin commands
- Update platform-specific command registration (Linux/other)

## Summary by CodeRabbit

* **Refactor**
  * Migrated the CLI to a dispatcher-based framework and re-registered commands.
  * Flattened many nested positional/rest argument groups into direct fields for simpler command usage.
  * Revised help/flag parsing and expanded debug-related subcommands.

* **Bug Fixes**
  * Improved CLI exit code and help behavior for clearer error reporting.

* **Chores**
  * Added a public Version() function.
  * Updated module dependencies to support the new CLI framework.


cli/commands/config_set_active.go
```diff
@@ -5,17 +5,15 @@ import (
 )
 
 func ConfigSetActive(ctx *Context, opts struct {
```
phinze, line 7: 💡 This is dead code, replaced w/ `cluster switch`

cli/commands/config_remove.go
```diff
@@ -5,11 +5,9 @@ import (
 )
 
 func ConfigRemove(ctx *Context, opts struct {
```
phinze, line 7: 💡 This is dead code, replaced w/ `cluster remove`

phinze (approved):
Looks good! Noticed a couple of old unregistered commands in the diff we can either remove now or after merge.

cli/commands/config_set_active.go
evanphx: Yeah, i'll remove it post merge.

PR #496, opened 2025-12-18
Fix stale pool reference when cached pool is at max size

phinze:
## Summary

- Fixes stale pool reference bug when cached pool has `DesiredInstances >= MaxPoolSize` (20)
- Adds verification that pool still exists before returning max size error
- Adds regression test that validates the fix

## Problem

When the activator has a cached pool at max size and that pool has been deleted from the entity store, it incorrectly returns "pool has reached maximum size of 20" instead of detecting that the pool no longer exists and clearing the stale cache.

This is because the max size check short-circuits before attempting any operation (like Patch) that would trigger `ErrNotFound` detection.

## Root Cause (from miren-club incident)

1. Pool `pool-CWDXw3RYWVicow14yasX4` was deleted at 11:22:35 UTC
2. Activator still had stale reference with `DesiredInstances=20`
3. New pool `pool-CWEQeWqXPFYPKwvDneqNM` was created by DeploymentLauncher at 15:43:05 UTC  
4. User requests at 16:07:32 UTC hit the stale max size check
5. Error "pool has reached maximum size of 20" returned incorrectly

## Fix

Before returning the max size error, verify the pool still exists in the entity store. If `ErrNotFound`, clear the stale cache and continue to re-query for pools.

This is a follow-up fix to PR #483 which added stale reference detection during Patch operations but didn't cover the max size check path.

## Test plan

- [x] Added `TestActivatorDeletedPoolAtMaxSize` regression test
- [x] Test fails without fix, passes with fix
- [x] All existing activator tests pass
- [x] Linter passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Added verification logic to detect when resources have been deleted while at maximum capacity, improving error handling by clearing stale cache references instead of returning misleading capacity-limit errors. The system now retries querying available resources.

* **Tests**
  * Added regression test ensuring the system correctly identifies deleted resources and handles them appropriately when operating at maximum capacity.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -501,9 +501,28 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
 			for attempt := 0; attempt < maxRetries; attempt++ {
 				a.mu.Lock()
 				if state.pool.DesiredInstances >= MaxPoolSize {
+					poolID := state.pool.ID
 					a.mu.Unlock()
+
+					// Before returning max size error, verify the pool still exists.
+					// The cached state may be stale if the pool was deleted and recreated.
+					_, getErr := a.eac.Get(ctx, poolID.String())
+					if getErr != nil {
+						if errors.Is(getErr, cond.ErrNotFound{}) {
+							a.log.Info("pool at max size was deleted, clearing stale reference",
+								"pool", poolID,
+								"service", service)
+							a.mu.Lock()
+							delete(a.pools, key)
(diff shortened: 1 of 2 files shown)


evanphx (approved):

phinze:
On deeper investigation I think there are better fixes for this than a check at the leaf of a conditional tree, #497 is one fix closer to the root cause, and I think Activator could use a `watchPools` goroutine to catch extraneous deletions. Closing in favor of those directions.

PR #500, opened 2025-12-19
install: Default to the branch the binary uses (or latest)

evanphx:
The code originally default to main, even when using a proper release binary. This updates the process to download the base code that matches the same release by default, while leaving the ability for the user to specific indicate they want main, latest, or a specific version by specifying it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved branch selection for release downloads and server installations: the CLI now derives the branch from runtime version metadata when not explicitly provided, and falls back to "latest" if no branch information exists. This replaces previous hardcoded defaults and ensures consistent behavior across download and install workflows.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -38,21 +38,13 @@ func GetInfo() Info {
 	return info
 }
 
-// Returns the branch that the version was built from
+// Branch returns the branch name if the binary was built from a branch (e.g. "main:abc123").
+// Returns empty string for tagged releases (e.g. "v0.2.0") or unknown versions.
 func Branch() string {
-	if Version == "unknown" {
-		return "unknown"
-	}
-
 	if branch, _, ok := strings.Cut(Version, ":"); ok {
 		return branch
 	}
-
-	if strings.HasPrefix(Version, "v") {
-		return Version
-	}
-
-	return "unknown"
+	return ""
 }
 
 // String returns the version info as a formatted string
--- a/version/version.go
+++ b/version/version.go
@@ -3,6 +3,7 @@ package version
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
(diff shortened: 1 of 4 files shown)


teresamychu (approved):

version/version.go
```diff
@@ -37,6 +38,23 @@ func GetInfo() Info {
...
+		return "unknown"
+	}
+
+	if branch, _, ok := strings.Cut(Version, ":"); ok {
+		return branch
+	}
+
+	if strings.HasPrefix(Version, "v") {
+		return Version
+	}
```
phinze: ⚠️ Combined with the defaulting behavior added to `server install` above, when you're on a version won't this always reinstall the same version?

e.g. when on `v0.2.0`, `version.Branch()` returns `"v0.2.0"`, which is set to `opts.Branch` above

phinze (changes requested):
Nice catch on the miss of this command! I think there's a looping behavior I called out inline unless I'm reading it wrong

phinze:
Pushed a fix in f96c2b4. Changed Branch() to return empty string for tagged releases instead of returning the tag itself.

phinze (approved):

teresamychu:
> Pushed a fix in [f96c2b4](https://github.com/mirendev/runtime/commit/f96c2b447484b522d131e88d938b9d8f84d44e51). Changed Branch() to return empty string for tagged releases instead of returning the tag itself.

LGTM!

version/version.go
evanphx: That was the intention, yup. Is there a reason that's the wrong behavior?

phinze: You could never upgrade - upgrade would just redownload the current tag

PR #511, opened 2025-12-24
Fix WebSocket over HTTP/2 by disabling extended CONNECT

phinze:
## Summary

- Fixes WebSocket connections failing with 502 Bad Gateway when clients use HTTP/2
- Sets `GODEBUG=http2xconnect=0` at server startup to disable RFC 8441 extended CONNECT advertisement
- Browsers fall back to HTTP/1.1 for WebSocket while keeping HTTP/2 for regular traffic

## Problem

When a browser connects via HTTP/2 (the default for modern browsers), WebSocket upgrade requests fail with 502 Bad Gateway. The same request works when forced to HTTP/1.1.

**Root cause:** Go's HTTP/2 server can advertise support for RFC 8441 extended CONNECT protocol. When browsers see this, they attempt WebSocket-over-HTTP/2 using the extended CONNECT method. However, Go's reverse proxy doesn't fully support this yet ([golang/go#53208](https://github.com/golang/go/issues/53208)), causing the proxy to fail.

Evidence:
```bash
# HTTP/2 (fails)
curl -v "https://example.com/socket.io/?transport=websocket" \
  -H "Connection: Upgrade" -H "Upgrade: websocket"
# Result: 502 Bad Gateway, server logs show "proxy error: EOF"

# HTTP/1.1 (works)  
curl --http1.1 "https://example.com/socket.io/?transport=websocket" \
  -H "Connection: Upgrade" -H "Upgrade: websocket"
# Result: HTTP/1.1 101 Switching Protocols ✓
```

## Solution

The fix sets `GODEBUG=http2xconnect=0` at server startup, which:

1. **Keeps HTTP/2 enabled** for all regular traffic (preserving multiplexing, header compression, etc.)
2. **Disables extended CONNECT advertisement** so browsers don't attempt WebSocket-over-HTTP/2
3. **Browsers fall back to HTTP/1.1** for WebSocket connections, which works correctly with our existing upgrade handling

This is the same approach used by Traefik ([traefik/traefik#11408](https://github.com/traefik/traefik/pull/11408)).

## Alternatives Considered

1. **Disable HTTP/2 entirely** - Would lose HTTP/2 benefits for all traffic
2. **Implement RFC 8441 support** - Go's ecosystem doesn't have production-ready support yet; would require significant custom work

## References

- [golang/go#71128](https://github.com/golang/go/issues/71128) - Go discussion on disabling extended CONNECT by default
- [golang/go#53208](https://github.com/golang/go/issues/53208) - RFC 8441 support proposal
- [traefik/traefik#11408](https://github.com/traefik/traefik/pull/11408) - Traefik's fix using same approach

## Test Plan

- [ ] Deploy to test environment
- [ ] Verify WebSocket connections work from browsers using HTTP/2
- [ ] Verify regular HTTP/2 traffic still works correctly
- [ ] Verify HTTP/1.1 WebSocket connections continue to work

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP/2 protocol compatibility and stability through server configuration adjustments.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -55,6 +55,14 @@ import (
 )
 
 func Server(ctx *Context, opts serverconfig.CLIFlags) error {
+	// Disable HTTP/2 extended CONNECT protocol (RFC 8441) to ensure WebSocket
+	// connections work correctly. When extended CONNECT is enabled, browsers
+	// attempt WebSocket-over-HTTP/2, but Go's reverse proxy doesn't support
+	// this yet. With it disabled, browsers fall back to HTTP/1.1 for WebSocket
+	// while still using HTTP/2 for regular traffic.
+	// See: https://github.com/golang/go/issues/71128
+	os.Setenv("GODEBUG", appendGodebug(os.Getenv("GODEBUG"), "http2xconnect=0"))
+
 	eg, sub := errgroup.WithContext(ctx)
 
 	// Load configuration from all sources with precedence:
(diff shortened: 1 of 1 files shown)


phinze:
On a fresh look, I think this was all a red herring based on Go 1.24 behavior that's since been cleaned up in Go 1.25. I believed I had a server exhibiting problems running main w/ the TimeoutHandler fix though... going to keep investigating.

PR #513, opened 2026-01-05
Panics during deployment should be marked as failed

teresamychu:
Deployments that panic currently do not get marked as "failed" status. As a result, the app is then locked from further deployments until the lock times out after 30 mins. 

The RPC server has panic recovery, but the panic recover actually repanics so the panic itself is not properly surfaced.

- Updated the RPC server to properly handle panics by logging the error and letting it propogate up to the client.
- Updated client to handle panics in streaming responses.
- Added specific panic error handling to return helpful error messages to the user. 
- Fixed inconsistency where non-streaming RPC calls return RemoteError instead of cond.Panic(errs) like `CallWithCaps` does. 
- Added RPC test for panic error handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-panic detection during deployment with clearer user-facing messages and updated deployment status handling
  * RPC handling now consistently propagates panic information so clients receive explicit panic errors
  * Server-side panics are logged with stack traces (no re-panicking), improving failure visibility

* **Tests**
  * Added tests to verify panic propagation from server to client

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -19,6 +20,7 @@ import (
 	"miren.dev/runtime/api/deployment/deployment_v1alpha"
 	"miren.dev/runtime/appconfig"
 	"miren.dev/runtime/clientconfig"
+	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/deploygating"
 	"miren.dev/runtime/pkg/git"
 	"miren.dev/runtime/pkg/progress/upload"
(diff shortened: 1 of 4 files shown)


phinze (approved):
Hey Teresa! Took a look at this w/ Claude.

At first I was concerned about unconditional panic recovery across the entirety of our RPC handlers - my instinct was we should "let it crash" to avoid running with potentially corrupted state.

But after digging around, I think skipping re-panic is a reasonable approach:

1. **http3 does the same thing** - quic-go recovers panics per-request. If the Go networking experts consider this safe, that's a good signal.

2. **Our RPC handlers are stateless** - audited the full surface area (EntityServer, DeploymentServer, Builder, Logs, etc.) and none hold shared mutable state during handler execution. They're all thin facades over transactional backends (etcd).

3. **The lock-heavy code is safe** - httpingress → activator is our hottest path with the most lock complexity, but that shared state isn't modified in the context of RPC handlers. The RPC calls made from that flow hit stateless endpoints.

The old re-panic was already broken anyway - http3 would catch it, preventing the crash while *also* preventing the error from reaching the client. Worst of both worlds!

We might want to hold for @evanphx to double check all this, but from my perspective LGTM! 🚢

teresamychu:
> Hey Teresa! Took a look at this w/ Claude.
> 
> At first I was concerned about unconditional panic recovery across the entirety of our RPC handlers - my instinct was we should "let it crash" to avoid running with potentially corrupted state.
> 
> But after digging around, I think skipping re-panic is a reasonable approach:
> 
> 1. **http3 does the same thing** - quic-go recovers panics per-request. If the Go networking experts consider this safe, that's a good signal.
> 2. **Our RPC handlers are stateless** - audited the full surface area (EntityServer, DeploymentServer, Builder, Logs, etc.) and none hold shared mutable state during handler execution. They're all thin facades over transactional backends (etcd).
> 3. **The lock-heavy code is safe** - httpingress → activator is our hottest path with the most lock complexity, but that shared state isn't modified in the context of RPC handlers. The RPC calls made from that flow hit stateless endpoints.
> 
> The old re-panic was already broken anyway - http3 would catch it, preventing the crash while _also_ preventing the error from reaching the client. Worst of both worlds!
> 
> We might want to hold for @evanphx to double check all this, but from my perspective LGTM! 🚢

Sounds good!

evanphx (approved):
Ah! sounds like the code we were sending before panic'ing probably wasn't getting data to the client anyway, as the re-panic would be seen by grpc and likely not use any of the data.

PR #516, opened 2026-01-06
Release disk leases when sandbox stops

phinze:
## Summary

Disk leases were not being released when sandboxes died, leaving orphaned BOUND leases pointing to dead sandboxes. When a new sandbox tried to use the same disk (e.g., after a version change), it would find the stale lease and fail.

**Root cause**: `stopSandbox()` released network IPs but not disk leases.

**Fix**: Release disk leases during sandbox cleanup, following the same pattern as network IP cleanup.

## Changes

- Add `releaseDiskLeases()` to transition leases to RELEASED status
- Call it from `stopSandbox()` alongside IP release
- Simplify `acquireDiskLease()` - no transfer logic needed since leases are properly released
- Add tests for lease acquisition and release

## Test plan

- [x] New unit tests for `acquireDiskLease()` covering:
  - Creating new lease when none exists
  - Reusing own lease (retry case)
  - Failing if another sandbox has a BOUND lease
  - Creating new lease when existing lease is RELEASED
  - Not reusing lease from different node
- [x] New unit tests for `releaseDiskLeases()` covering:
  - Releasing leases owned by sandbox
  - Not affecting leases owned by other sandboxes
- [ ] Manual testing with app version change

## Summary by CodeRabbit

* **New Features**
  * Sandboxes can acquire or take over disk leases and release their leases during shutdown.

* **Improvements**
  * Leases use generated unique IDs and binding is awaited before use.
  * Stale non-bound leases are ignored; leases aren’t reused across nodes.
  * Active lease conflicts are detected and surfaced; shutdown proceeds even if lease release errors occur.

* **Tests**
  * Added tests covering acquire, reuse, conflict, release, and node-specific lease behavior.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -43,6 +43,7 @@ import (
 	"miren.dev/runtime/api/core/core_v1alpha"
 	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
 	"miren.dev/runtime/api/network/network_v1alpha"
+	storage "miren.dev/runtime/api/storage/storage_v1alpha"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


phinze:
Pulling this back to draft to review strategy, I don't think we want this fix _inside_ `findOrCreateLease`

evanphx (approved):

teresamychu (approved):
Looks good!

I was originally digging through because I wasnt sure if there was a crash that it would call `stopSandbox()` but after some digging it looks like:

- if it crashes and the node recovers, then everything is fine it will just reuse the same leases. 
- if it crashes and the sandbox is dead, the`ReconcileController` watch that will ultimately route the delete event back to SandboxController's `Delete()`  which in turn calls `stopSandbox()` so it will be released anyways. 

Awesome possum!

PR #517, opened 2026-01-06
Add miren deploy cancel command for users to cancel stuck deployments

teresamychu:
### Summary

Originally we proposed that there should be a `force unlock` command for the user to unlock deployments. But within the context of the user, they understand that there is another deployment in-progress, not that they are locked from deploying. So this feature allows them to "cancel a deployment" which i think makes more sense from a user business-logic perspective? Happy to discuss! 

 Adds miren deploy cancel <deployment-id> command that allows users to cancel stuck in-progress deployments without waiting for the 30-minute lock expiry.

  Permission model:
  - Registered clusters (with miren.cloud auth): Only the deployment owner can cancel their deployment
  - Unregistered clusters (no auth): Anyone can cancel

  Changes:
  - Add CancelDeployment RPC method to deployment service
  - Create deploy_cancel.go CLI command
  - Add comprehensive test coverage (6 test cases)

This resolves [MIR-534](https://linear.app/miren/issue/MIR-534/deployment-lock-should-have-force-unlock)

## Summary by CodeRabbit

* **New Features**
  * Added a cancel-deployment operation and CLI command ("deploy cancel") to cancel in-progress deployments; returns success or error.
* **Bug Fixes**
  * Cancellation is a terminal state and is preserved (not overwritten by later failure updates).
* **UX**
  * App history shows deployment IDs and a styled "cancelled" status; errors now display for cancelled deployments.
* **Tests**
  * Added tests covering cancellation scenarios and post-cancellation state validation.

### DIFF
--- a/api/deployment/deployment_v1alpha/rpc.gen.go
+++ b/api/deployment/deployment_v1alpha/rpc.gen.go
@@ -1241,6 +1241,88 @@ func (v *DeploymentGetActiveDeploymentResults) UnmarshalJSON(data []byte) error 
 	return json.Unmarshal(data, &v.data)
 }
 
+type deploymentCancelDeploymentArgsData struct {
+	DeploymentId *string `cbor:"0,keyasint,omitempty" json:"deployment_id,omitempty"`
+	CallerUserId *string `cbor:"1,keyasint,omitempty" json:"caller_user_id,omitempty"`
+}
+
+type DeploymentCancelDeploymentArgs struct {
+	call rpc.Call
+	data deploymentCancelDeploymentArgsData
+}
+
+func (v *DeploymentCancelDeploymentArgs) HasDeploymentId() bool {
+	return v.data.DeploymentId != nil
+}
+
+func (v *DeploymentCancelDeploymentArgs) DeploymentId() string {
+	if v.data.DeploymentId == nil {
+		return ""
+	}
+	return *v.data.DeploymentId
+}
+
+func (v *DeploymentCancelDeploymentArgs) HasCallerUserId() bool {
+	return v.data.CallerUserId != nil
+}
+
+func (v *DeploymentCancelDeploymentArgs) CallerUserId() string {
+	if v.data.CallerUserId == nil {
+		return ""
+	}
+	return *v.data.CallerUserId
+}
+
+func (v *DeploymentCancelDeploymentArgs) MarshalCBOR() ([]byte, error) {
+	return cbor.Marshal(v.data)
+}
+
+func (v *DeploymentCancelDeploymentArgs) UnmarshalCBOR(data []byte) error {
+	return cbor.Unmarshal(data, &v.data)
+}
+
+func (v *DeploymentCancelDeploymentArgs) MarshalJSON() ([]byte, error) {
+	return json.Marshal(v.data)
+}
+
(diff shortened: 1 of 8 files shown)


phinze:
> Registered clusters (with miren.cloud auth): Only the deployment owner can cancel their deployment

I'm not sure about this permission gate... I think I would want you or Evan (or anybody with access to the cluster) to be able to cancel my stuck deployments.

IMHO it's better for us to keep things consistently permissive right now (access to the cluster implies full control) and then tackle per-operation authorization as its own holistic sweep vs making individual policy calls like this.

Happy to discuss further tho!

teresamychu:
> > Registered clusters (with miren.cloud auth): Only the deployment owner can cancel their deployment
> 
> I'm not sure about this permission gate... I think I would want you or Evan (or anybody with access to the cluster) to be able to cancel my stuck deployments.
> 
> IMHO it's better for us to keep things consistently permissive right now (access to the cluster implies full control) and then tackle per-operation authorization as its own holistic sweep vs making individual policy calls like this.
> 
> Happy to discuss further tho!

For sure! Especially since all other operations in runtime are cluster access = full control. Which is inline with current CI/CD behavior. I was still stuck in the old proposal thinking "admins should be able to force unlock" and trying to gate the lock bypass to a single user. But I guess if we're switching the context to "cancelling deployments" then it makes sense that anyone can cancel a deployment! 

I'll make this change!

evanphx (approved):
Looks good!

phinze (commented):
Two concerns, neither necessarily blocking:

1. The `app history` changes aren't using the central `pkg/ui` table control so they're inconsistent compared to other commands, also it seems like are some header alignment issues in some of the columns.

Image

2. Since we're calling this `cancel`, I think a user will expect this will actually kill a hung or long-running deployment. (Similar to how "Cancel" works in GitHub Actions, for example.) Currently all it does is change the status field - it doesn't look like we're currently touching any actual running deployment processes.

teresamychu:
> Two concerns, neither necessarily blocking:
> 
> 1. The `app history` changes aren't using the central `pkg/ui` table control so they're inconsistent compared to other commands, also it seems like are some header alignment issues in some of the columns.
> 
> Image
> 2. Since we're calling this `cancel`, I think a user will expect this will actually kill a hung or long-running deployment. (Similar to how "Cancel" works in GitHub Actions, for example.) Currently all it does is change the status field - it doesn't look like we're currently touching any actual running deployment processes.

OOO interesting! Mine looks very different on iterm...HUH

image

2. That is an EXCELLENT point. I actually thought that we had a poll on deployment status in `BuildFromTar()` but it looks like i was wildly mistaken. I'll add this change!

evanphx:
The header alignment is still there, probably because it's using a different render (see how the headers are indented). Those green lines are paul drawing on the image (I think!), not in the output! I thought it was in the output myself!

teresamychu:
OK i am rethinking this PR and im going to keep it simple and will create two separate PRs to: 

-  Add the `miren app history` changes (along with the pkg/ui issue that Paul found) in a separate PR.
-  Have `miren deploy cancel` actually kill the deploy process. 

Keeping this PR clean, it only updates the status right now but will fully kill the deploy process in a followup PR.

phinze (approved):

PR #519, opened 2026-01-07
Remove pkg/asm dependency injection framework

evanphx:
## Summary
- Replace reflection-based dependency injection (`pkg/asm`) with explicit dependency passing
- This simplifies the codebase by removing magic initialization and making dependencies visible at call sites
- Add `pkg/subsystem` for simple component lifecycle management
- Add `cli/commands/server_state.go` for explicit server dependencies
- Update all tests to use explicit `TestDeps` from `testutils`
- Fix lint issues in test files (SA5011 nil pointer warnings)

**51 files changed, 1650 insertions, 2205 deletions** (net -555 lines)

## Test plan
- [x] `make test` - all unit tests pass
- [x] `make test-e2e` - e2e tests pass
- [x] `make lint` - 0 issues

## Summary by CodeRabbit

* **Refactor**
  * Switched from implicit auto-wiring to explicit state- and dependency-driven initialization for clearer startup, shutdown, and lifecycle.
  * Many components now require/accept explicit constructors and dependency structs for predictable wiring.

* **New Features**
  * Centralized ServerState API with platform-specific stubs for consistent cross-platform behavior.
  * New configurable metrics and observability subsystems with explicit lifecycle control.

* **Tests**
  * Tests migrated to unified test dependency helpers and updated assertions for more reliable, self-contained setups.

* **Bug Fixes**
  * Metrics writer startup made idempotent to avoid duplicate starts.


components/runner/integration_test.go
```diff
@@ -70,8 +66,33 @@ func TestRunnerCoordinatorIntegration(t *testing.T) {
 	runnerCfg.Config = rcfg
 	runnerCfg.DataPath = t.TempDir()
 
+	res, _ := netresolve.NewLocalResolver()
```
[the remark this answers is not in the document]
evanphx: Those aren't errors.

pkg/testutils/reg.go
```diff
@@ -81,191 +136,143 @@ func Registry(extra ...func(*asm.Registry)) (*asm.Registry, func()) {
...
+		Client: bkClient,
+		Log:    log,
+	}
 
-		eac := entityserver_v1alpha.NewEntityAccessClient(client)
-		return eac, nil
-	})
+	// Create network service manager and resolver
+	netServ := network.NewServiceManager(log, nil)
+	resolver, _ := netresolve.NewLocalResolver()
```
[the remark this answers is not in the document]
evanphx: Those aren't error values.

teresamychu:
LOL i DO hate "magic" dependency injection cause I can never trace exactly whats going on. So this is a really nice change! 

I have a few minor initial thoughts that im going to leave here but we can discuss the PR further (love a walk-through in case I missed something):

- ServerState is a bit of a frankenstein and it feels like its just going to get bigger. Maybe we can separate these into different domains? ie. `ServerState` , `ContainerdState` , `MetricsState` 

- I might have missed it but need a cleanup somewhere to close the components on shutdown, `ServerState` could handle this by having a `Close()` method.

cli/commands/server_state.go
```diff
@@ -0,0 +1,159 @@
...
+	victoriaLogsAddr := os.Getenv("VICTORIALOGS_ADDR")
+	if victoriaLogsAddr == "" {
+		victoriaLogsAddr = "localhost:9428"
+	}
+
+	// Read VictoriaMetrics address from environment or use default
+	victoriaMetricsAddr := os.Getenv("VICTORIAMETRICS_ADDR")
+	if victoriaMetricsAddr == "" {
+		victoriaMetricsAddr = "localhost:8428"
+	}
```
phinze: 💡 The env var lookups here are end-running around `serverconfig` - these should be `MIREN_VICTORIALOGS_ADDRESS` and `MIREN_VICTORIAMETRICS_ADDRESS` to match the existing convention, and serverconfig already has defaults of `victorialogs:9428` / `victoriametrics:8428` (vs `localhost:*` here).

In practice these defaults always get overwritten in server.go by either the embedded component endpoint or `cfg.Victorialogs.GetAddress()` - so this is effectively dead code that could confuse someone trying to set `VICTORIALOGS_ADDR`.

I'd just remove the env var lookups entirely and let server.go initialize these fields from `cfg` like it already does.

cli/commands/server.go
```diff
@@ -337,25 +323,17 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 			return fmt.Errorf("victorialogs address specified but embedded victorialogs not started")
 		}
```
phinze, line 324: 💡 Not your code but this error message is wrong - error message says "address specified" but this branch triggers when address is empty.

controllers/sandbox/sandbox.go
```diff
@@ -102,9 +122,56 @@ type SandboxController struct {
...
+	}
+	if cfg.LogsMaintainer == nil {
+		return nil, fmt.Errorf("sandbox: LogsMaintainer is required")
+	}
+	if cfg.LogWriter == nil {
+		return nil, fmt.Errorf("sandbox: LogWriter is required")
+	}
+	if cfg.Resolver == nil {
+		return nil, fmt.Errorf("sandbox: Resolver is required")
+	}
```
phinze, line 159: 💡 `NodeId` isn't validated here but gets used in volume.go to build entity IDs like `entity.Id("node/" + c.NodeId)` - an empty string would give you `"node/"` which might cause subtle weirdness. Might be worth a quick check, but not blocking.

phinze (approved):
Love it - bye bye reflection magic, hello clarity. Much easier to follow now.

Maybe someday we'll bring back a fancy DI system that saves boilerplate and makes testing super slick... but today is not that day.

Few optional nits inline, nothing blocking.

Image

controllers/sandbox/sandbox.go
```diff
@@ -57,22 +57,42 @@ type containerPorts struct {
 	Ports []observability.BoundPort
 }
 
+// SandboxControllerConfig holds required dependencies for SandboxController.
+type SandboxControllerConfig struct {
```
phinze: 💡 Naming nit - AFAICT in this codebase we generally are using `*Config` to means user-facing serializable config (see `appconfig`, `serverconfig` - all have toml/env tags). But `SandboxControllerConfig` and `ServiceControllerConfig` are runtime deps with no serialization - more like `RunnerDeps` and `TestDeps`.

For consistency with existing conventions maybe `*Deps` for the controller structs? Not blocking, just noting the pattern mismatch.

cli/commands/server_state.go
evanphx: Oh yup, on it.

cli/commands/server.go
evanphx: mmm, yup.

controllers/sandbox/sandbox.go
evanphx: fair, on it

controllers/sandbox/sandbox.go
evanphx: let's validate!

evanphx:
@teresamychu Yeah, we'll probably want to break up the ServerState struct but I'm going to leave that til later. Cleanup wise, the old DI approach didn't have any cleanup that I recall, so this is 🍎 to 🍎 .

PR #520, opened 2026-01-07
Add configurable graceful shutdown timeout for apps during redeploy

phinze:
## Summary

- Apps now receive proper graceful shutdown time during redeploy instead of being force-killed after only 50ms
- Added configurable `shutdown_timeout` per-service (default 10 seconds)
- Refactored `destroySubContainers()` to implement 3-phase shutdown: SIGTERM → poll for exit → SIGKILL after timeout

## Configuration

```toml
# .miren/app.toml
[services.worker.concurrency]
mode = "fixed"
num_instances = 2
shutdown_timeout = "30s"  # Wait 30s for graceful shutdown (default: 10s)
```

## Changes

- `appconfig/appconfig.go` - Add `ShutdownTimeout` field with validation
- `api/core/schema.yml` - Add `shutdown_timeout` to service_concurrency
- `api/compute/schema.yml` - Add `shutdown_timeout` to container specs
- `servers/build/build.go` - Map config to entity schema
- `controllers/deployment/launcher.go` - Propagate to sandbox spec
- `controllers/sandbox/sandbox.go` - Implement graceful shutdown with per-container timeouts

## Test plan

- [x] Lint passes
- [x] Build passes
- [x] Unit tests pass (appconfig, build server)
- [x] CI passes
- [x] Manual test: deploy app with signal handler, verify graceful shutdown

## Known limitation

Graceful shutdown blocks reconciliation workers for the duration of the timeout. With the default 3 workers, this is acceptable for now but could be improved by making shutdown async if needed.

## Summary by CodeRabbit

* **New Features**
  * Per-service graceful shutdown timeout configuration added (default: 10s); services can set values like 10s or 30s.
  * Shutdown timeouts propagate to runtime and are validated with sensible defaults.
  * Runtime now coordinates phased, per-container graceful shutdowns before force-killing, improving cleanup and logs.
  * Added test app and example config demonstrating long shutdown behavior.


controllers/sandbox/sandbox.go
```diff
@@ -1896,121 +1931,142 @@ func (c *SandboxController) destroySubContainers(ctx context.Context, id entity.
...
-			return fmt.Errorf("context cancelled while destroying subcontainers: %w", ctx.Err())
-		default:
+	for i := range containers {
+		info := &containers[i]
+		task, err := info.container.Task(ctx, nil)
+		if err != nil {
+			c.Log.Debug("no task found for container", "id", info.id)
+			continue
 		}
+		info.task = task
```
evanphx, line 1995: One idea is that after this code, you call container.Wait() to get a channel that you can wait on for the container to stop. Then you can use them in the for+select loop below to detect much more quickly that the task has stopped.

evanphx (approved):
Looks good! Left a comment about an optional optimization using Wait() to detect container exits more quickly.

controllers/sandbox/sandbox.go
phinze: Ooo nice will do that.

evanphx (approved):

controllers/sandbox/sandbox.go
phinze: Done in 8ea434f085b24dd211708bdb2cb51bd9084eb67a

PR #522, opened 2026-01-07
Poll the deployment status for manual cancellation

teresamychu:
### Teresa's Notes

I think the proper way to do this would be to move deployments to a server-side managed entity and have the build server watch the deployment entity and abort the build if it detects the `status = cancelled`

However, the above change is a larger architectural change that i will do in a separate PR. For now, we'll poll the deployment status every 2 seconds and kill the deployment when the status changes.

The polling solution is temporary until it can be replaced by the larger architectural change described above. 

--------------------------------------------
This fixes [MIR-608](https://linear.app/miren/issue/MIR-608/manually-cancelling-a-deployment-should-kill-the-build-process)

### Summary

  - Added deployment cancellation polling in cancellationPoller component
  - Add unit tests for the polling logic (4 test cases)
  - Refactor deploy.go to use the extracted component

  Test plan

  - make lint passes
  - Manual test: run miren deploy, then miren deploy cancel in another terminal - verify deploy exits with "Deploy cancelled" message

## Summary by CodeRabbit

* **New Features**
  * External cancellation support in the deploy flow with background monitoring and propagation through build and deploy phases.
  * Deployment status handling updated to avoid redundant or misleading status updates when cancellation is external.
  * App config include-pattern validation with invalid-pattern reporting during deploy.

* **Tests**
  * Added comprehensive tests for cancellation polling covering detection, errors, immediate cancellation, and context-cancel scenarios.

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -240,45 +240,18 @@ func Deploy(ctx *Context, opts struct {
 	buildCtx, cancelBuild := context.WithCancel(ctx.Context)
 	defer cancelBuild()
 
-	// Track if we were cancelled externally (vs user pressing CTRL-C)
-	var externallyCancelled bool
-	var cancelMu sync.Mutex
-
-	// Start goroutine to poll for external cancellation
+	// Start goroutine to poll for external cancellation using the cancellationPoller
+	statusGetter := newDepClientStatusGetter(depClient, ctx.Log)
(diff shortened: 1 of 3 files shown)


evanphx (approved):
Looks great! Makes sense to do it this way while we've got the deployment entity visibility only in the client.

PR #523, opened 2026-01-07
Add self-signed TLS option for dev environment

phinze:
## Summary

- Adds `--self-signed-tls` config option to use self-signed certificates instead of ACME/autocert
- Fixes TLS handshake errors in dev environment (autocert doesn't work with localhost)
- Enables self-signed TLS by default in `hack/dev-server`

## Changes

- `pkg/serverconfig/schema.yml` - Added `self_signed` field to TLSConfig
- `components/autotls/selfsigned.go` - New file with `ServeTLSSelfSigned()` function
- `cli/commands/server.go` - Check for self-signed mode before ACME
- `hack/dev-server` - Add `--self-signed-tls` as default flag

## Test plan

- [x] `make dev` and verify server starts without TLS errors
- [x] `curl -k https://localhost` works from inside dev container

## Summary by CodeRabbit

* **New Features**
  * Added self-signed TLS mode for secure local HTTPS (one-year cert) and a new --self-signed-tls flag / MIREN_TLS_SELF_SIGNED env var.
  * Development server startup now enables self-signed TLS by default via dev settings.

* **Behavior Change**
  * TLS setup is now explicit: DNS-challenge TLS is used only when a DNS provider is configured; otherwise ACME/HTTP (autocert) is used.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -793,6 +793,7 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 			// Use self-signed certificate (for development/testing)
 			if err := autotls.ServeTLSSelfSigned(sub, ctx.Log, hs); err != nil {
 				ctx.Log.Error("failed to enable self-signed TLS", "error", err)
+				return err
 			}
 		} else {
 			email := cfg.TLS.GetAcmeEmail()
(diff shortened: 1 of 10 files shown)


evanphx (approved):
very nice.

PR #531, opened 2026-01-09
Add debug system command for gathering system info

evanphx:
## Summary

Adds `m debug system` command that collects debugging information into a tar.gz archive for troubleshooting production issues.

**Archive contents:**
- `info.txt` - hostname, timestamp, disk usage (df), memory usage (free)
- `processes.txt` - all system processes with CPU% via gopsutil
- `containers.txt` - containerd containers with status, PID, age, entity-id
- `server-logs.txt` - journalctl logs for miren service + docker logs
- `docker-processes.txt` - ps, df, free, nerdctl output from inside Docker container (if running)
- `docker-inspect.json` - docker inspect output (if Docker container exists)

**Options:**
- `-o/--output` - output file path (default: `miren-debug.tar.gz`)
- `-s/--since` - log time range (default: `1 day ago`)
- `-d/--docker-container` - Docker container name to inspect (default: `miren`)
- `--namespace` - containerd namespace (default: `miren`)
- `--socket` - containerd socket path

The command automatically detects if miren is running in Docker and gathers additional info from inside the container via `docker exec`.

Closes MIR-615

## Test plan

- [x] Run `m debug system` and verify archive is created
- [x] Extract archive and verify all expected files are present
- [x] Test with Docker container running to verify docker-specific collection
- [x] Test `-s "2 hours ago"` flag for custom time range


cli/commands/debug_bundle.go
```diff
@@ -0,0 +1,505 @@
...
+	// Add memory usage (free)
+	fmt.Fprintf(&buf, "\n%s\n", strings.Repeat("=", 80))
+	fmt.Fprintf(&buf, "MEMORY USAGE (free -h)\n")
+	fmt.Fprintf(&buf, "%s\n", strings.Repeat("=", 80))
+	freeOutput, err := exec.Command("free", "-h").Output()
+	if err != nil {
+		fmt.Fprintf(&buf, "Error running free: %v\n", err)
+	} else {
+		buf.Write(freeOutput)
+	}
```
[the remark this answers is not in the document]
evanphx: That's ok because we also run it inside docker in that case.

cli/commands/debug_system.go
```diff
@@ -0,0 +1,505 @@
...
+	// Close archive
+	if err := tw.Close(); err != nil {
+		return fmt.Errorf("closing tar writer: %w", err)
+	}
+	if err := gw.Close(); err != nil {
+		return fmt.Errorf("closing gzip writer: %w", err)
+	}
+
+	// Write to file
+	if err := os.WriteFile(opts.OutputFile, buf.Bytes(), 0644); err != nil {
```
phinze: 💡 The logs could get hefty, might be worth streaming to disk instead of spooling into memory? Especially if this is likely to be run on a distressed system.

cli/commands/debug_bundle.go
```diff
@@ -0,0 +1,505 @@
...
+)
+
+func DebugSystem(ctx *Context, opts struct {
+	OutputFile      string `short:"o" long:"output" description:"Output file path" default:"miren-debug.tar.gz"`
+	Since           string `short:"s" long:"since" description:"Include logs since this time" default:"1 day ago"`
+	Namespace       string `long:"namespace" description:"containerd namespace" default:"miren"`
+	Socket          string `long:"socket" description:"path to containerd socket"`
+	DockerContainer string `short:"d" long:"docker-container" description:"Docker container name to get logs from" default:"miren"`
+}) error {
+	ctx.Info("Gathering system debug information...")
```
phinze, line 45: 💡 Might be worth adding timeouts to the gather operations - if Docker or journalctl is wedged (exactly when we'd be running this), the command could hang indefinitely. Could be a top-level timeout for the whole command, or per-operation timeouts so we still get partial output if one thing hangs.

phinze (approved):
Naming bikeshed: would `debug bundle` be more intuitive than `debug system`? "Support bundle" is pretty standard terminology for this kind of thing, and "system" reads a bit like we're debugging the system rather than collecting debug info about it.

teresamychu (changes requested):
im getting an error on listing containers probably because `containerd` needs `sudo` on linux:

_Gathering system debug information...
  Collecting system information...
  Collecting process information...
  Collecting container information...
W Failed to gather containers: listing containers: connection error: desc = "transport: Error while dialing: dial unix /var/lib/miren/containerd/containerd.sock: connect: permission denied"_

Running it with`sudo` works though. So we can probably warn them to run it with sudo or prompt them to `retry with sudo` on error.

evanphx:
@teresamychu ah! My intention was that someone would run it as root! I'll add a warning when that happens to re-run as root.

cli/commands/debug_bundle.go
evanphx: yaaas.

cli/commands/debug_system.go
evanphx: oh oh, very good call.

evanphx:
@phinze renamed to `debug bundle`, that's the right name.

teresamychu:
> @teresamychu ah! My intention was that someone would run it as root! I'll add a warning when that happens to re-run as root.

Awesome thanks so much! ✅

teresamychu (approved):

PR #532, opened 2026-01-09
cosmetic changes to clean up app history table.

teresamychu:
### Teresa's Notes

This PR cleans up the app history table to make it more readable with some minor changes. 

Also found a bug that the limit (-n) flag was actually never working because it was declared as `int32` 

Heres what it looks in this PR: 

image

This concludes the fixes for [MIR-607](https://linear.app/miren/issue/MIR-607/miren-app-history-table-needs-to-be-cleaned-up)

-----------------------------------------

#### Cosmetic Changes
  - Change default deployment history `limit` to 10
  - Sort deployments with `active` deployment at the top, then by time (most recent first)
  - Hide `DEPLOYED BY` column when cluster has no identity configured

#### Bug Fixes

  - Fix `--limit` flag 
  - Fix `--all` flag to show all deployments for this cluster only. 

#### Test plan

  - Run miren app history <app> and verify default shows 10 entries
  - Run miren app history <app> -n 5 and verify limit works
  - Verify active deployment appears at top of list
  - On cluster without identity: verify DEPLOYED BY column is hidden
  - On cluster with identity: verify DEPLOYED BY column is shown

### DIFF
--- a/cli/commands/app_history.go
+++ b/cli/commands/app_history.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -24,15 +25,16 @@ var statusStyles = map[string]struct {
 }
 
 type historyDisplayOpts struct {
-	detailed bool
+	detailed    bool
+	hasIdentity bool
 }
 
 func AppHistory(ctx *Context, opts struct {
 	AppCentric
 
-	Limit      int32  `short:"n" long:"limit" description:"Maximum number of deployments to show" default:"20"`
+	Limit      int    `short:"n" long:"limit" description:"Maximum number of deployments to show" default:"10"`
+	All        bool   `short:"a" long:"all" description:"Show all deployments (ignore limit)"`
 	Status     string `short:"s" long:"status" description:"Filter by status (active, failed, rolled_back)"`
-	All        bool   `long:"all" description:"Show deployments from all clusters"`
 	HideFailed bool   `long:"hide-failed" description:"Hide failed deployments"`
 	Detailed   bool   `long:"detailed" description:"Show all columns including git information"`
 }) error {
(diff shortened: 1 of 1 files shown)


cli/commands/app_history.go
```diff
@@ -24,15 +25,16 @@ var statusStyles = map[string]struct {
...
+	detailed    bool
+	hasIdentity bool
 }
 
 func AppHistory(ctx *Context, opts struct {
 	AppCentric
 
-	Limit      int32  `short:"n" long:"limit" description:"Maximum number of deployments to show" default:"20"`
+	Limit      int    `short:"n" long:"limit" description:"Maximum number of deployments to show" default:"10"`
+	All        bool   `short:"a" long:"all" description:"Show all deployments (ignore limit)"`
```
phinze: ⚠️ This short flag collides with `--app` so you can't set app with `-a`. Maybe we just keep the long one for this?

```
❯ bin/miren app history -a meet
ERROR: error validating options: app is required
```

cli/commands/app_history.go
```diff
@@ -142,24 +179,20 @@ func buildDeploymentRow(dep *deployment_v1alpha.DeploymentInfo, opts historyDisp
...
-			formatUser(dep),
-			formatDeploymentTime(dep),
-		)
 		gitSha, gitBranch, gitMessage := formatGitInfo(dep)
 		row = append(row, gitSha, gitBranch, gitMessage)
 	} else {
-		row = append(row,
-			formatVersion(dep.AppVersionId(), status),
-			formatUser(dep),
-			formatDeploymentTime(dep),
```
phinze, line 159: Nit (outside this PR's scope): We have a similar `humanFriendlyTimestamp` in `sandbox_list.go` that's more comprehensive (handles seconds, weeks, months, years). Might be worth consolidating these into a shared helper in `pkg/ui` or at least reusing the existing one. Not blocking for this PR though!

cli/commands/app_history.go
```diff
@@ -113,20 +114,56 @@ func filterDeployments(deps []*deployment_v1alpha.DeploymentInfo, keep func(*dep
 	return filtered
 }
 
+func sortDeployments(deployments []*deployment_v1alpha.DeploymentInfo) {
+	sort.Slice(deployments, func(i, j int) bool {
+		// Active deployments come first
```
phinze: 💭 Active deployment showing up before failed deployment and the timestamps being out of order seems like it could be slightly confusing.

Like if I see:

```
 * active     Jan 9
 * failed     Jan 9 
 * succeeded  Jan 7
 * succeeded  Jan 7
```

I would assume the failed deployment was _before_ the active one, even though it might have been after.

I think ordering by timestamp desc is a little more straightforward - the `active` status coloring will be obvious enough when scanning visually.

phinze (changes requested):
One flag collision to address and a couple of nonblocking Qs for ya!

cli/commands/app_history.go
teresamychu: AH yes! I also saw that @evanphx also has a `formatAge` that is also very similar in his `debug system` PR. So this is a good note! 

I've created a follow-up ticket here:

https://linear.app/miren/issue/MIR-618/create-a-shared-function-in-pkgui-for-formatting-timeage-etc

cli/commands/app_history.go
teresamychu: OOOO i didnt see this cause i was testing within the application folder so i didnt need to pass in the app argument. 

Yes! I will remove the short form since `-a` and `-all` is not that far off in terms of keystrokes haha

cli/commands/app_history.go
teresamychu: Yes this one i wanted get yours/Evans opinions on cause it also felt weird going down the list to find the live version. 

But yeah if it doesnt feel right to you, then i will absolutely change it.

phinze (approved):

PR #535, opened 2026-01-20
Fix cluster switch for multi-identity setups and improve identity UX

phinze:
## Summary
Fixes `miren cluster switch` failing with "cluster not found in organization" when the 
cluster was added with a different local name than its cloud name. The root cause was 
that we weren't storing the cluster's cloud XID, forcing a name-based lookup that failed 
when names differed.

While investigating, discovered several related UX issues with multi-identity workflows:
- `miren login` silently overwrote existing identities
- `miren logout` with multiple identities required `--identity` flag with no guidance
- `miren cluster add` required `--force` for overwrites with no interactive option
- `miren config info` lumped all leaf configs as "cluster configs"

## Changes
- Store cluster XID when adding clusters; use it for permission checks in `cluster switch`
- Add `--force` flag to login; prompt interactively when identity exists
- Add identity picker to logout when multiple identities configured
- Add overwrite prompt to cluster add in interactive mode
- Categorize config info output by type (clusters, identities, keys)
- Modernize `config load` to write leaf configs to `clientconfig.d/`
- Remove dead code (`config_bind.go`)

## Test plan
- [ ] `miren login` with existing identity prompts for action
- [ ] `miren logout` with multiple identities shows picker
- [ ] `miren cluster add` with existing cluster prompts to overwrite
- [ ] `miren cluster switch` works for clusters with different local vs cloud names
- [ ] `miren config info` shows separate sections for clusters/identities/keys


cli/commands/login.go
```diff
@@ -706,3 +764,52 @@ func autoConfigureCluster(ctx *Context, identityName, cloudURL string, keyPair *
...
+	}
+
+	// Parse claims to get user info
+	claims, err := auth.ParseUnverifiedClaims(token)
+	if err != nil || claims == nil {
+		return ""
+	}
+
+	// Prefer UserName (display name), fall back to UserID
+	if claims.UserName != "" {
```
teresamychu, line 814: 💭 This missing username  from the claims has come up a few times and i think its a big enough problem that we should probably just add username to the cloud side when its generating the token. 

But not in this PR. Just a thought.

teresamychu (approved):
woohoo for removing dead code!

cli/commands/login.go
teresamychu: I made a linear issue for this: https://linear.app/miren/issue/MIR-633/add-username-and-name-to-jwt-claims-on-cloud

cli/commands/cluster_switch.go
```diff
@@ -119,27 +119,31 @@ func checkClusterAccess(ctx *Context, cfg *clientconfig.Config, clusterName stri
...
+				break
+			}
 		}
-	}
 
-	if clusterXID == "" {
-		return fmt.Errorf("cluster %q not found in organization", clusterName)
+		if clusterXID == "" {
+			return fmt.Errorf("cluster %q not found in organization", clusterName)
+		}
```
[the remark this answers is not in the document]
phinze: Good observation, but I think this isn't worth the added complexity. This fallback path is transitional - new clusters now get XIDs stored automatically. For legacy clusters without XIDs, users can simply re-add the cluster to get the XID stored. The hostname matching logic would add significant complexity (normalizing addresses, handling multiple addresses, port/scheme variations) for a path that will be used less over time.

The current behavior already warns and proceeds if we can't verify, which seems reasonable for legacy configs.

cli/commands/login.go
phinze: Agreed! Good to capture.

PR #540, opened 2026-01-20
Fix flaky TestEtcdStore_ReplaceEntity revision test

phinze:
## Summary
- Fix flaky tests that used hardcoded revision `9999` as the "wrong" revision
- Since etcd revisions are global counters, tests would fail if the revision happened to equal 9999
- Now uses `initial.GetRevision() + 1` which is guaranteed to be wrong

Affected tests:
- `TestEtcdStore_ReplaceEntity/replace_with_wrong_revision_fails`
- `TestEtcdStore_PatchEntity/patch_with_wrong_revision_fails`

## Test plan
- [x] Both wrong revision tests pass

### DIFF
--- a/pkg/entity/store_test.go
+++ b/pkg/entity/store_test.go
@@ -1326,10 +1326,13 @@ func TestEtcdStore_ReplaceEntity(t *testing.T) {
 		))
 		require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


evanphx (approved):
not wrong enough!

teresamychu (approved):

PR #541, opened 2026-01-21
Fix DNS server not watching sandbox entities

phinze:
## Summary
- Pass EntityAccessClient to `InitNetServ` so the DNS server can watch sandbox entities
- Add `testdata/dns-test` app for testing cross-service DNS resolution

## Problem
DNS queries for `*.app.miren` domains returned "no such host" because `InitNetServ` was passing `nil` for the EAC, so `server.Watch(ctx)` was never called and the DNS server never built the service→IP mappings.

Regression from #519 (not yet in a release).

## Fix
Pass the `eac` created in `server.go:657` through to `InitNetServ`, which passes it to `NewServiceManager`.

## Testing
Deployed `testdata/dns-test` which has two services (web and backend). Before fix, `/dns-lookup` returned "no such host". After fix, it resolves correctly and `/call-backend` successfully reaches the backend service.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -677,7 +677,7 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 	hs := httpingress.NewServer(ctx, ctx.Log, ingressConfig, client, aa, httpMetrics, logWriter)
 
 	// Initialize remaining ServerState components
-	ctx.ServerState.InitNetServ(ctx.Log)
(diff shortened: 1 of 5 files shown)


teresamychu (approved):
minor non-blocking nit:

The `dns-test` app has both `app.toml` and a `Procfile`? Was this on purpose?

evanphx (approved):

phinze:
> The dns-test app has both app.toml and a Procfile? Was this on purpose?

Good catch! That's redundant. Will remove Procfile

PR #549, opened 2026-01-25
chown during copy to avoid OCI layer duplication

phinze:
Turns out our `chown -R app:app /app` after copying code was silently doubling our layer size.

In overlayfs, changing file metadata (even just ownership) [triggers a full copy-up of the file contents](https://docs.kernel.org/filesystems/overlayfs.html#non-standard-behavior):

> The copy_up process... creates the object with the same metadata... and then if the object is a file, the data is copied from the lower to the upper filesystem.

So we were storing every app file twice - once from the copy, once from the chown.

The fix: use buildkit's `ChownOpt` to set ownership during the copy itself. One operation, one layer.

There's a kernel feature called `metacopy` (4.19+) that would let overlayfs do metadata-only changes, but it's not on by default in container runtimes. See [containerd#6310](https://github.com/containerd/containerd/issues/6310) for the saga.

**What changed:**
- `copyApp()` now sets ownership via `ChownOpt`
- Removed post-copy `chownApp()` from all stacks
- Moved `addAppUser()` before `copyApp()` in Go (it was after, oops)
- Kept `chownApp()` for Python/uv - it needs to chown before `uv sync` runs as the app user

Stacked on #548.

### DIFF
--- a/pkg/stackbuild/bun.go
+++ b/pkg/stackbuild/bun.go
@@ -100,8 +100,6 @@ func (s *BunStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, error
 
 	state = s.applyOnBuild(state, opts)
 
-	state = s.chownApp(state)
-
 	return &state, nil
 }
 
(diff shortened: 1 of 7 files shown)


evanphx (approved):
What a find.

PR #554, opened 2026-01-26
Add LSVD component that runs standalone

evanphx:


components/lsvd/server/volume_controller.go
```diff
@@ -0,0 +1,382 @@
...
+	if err := c.ReconcileWithEntities(ctx); err != nil {
+		c.log.Error("initial volume reconciliation failed", "error", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			if err := c.ReconcileWithEntities(ctx); err != nil {
```
phinze: The polling loop only calls `ReconcileWithEntities`, not `ReconcileWithSystem`. This means if a volume directory gets deleted or NBD wedges after startup, we won't detect it until lsvd-server restarts.

Given that NBD issues were the motivation for this work, should we periodically run system verification too? Maybe every N ticks or on a separate slower timer?

components/lsvd/server/volume_controller.go
```diff
@@ -0,0 +1,382 @@
...
+// reconcileVolumePresent ensures the volume exists
+func (c *VolumeController) reconcileVolumePresent(ctx context.Context, volume *storage_v1alpha.LsvdVolume) error {
+	entityId := string(volume.ID)
+
+	// Check if volume already exists in our state
+	if existing := c.state.GetVolume(entityId); existing != nil {
+		// Volume already exists, verify it's in a good state
+		if volume.ActualState == storage_v1alpha.VOL_READY {
+			c.log.Debug("volume already ready", "entity_id", entityId)
+			return nil
```
phinze, line 107: When the entity says `VOL_READY` and we have local state, we return without verifying the volume directory still exists. If the directory gets deleted (disk failure, accidental cleanup, etc.), we'll keep reporting ready while mounts fail with confusing errors.

Worth adding a quick `VolumePathExists` check here, similar to the crash-recovery path below?

components/lsvd/server/mount_controller.go
```diff
@@ -0,0 +1,775 @@
...
+		return nil
+	case storage_v1alpha.MNT_ATTACHED:
+		// Attached but not mounted, mount it
+		return c.mountVolume(ctx, mount)
+	case storage_v1alpha.MNT_MOUNTING:
+		// Already mounting, wait
+		return nil
+	case storage_v1alpha.MNT_MOUNTED:
+		// Already mounted, nothing to do
+		return nil
```
phinze, line 231: Similar pattern here - when entity says `MNT_MOUNTED`, we trust it without verifying NBD is still connected or the filesystem is actually mounted. If NBD wedges after startup (the exact scenario that motivated this work!), the polling loop won't catch it.

The `ReconcileWithSystem` logic already knows how to check `NBDStatus` and `IsMounted` - could we call a lightweight version of that here?

components/lsvd/server/volume_controller.go
```diff
@@ -0,0 +1,382 @@
...
+	for _, volState := range c.state.ListVolumes() {
+		entityId := volState.EntityId
+		// Verify volume directory exists
+		if volState.DiskPath != "" {
+			if !c.ops.VolumePathExists(volState.DiskPath) {
+				c.log.Warn("volume directory missing",
+					"entity_id", entityId,
+					"path", volState.DiskPath,
+				)
+				// TODO: Mark as error in etcd
```
phinze: This TODO seems important - if we detect a missing volume but don't update the entity, we're left in a split-brain state where local state and entity state disagree. Should we implement this before merging, or is it okay to leave as a follow-up?

components/lsvd/server/mount_controller.go
```diff
@@ -0,0 +1,775 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			// Cleanup handlers on shutdown
+			c.handlersMu.Lock()
+			for entityId, h := range c.handlers {
+				c.log.Info("cleaning up handler on shutdown", "entity_id", entityId)
+				if h.cancel != nil {
+					h.cancel()
```
phinze: On shutdown, we cancel handlers but don't explicitly close the disks. The disks get closed in `unmountAndDetach`, but on graceful shutdown we're just canceling contexts here. Should we add `h.disk.Close(ctx)` to ensure LSVD flushes any buffered writes?

phinze (approved):
Looking good! At first I was surprised by how much code is here given this is essentially an extraction, but I came around to the additions for introspection/debuggability. Just a couple of inline optionals.

components/lsvd/server/volume_controller.go
evanphx: Done!

components/lsvd/server/volume_controller.go
evanphx: Yup! addressed.

components/lsvd/server/mount_controller.go
evanphx: Yup, yup. fixed.

components/lsvd/server/mount_controller.go
evanphx: Fixed!

components/lsvd/server/volume_controller.go
evanphx: Done!

phinze:
Heyo! Claude and Paul here, broadcasting from our review session 📡

We caught up on all the latest commits since the last review - it's a lot of code! Which freaked us out at first, but after walking through it together it seems warranted. The big delete of `lsvd_client.go` is the point, not scope creep. And the robustness stuff (NBD verification, volume directory checks, proper shutdown sequencing) all landed well.

We also dug into the outboard package design. The FIFO + slogreplay trick is clever - logs appearing in the parent's stream with full fidelity is nice. The config-file-as-rendezvous for the RPC address is clean. We talked about whether this could eventually replace the containerd harness too, but that's a future thing - would need some refactoring to make the RPC layer pluggable since containerd speaks gRPC.

**One question about orphan handling:**

When miren dies hard (SIGKILL, OOM), we traced through what happens:

1. New miren starts, launches new lsvd-server
2. New lsvd-server loads persisted state, `ReconcileWithSystem` reconnects NBD devices
3. Old orphaned lsvd-server dies from... SIGPIPE? (FIFO read end closed when old miren died)

Is that implicit SIGPIPE death the intended mechanism? It's kinda elegant if so - just want to confirm it's deliberate and not accidental. The 30-second reconciliation logging means it'd notice pretty quickly.

No rush, not blocking - just curious!

<3 your pals claude & phinze

PR #558, opened 2026-01-27
Fix tag-release workflow and add manual trigger

phinze:
## Summary
- Configure `github-actions[bot]` git identity before creating annotated tags (turns out they don't get one by default)
- Add `workflow_dispatch` trigger so we can manually re-run for v0.3.0


phinze:
This'll fix the failure we [saw here](https://github.com/mirendev/runtime/actions/runs/21414607056/job/61659906934#step:3:22)

PR #562, opened 2026-01-30
Default app name from .miren/app.toml for CLI commands

evanphx:
## Summary
- Make `app delete`, `route set`, and `route set-default` commands default to the app name from `.miren/app.toml` when not provided as a positional argument
- Maintains backwards compatibility - positional arguments still work as before
- Improves UX when running commands from within an app directory

## Test plan
- [x] `make lint` passes
- [x] `hack/it ./cli/commands/...` tests pass
- [x] Manual testing in dev environment:
  - Create `.miren/app.toml` with `name = "testapp"`
  - Run `m route set example.com` (should use testapp)
  - Run `m route set-default` (should use testapp)
  - Run `m app delete` (should prompt for testapp)

### DIFF
--- a/cli/commands/app_delete.go
+++ b/cli/commands/app_delete.go
@@ -15,7 +15,7 @@ func AppDelete(ctx *Context, opts struct {
 }) error {
 	appName := opts.AppName
 	if appName == "" {
-		if ac, err := appconfig.LoadAppConfig(); err == nil && ac.Name != "" {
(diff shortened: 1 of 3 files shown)


phinze (approved):

phinze:
^^ Maybe `LoadAppConfig()` should return an empty config object instead of nil when nothing's there?

evanphx:
> ^^ Maybe `LoadAppConfig()` should return an empty config object instead of nil when nothing's there?

I'm thinking we do a pass at the app config api writ large.

PR #564, opened 2026-01-30
Fix miren upgrade permission errors for non-root users

evanphx:
## Summary

- Add permission checking before downloading during `miren upgrade`
- In interactive mode: show picker menu to choose between sudo or user install
- In non-interactive mode: print clear error message with command suggestions
- Add `--user` flag for explicit user-directory installation to `~/.miren/release/miren`

Fixes MIR-642

## Test plan

- [x] Test upgrade from `/usr/local/bin/miren` as normal user - should see picker options
- [x] Test `miren upgrade --user` - should install to `~/.miren/release/miren`
- [x] Test upgrade from `~/.miren/release/miren` - should work without sudo
- [x] Test non-interactive mode (pipe to /dev/null) - should show text options
- [x] Run `make lint` to verify code quality

### DIFF
--- a/cli/commands/upgrade.go
+++ b/cli/commands/upgrade.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


cli/commands/upgrade.go
```diff
@@ -56,6 +49,72 @@ func Upgrade(ctx *Context, opts struct {
...
+		if err != nil {
+			return fmt.Errorf("failed to determine user install path: %w", err)
+		}
+		installPath = userPath
+	} else {
+		installPath = exe
+	}
+
+	// Check permissions before downloading
+	if err := checkInstallPermissions(installPath); err != nil {
```
phinze: The permission check is great for the "upgrade in place" case - offering sudo vs user-install when you can't write to `/usr/local/bin` makes sense.

But when `--user` is already specified, we still check `~/.miren/release` which probably doesn't exist yet. `unix.Access()` returns ENOENT, we treat it as a permission error, and the user gets offered "try sudo" or "install to user dir" when they already asked for `--user` 🙃

Thinking we just skip the check for `--user`:

```go
if !opts.User {
    if err := checkInstallPermissions(installPath); err != nil {
        // ... existing handler ...
    }
}
```

If creating dirs in $HOME fails, that's a legit error to surface directly.

cli/commands/upgrade.go
```diff
@@ -56,6 +49,72 @@ func Upgrade(ctx *Context, opts struct {
...
+		}
+	}
+
+	// Ensure the install directory exists for user installs
+	needsPathUpdate, err := ensureUserInstallDir(installPath)
+	if err != nil {
+		return err
+	}
+
+	if needsPathUpdate {
```
phinze: PATH warning is good. One edge case: if someone has `/usr/local/bin/miren` and does `--user`, they'll end up with two binaries. Even if they update PATH, the system one might still win depending on order.

Could be worth a heads up like "btw /usr/local/bin/miren still exists and might take precedence" - save some confusion when `miren --version` still shows the old one.

Not a big deal, just a thought.

phinze (approved):
Solid UX improvement - way better to check permissions upfront than bail mid-download.

A couple minor thoughts inline, nothing blocking.

cli/commands/upgrade.go
evanphx: oh, great call.

cli/commands/upgrade.go
evanphx: Yeah, not sure how to handle that one. I have another issue that we should probably move to installing miren into ~/.local/bin instead, which most setups prepend on the path.

PR #566, opened 2026-01-30
Clarify scaling mode in CLI output

evanphx:
## Summary

- Display `(auto)` or `(fixed)` suffix on instance counts in CLI output to help users distinguish between autoscaling apps that are idle vs fixed mode apps with 0 instances configured
- Rename STATUS column to SCALE in `app list`
- Add MODE column to `sandbox-pool list`
- Add `scaling_mode` field to JSON output

## Test plan

- [x] Deploy an app with autoscaling mode, verify `m app list` shows `N (auto)` when running and `💤 idle` when scaled to zero
- [x] Deploy an app with fixed mode, verify `m app list` shows `N (fixed)` or `⚠️ 0 (fixed)`
- [x] Run `m app <name>` and verify pool display shows scaling mode
- [x] Run `m sandbox-pool list` and verify MODE column appears
- [x] Run `m app list --json` and verify `scaling_mode` field is present

### DIFF
--- a/cli/commands/app_list.go
+++ b/cli/commands/app_list.go
@@ -160,6 +160,7 @@ func AppList(ctx *Context, opts struct {
 		inCooldown   bool
 		crashCount   int64
 		cooldownLeft time.Duration
+		isAutoscale  bool // true if any service uses autoscaling mode
 	}
 	poolStateMap := make(map[string]*appPoolState)
 	now := time.Now()
@@ -169,7 +170,9 @@ func AppList(ctx *Context, opts struct {
 
 		appName := ui.CleanEntityID(pool.App.String())
 		if poolStateMap[appName] == nil {
-			poolStateMap[appName] = &appPoolState{}
+			poolStateMap[appName] = &appPoolState{
+				isAutoscale: true, // default to autoscale, set to false if ANY service uses fixed mode
+			}
 		}
 		state := poolStateMap[appName]
 		state.ready += int(pool.ReadyInstances)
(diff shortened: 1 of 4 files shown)


cli/commands/app_list.go
```diff
@@ -179,6 +182,15 @@ func AppList(ctx *Context, opts struct {
 			state.crashCount = pool.ConsecutiveCrashCount
 			state.cooldownLeft = pool.CooldownUntil.Sub(now)
 		}
+
+		// Check concurrency mode from the app's active version config
+		if version, ok := versionMap[pool.SandboxSpec.Version.String()]; ok {
+			for _, svc := range version.Config.Services {
```
phinze, line 188: Note: if an app has multiple services with mixed concurrency modes, this labels the whole app as "fixed" since any fixed service sets `isAutoscale = false`. We may want to consider a way to display per-service info in the app list at some point, but this is an improvement for now.

cli/commands/sandbox_pool_list.go
```diff
@@ -30,6 +31,25 @@ func SandboxPoolList(ctx *Context, opts struct {
 		return err
 	}
 
+	// Fetch app versions to determine concurrency mode
```
phinze, line 34: Minor: this version lookup pattern (LookupKind → List → build map) is now duplicated between here and `app_list.go`. Not a big deal, but if we touch this area again we might want to extract a helper.

phinze (approved):
Looks good! This makes the CLI output much clearer - knowing whether 0 instances means "idle and will scale" vs "fixed at zero" is useful information.

A couple of minor notes below, nothing blocking.

cli/commands/app_list.go
evanphx: yeah, I think we should pick up that when we do a wider `app list`, `app status` cleanup.

cli/commands/sandbox_pool_list.go
evanphx: agree.

PR #568, opened 2026-02-02
Add distributed runners Phase 1: registration and join flow

phinze:
## Summary

This implements the foundation for distributed runners, allowing runner nodes to register with a coordinator and connect via mTLS.

- **Join code system**: Mnemonic codes (e.g., `sunset-piano-7x2k`) for secure one-time registration
- **Runner registration RPC**: CreateInvite, Join, ListInvites, RevokeInvite, ListRunners
- **CLI commands**: `miren runner invite/join/list/revoke`
- **Feature flagged**: Behind `MIREN_LABS=distributedrunners`

### What this PR does (Phase 1)
- Runner invitation and join flow
- mTLS certificate issuance for runners
- Node entity creation for registered runners
- CLI commands to manage invites and list runners

### What this PR does NOT do (future phases)
- Actually schedule or distribute work to runners
- Runner-side server mode (`miren server --runner`)
- Sandbox execution on remote runners
- Health checking or heartbeat monitoring
- Runner disconnection/reconnection handling

This is purely the registration infrastructure - runners can join and get certificates, but no workloads will be scheduled to them yet.

### New packages
- `pkg/joincode` - Generate/validate mnemonic join codes
- `pkg/runnerconfig` - Runner configuration file management  
- `servers/runner` - RunnerRegistration RPC server

### Schema changes
- New `runner_invite` entity kind
- Added `runner_id`, `version`, `registered_at` fields to `node` entity

### Usage (when feature enabled)

```bash
# Coordinator: create invite
export MIREN_LABS=distributedrunners
miren runner invite --labels=compute:gpu
# Output: sunset-piano-7x2k

# Runner: join coordinator
miren runner join coordinator.example.com:8443
# Enter code when prompted
```

## Test plan

- [x] `pkg/joincode` unit tests pass
- [x] `make lint` passes
- [ ] CI passes
- [ ] Manual test in dev environment


cli/commands/runner_join.go
```diff
@@ -0,0 +1,111 @@
...
+
+	cs, err := rpc.NewState(ctx, rpc.WithSkipVerify, rpc.WithLogger(ctx.Log))
+	if err != nil {
+		return fmt.Errorf("failed to create RPC state: %w", err)
+	}
+
+	client, err := cs.Connect(coordinator, "dev.miren.runtime/runner")
+	if err != nil {
+		return fmt.Errorf("failed to connect to coordinator: %w", err)
+	}
```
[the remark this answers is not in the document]
phinze: Keeping this as-is intentionally. This is a bootstrap flow where the runner doesn't yet have the coordinator's CA certificate - that's exactly what we're fetching during join. The security model here is trust-on-first-use (similar to SSH's first connection), with the join code providing the authentication. The code is single-use, validated server-side, and short-lived.

Adding `--insecure` would just be friction without security benefit - if an attacker has the join code, they'd just use the flag anyway.

cli/commands/runner_join.go
```diff
@@ -0,0 +1,112 @@
...
+
+	ctx.Printf("Joining coordinator at %s\n", coordinator)
+	ctx.Printf("\nEnter join code: ")
+
+	reader := bufio.NewReader(os.Stdin)
+	code, err := reader.ReadString('\n')
+	if err != nil {
+		return fmt.Errorf("failed to read code: %w", err)
+	}
```
evanphx: Let's allow the code to be passed as an argument, will make it easier to enable this to happen automated.

cli/commands/runner_join.go
```diff
@@ -0,0 +1,112 @@
...
+	if !joincode.Validate(code) {
+		return fmt.Errorf("invalid join code format")
+	}
+
+	cs, err := rpc.NewState(ctx, rpc.WithSkipVerify, rpc.WithLogger(ctx.Log))
+	if err != nil {
+		return fmt.Errorf("failed to create RPC state: %w", err)
+	}
+
+	client, err := cs.Connect(coordinator, "dev.miren.runtime/runner")
```
evanphx: 💡 we should probably put the "dev.miren.runttime/runner", etc strings behind some type-safe registry.

cli/commands/runner_list.go
```diff
@@ -0,0 +1,229 @@
...
+
+	res, err := rc.ListRunners(ctx)
+	if err != nil {
+		return err
+	}
+
+	runners := res.Runners()
+
+	if opts.IsJSON() {
+		type RunnerJSON struct {
```
evanphx, line 34: 💡 we should maybe make CLAUDE.md instructions to be sure that we include this in all commands!

pkg/joincode/wordlist.go
```diff
@@ -0,0 +1,60 @@
+package joincode
+
+var adjectives = []string{
```
evanphx, line 3: ❤️  I love words!

servers/runner/registration.go
```diff
@@ -0,0 +1,433 @@
...
+		"expires_at", expiresAt.Format(time.RFC3339),
+		"label_count", len(invite.Labels))
+
+	results.SetCode(code)
+	results.SetExpiresAt(standard.ToTimestamp(expiresAt))
+
+	return nil
+}
+
+func (s *RegistrationServer) Join(ctx context.Context, req *runner_v1alpha.RunnerRegistrationJoin) error {
```
evanphx, line 104: ❓  Maybe I missed it, but do we need something on the rpc side to allow this to be called by an anonymous client? That's a good feature to enable anyway.

evanphx (approved):
Image

cli/commands/runner_join.go
phinze: Done in 589db6d - join code can now be passed as a positional arg: `miren runner join <coordinator> <join-code>`. Also switched to `ui.PromptForInput` for consistent interactive UX when prompting.

cli/commands/runner_join.go
phinze: Added `pkg/rpc/service.go` with `ServiceID` type alias and `rpc.ServiceRunner` constant. Using a type alias for now so existing string call sites still work - when we do the full sweep we can convert to a distinct type and update function signatures. Incremental path forward!

servers/runner/registration.go
phinze: This is handled by a public methods feature we built on a separate branch (rpc-public-methods). The `Join` method is marked as public in the schema, allowing unauthenticated calls for the bootstrap flow. Will PR that shortly!

cli/commands/runner_list.go
phinze: Good idea! Filed [MIR-662](https://linear.app/miren/issue/MIR-662) to audit existing commands and add CLAUDE.md guidance. Most list commands already follow the pattern (7/11), but a few gaps to clean up.

servers/runner/registration.go
```diff
@@ -0,0 +1,433 @@
...
+	nodeAttrs := node.Encode()
+	nodeEntity := &entityserver_v1alpha.Entity{}
+	nodeEntity.SetAttrs(nodeAttrs)
+
+	nodePutResp, err := s.EAC.Put(ctx, nodeEntity)
+	if err != nil {
+		s.Log.Error("Failed to create node entity", "error", err, "runner_id", runnerID)
+		results.SetError("failed to register runner")
+		return nil
+	}
```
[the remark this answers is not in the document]
phinze: That's a lot of complexity for an edge case - skipping this for now.

servers/runner/registration.go
```diff
@@ -0,0 +1,433 @@
...
+
+	if time.Now().After(invite.ExpiresAt) {
+		results.SetError("join code has expired")
+		return nil
+	}
+
+	runnerID := args.RunnerId()
+	if runnerID == "" {
+		runnerID = uuid.New().String()
+	}
```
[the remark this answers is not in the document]
phinze: This will get cleaned up in phase 2

PR #569, opened 2026-02-02
Add distributed runners Phase 2: WireGuard networking and scheduling

phinze:
## Summary

This PR adds Phase 2 of the distributed runners feature, building on Phase 1 (#556):

- **Configurable network backend**: Grunge now supports both vxlan and wireguard backends
- **Network config in join response**: Runners receive etcd endpoints and network backend type when joining
- **Runner network initialization**: Distributed runners join the Flannel mesh on startup
- **Stateful/stateless scheduling**: 
  - Sandboxes with miren disk volumes are scheduled to the coordinator
  - Stateless sandboxes prefer runner nodes when available
- **etcd mTLS security**: When `DistributedRunners` feature flag is enabled, all etcd connections use mutual TLS

## Security Architecture

The etcd security concern from the initial implementation has been addressed:

| Component | etcd Auth |
|-----------|-----------|
| etcd server | Starts with `--client-cert-auth` and `--trusted-ca-file` |
| Coordinator | Uses client cert issued by cluster CA |
| Grunge (Flannel) | Uses same client cert for network coordination |
| Runners | Will use join-issued cert (when `runner start` implemented) |

All certificates are issued by the same CA used for RPC authentication, so the PKI is unified.

**Remaining consideration**: Runners still have direct etcd access (read/write). For now this is acceptable since runners are authenticated. Future work could proxy Flannel coordination through the Miren RPC layer for finer-grained access control.

## Test plan

- [x] Lint passes
- [x] Unit tests for IsUUID helper
- [x] Unit tests for stateful/stateless scheduling
- [x] etcd TLS config builds and integrates correctly
- [ ] Manual testing with coordinator + runner
- [ ] Test WireGuard connectivity across nodes
- [ ] Implement `runner start` command to test full distributed flow

### DIFF
--- a/api/runner/rpc.yml
+++ b/api/runner/rpc.yml
@@ -67,6 +67,16 @@ interfaces:
           - name: runner_id
             type: string
             doc: The assigned or confirmed runner ID
+          - name: etcd_endpoints
+            type: list
+            element: string
+            doc: etcd endpoints for Flannel coordination
+          - name: etcd_prefix
+            type: string
+            doc: etcd prefix for Flannel network
+          - name: network_backend
+            type: string
+            doc: Network backend type (vxlan or wireguard)
           - name: error
             type: string
             doc: Error message if join failed
--- a/api/runner/rpc.yml
+++ b/api/runner/rpc.yml
@@ -32,7 +32,7 @@ interfaces:
 
       - name: Join
         index: 1
-        public: true
+        public: true  # Allow unauthenticated access - join code provides authentication
         doc: Join a runner to the coordinator using an invite code
         parameters:
           - name: code
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -3,6 +3,7 @@ package coordinate
 import (
 	"context"
 	"crypto/sha1"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/hex"
 	"encoding/pem"
@@ -62,6 +63,13 @@ import (
 	"miren.dev/runtime/version"
 )
 
+// EtcdTLSConfig holds TLS configuration for connecting to etcd with mTLS.
+type EtcdTLSConfig struct {
+	CertPEM []byte // Client certificate PEM
+	KeyPEM  []byte // Client private key PEM
+	CACert  []byte // CA certificate PEM for verifying server
+}
+
 type CoordinatorConfig struct {
 	Address         string              `json:"address" yaml:"address"`
 	EtcdEndpoints   []string            `json:"etcd_endpoints" yaml:"etcd_endpoints"`
@@ -82,6 +90,10 @@ type CoordinatorConfig struct {
 
 	// NoAuth disables authentication entirely (for testing only)
 	NoAuth bool `json:"no_auth" yaml:"no_auth"`
+
+	// EtcdTLS holds mTLS configuration for etcd connections (optional).
+	// When set, the coordinator will use mTLS to connect to etcd.
+	EtcdTLS *EtcdTLSConfig `json:"etcd_tls" yaml:"etcd_tls"`
 
 	Mem       *metrics.MemoryUsage
 	Cpu       *metrics.CPUUsage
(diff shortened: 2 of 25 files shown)


components/runner/runner.go
```diff
@@ -293,6 +311,46 @@ func (r *Runner) Start(ctx context.Context) error {
...
+	}
+
+	// Update deps with the leased IP
+	lease := gn.Lease()
+	r.deps.IPv4Routable = lease.IPv4()
+
+	r.Log.Info("Joined Flannel network", "ipv4", lease.IPv4().String())
+
+	return nil
+}
```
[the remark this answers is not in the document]
phinze: This is a false positive. Looking at `grunge.go`:

- Line 299: `bn, err := be.RegisterNetwork(ctx, &wg, config)` acquires the lease synchronously
- Line 315: `n.lease = bn.Lease()` sets the lease **before** the `eg.Go()` goroutine starts
- Lines 317-340: The goroutine runs `CompleteLease()` which handles **lease renewal**, not initial acquisition

The lease is populated synchronously before `Start()` returns. The code correctly accesses `gn.Lease()` after `Start()` completes.

controllers/scheduler/scheduler.go
```diff
@@ -92,6 +122,49 @@ func (c *Controller) Reconcile(ctx context.Context, sandbox *compute_v1alpha.San
...
+// filterRunnerNodes returns only nodes that are distributed runners (have UUID runner_id).
+func (c *Controller) filterRunnerNodes(nodes []*compute_v1alpha.Node) []*compute_v1alpha.Node {
+	var runners []*compute_v1alpha.Node
+	for _, node := range nodes {
+		if joincode.IsUUID(node.RunnerId) {
+			runners = append(runners, node)
+		}
+	}
+	return runners
+}
```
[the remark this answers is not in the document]
phinze: The UUID-based detection is intentional - distributed runners are assigned UUIDs during registration while the coordinator uses the cluster ID. This is enforced at the registration layer.

Adding an explicit coordinator field is something we could consider if the heuristic causes issues, but it's working well for our current needs.

servers/runner/registration.go
```diff
@@ -234,6 +240,17 @@ func (s *RegistrationServer) Join(ctx context.Context, req *runner_v1alpha.Runne
...
 
+	// Provide network configuration for distributed runners
+	if len(s.EtcdEndpoints) > 0 {
+		results.SetEtcdEndpoints(s.EtcdEndpoints)
+	}
+	if s.EtcdPrefix != "" {
+		results.SetEtcdPrefix(s.EtcdPrefix + "/sub/flannel")
+	}
+	if s.NetworkBackend != "" {
+		results.SetNetworkBackend(s.NetworkBackend)
```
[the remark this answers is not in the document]
phinze: The configuration doesn't use trailing slashes in practice - we've kept the pattern consistent with the coordinator code in `cli/commands/server.go`. Happy to add normalization if this becomes an issue, but keeping it simple for now.

api/runner/rpc.yml
```diff
@@ -66,6 +66,16 @@ interfaces:
...
+          - name: etcd_endpoints
+            type: list
+            element: string
+            doc: etcd endpoints for Flannel coordination
+          - name: etcd_prefix
+            type: string
+            doc: etcd prefix for Flannel network
+          - name: network_backend
+            type: string
+            doc: Network backend type (vxlan or wireguard)
```
evanphx, line 79: 💡 I think I want to do do a follow-on after this that moves us away from flannel to our own mechanism. I'd rather have control over it and have it flow through rpc rather than directly to etcd.

components/coordinate/coordinate.go
```diff
@@ -103,6 +116,119 @@ const (
...
+		if err != nil {
+			return nil, fmt.Errorf("missing key for CA: %w", err)
+		}
+
+		ca, err = caauth.LoadFromPEM(data, key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load CA: %w", err)
+		}
+	} else {
+		log.Info("generating new CA for etcd TLS", "path", certPath)
```
evanphx: we're reusing the same ca that we used to generate other certs, so I don't think we'll ever hit this path and can just remove it?

controllers/scheduler/scheduler.go
```diff
@@ -92,6 +122,49 @@ func (c *Controller) Reconcile(ctx context.Context, sandbox *compute_v1alpha.San
 	return nil
 }
 
+// isStateful determines if a sandbox is stateful based on its volume configuration.
+// A sandbox is considered stateful if it has volumes with provider="miren" (disk volumes).
+func (c *Controller) isStateful(sandbox *compute_v1alpha.Sandbox) bool {
```
evanphx, line 126: should we figure out if the app is using /miren/data/local and make it stateful? I'm not sure we want to bite off making /miren/data/local actually not local for this release.

controllers/scheduler/scheduler.go
```diff
@@ -92,6 +122,49 @@ func (c *Controller) Reconcile(ctx context.Context, sandbox *compute_v1alpha.San
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// findCoordinatorNode finds the coordinator node among the available nodes.
+// The coordinator is identified by having a non-UUID runner_id (e.g., "miren" or empty).
+func (c *Controller) findCoordinatorNode(nodes []*compute_v1alpha.Node) *compute_v1alpha.Node {
```
evanphx, line 146: 💡 totes fine but we should probably add an attribute to nodes that indicates it's a coordinator.

pkg/serverconfig/schema.yml
```diff
@@ -161,6 +161,17 @@ configs:
         env: MIREN_SERVER_STOP_SANDBOXES_ON_SHUTDOWN
         toml: stop_sandboxes_on_shutdown
 
+      network_backend:
+        type: string
+        default: vxlan
```
evanphx, line 166: ❓ I'm ok with defaulting to vxlan, but curious why we're no defaulting to wireguard if we feel like it's the better default.

evanphx (approved):
Image

api/runner/rpc.yml
phinze: We discussed this and agreed it's something we'll plan next, but that it's okay to lean on flannel for v1.

controllers/scheduler/scheduler.go
phinze: Hm yes good point... but how to figure that out... will give it some thought but might need to change the host shared volume to be opt-in via config to support this 🤔

controllers/scheduler/scheduler.go
phinze: Yeah that would be much cleaner, will do

pkg/serverconfig/schema.yml
phinze: I'm thinking this will change once we're not feature flagged

components/coordinate/coordinate.go
phinze: Good call — removed the CA generation branch. SetupEtcdTLS now requires the CA to already exist (created by LoadCA) and returns an error if it's missing.

controllers/scheduler/scheduler.go
phinze: Done — added `is_coordinator` bool field to the node entity. The coordinator sets it during registration, and the scheduler uses it instead of the UUID heuristic.

controllers/scheduler/scheduler.go
phinze: Dug into this — `/miren/data/local` is auto-mounted as a host bind mount for every app sandbox, so there's no way to detect usage at scheduling time. I think the right framing is that "local" data is node-local by definition: each runner node gets its own `/miren/data/local`, and apps needing portable storage should use Miren Disks. We could document this clearly when distributed runners goes GA.

controllers/scheduler/scheduler.go
phinze: Done — went with using the existing `constraints` labels instead of adding a new field. The coordinator now sets `role=coordinator` in its constraints, and the scheduler checks for that label. This way we can extend with `role=ingress`, capability labels, etc. without schema changes.

pkg/rpc/authenticator.go
```diff
@@ -32,14 +33,25 @@ func (n *NoOpAuthenticator) NoAuthorization(ctx context.Context, r *http.Request
...
 	return l.NoAuthorization(ctx, r)
 }
 
 func (l *LocalOnlyAuthenticator) NoAuthorization(ctx context.Context, r *http.Request) (bool, string, error) {
+	// Check if this path allows unauthenticated access
+	for _, prefix := range l.AllowUnauthenticatedPaths {
+		if strings.HasPrefix(r.URL.Path, prefix) {
+			return true, "runner-join", nil
+		}
+	}
```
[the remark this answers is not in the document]
phinze: This code (`AllowUnauthenticatedPaths` / `NoAuthorization`) doesn't exist in the current codebase — looks like a phantom finding.

cli/commands/app_delete.go
```diff
@@ -4,15 +4,24 @@ import (
...
 	ConfigCentric
 }) error {
 	appName := opts.AppName
+	if appName == "" {
+		if ac, err := appconfig.LoadAppConfig(); err == nil && ac != nil && ac.Name != "" {
+			appName = ac.Name
+		}
+	}
+	if appName == "" {
+		return fmt.Errorf("app is required")
```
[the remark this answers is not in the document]
phinze: Not part of this PR's changeset — this is existing code unrelated to distributed runners.

cli/commands/route_set_default.go
```diff
@@ -5,12 +5,23 @@ import (
...
 	ConfigCentric
 }) error {
+	appName := opts.AppName
+	if appName == "" {
+		if ac, err := appconfig.LoadAppConfig(); err == nil && ac != nil && ac.Name != "" {
+			appName = ac.Name
+		}
+	}
+	if appName == "" {
+		return fmt.Errorf("app is required")
```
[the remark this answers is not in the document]
phinze: Not part of this PR's diff.

cli/commands/route_set.go
```diff
@@ -5,23 +5,34 @@ import (
...
 	ConfigCentric
 }) error {
+	appName := opts.AppName
+	if appName == "" {
+		if ac, err := appconfig.LoadAppConfig(); err == nil && ac != nil && ac.Name != "" {
+			appName = ac.Name
+		}
+	}
+	if appName == "" {
+		return fmt.Errorf("app is required")
```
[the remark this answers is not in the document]
phinze: Not part of this PR's diff.

cli/commands/sandbox_pool_list.go
```diff
@@ -30,6 +31,25 @@ func SandboxPoolList(ctx *Context, opts struct {
...
+	// Fetch app versions to determine concurrency mode
+	versionKindRes, err := eac.LookupKind(ctx, "app_version")
+	if err != nil {
+		return err
+	}
+
+	versionsRes, err := eac.List(ctx, versionKindRes.Attr())
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
phinze: Not part of this PR's diff — existing code unrelated to distributed runners.

pkg/saga/executor.go
```diff
@@ -0,0 +1,512 @@
...
+		}
+		node := def.Actions[actionName]
+		if node == nil {
+			continue
+		}
+		// Extract output keys from the result
+		if err := extractOutputs(node, result.Output, outputs); err != nil {
+			log.Warn("failed to extract outputs from prior action", "action", actionName, "error", err)
+		}
+	}
```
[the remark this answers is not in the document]
phinze: Not part of this PR's diff — existing code unrelated to distributed runners. Worth tracking separately though.

PR #570, opened 2026-02-04
Capture BuildKit output in build logs

phinze:
Fixes MIR-658

Build logs were only showing the detection phase stuff, but all the
juicy BuildKit output (layer pulls, cache status, command output) that
you see during `miren deploy` was missing from `miren logs --build`.

Turns out we had two `WithStatusUpdates` callbacks - one for persisting
logs, one for streaming to the client - but `WithStatusUpdates` just
overwrites, it doesn't chain. So the streaming callback was clobbering
the persistence one.

Combined them into a single callback that does both. Now build logs
include everything: layer info, phase markers, cache status, and
importantly the actual command output (like `bun install` or `go build`
output, and onbuild commands).

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -723,20 +723,9 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
 		WithBuildArg("MIREN_VERSION", mrv.Version),
 	)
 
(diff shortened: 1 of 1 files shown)


evanphx (approved):
aaaaaaah!

PR #571, opened 2026-02-04
admin: Finish admin API with endpoint, docs, CLI, and test improvements

evanphx:
## Summary

- Move admin well-known path from `/.well-known/miren-admin` to `/.well-known/miren/admin` to match the existing `/.well-known/miren/health` convention
- Update docs to specify JSON-RPC 2.0 protocol (matches actual wire format — the server already sends `"jsonrpc": "2.0"`)
- Add documentation for how `$methods` introspection works
- Add `--func-help` CLI flag for per-method help (e.g. `miren admin get-user --func-help`)
- Add type-aware CLI parameter parsing using `$methods` type info (string, number, boolean, object, array)
- Add `DescribeMethods` RPC for fetching info about specific methods without listing all
- Extract shared `fetchMethods` helper on the server to reduce duplication
- Add comprehensive test suite (23 tests) covering Invoke, ListMethods, and DescribeMethods

## Test plan

- [x] All 23 new admin server tests pass via `./hack/it ./servers/admin/`
- [x] Full test suite (`make test`) passes with no new failures
- [x] Manual: `miren admin --list` still discovers methods
- [x] Manual: `miren admin <method> --func-help` shows method details
- [x] Manual: Type-aware param parsing coerces values correctly (e.g. number params parsed as numbers)


cli/commands/commands.go
```diff
@@ -89,6 +89,9 @@ func RegisterAll(d *mflags.Dispatcher) {
 	// Auth commands
 	d.Dispatch("auth generate", Infer("auth generate", "Generate authentication config file", AuthGenerate))
 
+	// Admin commansd
```
phinze, line 92:
```suggestion
	// Admin commands
```

servers/admin/admin.go
```diff
@@ -0,0 +1,466 @@
...
+	wanted := make(map[string]struct{}, len(args.Methods()))
+	for _, name := range args.Methods() {
+		wanted[name] = struct{}{}
+	}
+
+	// Filter to only requested methods
+	var matched []*admin_v1alpha.AdminMethod
+	for _, m := range allMethods {
+		if _, ok := wanted[m.Name()]; ok {
+			matched = append(matched, m)
```
phinze, line 307: Nit: Map iteration order is non-deterministic, so params will appear in random order in `--list` output. Might want to sort by name for consistent output:

```go
// after building params slice
sort.Slice(params, func(i, j int) bool {
    return params[i].Name() < params[j].Name()
})
```

phinze (approved):
Looking good - will be fun to play with this for our own apps.

Two small nits inline from me.

The rabbit is right about:
- **ADMIN_TOKEN override** - users shouldn't be able to clobber the system token via config env vars
- **`lipgloss.Width()` for label alignment** - otherwise multi-byte chars will misalign

The other CodeRabbit comments (deprecated `strings.Title`, docs for `--func-help`) are worth addressing but not blocking.

servers/admin/admin.go
evanphx: ooooh yeah.

go.mod
```diff
@@ -113,6 +113,7 @@ require (
 	google.golang.org/protobuf v1.36.10
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/klog/v2 v2.130.1
+	miren.dev/jsonrpc3/go/jsonrpc3 v0.0.0-20260106052505-c98e2702b093
```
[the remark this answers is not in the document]
evanphx: That's fine, our jsonrpc 3.0 supports 2.0 fine and has some nice bits we'll use.

servers/admin/admin.go
evanphx: this calls for my `mapx.StableOrder` iterator!

PR #573, opened 2026-02-05
Add schema-driven public method support to RPC layer

phinze:
For distributed runners, we need the `Join` RPC method to be callable without a TLS client cert - it's a bootstrap flow where the runner doesn't have certs yet (that's the whole point of joining).

Previously all RPC methods required TLS client certs. Now you can mark individual methods as `public: true` in the schema and they'll allow unauthenticated callers. Everything else stays locked down, and capability-based auth (Ed25519 signatures) is still always enforced.

```yaml
methods:
  - name: Join
    public: true  # bootstrap flow, no certs yet
  - name: CreateInvite
    # default: requires TLS client cert
```

## Authenticator/Authorizer Interface Refactor

This PR also refactors the auth interfaces to cleanly separate authentication from authorization:

**Authentication** (`rpc.Authenticator`):
- Single method: `Authenticate(ctx, req) (*Identity, error)`
- Returns `*Identity` for authenticated callers, `nil` for no credentials
- `Identity` includes Subject, Groups, Method (cert/jwt/token/anonymous), and Metadata

**Authorization** (`rpc.Authorizer`):
- Single method: `Authorize(ctx, identity, resource, action) error`
- Matches Miren Cloud RBAC semantics (resource + action)
- For RPC methods: resource = interface name, action = method name (both lowercase)

**Auth flow is now:**
1. `ServeHTTP` calls `Authenticate` once, stores Identity in context
2. `handleCalls` checks if Identity is present for non-public methods
3. If authorizer configured, calls `Authorize(identity, resource, action)`
4. `NoOpAuthenticator` returns anonymous identity (tests always pass)
5. `LocalOnlyAuthenticator` returns cert identity or nil (no authorizer needed)
6. `cloudauth.RPCAuthenticator` implements both - JWT validation + RBAC

Cert-authenticated callers bypass RBAC (local/internal trust model).

---

**Also fixes a pre-existing bug:** Several RPC error paths were setting HTTP headers *after* calling `WriteHeader()`, which meant clients never received the `rpc-status` and `rpc-error` headers on error responses. This affected unknown method, unknown capability, and auth rejection paths in both `handleCalls` and `startCallStream`. Now headers are set before `WriteHeader()` so clients get proper error details.

### DIFF
--- a/pkg/rpc/server.go
+++ b/pkg/rpc/server.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"runtime/debug"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -1035,6 +1036,20 @@ func (s *Server) handleCalls(w http.ResponseWriter, r *http.Request) {
 				w.Header().Add("rpc-error", "authentication required")
 				w.WriteHeader(http.StatusUnauthorized)
 				return
+			}
+
+			// Check authorization if an authorizer is configured
+			if s.state.authorizer != nil {
+				resource := strings.ToLower(mm.InterfaceName)
+				action := strings.ToLower(mm.Name)
+				if err := s.state.authorizer.Authorize(ctx, identity, resource, action); err != nil {
+					s.state.log.Warn("authorization denied",
+						"resource", resource, "action", action, "subject", identity.Subject, "error", err)
+					w.Header().Add("rpc-status", "forbidden")
+					w.Header().Add("rpc-error", err.Error())
+					w.WriteHeader(http.StatusForbidden)
+					return
+				}
 			}
 		}
 
--- a/pkg/rpc/server.go
+++ b/pkg/rpc/server.go
@@ -303,46 +303,43 @@ func (s *Server) setupMux() {
 }
 
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	// If no authenticator configured, allow all requests
 	if s.state.authenticator == nil {
 		s.mux.ServeHTTP(w, r)
 		return
 	}
 
-	// Check for JWT token in Authorization header
-	authHeader := r.Header.Get("Authorization")
-	if authHeader != "" {
-		// Try JWT authentication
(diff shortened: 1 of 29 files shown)


pkg/rpc/server.go
```diff
@@ -1055,6 +1058,20 @@ func (s *Server) handleCalls(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		// Enforce authentication for non-public methods
+		// Public methods can be called without a TLS client certificate
+		if !mm.Public {
+			hasCert := r.TLS != nil && len(r.TLS.PeerCertificates) > 0
```
evanphx: I don't think this will work correctly because when users use miren.cloud authentication to talk to a miren server, the client doesn't present a peer certificate because we use the miren cloud authenticator which validates the JWT the client presents instead.

evanphx (changes requested):

pkg/rpc/server.go
phinze: Yeah I didn't spend enough time thinking about the cloud-y auth path. Lemme dig in here

phinze: K now checks for Authorization header too, so JWT auth works. Added tests covering the different auth scenarios.

pkg/rpc/server.go
```diff
@@ -1055,6 +1058,20 @@ func (s *Server) handleCalls(w http.ResponseWriter, r *http.Request) {
...
+			if !hasCert {
+				s.state.log.Warn("authentication required for non-public method",
+					"method", method, "interface", mm.InterfaceName)
+				w.WriteHeader(http.StatusUnauthorized)
+				w.Header().Add("rpc-status", "unauthorized")
+				w.Header().Add("rpc-error", "authentication required")
+				return
+			}
+		}
```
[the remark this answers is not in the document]
phinze: Addressed in a more recent commit - double check me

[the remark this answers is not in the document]
evanphx: The robot is right, we need to define what happens with Public-ness inside startCallStream.

evanphx (approved):
Just need to sort out startCallStream

phinze:
Added the same auth check to `startCallStream` - streaming methods now enforce Public flag + Identity + Authorization consistently with regular RPC methods.

PR #576, opened 2026-02-06
Fix DNS failures during sandbox startup

evanphx:
## Summary

- When a sandbox starts, its containers begin making DNS queries before the entity watcher processes the RUNNING status update, causing DNS to return empty responses from unknown source IPs
- Added lazy resolution: when the DNS server sees an unknown source IP, it sleeps 200ms to let the watcher catch up, then falls back to searching the entity store for a matching sandbox
- The entity lookup registers sandboxes regardless of status (not just RUNNING), since the container is clearly running if it's making DNS queries

## Test plan

- [x] `TestResolveUnknownIPWithoutEntityClient` — graceful no-op when entity client is nil
- [x] `TestResolveUnknownIPFindsAndRegistersSandbox` — full integration test with in-memory entity server, creates app/version/sandbox entities and verifies lazy lookup registers all DNS mappings
- [x] `TestResolveUnknownIPNoMatchingIP` — returns false when no sandbox matches the IP
- [x] All existing DNS tests continue to pass
- [x] `make lint` clean

### DIFF
--- a/api/compute/compute.go
+++ b/api/compute/compute.go
@@ -1,3 +1,17 @@
 package compute
 
+import "miren.dev/runtime/api/compute/compute_v1alpha"
+
 //go:generate go run ../../pkg/entity/cmd/schemagen -input schema.yml -output compute_v1alpha/schema.gen.go -pkg compute_v1alpha
+
+// SandboxActive reports whether a sandbox status indicates the sandbox
+// may be actively running (PENDING, NOT_READY, or RUNNING).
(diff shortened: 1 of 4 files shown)


phinze (approved):
Talked this over live - should also fix the missing PENDING status here but good fix overall https://github.com/mirendev/runtime/commit/0471d2c54b09d7294df7c56cf63c3ac4d47d4f59

PR #579, opened 2026-02-06
Re-read App from store in Launcher.Reconcile to coalesce rapid deploys

evanphx:
When multiple AppVersions are created in quick succession (rapid deploys, batch env var sets, programmatic API calls), the controller framework dispatches a watch event for each change. Each event embeds the App entity snapshot from dispatch time, so processing the v2 event always sees ActiveVersion=v2 even if v3 is already committed.

This causes the launcher to create an intermediate SandboxPool for every version, only to immediately scale each to 0 and replace it with the next. N rapid versions = N pools created, N-1 immediately torn down.

Fix by re-reading the App from the entity store at the start of Reconcile(), ignoring the stale event snapshot. Every reconciliation now sees the latest ActiveVersion. When events for v1, v2, v3 queue up, the first reconcile re-reads and sees v3, creates one pool. The subsequent reconciles also re-read, see v3, find the pool already exists, and no-op.

The extra Get has negligible cost — the launcher already does a Get for the AppVersion and the App metadata on every reconcile.

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"sync"
 
 	"miren.dev/runtime/api/compute/compute_v1alpha"
 	"miren.dev/runtime/api/core/core_v1alpha"
@@ -23,6 +24,8 @@ import (
 type Launcher struct {
 	Log *slog.Logger
 	EAC *entityserver_v1alpha.EntityAccessClient
+
+	appMu sync.Map // per-app mutexes: app ID -> *sync.Mutex
 }
 
 // PoolWithEntity wraps a SandboxPool with its entity, allowing updates without re-fetching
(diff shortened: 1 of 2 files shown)


controllers/deployment/launcher.go
evanphx, line 83: Should we put a per-id mutex RIGHT HERE so that we're only attempting to reconcile an app in a bubble?

controllers/deployment/launcher.go
```diff
@@ -42,13 +44,29 @@ func (l *Launcher) Init(ctx context.Context) error {
 }
 
 func (l *Launcher) Reconcile(ctx context.Context, app *core_v1alpha.App, meta *entity.Meta) error {
-	if app.ActiveVersion == "" {
-		l.Log.Debug("app has no active version, skipping", "app", app.ID)
+	// Re-read from store to get latest state, coalescing rapid updates.
```
evanphx, line 56: The per-app-id mutex should go here.

controllers/deployment/launcher.go
```diff
@@ -44,6 +47,12 @@ func (l *Launcher) Init(ctx context.Context) error {
 }
 
 func (l *Launcher) Reconcile(ctx context.Context, app *core_v1alpha.App, meta *entity.Meta) error {
+	// Serialize reconciles for the same app to avoid races between rapid deploys.
+	val, _ := l.appMu.LoadOrStore(app.ID, &sync.Mutex{})
+	mu := val.(*sync.Mutex)
+	mu.Lock()
+	defer mu.Unlock()
```
phinze, line 54: Dying for a close-at-hand metrics subsystem that we could use to toss some visibility into the lock behavior here. 🔜 🚀

phinze (approved):

controllers/deployment/launcher.go
evanphx: yeeeahhh

PR #580, opened 2026-02-06
Fix adding [[env]] to app.toml wiping existing env vars

evanphx:
When a user added their first [[env]] entry to app.toml and deployed, all existing environment variables with Source="" were silently lost.

The root cause was in the merge logic: mergeVariablesFromAppConfig() and mergeServiceEnvVars() treated empty-source vars as Source="config". When app.toml had no [[env]] section the merge short-circuited and all vars were preserved. But as soon as any [[env]] entry appeared, the full merge ran — classifying Source="" vars as "config" and dropping them because they weren't in the new app.toml config.

Env vars created before source tracking was added, or through code paths that don't explicitly set Source, all have Source="". Changing the backward-compatibility default from "config" to "manual" ensures these unknown-source vars are preserved rather than silently deleted.

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -116,7 +116,7 @@ func mergeServiceEnvVars(existingEnvs []core_v1alpha.Env, newEnvs []core_v1alpha
 	for _, e := range existingEnvs {
 		source := e.Source
 		if source == "" {
-			source = "config" // backward compatibility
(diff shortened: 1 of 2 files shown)


phinze (approved):
Goooood bug and nice find 🔧 🔍

PR #584, opened 2026-02-07
Minor cleanup of $methods handling in admin server

evanphx:
Minor cleanup of handling the response to $methods

### DIFF
--- a/servers/admin/admin.go
+++ b/servers/admin/admin.go
@@ -375,18 +375,20 @@ func (s *Server) fetchMethods(ctx context.Context, appName string) ([]*admin_v1a
 		return nil, "", fmt.Errorf("method discovery failed: %s", rpcResp.Error.Message)
 	}
 
-	var methodInfos []struct {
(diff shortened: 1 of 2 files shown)


phinze (changes requested):
I think we should keep the parser strict here rather than making it forgiving. Our docs already specify `$methods` returns a flat array, and the producer-side fix is landing separately. Adding tolerance for the wrapper format and name→info map means we're permanently supporting undocumented formats, which makes the protocol fuzzier over time.

If a user hand-rolls a `$methods` handler and gets the format wrong, a clear error is more useful than us silently accepting it (and maybe quietly dropping methods in the map case).

I'd suggest keeping the strict array-only parsing and letting the producer-side fix land. The cleanup of the redundant `json.Marshal` round-trip on `RawMessage` is nice though — worth keeping that part.

evanphx:
Ok, fair!

phinze (approved):
PR description still has the old behavior but otherwise looks good!

PR #586, opened 2026-02-09
Implement ConfigSpec to hold app version config

evanphx:
## Summary

This PR introduces a ConfigVersion entity to hold the configuration that we previously held on an AppVersion. This lets us directly share the ConfigVersion between AppVersions if needed, as well as let's us setup Config before AppVersions exist.

- Eliminate the conversion layer between `ConfigSpec` (stored in `ConfigVersion` entities) and the old `Config` type
- `ResolveConfig` now returns `*ConfigSpec` directly instead of converting back to `Config`
- All readers (launcher, activator, exec, exec_proxy, CLI) use `ConfigSpec` fields natively (`Services[].Command`, `Services[].Concurrency`, `Variables`, etc.)
- All writers (build server, app server) build `ConfigSpec` natively instead of building `Config` then converting
- Add `ConfigVersion` schema with standalone components to avoid type name collisions with inline types
- Add migration helper (`MigrateAppVersionToConfigVersion`) for existing `AppVersion` entities without `ConfigRef`
- Add `EAC()` accessor to `entityserver.Client` for direct entity access needed by `ResolveConfig`

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./api/core/...` passes
- [x] `go test ./servers/build/...` passes
- [x] `go test ./servers/app/...` passes
- [x] `go test ./servers/exec/...` passes
- [x] `go test ./servers/exec_proxy/...` passes
- [x] `go test ./controllers/deployment/...` passes
- [x] `go test ./controllers/sandboxpool/...` passes
- [x] `go test ./components/activator/...` passes (unit tests; `TestConcurrentPoolIncrement` excluded - pre-existing etcd infra issue)
- [x] `go test ./cli/commands/...` passes
- [x] `make lint` — 0 issues

### DIFF
--- a/api/core/config_version.go
+++ b/api/core/config_version.go
@@ -1,0 +1,95 @@
+package compute
+
+import (
+	"context"
+	"fmt"
+
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+)
+
+// ConfigSpecFromConfig converts the existing inline Config into a
+// ConfigSpec suitable for a ConfigVersion entity. This is used during
+// dual-write to create ConfigVersion entities alongside inline Config.
+func ConfigSpecFromConfig(cfg *core_v1alpha.Config) core_v1alpha.ConfigSpec {
+	spec := core_v1alpha.ConfigSpec{
+		Entrypoint:     cfg.Entrypoint,
+		StartDirectory: cfg.StartDirectory,
+	}
+
+	// Build a map of service -> command for merging into services
+	cmdMap := make(map[string]string, len(cfg.Commands))
+	for _, cmd := range cfg.Commands {
+		cmdMap[cmd.Service] = cmd.Command
+	}
+
+	// Convert variables
+	for _, v := range cfg.Variable {
+		spec.Variables = append(spec.Variables, core_v1alpha.ConfigSpecVariables(v))
+	}
+
+	// Convert services, merging in commands
+	for _, svc := range cfg.Services {
+		s := core_v1alpha.ConfigSpecServices{
+			Name:     svc.Name,
+			Command:  cmdMap[svc.Name],
+			Port:     svc.Port,
+			PortName: svc.PortName,
+			PortType: svc.PortType,
+			Image:    svc.Image,
+			Concurrency: core_v1alpha.ConfigSpecServicesConcurrency{
+				Mode:                svc.ServiceConcurrency.Mode,
+				NumInstances:        svc.ServiceConcurrency.NumInstances,
+				RequestsPerInstance: svc.ServiceConcurrency.RequestsPerInstance,
+				ScaleDownDelay:      svc.ServiceConcurrency.ScaleDownDelay,
+				ShutdownTimeout:     svc.ServiceConcurrency.ShutdownTimeout,
+			},
+		}
+
+		// Convert service-level env vars
+		for _, e := range svc.Env {
+			s.Env = append(s.Env, core_v1alpha.ConfigSpecServicesEnv(e))
+		}
+
+		// Convert disks
+		for _, d := range svc.Disks {
+			s.Disks = append(s.Disks, core_v1alpha.ConfigSpecServicesDisks(d))
+		}
+
+		spec.Services = append(spec.Services, s)
+	}
+
+	return spec
+}
+
+// ResolveConfig loads the configuration for an AppVersion. If the version has
+// a ConfigVersion, it loads the ConfigVersion entity and returns its spec directly.
+// Otherwise, it falls back to the inline Config field and converts it.
+func ResolveConfig(ctx context.Context, eac *entityserver_v1alpha.EntityAccessClient, ver *core_v1alpha.AppVersion) (*core_v1alpha.ConfigSpec, error) {
+	if ver.ConfigVersion != "" {
+		ret, err := eac.Get(ctx, ver.ConfigVersion.String())
+		if err != nil {
+			return nil, fmt.Errorf("failed to get config version %s: %w", ver.ConfigVersion, err)
+		}
+
+		var cv core_v1alpha.ConfigVersion
+		cv.Decode(ret.Entity().Entity())
+
+		return &cv.Spec, nil
+	}
+
+	// Fallback to inline config, converting to ConfigSpec
+	spec := ConfigSpecFromConfig(&ver.Config)
+	return &spec, nil
+}
+
+// GetServiceConcurrency returns the concurrency configuration for a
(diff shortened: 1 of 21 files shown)


api/core/schema.yml
```diff
@@ -197,6 +204,13 @@ kinds:
             source:
               type: string
               doc: The source of the variable (config or manual). Defaults to config for backward compatibility.
+            origin:
+              type: string
+              doc: The provenance of the variable (user, file, generated, detected).
+
+    config_ref:
```
phinze: `config_ref` breaks our naming convention for ref fields. Looking across our schemas, refs are named after the target entity: `app`, `artifact`, `active_version`, `project`, `sandbox`, `disk_id`, etc. No other ref uses a `_ref` suffix. I realize we can't use `config` here since the inline config component still occupies that name, but `config_version` would match the target kind and our conventions. (And once inline config is eventually removed, this could become just `config` per the RFD.)

servers/app/app.go
```diff
@@ -143,12 +143,17 @@ func (r *AppInfo) SetConfiguration(ctx context.Context, state *app_v1alpha.CrudS
...
 	var appVer core_v1alpha.AppVersion
+	var spec core_v1alpha.ConfigSpec
 
 	if appRec.ActiveVersion != "" {
 		err = r.EC.GetById(ctx, appRec.ActiveVersion, &appVer)
 		if err != nil {
 			return err
 		}
+		resolvedCfg, err := coreutil.ResolveConfig(ctx, r.EC.EAC(), &appVer)
+		if err == nil {
```
phinze: +1 to CodeRabbit's flag here. The `if err == nil` pattern silently proceeds with an empty `spec` when ResolveConfig fails. Since we then mutate and persist as a new ConfigVersion, this can silently wipe existing config. Should return the error. Same pattern in `SetEnvVar` (~line 424) and `DeleteEnvVar` (~line 528).

api/core/config_version.go
```diff
@@ -0,0 +1,105 @@
...
+// Otherwise, it falls back to the inline Config field and converts it.
+func ResolveConfig(ctx context.Context, eac *entityserver_v1alpha.EntityAccessClient, ver *core_v1alpha.AppVersion) (*core_v1alpha.ConfigSpec, error) {
+	if ver.ConfigRef != "" {
+		ret, err := eac.Get(ctx, ver.ConfigRef.String())
+		if err != nil {
+			return nil, fmt.Errorf("failed to get config version %s: %w", ver.ConfigRef, err)
+		}
+
+		var cv core_v1alpha.ConfigVersion
+		cv.Decode(ret.Entity().Entity())
```
phinze, line 76: `cv.Decode()` has no error return, so if the entity data is malformed or the spec component is missing, we silently return an empty ConfigSpec. Might be worth a defensive check - e.g., if `cv.Spec.Empty()` after decode when we fetched a real entity, that's suspicious enough to log a warning.

components/activator/activator.go
```diff
@@ -725,12 +726,18 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
...
+				a.mu.Unlock()
+				return nil, fmt.Errorf("failed to resolve config: %w", err)
+			}
+			svcConcurrency, err := coreutil.GetServiceConcurrency(spec, service)
 			if err != nil {
 				a.mu.Unlock()
 				return nil, fmt.Errorf("failed to get service concurrency: %w", err)
 			}
-			strategy := concurrency.NewStrategy(&svcConcurrency)
+			sc := core_v1alpha.ServiceConcurrency(svcConcurrency)
```
phinze, line 739: The `core_v1alpha.ServiceConcurrency(svcConcurrency)` type conversion works because both structs happen to have identical field layouts, but it's brittle if either generated type drifts. This pattern repeats in `sandboxpool/manager.go` and multiple places in this file. Could we either update `NewStrategy` to accept `ConfigSpecServicesConcurrency` directly, or add a small conversion helper to make the intent explicit?

servers/build/build.go
```diff
@@ -97,29 +98,30 @@ func NewBuilder(log *slog.Logger, eas *entityserver_v1alpha.EntityAccessClient,
...
 	// Build result by merging
-	envMap := make(map[string]core_v1alpha.Env)
+	envMap := make(map[string]core_v1alpha.ConfigSpecServicesEnv)
 
 	// Keep manual vars - they shadow config vars with the same key
 	for _, e := range existingEnvs {
 		source := e.Source
 		if source == "" {
-			source = "config" // backward compatibility
+			source = "literal" // backward compatibility
```
phinze: Calling out: the `source` field is still carrying "manual"/"config" values throughout the codebase while the new ConfigSpec schema docs describe storage-type semantics ("literal", "vault", "aws-sm"). Not a problem for this PR, but aligning these is likely the next phase per the RFD's direction on separating `source` (storage type) from `origin` (provenance).

phinze (approved):
Paul & Claude here from a review session. We walked through the full diff, compared it against RFD-55, discussed migration safety (Config.Port, Config.Commands, atomicity), confirmed the schema design is sound, and landed on approve with a few nits.

Nice work extracting ConfigVersion as a first-class entity. The ResolveConfig abstraction with inline fallback gives a clean migration path, and the schema design (standalone component for ConfigSpec) is solid. The conversion and migration coverage looks thorough.

CodeRabbit caught a couple things we agree with: the error swallowing in app.go mutations (we flagged the same thing) and the Origin field not being copied in ConfigSpecFromConfig. The ident collision concern in migrate.go is worth a look too.

api/core/schema.yml
evanphx: Good call, fixing.

servers/app/app.go
evanphx: agreed!

api/core/config_version.go
evanphx: hmmm, ok.

components/activator/activator.go
evanphx: What's funny is that our linting complains if we DON'T do that. if we add stuff on either side in the future, Go will make this a compiler error and we'll need to fix it at that point, which is fine.

servers/build/build.go
evanphx: ah shoot. Claude changed those but I thought I changed them all back.

components/activator/activator.go
phinze: ah cool that works!

api/core/config_version.go
phinze: yeah idk fine to skip too - honestly I want to rework the standard ceremony for putting stuff in and out of the entity system... it feels a bit noisy and awkward in its current form

PR #587, opened 2026-02-10
Implement OIDC authentication for HTTP routes (MIR-675)

evanphx:
## Summary

- Add OIDC authentication support for HTTP routes, allowing routes to require OpenID Connect login before proxying to the backend app
- Implement full OAuth2 authorization code flow with PKCE (RFC 7636) and resource indicators (RFC 8707)
- Support both named host routes and the default route via `--default` flag
- Add OIDC provider as a separate entity for reuse across routes
- Gate behind `routeoidc` labs feature flag

### OIDC Core (`pkg/oidc/`)
- `Client` - OIDC client with provider discovery, authorization URL generation, code exchange, and ID token parsing
- `SessionManager` - Cookie-based session management with encrypted state for OIDC flow

### HTTP Ingress (`servers/httpingress/`)
- OIDC middleware that intercepts requests, redirects unauthenticated users to the OIDC provider, handles callbacks, and injects claims as headers
- Supports both named routes and default routes (derives redirect URL from request Host header)

### Entity Schema (`api/ingress/`)
- New `oidc_provider` entity with name, provider URL, client credentials, and scopes
- Extended `http_route` with `oidc_provider` reference and `claim_mappings`
- Route-based attach/detach methods (`AttachOIDCProviderToRoute`, `DetachOIDCProviderFromRoute`)

### CLI Commands (`cli/commands/`)
- `route oidc enable [host] --default` - Enable OIDC on a route with inline provider creation
- `route oidc show [host] --default` - Display OIDC configuration
- `route oidc disable [host] --default` - Remove OIDC from a route
- `server docker install --labs` - Pass labs feature flags as `MIREN_LABS` env var to the container
- Updated output to use `pkg/ui` (NamedValueList, Table with title)

### UI (`pkg/ui/`)
- Added `WithTableTitle` option for displaying a bold title above tables

### Bug Fixes
- Fix nil pointer crash in route watcher when processing entity delete events (delete ops don't carry an entity payload)

## Test plan

- [x] Enable routeoidc labs flag and configure an OIDC provider on a named route
- [x] Enable OIDC on the default route using `--default` flag
- [x] Verify unauthenticated requests redirect to the OIDC provider with correct redirect_uri and resource parameter
- [x] Verify callback completes authentication and redirects back to the original path
- [x] Verify claim-header mappings inject claims into proxied requests
- [x] Verify `route oidc show` and `route oidc disable` work for both named and default routes
- [x] Verify `server docker install --labs routeoidc` passes the env var to the container
- [x] Verify route deletion no longer crashes the route watcher

### DIFF
--- a/api/ingress/routewatcher.go
+++ b/api/ingress/routewatcher.go
@@ -63,6 +63,12 @@ func (rw *RouteWatcher) Watch(ctx context.Context) error {
 
 		routeID := op.EntityId()
 
+		// Delete events don't carry an entity payload
+		if op.OperationType() == entityserver_v1alpha.EntityOperationDelete {
+			rw.removeRouteHost(routeID)
+			return nil
+		}
+
 		var route ingress_v1alpha.HttpRoute
 		route.Decode(op.Entity().Entity())
 
@@ -78,9 +84,6 @@ func (rw *RouteWatcher) Watch(ctx context.Context) error {
 
 		case entityserver_v1alpha.EntityOperationUpdate:
 			rw.updateRouteHost(routeID, host)
-
-		case entityserver_v1alpha.EntityOperationDelete:
-			rw.removeRouteHost(routeID)
 		}
 
 		return nil
--- a/pkg/oidc/client.go
+++ b/pkg/oidc/client.go
@@ -103,39 +103,19 @@ func (c *Client) ExchangeCode(ctx context.Context, code, pkceVerifier string) (*
 	return token, nil
 }
 
-// ParseIDToken parses and validates an ID token JWT without signature verification.
-// Signature verification should be done separately if needed using a JWT validator.
-func (c *Client) ParseIDToken(idToken string) (map[string]interface{}, error) {
-	// Parse without verification - just extract claims
-	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
-	token, _, err := parser.ParseUnverified(idToken, jwt.MapClaims{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse ID token: %w", err)
-	}
-
-	claims, ok := token.Claims.(jwt.MapClaims)
-	if !ok {
-		return nil, fmt.Errorf("failed to extract claims")
-	}
-
-	return claims, nil
-}
-
-// VerifyIDToken validates an ID token's signature and claims.
-// This performs full JWT validation including signature verification.
-func (c *Client) VerifyIDToken(ctx context.Context, idToken string) (map[string]interface{}, error) {
+// ParseIDToken validates an ID token's signature via the provider's JWKS
+// and checks standard claims (issuer, audience).
+func (c *Client) ParseIDToken(ctx context.Context, idToken string) (map[string]interface{}, error) {
 	discovery, err := c.getDiscovery(ctx)
 	if err != nil {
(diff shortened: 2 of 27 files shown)


pkg/oidc/client.go
```diff
@@ -0,0 +1,339 @@
...
+
+	return token, nil
+}
+
+// ParseIDToken parses and validates an ID token JWT without signature verification.
+// Signature verification should be done separately if needed using a JWT validator.
+func (c *Client) ParseIDToken(idToken string) (map[string]interface{}, error) {
+	// Parse without verification - just extract claims
+	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+	token, _, err := parser.ParseUnverified(idToken, jwt.MapClaims{})
```
phinze: `ParseIDToken` uses `ParseUnverified` and is the only code path used — both for session validation and in the callback after receiving a fresh token. `VerifyIDToken` exists with signature verification but is never called (and its JWKS key function has the issue CodeRabbit flagged). Fine for iterating under labs, but before prod we should wire up actual token verification.

servers/httpingress/oidc.go
```diff
@@ -0,0 +1,322 @@
...
+	if route.Default {
+		resource = "cluster:default"
+	} else {
+		resource = route.Host
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Build base URL from the request host, which works for both
+		// named routes and the default route (which has no static host)
+		baseURL := fmt.Sprintf("https://%s", r.Host)
```
phinze: `baseURL` is always `fmt.Sprintf("https://%s", r.Host)`, which means `cookieSecure` is always `true`. Browsers won't send Secure cookies over plain HTTP, so this could block local dev/testing. Might want to derive the scheme from the request or make it configurable.

api/ingress/routewatcher.go
```diff
@@ -63,6 +63,12 @@ func (rw *RouteWatcher) Watch(ctx context.Context) error {
 
 		routeID := op.EntityId()
 
+		// Delete events don't carry an entity payload
```
phinze, line 66: Nice catch on this bug fix — moving the delete check before the entity decode prevents the nil pointer crash cleanly. Delete events don't carry an entity payload, so the old code path through `Decode` would panic.

phinze (approved):
Paul & Claude here from a review session where we walked through the full diff, discussed the security model, and cross-referenced CodeRabbit's findings. We also contributed two commits: a header spoofing fix in `injectClaims` (f4e3023) and a Route OIDC guide doc with labs callout (f7f01ee).

Really nice feature - OIDC auth at the route layer is a great capability. The architecture is solid: separate provider entity for reuse, claim mappings for flexibility, PKCE for security, and labs gating for safe iteration. Approving as-is for labs with notes on things to address before prod.

**Agrees with CodeRabbit on:** session cookie needs actual encryption (base64 != encryption), per-request handler creation breaks discovery cache and perf, JWKS key function returns raw map instead of crypto key.

pkg/oidc/client.go
evanphx: oooph, yeah. i'll get that wired in now.

servers/httpingress/oidc.go
evanphx: ha! I actually hit this while testing.

PR #592, opened 2026-02-11
Fix route updates

evanphx:
The code was incorrectly creating entites that look like http_route/http_route/<host> because it was passing in the full id to the path. Switch to using Update.


evanphx:
Minor, merging to test it out.

PR #593, opened 2026-02-11
Move -v for version to -V to avoid colliding with -v for verbose

evanphx:


phinze (approved):
Whoops good catch! I wonder if we can write a test that will catch collisions for us 🤔

evanphx:
This will catch it in the future! https://github.com/mirendev/mflags/pull/5

PR #595, opened 2026-02-11
Replace httpingress per-request logs with OTel tracing

phinze:
The httpingress logs every request at INFO level — "routing request" and
"using existing lease" — which is noise at scale.

Per RFD-54, OTel tracing is the right replacement. The OTel SDK deps
were already in go.mod and the RPC layer already had span instrumentation,
but it was all going to the no-op global provider.

This wires up real OTLP export (auto-enabled via `OTEL_EXPORTER_OTLP_ENDPOINT`)
and instruments httpingress with three span layers: root request, lease
acquisition, and trace context injection into outbound proxy requests. That
last bit means W3C `traceparent` headers now flow into user sandboxes,
enabling end-to-end traces when users instrument their apps.

The noisy per-request INFO logs are removed — same info lives in span
attributes now. Error/Warn logs are untouched.


evanphx (approved):
Yer boy loves a structured attribute set!

PR #596, opened 2026-02-12
Add per-app cluster pinning

evanphx:
## Summary
- Store per-app cluster state in `~/.config/miren/app-state.toml` so each app remembers which cluster it targets, removing the need to constantly switch the global cluster or pass `-C` on every command
- Cluster resolution for AppCentric commands: `-C` flag > per-app pinned cluster > global default
- `cluster switch` updates both global and per-app state when run inside an app directory
- `cluster list` shows `*` for the effective cluster and `g` for the global default when they differ
- `cluster current` shows which cluster an app is targeting

## Test plan
- [x] Unit tests for app state load/save round-trip, multiple apps, missing file, overwrite
- [x] `make lint` passes with 0 issues
- [ ] Manual: run `miren cluster switch` in an app dir, verify `app-state.toml` is created
- [ ] Manual: run app command, verify pinned cluster is used over global default
- [ ] Manual: use `-C other` flag, verify state updates to new cluster

### DIFF
--- a/appconfig/appstate.go
+++ b/appconfig/appstate.go
@@ -1,0 +1,118 @@
+package appconfig
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+
+	toml "github.com/pelletier/go-toml/v2"
+	"golang.org/x/sys/unix"
+)
+
+const appStateFile = "app-state.toml"
+
+type appStateStore struct {
+	Apps map[string]AppState `toml:"apps"`
+}
+
+type AppState struct {
+	Cluster string `toml:"cluster"`
+}
+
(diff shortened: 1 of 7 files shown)


phinze (changes requested):
I wanna talk a little more about the config before we commit to the 1:1 app:cluster mapping here.

I think this could instead be a baby step towards a `deploy.yaml` whose underlying data model is more robust... I haven't thought it through extensively, but I'm thinking "deploy targets" as the core concept?

Let's chat about this today!

phinze (approved):
Talking this through live - I didn't realize this was client config! This is a totally good next step, we can keep noodling on the rest.

PR #597, opened 2026-02-12
Improve disk mount lifecycle robustness and add integration tests

evanphx:
## Summary
- Improve disk mount lifecycle handling: recover from error/detached states, skip unmount when another mount entity shares the same path, and add a mount-watch controller for disk lease awareness
- Run LSVD controllers in-process when SkipLSVD is set, enabling full integration testing without a separate LSVD server
- Add comprehensive integration test harness with tests covering lifecycle, chaos/failure injection, overlapping mounts, redeployment, and reconciliation scenarios
- Integration tests cause an incorrect use of Labels.Equal which was changed to Labels.SubsetOf, as that's the correct behavior for labels rather than exact equality. 

## Test plan
- [x] Existing unit tests pass (`make test`)
- [x] New integration tests in `controllers/integration/` pass
- [x] Verify disk mount recovery from error and detached states
- [x] Verify overlapping mount cleanup during rapid sandbox replacement

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1021,7 +1021,7 @@ func (c *SandboxController) updateServices(
 		var srv network_v1alpha.Service
 		srv.Decode(ent.Entity())
 
-		if !srv.Match.Equal(md.Labels) {
+		if !srv.Match.SubsetOf(md.Labels) {
 			c.Log.Debug("skipping service, labels do not match", "service", srv.ID, "labels", srv.Match, "entity", md.Labels)
 			continue
 		}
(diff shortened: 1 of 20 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -1021,7 +1021,7 @@ func (c *SandboxController) updateServices(
 		var srv network_v1alpha.Service
 		srv.Decode(ent.Entity())
 
-		if !srv.Match.Equal(md.Labels) {
+		if !srv.Match.SubsetOf(md.Labels) {
```
phinze, line 1024: This fix is correct — `SubsetOf` is the right semantic for label selectors — but it's easy to miss why it's in a disk-robustness PR. Might be worth a one-liner in the PR description noting this was caught by the integration tests exercising the full deployment flow.

phinze (approved):
Paul & Claude here from a review session where we traced through the rapid sandbox replacement flow to understand how the behavioral changes compose. Here's where we landed:

The three core changes — early lease release in `stopSandbox`, immediate `activeLeases` clear in `handleReleasedLease`, and PENDING-not-FAILED on lease conflict — form a tight unit solving the same problem: making disk handoff fast during sandbox replacement while surviving reconciliation ordering races. We walked through the happy path and the race window and both converge correctly.

The `Labels.SubsetOf` change looked like an unrelated drive-by at first, but it's a real bugfix surfaced by the integration tests. Services have `Match: {app: "foo"}` while sandboxes get `{app: "foo", service: "web", pool: "...", instance: "1"}` from the pool manager. The old `Equal` check fails on the length mismatch, so `updateServices` was silently skipping service matching. Worth a note in the PR description since the connection isn't obvious.

The integration test suite is impressive — the chaos test with weighted fault injection and invariant validation is a nice way to build confidence in the convergence properties. CodeRabbit flagged some test-quality nits (error swallowing in reconcile loops, unclosed FDs in mock NBD cleanup) that seem worth addressing as follow-ups but don't block.

LGTM.

controllers/sandbox/sandbox.go
evanphx: thumbsup

PR #598, opened 2026-02-12
Add nested saga support and deterministic execution ordering

evanphx:
## Summary

- Add `RunNested()` for executing child sagas from within a parent saga action, enabling complex multi-step workflows to compose smaller sagas
- Make topological sort deterministic by sorting actions alphabetically at each dependency level, and expose `ExecutionOrder()` for testing
- Add `ExecutionOutputs()` to read completed saga results without a capture struct

## Test plan

- [x] Unit tests for nested saga execution (single-level and multi-level nesting)
- [x] Tests for deterministic execution ordering
- [x] Tests for `ExecutionOutputs()` retrieval

### DIFF
--- a/api/saga/saga_v1alpha/schema.gen.go
+++ b/api/saga/saga_v1alpha/schema.gen.go
@@ -12,6 +12,7 @@ const (
 	SagaExecutedActionsId   = entity.Id("dev.miren.saga/saga.executed_actions")
 	SagaExecutionOrderId    = entity.Id("dev.miren.saga/saga.execution_order")
 	SagaInitialInputsId     = entity.Id("dev.miren.saga/saga.initial_inputs")
+	SagaParentExecutionIdId = entity.Id("dev.miren.saga/saga.parent_execution_id")
 	SagaStatusId            = entity.Id("dev.miren.saga/saga.status")
 	SagaStatusPendingId     = entity.Id("dev.miren.saga/status.pending")
 	SagaStatusRunningId     = entity.Id("dev.miren.saga/status.running")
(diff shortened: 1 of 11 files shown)


phinze (commented):
whoops sorry premature post

phinze (approved):
Paul & Claude here from a review session where we read through the nested saga implementation and contributed a deterministic ordering test (3f6315d). Here's where we landed:

Overall this looks great — the nesting API is clean and the builder integration is natural. We have a few questions about recovery semantics for nested sagas that are worth considering as follow-ups.

We agree with CodeRabbit's note on `createChildExecution` — distinguishing real storage errors from "not found" would be good to tighten up.

The bigger questions are around what happens during `Recover()` with nested executions:

**1. Recovery could double-execute child sagas.** `ListIncomplete()` returns both parent and child executions. Recovery resumes the parent (whose action calls `RunNested` again) *and* independently resumes the child. There's no guard against concurrent execution of the same saga. One option: skip executions with a `ParentExecutionID` during recovery, since the parent's re-execution will drive the child.

**2. Without `WithNestedID`, child IDs are non-deterministic.** If a parent action is re-executed during recovery, `generateID()` produces a new child ID, so the idempotency check in `createChildExecution` is bypassed. The old child becomes an orphan. Deriving a deterministic child ID by default (e.g., from parent execution ID + saga name) would make retries naturally idempotent.

**3. `ParentExecutionID` is write-only.** It's stored but never queried in any code path. If we address (1), we'd need to filter on it during recovery. Worth deciding if this should be an indexed relationship or stay as debug metadata.

None of these are blocking — nesting works correctly for the normal path and the tests cover the key scenarios well. Just worth thinking through before we lean on nesting for production workflows.

evanphx:
@phinze all good notes! incorporated them.

PR #601, opened 2026-02-12
OTel Phase 2: inbound trace propagation, containerd instrumentation, observability docs

phinze:
Phase 1 (#595) got OTLP export and httpingress spans flowing. This is the follow-up with three small pieces:

**Inbound propagation** — httpingress was ignoring `traceparent` headers from callers and always starting root spans. Now it extracts inbound trace context before creating the span, so external callers (or a user's instrumented frontend) get one connected trace through Miren. Same pattern we already use in the RPC server.

**Containerd gRPC instrumentation** — Added `otelgrpc` stats handler to the containerd client. Container operations like image pulls and task creation now show up as child spans, so you can see the full request→lease→containerd path in a single trace.

**Observability docs** — User-facing docs page explaining what Miren traces, how propagation works both directions, and how to connect your app's traces with auto-instrumentation examples for Python and Node.

### DIFF
--- a/docs/docs/observability.md
+++ b/docs/docs/observability.md
@@ -27,9 +27,11 @@ Miren participates in [W3C Trace Context](https://www.w3.org/TR/trace-context/) 
 
 ## Connecting Your App's Traces
 
-To participate in Miren's traces, configure your app with an OpenTelemetry SDK and point it at an OTLP-compatible backend. The SDK will automatically read the `traceparent` header from incoming requests and create child spans.
+Miren's tracing and your app's tracing are configured independently — Miren handles its own trace export, and your app handles its own. The `traceparent` header is what connects them: when both sides send traces to the same backend, they show up as one unified trace because they share the same trace ID.
 
-Set these environment variables on your app:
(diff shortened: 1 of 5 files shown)


docs/docs/observability.md
```diff
@@ -0,0 +1,126 @@
...
+[[env]]
+key = "OTEL_EXPORTER_OTLP_HEADERS"
+value = "Authorization=Bearer your-api-key"
+
+[[env]]
+key = "OTEL_SERVICE_NAME"
+value = "my-app"
+```
+
+This works with any OTel-compatible backend: Grafana Tempo, Honeycomb, Datadog, Jaeger, and others.
```
evanphx: Does this mean that the internal traces pick up the app's config? or that we're telling the user that if they configure server traces AND app traces, they'll get them all in one place?

evanphx (approved):
Just a question about how the app's traces interact with the traces that the server generates on the users otlp service.

docs/docs/observability.md
phinze: The second one — Miren's tracing and the app's tracing are totally independent configs. The `traceparent` header is just the thread that connects them. I updated the text to make this clearer: the app needs its own OTel SDK pointed at its own backend, and as long as traces with the same trace ID end up in the same place, they show up as one unified trace.

PR #602, opened 2026-02-13
Add OTel tracing to the deploy/build path

phinze:
Building on the OTel infrastructure work, this lights up tracing for
the deploy path. With a collector running, `miren deploy` produces a
nice end-to-end trace showing how time breaks down across CLI, RPC,
server-side build phases, and BuildKit internals.

The CLI acts as the trace root and sets up tracing when
`OTEL_EXPORTER_OTLP_ENDPOINT` is set. The RPC framework propagates
context to the server automatically from there. On the server side,
`BuildFromTar` gets spans for each phase: tar receive, setup/stack
detection, buildkit build, artifact lookup, version creation, and
activation.

BuildKit gets two layers of tracing — the client gets the global
TracerProvider for gRPC spans, and the daemon container gets OTEL env
vars forwarded so it can export its own internal per-step spans
(dockerfile parse, layer pulls, RUN steps, etc).

The launcher controller gets standalone `launcher.reconcile` spans with
app/version attributes. These aren't connected to the deploy trace yet
(would need a data model change to thread context through), but the
attributes make correlation easy enough for now.

Everything is a no-op when `OTEL_EXPORTER_OTLP_ENDPOINT` isn't set.

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -14,6 +14,11 @@ import (
 	"github.com/charmbracelet/lipgloss"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/progress/progresswriter"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/term"
 
 	"miren.dev/runtime/api/app/app_v1alpha"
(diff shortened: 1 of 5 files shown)


cli/commands/deploy.go
```diff
@@ -181,6 +189,32 @@ func Deploy(ctx *Context, opts struct {
 		return nil
 	}
 
+	// Set up OTel tracing if an OTLP endpoint is configured
+	if os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") != "" {
```
evanphx, line 193: I wonder if we should let the server expose it's version of this value back to the CLI rather than the CLI having to have it's own version. Fine for now though!

evanphx (approved):

cli/commands/deploy.go
phinze: Gooood point let me think about whether there's a way to sneak that in

phinze: Filed MIR-707 to explore a proxy exporter approach — CLI would ship its spans through the existing RPC connection to the server, which re-exports them to the collector. That way the CLI gets tracing without needing its own collector config. Keeping the CLI span code as-is for now since it'll serve as the scaffold for that.

evanphx: ah! yeah, that's definitely a new task. sounds good!

PR #603, opened 2026-02-13
Add addon system with PostgreSQL provider

evanphx:
## Summary

- Adds the addon system (RFD-59) with provider interface, registry, framework, and controller for provisioning/deprovisioning addons
- Implements PostgreSQL addon provider with shared and dedicated variants using the saga framework
- Integrates addon support into app.toml config, CLI (`m addon list/add/remove/info/providers`), RPC server, and server startup
- Preserves addon env vars across deploys with source="addon" merge strategy
- Fixes addon controller to use ConfigVersion entities instead of inline Config (compatible with recent ConfigVersion refactor)
- Fixes coordinator node registration in test server, nil pointer in RouteWatcher on deletes, and versionless sandbox pool handling

## Test plan

- [ ] Integration tests pass for shared PostgreSQL provisioning (`pkg/addon/postgresql/shared_integration_test.go`)
- [ ] Unit tests pass for addon registry, controller, provider, and build server merge logic
- [ ] Addon env vars appear correctly in sandbox environment after provisioning
- [ ] Addon env vars survive app redeploys (source="addon" persisted in ConfigVersion)
- [ ] `m addon add/remove/list/info/providers` CLI commands work end-to-end


pkg/addon/provider.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// App identifies the application an addon is being attached to.
+type App struct {
+	ID   entity.Id
+	Name string
+}
+
+// Variant describes the variant selected for provisioning.
+type Variant struct {
```
phinze, line 30: Yes you hit on the perfect name for these! Much better than "plan" or "flavor".

servers/app/addons.go
```diff
@@ -0,0 +1,203 @@
...
+			)
+			return nil
+		}
+	}
+
+	return fmt.Errorf("addon %q is not attached to app %q", addonName, appName)
+}
+
+// addonNameFromRef extracts the addon name from an entity ref like "addon/miren-postgresql".
+func addonNameFromRef(ref entity.Id) string {
```
phinze: Nit: this function is duplicated in `controllers/addon/controller.go`. Since both packages need it, consider extracting it to `pkg/addon` (e.g., `addon.NameFromRef`) or the entity package. Small thing but easy to let diverge.

pkg/addon/postgresql/shared.go
```diff
@@ -0,0 +1,827 @@
...
+
+func IncrementAssociationCount(ctx context.Context, in IncrementAssociationCountIn) (IncrementAssociationCountOut, error) {
+	fw := saga.Get[*addon.ProviderFramework](ctx)
+
+	var server addon_v1alpha.PostgresServer
+	if err := fw.EC.GetById(ctx, in.ServerID, &server); err != nil {
+		return IncrementAssociationCountOut{}, fmt.Errorf("getting server for count increment: %w", err)
+	}
+
+	server.AssociationCount++
```
phinze: The `IncrementAssociationCount` and `DecrementAssociationCount` actions do a read-modify-write on `PostgresqlSharedData` via `EC.Update`, which calls `Put` with no revision check. If two apps provision against the same shared server concurrently, both read count=N, both write count=N+1, and one increment is lost.

`EC.Patch` supports revision-based OCC — switching to a patch with the revision from the read would let the entity store reject stale writes. Worth filing as a follow-up; not blocking for an initial implementation where concurrent shared provisions are unlikely, but it's a correctness gap for production.

api/addon/addon.go
```diff
@@ -0,0 +1,4 @@
+// Package addon provides entity definitions for the addon system.
+package addon
+
+//go:generate go run ../../pkg/entity/cmd/schemagen -input schema.yml -output addon_v1alpha/schema.go -pkg addon_v1alpha
```
phinze: The schemagen output target is `addon_v1alpha/schema.go`, but the codebase convention is to use the `.gen.go` suffix for generated files (e.g., `core_v1alpha/schema.gen.go`, `app_v1alpha/schema.gen.go`). Should be `addon_v1alpha/schema.gen.go` to match.

examples/bun-postgres/node_modules/pg/lib/client.js
```diff
@@ -0,0 +1,733 @@
+'use strict'
```
phinze: These `node_modules/` shouldn't be checked in — the deploy will run `bun install`. This adds ~100 vendored files to the repo. Should be removed and added to `.gitignore` for this example.

phinze (commented):
Claude and I sat down with this one and read through the full addon system end-to-end — the provider abstraction, saga-based provisioning, controller lifecycle, env var merge semantics, and the integration points with the deployment launcher and build server.

Overall this is a really well-structured addition. The saga approach for provisioning gives us durable multi-step operations with proper compensation, and the env var source-tracking (`manual`/`config`/`addon`) ensures addon vars survive deploys without clobbering user config. The deployment launcher integration (deferring pool creation until addons are ready) is a nice detail that prevents the race where sandboxes start before DATABASE_URL exists.

The bug fixes bundled in (RouteWatcher nil pointer on deletes, versionless sandbox pool handling, scheduler debug logging) are solid and well-motivated.

A few things we want to flag:

**Consider gating behind a Miren Labs flag** — The addon system runs unconditionally on coordinator startup (entity schema writes, controllers, RPC). A labs flag (`addons`, default false) would give a clean kill switch for rollout/rollback, similar to how `AdminAPI` is gated.

**Shared server association count has a TOCTOU race** — `IncrementAssociationCount` / `DecrementAssociationCount` do read-modify-write via `EC.Update` (which calls `Put` with no revision), so concurrent provisions on the same shared server could lose an increment. Worth switching to `Patch` with a revision, or adding OCC to `Update`.

**CodeRabbit overlap** — We agree with CodeRabbit's flags on the `.gen.go` suffix for the generated schema file, and the controller deleting the association even when `removeEnvVars` fails (though we understand the intent to not leave associations stuck). The `buildDatabaseURL` credential encoding flag is a false positive — `url.UserPassword` already handles that.

Details in inline comments below.

phinze (approved):

examples/bun-postgres/node_modules/pg/lib/client.js
phinze: Also let's move this app to `testdata` where the other example apps lie rather than establishing a new top level dir. If we wanna do `examples` we can talk about that as a separate move

servers/app/addons.go
evanphx: agreed!

pkg/addon/postgresql/shared.go
evanphx: ooooh yeah.

api/addon/addon.go
evanphx: fully, adding!

examples/bun-postgres/node_modules/pg/lib/client.js
evanphx: oh that was a big oopsie! I'll rebase that directory out.

PR #604, opened 2026-02-13
Align snapshot keys with container IDs

evanphx:
## Summary
- Snapshot keys for pause containers now use `pauseContainerId()` instead of raw `sb.ID.String()`, producing `sandbox.<uuid>_pause` to match the container ID
- Snapshot keys for subcontainers now use `containerPrefix()` instead of raw `sb.ID`, producing `sandbox.<uuid>-<name>` to match the container ID
- No migration needed — existing snapshot cleanup reads keys from `container.Info()`, not from code

## Test plan
- [ ] `make test` passes (full test suite with containerd)


phinze (approved):
lgtm

PR #605, opened 2026-02-13
Add automatic reindexing system for schema changes

evanphx:
## Summary

- Automatically detects when indexed attributes are added/removed from schemas and rebuilds missing index entries at startup, eliminating the need to manually run `m debug reindex`
- Extracts reindex logic from the entityserver RPC into a shared `EtcdStore.Reindex()` method, removing ~200 lines of duplicated code
- Computes a deterministic blake2b-256 hash of all indexed attribute IDs from the schema registry, stored in etcd at `{prefix}/meta/index-hash` to detect changes across restarts

## Test plan

- [x] `hack/it ./pkg/entity/schema/...` — schema hash tests (2/2 pass)
- [x] `hack/it ./pkg/entity/...` — entity + reindex tests (234/234 pass)
- [x] `hack/it ./servers/entityserver/...` — RPC reindex still works (19/19 pass)
- [x] `make lint` — 0 issues
- [ ] Manual: `make dev`, verify server starts, check logs for reindex hash comparison

### DIFF
--- a/pkg/entity/reindex.go
+++ b/pkg/entity/reindex.go
@@ -1,0 +1,193 @@
+package entity
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/mr-tron/base58"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"miren.dev/runtime/pkg/cond"
+)
+
+// ReindexStats holds statistics about a reindex operation.
+type ReindexStats struct {
+	EntitiesProcessed        int64
+	IndexesRebuilt           int64
+	CollectionEntriesScanned int64
+	StaleEntriesFound        int64
+	StaleEntriesRemoved      int64
+}
+
+// ReindexOptions controls the behavior of a reindex operation.
+type ReindexOptions struct {
+	DryRun       bool
+	CleanupStale bool
+}
+
+// Reindex rebuilds all index (collection) entries for every entity in the store.
+// If opts.CleanupStale is true, it also scans for and removes stale collection entries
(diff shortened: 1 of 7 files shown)


pkg/entity/reindex.go
```diff
@@ -0,0 +1,193 @@
...
+			indexedAttrs[attr.ID] = append(indexedAttrs[attr.ID], attr)
+		}
+	}
+	return indexedAttrs
+}
+
+// addToCollectionDirect writes a single collection entry for the given entity and collection key.
+func (s *EtcdStore) addToCollectionDirect(ctx context.Context, ent *Entity, collection string) error {
+	key := base58.Encode([]byte(ent.Id()))
+	colKey := strings.NewReplacer("/", "_", ":", "_").Replace(collection)
```
phinze: Optional nit: this allocates a new `strings.NewReplacer` on every call. There's already a package-level `var tr` in `store.go` that does the same `"/" → "_", ":" → "_"` replacement (used by `addToCollectionOp`). Could reuse that here to avoid per-entry allocations during a large reindex.

phinze (approved):
We sat down and traced through the full startup flow and the reindex extraction together. This is a clean design — the hash-based change detection is simple and deterministic, the reindex logic consolidation removes real duplication, and the placement in the startup sequence (after schema apply + migration, before entity server creation) is correct.

One thing worth noting: the automatic reindex runs synchronously during boot, so the first startup after this lands (or any startup where indexes change) will block until reindex completes (bounded by a 5-minute timeout). Seems like the right tradeoff for correctness.

One optional nit inline, otherwise looks great.

PR #606, opened 2026-02-13
Fix lsvd_volume entities using wrong entity ID format

evanphx:
## Summary
- Fixed volume entity ID creation in `disk_controller.go` which used `KindLsvdVolume.String()` as the ID prefix, producing IDs like `dev.miren.storage/kind.lsvd_volume/lsvd-vol-xxx` instead of the correct `lsvd_volume/lsvd-vol-xxx`
- Updated test volume entity IDs in `reconcile_system_test.go` to use the correct format
- Cleaned up stale old-format fallbacks in the integration test harness

## Test plan
- [x] `hack/it ./controllers/disk/...` — 44 tests pass
- [x] `hack/it ./controllers/integration/...` — 29 tests pass
- [x] `hack/it ./components/lsvd/server/...` — 114 tests pass

Closes MIR-706

### DIFF
--- a/controllers/disk/disk_controller.go
+++ b/controllers/disk/disk_controller.go
@@ -222,7 +222,7 @@ func (d *DiskController) handleProvisioning(ctx context.Context, disk *storage_v
 	// Build entity with id and encoded attributes
 	volumeId := idgen.GenNS("lsvd-vol")
 	createAttrs := entity.New(
-		entity.DBId, entity.Id(storage_v1alpha.KindLsvdVolume.String()+"/"+volumeId),
+		entity.DBId, entity.Id("lsvd_volume/"+volumeId),
 		lsvdVolume.Encode,
 	).Attrs()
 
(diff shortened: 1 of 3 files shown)


phinze (approved):
Claude and I sat down and traced through the entity ID lifecycle end-to-end to check for migration and backcompat concerns — the main thing we wanted to verify.

Looks good. The fix is clean, and we're satisfied there's no migration risk:

- `getLsvdVolumeForDisk()` queries by `disk_id` index, not by entity ID, so existing old-format entities in etcd will still be found — no risk of duplicate volume creation.
- `ReconcileWithEntities` compares entity IDs from the store against local `lsvd-state.json`, and both sides will continue to agree for existing volumes (old IDs stay old).
- Path derivation uses `strings.LastIndex("/")` so it handles either format identically.
- `disk.LsvdVolumeId` stores the raw volume ID (`"lsvd-vol-xxx"`), not the entity ID, so it's format-agnostic.

One curiosity: the harness also dropped the `"dev.miren.storage/kind.lsvd_mount/"` fallback — did lsvd_mount have a similar issue at some point, or was that added preemptively?

evanphx:
The long lsvd_mount path was never actually used turns out!

PR #609, opened 2026-02-13
Add CLI tracing proxy exporter

phinze:
## Summary

- CLI deploy spans now ship through the server's existing RPC connection automatically — no `OTEL_EXPORTER_OTLP_ENDPOINT` needed on the client side
- New `Telemetry.reportSpans(bytes)` RPC endpoint receives serialized OTLP protobuf and forwards it to the server's configured collector
- Server silently drops spans when no collector is configured; RPC errors are swallowed so tracing never blocks deploys

## Design

Three layers, all simple:

| Layer | File(s) | What it does |
|-------|---------|-------------|
| Proxy client | `pkg/otelproxy/proxy.go` | Implements `otlptrace.Client`, marshals spans to protobuf, sends via RPC |
| RPC endpoint | `api/telemetry/` + `servers/telemetry/server.go` | Receives bytes, HTTP POSTs to `{OTEL_EXPORTER_OTLP_ENDPOINT}/v1/traces` |
| CLI integration | `cli/commands/deploy.go` | Connects to telemetry service, sets up proxy tracing. Falls back to no-op if server unreachable |

The proxy client lives in `pkg/otelproxy/` (not `pkg/rpc/`) to avoid an import cycle through the generated telemetry RPC code.

## Test plan

- [x] `go generate ./api/telemetry/...` succeeds
- [x] `make bin/miren` builds clean
- [x] `make test` passes (3287 tests, 20 pre-existing skips)
- [ ] Manual: start dev server with `OTEL_EXPORTER_OTLP_ENDPOINT` pointed at a collector, run `m deploy` without any env vars on the CLI side, verify deploy+build spans appear in the collector

### DIFF
--- a/api/telemetry/rpc.yml
+++ b/api/telemetry/rpc.yml
@@ -1,0 +1,10 @@
+apiVersion: miren.dev/rpc/v1
+kind: IDL
+
+interfaces:
+  - name: Telemetry
+    methods:
+      - name: reportSpans
+        parameters:
+          - name: span_data
+            type: bytes
--- a/api/telemetry/telemetry.go
+++ b/api/telemetry/telemetry.go
@@ -1,0 +1,3 @@
+package telemetry
+
+//go:generate go run ../../pkg/rpc/cmd/rpcgen -pkg telemetry_v1alpha -input rpc.yml -output telemetry_v1alpha/rpc.gen.go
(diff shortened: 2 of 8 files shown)


evanphx (approved):
Image

PR #613, opened 2026-02-16
Release disk leases in periodic sandbox cleanup

phinze:
We discovered in one of our clusters that a handful of apps had sandboxes stuck pending for weeks. The root cause was that `Periodic()` — the background job that cleans up old dead sandbox entities — was calling `EAC.Delete()` directly without first releasing disk leases. The normal teardown path releases leases before cleanup, but the periodic path never got the same treatment.

This meant disk leases stayed in `status.bound` pointing at sandbox entities that no longer existed. When new sandboxes tried to start and needed those volumes, they'd see the active lease and stay pending forever, causing timeouts all the way up through the activator and HTTP ingress.

The fix is straightforward: call `releaseDiskLeases()` before `EAC.Delete()` in `Periodic()`, matching what the normal teardown already does. Added tests that exercise the periodic cleanup path with disk leases to make sure they get released, that sandboxes without leases still clean up fine, and that running sandboxes aren't touched.

Closes MIR-717

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -2480,6 +2480,11 @@ func (c *SandboxController) Periodic(ctx context.Context, timeHorizon time.Durat
 					"updated_at", updatedAt.Format(time.RFC3339),
 					"age", now.Sub(updatedAt).String())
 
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Image

Whooopsie!

PR #616, opened 2026-02-17
Fix activator pool cache permanent lockout at MaxPoolSize

phinze:
The activator keeps an in-memory cache of each pool's `DesiredInstances` to avoid hitting the entity store on every request. Problem is, that cache only ever went up — once it reached `MaxPoolSize` (20), the activator would bail immediately without consulting the store. Since no store interaction meant no OCC conflict, the cache never got a chance to refresh. If the pool manager reset `DesiredInstances` to 0 externally (say, after a scale-down), the activator would never find out. Permanent deadlock.

Fixed with two complementary changes, defense-in-depth style. First, when the cache says we're at max, we now re-read from the entity store before giving up — if the fresh state is lower, we update the cache and continue the increment. This directly breaks the deadlock on the hot path. Second, `watchPools` now handles Update events in addition to Delete events, so `DesiredInstances` changes flow into the cache proactively. The re-read is a safety net for any remaining staleness window.

Closes MIR-721

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -533,12 +533,55 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
 			for attempt := 0; attempt < maxRetries; attempt++ {
 				a.mu.Lock()
 				if state.pool.DesiredInstances >= MaxPoolSize {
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Image

PR #617, opened 2026-02-17
Respect Dockerfile WORKDIR when running app containers

phinze:
When you set `WORKDIR /src` in a Dockerfile, you'd expect your app to start in `/src`. Instead, we were always landing in `/app` — or `/` in some cases — because the working directory got overridden at every layer on its way down to the container runtime.

The problem was a chain of well-intentioned defaults. The build server would stuff `/app` into `StartDirectory` when the build result didn't have a working dir. The deployment launcher and exec proxy would each independently fall back to `/app` if it was still empty. And at the bottom of the stack, the sandbox controller always called `oci.WithProcessCwd()` — even with a default of `/` — which clobbered the correct WORKDIR that `oci.WithImageConfig()` had already set from the image metadata.

The fix is to let empty mean "use what the image says." We stop inserting `/app` as a default at the build, deployment, and exec layers, and the sandbox controller now only calls `WithProcessCwd` when an explicit directory is configured. This lets `oci.WithImageConfig()` do its thing, which reads WORKDIR from the image per the OCI spec and defaults to `/` when none is set.

Existing apps that already have a `StartDirectory` stored in their config entity are unaffected — their value flows through as before. Images without a WORKDIR still get `/` per the OCI spec, which is correct.

Closes MIR-715

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1940,11 +1940,6 @@ func (c *SandboxController) buildSubContainerSpec(
 		})
 	}
 
-	dir := co.Directory
-	if dir == "" {
-		dir = "/"
-	}
-
 	// Extract instance number from metadata labels and inject MIREN_INSTANCE_NUM
 	envVars := co.Env
 	var md core_v1alpha.Metadata
@@ -1960,7 +1955,6 @@ func (c *SandboxController) buildSubContainerSpec(
 		oci.WithDefaultUnixDevices,
 		oci.WithoutMounts("/sys"),
 		oci.WithMounts(mounts),
-		oci.WithProcessCwd(dir),
 		oci.WithLinuxNamespace(specs.LinuxNamespace{
 			Type: specs.NetworkNamespace,
 			Path: fmt.Sprintf("/proc/%d/ns/net", sbPid),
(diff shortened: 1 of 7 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -1976,6 +1970,10 @@ func (c *SandboxController) buildSubContainerSpec(
 		oci.WithEnv(envVars),
 	}
 
+	if co.Directory != "" {
+		specOpts = append(specOpts, oci.WithProcessCwd(co.Directory))
```
evanphx, line 1974: I think we still need an else branch for this, I don't think the sandbox will boot without some cwd set.

evanphx: Or maybe not? Maybe it uses the directory in the OCI spec for the container?

evanphx (approved):

controllers/sandbox/sandbox.go
phinze: Yep, exactly right — `oci.WithImageConfig(img)` (line 1955 in the new code) reads the image's WORKDIR and sets `Process.Cwd` on the OCI spec. When the image has no WORKDIR, it defaults to `/` per the OCI spec. So by the time we get here, there's always a valid Cwd set — we only override it if someone explicitly configured a `Directory`.

phinze: To be precise: the OCI runtime spec says `process.cwd` is REQUIRED but doesn't define a default, and the image spec says `WorkingDir` is optional. It's containerd's `WithImageConfigArgs` that bridges the gap — when the image has no WorkingDir, it fills in `"/"`: https://github.com/containerd/containerd/blob/main/pkg/oci/spec_opts.go#:~:text=cwd%20%3A%3D%20config.WorkingDir

phinze: ⚠️ Welp this was wrong, we wanted it to fill in `/app`, fixing now and cutting a bugfix release!

PR #621, opened 2026-02-18
Restore /app default for start directory

phinze:
The WORKDIR fix (f24f27d7) did the right thing for new deploys — detect the image's WORKDIR at build time and store it in the config. But it also removed the runtime `/app` fallback from the deployment launcher, exec proxy, and build server. That's a problem for every existing app version that was built before the fix: they have no `start_directory` in their config at all, so they end up booting with CWD=`/` instead of `/app` and immediately crash.

This restores the `/app` default in all three places. The build server defaults new builds to `/app` when no WORKDIR is detected, and the launcher and exec proxy fall back to `/app` when the stored config is empty. The sandbox controller's conditional `WithProcessCwd` is left as-is since it's the correct behavior.

The launcher and exec proxy defaults are backward-compat scaffolding — once we run a boot migration to backfill `start_directory` on existing configs, those can be cleaned up.

Fixes MIR-727


phinze:
Merging this right away to get this fix out there.

PR #626, opened 2026-02-18
Add --yes flag to release scripts

phinze:
Our release scripts have interactive `read -p` confirmation prompts, which is fine for humans but blocks non-interactive callers from running them directly. This adds `--yes`/`-y` to both `hack/release.sh` and `hack/release-tag.sh` to skip the confirmation prompt. All the actual safety checks (clean working directory, on main, up to date with origin, changelog validation) still run — we're only skipping the "are you sure?" prompt, not the guardrails.

### DIFF
--- a/hack/release-tag.sh
+++ b/hack/release-tag.sh
@@ -2,10 +2,11 @@
 set -euo pipefail
 
 # Push a release tag after the release PR has been merged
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Image

PR #631, opened 2026-02-25
Implement GitHub OIDC token auth for CI deploys

evanphx:
## Summary

- Add external OIDC token authentication (MIR-724) so CI systems like GitHub Actions can authenticate with short-lived OIDC identity tokens instead of long-lived API secrets
- New `oidc_binding` entity kind ties an OIDC issuer + subject pattern to an app, with optional claim conditions for fine-grained control (allowed events, refs, etc.)
- Composite authenticator/authorizer chains existing cloud auth with OIDC, giving OIDC callers a scoped "deploy" role
- `rpc.AllowApp()` enforces app-scoping across all app-operating handlers (deployment, builder, logs, app status, crud)
- CLI commands: `miren oidc add --github OWNER/REPO -a APP`, `miren oidc list`, `miren oidc remove`
- GitHub Actions auto-detection: when no identity is configured and `ACTIONS_ID_TOKEN_REQUEST_URL` is set, automatically requests and uses an OIDC token

## Test plan

- [x] Unit tests for OIDC token validator (discovery, JWKS, key rotation, expiry, wrong issuer/audience)
- [x] Unit tests for OIDCAuthenticator with in-memory entity server (binding matching, subject/claim filtering, audience resolution)
- [x] Unit tests for CompositeAuthenticator/Authorizer (primary fallback, cert bypass, OIDC role allow/deny, JWT delegation)
- [x] Unit tests for GitHub Actions detection and token request
- [x] Full test suite passes (3432 tests, 0 failures)
- [ ] Manual integration test: `miren oidc add --github org/repo -a app` in dev environment
- [ ] End-to-end test with GitHub Actions workflow


servers/logs/logs.go
```diff
@@ -114,6 +119,12 @@ func (s *Server) StreamLogs(ctx context.Context, state *app_v1alpha.LogsStreamLo
 	send := args.Logs()
 	target := args.Target()
 
+	if target.HasApp() && target.App() != "" {
```
phinze, line 122: Nit: `AllowApp` is gated on `target.HasApp()`, so a sandbox-only target skips the check. Probably fine since it's read-only log access from an authenticated caller, but could be worth tightening for defense in depth down the road.

phinze (approved):
Claude and I reviewed this together, focusing on the end-to-end auth flow, the authorization model, and how the CLI UX fits alongside the existing `route oidc` commands from #587.

Really excited about this one — the zero-config GitHub Actions experience (just `permissions: id-token: write` and `miren deploy`) is going to make CI integration so much cleaner. No more long-lived secrets to rotate!

The design is great. The layered approach — composite authenticator → role-based authorizer → per-handler `AllowApp()` scoping — is easy to follow, and the test coverage is solid. The `oidcDeployRole` whitelist being separate from the app-scoping enforcement is a nice separation of concerns.

CodeRabbit flagged a few real things (the `r.TLS` nil-pointer, URL construction in `github.go`, JWKS retry scope, missing `WithValidMethods`) — we traced through each of those and agree they're worth fixing.

Two additional notes from our side:

- **Sandbox log targets**: `AllowApp` in the log streaming handlers only runs when `target.HasApp()`. In theory an OIDC caller could pass a sandbox ID directly and skip the app-scoping check. Probably low-impact since it's read-only log access from an already-authenticated caller, but might be worth tightening for defense in depth at some point.

- **CLI UX**: The top-level `miren oidc` (CI bindings) and `miren route oidc` (user-facing auth, behind labs flag) are differentiated well enough by their command tree position today. Worth keeping an eye on wording and help text as both features mature to make sure they stay clearly separated for users.

servers/logs/logs.go
evanphx: Fixed!

PR #634, opened 2026-02-26
Replace http.TimeoutHandler with transport-level timeouts

phinze:
`http.TimeoutHandler` buffers the entire response in memory and only sends it when the handler returns. That's fine for normal request/response, but it completely breaks SSE, chunked downloads, long-polling, and anything that relies on incremental flushing. The previous iteration replaced it with `context.WithTimeout`, which fixed the buffering but still needed a special `isUpgradeRequest` check to let WebSocket connections through.

This switches to transport-level timeouts — the same approach used by nginx (`proxy_read_timeout`) and Traefik (`responseHeaderTimeout`). A custom `http.Transport` with two complementary mechanisms handles all the cases: `ResponseHeaderTimeout` catches backends that never start responding (crashed, deadlocked, stuck boot), and an idle read timeout via `SetReadDeadline` on a wrapping `idleTimeoutConn` catches backends that stop sending mid-stream. Both use the configured `RequestTimeout` (default 60s). Active streams stay alive because each chunk of data resets the idle timer — apps just need standard keepalive (SSE `:` comments, WebSocket ping/pong), which they'd already need behind any production reverse proxy.

The nice thing about this approach is that it eliminates all protocol-specific branching. No more `isUpgradeRequest` check, no more upgrade-vs-normal code paths — every request flows through the same transport.

Closes MIR-740

### DIFF
--- a/servers/httpingress/httpingress.go
+++ b/servers/httpingress/httpingress.go
@@ -38,6 +38,21 @@ import (
 	"miren.dev/runtime/pkg/rpc"
 )
 
+// idleTimeoutConn wraps a net.Conn and sets a read deadline before each
+// Read call. If no data arrives within the idle timeout, the read fails
+// with a timeout error. Each successful read resets the deadline, so
+// active streams (SSE, WebSocket, chunked) are unaffected as long as
+// data keeps flowing.
+type idleTimeoutConn struct {
+	net.Conn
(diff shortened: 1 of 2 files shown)


evanphx (approved):
I like it! A good solution for the problem, using SetReadDeadline

PR #636, opened 2026-02-27
Add CI/CD deployment with OIDC documentation

evanphx:
## Summary

- Create `docs/docs/oidc-deploy.md` — full documentation for deploying from CI/CD pipelines using OIDC token authentication
- Covers GitHub Actions quick start with `MIREN_CLUSTER` env var and `mirendev/actions/deploy` action
- Documents restricting access (subject patterns, allowed events/refs, claim conditions)
- Includes generic OIDC provider setup with example shell script for non-GitHub CI platforms
- Documents `miren oidc add/list/remove` CLI commands and `miren cluster export-address`
- Adds OIDC deploy to sidebar, CLI reference, and disambiguation callouts on route-oidc page

## Test plan

- [ ] Verify sidebar renders correctly with new `oidc-deploy` entry
- [ ] Verify cross-links between `oidc-deploy`, `route-oidc`, and `cli-reference` work
- [ ] Review GitHub Actions workflow YAML is valid and complete

### DIFF
--- a/docs/docs/ci-deploy.md
+++ b/docs/docs/ci-deploy.md
@@ -1,0 +1,318 @@
+---
+sidebar_position: 9
+---
+
+# CI/CD Deployment
+
+Deploy to Miren from CI/CD pipelines without storing long-lived secrets. Instead of managing API keys or certificates, your CI system presents a short-lived OIDC identity token that Miren validates directly with the identity provider.
+
+GitHub Actions works out of the box. Other OIDC-capable CI systems (GitLab CI, CircleCI, etc.) are supported with manual configuration.
+
+:::info Looking for Route OIDC?
+This page covers OIDC for **CI/CD deployment authentication** — letting pipelines deploy without stored secrets. For protecting your **application's HTTP routes** with single sign-on, see [Route OIDC Authentication](/route-oidc).
+:::
+
+## How It Works
+
+1. **Create an OIDC binding** on your Miren cluster that links an app to an identity provider and subject pattern.
+2. **Your CI job runs** and the Miren CLI auto-detects the CI environment (e.g., GitHub Actions sets `ACTIONS_ID_TOKEN_REQUEST_URL`).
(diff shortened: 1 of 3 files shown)


docs/docs/ci-deploy.md
```diff
@@ -0,0 +1,318 @@
...
+def456     generic    https://gitlab.com                            project_path:acme/web-*
+```
+
+### Remove a Binding
+
+```bash
+miren oidc remove abc123
+```
+
+## CLI Reference
```
phinze, line 228: This repeats the other CLI reference doc - I'd suggest either removing or cross referencing.

docs/docs/oidc-deploy.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+The value includes the cluster address and a TLS certificate fingerprint for verification. Store this as a GitHub Actions secret named `MIREN_CLUSTER`.
+
+### Step 2: Add an OIDC Binding
+
+Create a binding that allows your GitHub repository to deploy:
+
+```bash
+miren oidc add myapp --github acme/web-app
```
phinze: Rabbit is right below, there's no positional arg I think? Need `-a`

phinze (approved):
Looking good! A couple of nits inline.

💭 As a secondary thing, I'm getting a little 😵‍💫 with how often I'm reading OIDC and vaguely thinking about whether there's a more user-centric way for us to talk about the two OIDC-related features that downplays the protocol (e.g. should we just call them App Auth / CI Deploy?). But that's a separate conversation we can have down the line.

evanphx:
@phinze My thoughts exactly on the naming. As I was working on these docs, I was like "we're calling OIDC the feature, it should be a way you use the feature, not the feature itself."

docs/docs/oidc-deploy.md
evanphx: `miren auth ci --github acme/web-app`

PR #639, opened 2026-03-02
Implement RFD 64: Universal and Accelerator mode disk system

evanphx:
## Summary

- **Replace the LSVD subsystem** with a new Universal mode disk system using loop devices and a simpler architecture
- **Add accelerator mode** with LBD log-structured writes and cloud segment upload/download/replay
- **Add disk backup/restore** via the new `pkg/snapshot` package
- **Add automatic LSVD → Universal migration** path for existing volumes
- **Add cloud integration tests** exercising the full segment round-trip against a mock cloud server

### Phase 1: Universal mode (`7029049`)
Replaces the old LSVD/NBD-based disk system with loop-device mounts backed by raw or qcow2 disk images. New `components/diskio` package with volume controller, mount controller, and OS-level ops abstraction.

### Directory mode removal (`767423c`)
Removes the legacy directory-based volume mode that is no longer needed.

### LSVD → Universal migration (`acd6d09`)
Adds automatic migration from LSVD volumes to Universal mode, including integration tests that verify the upgrade path.

### Disk backup/restore (`ab9839e`)
New `pkg/snapshot` package supporting backup and restore of disk images with streaming tar format.

### Cloud accelerator mode (`b76eee3`)
Adds cloud lease management, LBD segment upload via presigned URLs, segment download/replay, and a LogWatcher that uploads completed log segments.

### Cloud integration tests (`e4a2361`)
Mock cloud HTTP server exercising `CloudSegmentUploader`, `cloudDiskClient`, `LogWatcher`, and `replayMissingSegments` end-to-end with 5 test scenarios.

## Test plan

- [x] All `components/diskio/` unit tests pass
- [x] Cloud integration tests pass (`TestCloudIntegration*`)
- [x] `controllers/disk/` integration tests pass
- [x] `controllers/integration/` migration tests pass
- [x] `pkg/snapshot/` tests pass
- [x] Linter passes (`make lint` — 0 issues)


components/diskio/disk_mount_controller.go
```diff
@@ -0,0 +1,722 @@
...
+		if mountState == nil {
+			c.log.Warn("entity says DM_MOUNTED but no local state found", "entity_id", entityId)
+			return nil
+		}
+		if mountState.MountPath != "" && !c.ops.IsMounted(mountState.MountPath) {
+			c.log.Warn("entity reports mounted but mount not found on system, recovering",
+				"entity_id", entityId,
+				"mount_path", mountState.MountPath,
+			)
+			return c.attachAndMount(ctx, mount)
```
phinze, line 163: For accelerator mode, this recovery path calls `attachAndMount` which allocates a new device without detaching the old one from `mountState.DevicePath`. The volume controller's `ensureVolumeMount` now handles this for universal mode (detaches stale device before re-attaching) — could we add the same pattern here for the accelerator path?

components/diskio/disk_mount_controller.go
```diff
@@ -0,0 +1,722 @@
...
+	formatted, err := c.ops.IsFormatted(ctx, mountState.DevicePath, filesystem)
+	if err != nil {
+		c.setMountError(ctx, mount.ID, fmt.Sprintf("failed to check if formatted: %v", err))
+		return fmt.Errorf("failed to check if formatted: %w", err)
+	}
+
+	if !formatted {
+		c.log.Info("formatting device", "device", mountState.DevicePath, "filesystem", filesystem)
+
+		formatDeadline := time.Now().Add(10 * time.Minute)
```
phinze: This 10-minute format retry loop runs inside the reconcile callback with `workers: 1`. A single flaky device would block all other mount reconciliations. Same pattern exists in `disk_volume_controller.go:ensureVolumeMount`. Worth thinking about whether to bump workers or cap the retry shorter — but also possible this is fine given mounts are rare events. Curious what you think.

components/diskio/segment_replay.go
```diff
@@ -0,0 +1,255 @@
...
+			return "", nil
+		}
+		return "", err
+	}
+	return strings.TrimSpace(string(data)), nil
+}
+
+func writeLogHorizon(diskPath, label string) error {
+	path := filepath.Join(diskPath, logHorizonFile)
+	return os.WriteFile(path, []byte(label+"\n"), 0644)
```
phinze: `writeLogHorizon` uses `os.WriteFile` (truncate + write) — a crash mid-write leaves a corrupt/empty file and causes full re-replay. Recoverable since replays are idempotent, but `State.Save()` already uses temp-file-then-rename — could we reuse that pattern here?

phinze (commented):
Claude and I spent a couple days with this PR — reading through the full code paths and tracing the entity lifecycle end-to-end. We know you're still actively working on it, so this is more "approve the direction and add a few things to the list" than a final review. Happy to be told any of this is stale or already planned.

The architecture is really clean. The four-entity decomposition with watch controllers bridging the layers is a big step up from the LSVD server approach. Mounting universal volumes at creation time so leases become pure bookkeeping is a nice simplification, and consolidating to persistent state only (killing the in-memory mounts map) landed well.

A couple of things that aren't easy to comment inline:

**`updateLeaseDetails` is a no-op** (`disk_lease_controller.go:608-613`) — acquires the lock, reads from the map into `_`, returns. Stub or dead code?

**`acquireDiskLease` scans all leases by kind** (`controllers/sandbox/volume.go:252`) — uses `entity.Ref(entity.EntityKind, storage.KindDiskLease)` instead of the `DiskLeaseDiskIdId` index. This is on the sandbox boot path. Same pattern in `disk_resolver.go:FindDisk`. CodeRabbit flagged the resolver one too — we agree.

phinze (approved):

phinze:
This might be a good thing to run through a bit of a demo slash manual test session before merging.

components/diskio/disk_mount_controller.go
evanphx: Yup!

components/diskio/disk_mount_controller.go
evanphx: Yeah, I think I'm going to bump the workers off 1 too.

components/diskio/segment_replay.go
evanphx: very true!

evanphx:
@phinze Totally agree on the demo / manual, doing heavy pass at that today.

cli/commands/commands.go
```diff
@@ -810,6 +810,11 @@ Warning: These commands are intended for advanced users and developers. They may
 	d.Dispatch("debug entity patch", Infer("debug entity patch", "Patch an existing entity", EntityPatch))
 	d.Dispatch("debug entity ensure", Infer("debug entity ensure", "Ensure an entity exists", EntityEnsure))
 
+	// Disk commands
+	d.Dispatch("disk migrate", Infer("disk migrate", "Migrate LSVD volume to raw disk image", DiskMigrate))
```
evanphx: Remove this guy, it's automatic.

cli/commands/commands.go
```diff
@@ -827,17 +832,6 @@ Warning: These commands are intended for advanced users and developers. They may
...
-	// Debug LSVD commands
-	d.Dispatch("debug lsvd", Section("debug lsvd", "LSVD server debug commands", ""))
-	d.Dispatch("debug lsvd info", Infer("debug lsvd info", "Show combined LSVD server volumes, mounts, and metrics", DebugLsvdInfo))
-	d.Dispatch("debug lsvd volumes", Infer("debug lsvd volumes", "List volumes managed by LSVD server", DebugLsvdVolumes))
-	d.Dispatch("debug lsvd mounts", Infer("debug lsvd mounts", "List mounts managed by LSVD server", DebugLsvdMounts))
-	d.Dispatch("debug lsvd metrics", Infer("debug lsvd metrics", "Show LSVD reconciliation metrics", DebugLsvdMetrics))
-
-	// Debug outboard commands
-	d.Dispatch("debug outboard", Section("debug outboard", "Outboard process debug commands", ""))
-	d.Dispatch("debug outboard health", Infer("debug outboard health", "Check health of an outboard process", DebugOutboardHealth))
```
evanphx, line 839: Keep these in.

cli/commands/disk_restore.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// DiskRestore restores a disk from a compressed snapshot file.
+func DiskRestore(ctx *Context, opts struct {
+	ConfigCentric
+	Snapshot string `short:"s" long:"snapshot" description:"Path to snapshot file" required:"true"`
+	Name     string `short:"n" long:"name" description:"Disk name to restore to (default: original name from snapshot)"`
+	DataPath string `long:"data-path" description:"Path to miren data directory" default:"/var/lib/miren"`
+	Force    bool   `short:"f" long:"force" description:"Overwrite existing disk image without confirmation"`
+}) (retErr error) {
```
evanphx, line 21: Enforce that this has to happen on the server, not remotely over RPC.

cli/commands/disk_backup.go
```diff
@@ -0,0 +1,94 @@
...
+	"miren.dev/runtime/pkg/snapshot"
+)
+
+// DiskBackup backs up a disk to a compressed snapshot file.
+func DiskBackup(ctx *Context, opts struct {
+	ConfigCentric
+	Name     string `short:"n" long:"name" description:"Disk name to backup" required:"true"`
+	Output   string `short:"o" long:"output" description:"Output snapshot path (default: DISK-YYYYMMDD-HHMMSS.miren.zst)"`
+	DataPath string `long:"data-path" description:"Path to miren data directory" default:"/var/lib/miren"`
+}) (retErr error) {
```
evanphx, line 18: Enforce that this is run from the server.

components/runner/runner.go
```diff
@@ -239,6 +241,61 @@ func (r *Runner) ContainerdNamespace() string {
 	return r.namespace
 }
 
+// pauseAllContainers pauses all running containerd tasks that are user
+// workloads (not infrastructure). Containers with IDs starting with "miren-"
+// are infrastructure (etcd, etc.) and must not be paused.
+// Returns the list of tasks that were successfully paused so they can be
+// resumed later. Containers that fail to pause are logged but skipped —
+// migration proceeds regardless since it must happen to complete the upgrade.
+func (r *Runner) pauseAllContainers(ctx context.Context, log *slog.Logger) []containerd.Task {
```
evanphx: This needs to be stop because we need to release any file descriptors.

controllers/disk/disk_controller.go
```diff
@@ -15,44 +15,42 @@ import (
...
-	if os.Getenv("MIREN_DISABLE_NBD") != "" {
-		return false
-	}
-
-	// Check if NBD module is loaded
-	if _, err := os.Stat("/sys/module/nbd"); err == nil {
-		return true
+// detectDiskMode determines which disk I/O mode to use.
+func detectDiskMode() storage_v1alpha.DiskMode {
+	if mode := os.Getenv("MIREN_DISK_MODE"); mode != "" {
```
evanphx: Move MIREN_DISK_MODE to server config package

pkg/snapshot/backup.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/klauspost/compress/zstd"
+)
+
+// Backup reads the disk image from src, computes its SHA-256 checksum,
+// and writes a complete snapshot (header + zstd-compressed data) to dst.
+// It returns the checksum of the uncompressed image data.
+func Backup(dst io.Writer, src io.ReadSeeker, name string, sizeBytes int64, filesystem string) (checksum string, err error) {
+	// First pass: compute SHA-256 checksum
+	hasher := sha256.New()
```
evanphx: Move this to be a Tee with the zstd so both compression AND hashing see the same data for later comparison.

PR #611, opened 2026-02-13
Rename network Service entity to Target

evanphx:
## Summary

- Renames the network "Service" entity to "Target" to eliminate naming confusion with app-level services (web, worker processes in AppVersion configs)
- Renames `controllers/service` package to `controllers/target`, updating all types (`ServiceController` → `TargetController`, etc.)
- Updates all consumers: runner, sandbox controller, ipalloc, CLI config (`ServicePrefixes` → `TargetPrefixes`)
- Adds `MigrateServiceToTarget()` migration to rename stored entities in etcd (attribute IDs, kind refs, endpoints references), using `Replace` for full entity replacement

## What's NOT changed

- nftables chain/set names (`service_ip4s`, `services` chain) — internal kernel identifiers
- `network.ServiceManager` — unrelated DNS/network management concept
- App-level `services` in AppVersion configs

## Test plan

- [x] Migration unit tests pass (basic rename, port component, already-migrated, empty store)
- [ ] `make lint` passes
- [ ] `make test` passes

PR #641, opened 2026-03-02
Support multiple ports and non-HTTP services per app

phinze:
Users have been asking to deploy non-HTTP services — the immediate motivation was wanting to run an IRC server on Miren with ports 6667 and 6697 exposed. Today the app config layer only supports a single port per service, and there's no way to specify `node_port` or `protocol`.

The fun part is that all the hard infrastructure work was already done. The sandbox container spec, firewall rules, and service controller all handle multi-port just fine. The bottleneck was the app config layer: one `port` integer per service, no array, no `node_port`, no `protocol`.

## Config layer (`appconfig`)

Adds a `ports[]` array to the config spec schema and wires it through the full pipeline — TOML parsing with validation in appconfig, mapping through the build server's `buildServicesConfig`, config version conversion, and the deployment launcher's `buildSandboxSpec`. The existing scalar `port`/`port_name`/`port_type` fields are preserved for backward compat; `ports[]` takes precedence when present, and mixing the two styles is a validation error. The `PORT` env var picks the first HTTP-typed port, or falls back to the first port in the array.

Also adds a deploy-time guardrail: since `node_port` is now a first-class config field, two apps could silently claim the same host port and only find out when nftables blew up at runtime. The new `validateNodePorts` check runs during `BuildFromTar` and catches both intra-app duplicates and cross-app conflicts with clear error messages.

Empty protocol is now normalized to `"tcp"` before duplicate `(port, protocol)` validation.

## Service entity lifecycle (`controllers/deployment/launcher`)

HTTP services are proxied at L7 via httpingress, but non-HTTP services (TCP/UDP) need L4 routing through the ipalloc → ServiceController → nftables pipeline. That pipeline requires a `network_v1alpha.Service` entity to exist — without one, no IP is allocated and no firewall rules are created.

The launcher now manages Service entities for app services that have at least one non-HTTP port:

- **`ensureServiceForPorts`** — creates or updates a Service entity with the correct ports, match labels, and metadata. On update, preserves existing IPs so ipalloc doesn't need to re-allocate.
- **`cleanupStaleServices`** — deletes orphaned Service entities when services are removed from config or become HTTP-only.
- Deterministic entity IDs (`svc/<appname>-<servicename>`) for stable identity across deploys.

## ServiceController per-port endpoint chain fix (`controllers/service`)

The ServiceController had a bug where `Create()` used `srv.Port[0]` to determine the DNAT target port for **all** endpoint chains. For multi-port services, this meant all traffic was forwarded to the first port regardless of the original destination port — e.g., TCP traffic arriving on port 7000 would get DNAT'd to port 3000.

Moved endpoint chain setup inside the per-port loop so each service port gets its own endpoint chains targeting the correct port.

## Test app

Adds `testdata/tcp-echo/` — a minimal Go app with a TCP echo server on port 7000 and an HTTP health endpoint on port 3000, for validating multi-port service behavior end-to-end.

Example app config for an IRC server:
```toml
[services.irc]
command = "./ircd"

[[services.irc.ports]]
port = 6667
name = "irc"
type = "tcp"

[[services.irc.ports]]
port = 6697
name = "irc-tls"
type = "tcp"
node_port = 6697
```

Closes MIR-745

### DIFF
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -226,7 +226,7 @@ func (ac *AppConfig) Validate() error {
 		// Validate ports configuration
 		if len(svcConfig.Ports) > 0 {
 			// Mutual exclusion: cannot use both ports[] and scalar port fields
-			if svcConfig.Port > 0 || svcConfig.PortName != "" || svcConfig.PortType != "" {
+			if svcConfig.Port != 0 || svcConfig.PortName != "" || svcConfig.PortType != "" {
 				return fmt.Errorf("service %s: cannot use both 'ports' array and scalar port/port_name/port_type fields", serviceName)
 			}
 
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -48,7 +48,6 @@ type DiskConfig struct {
 type PortConfig struct {
 	Port     int    `toml:"port"`
 	Name     string `toml:"name"`
-	Protocol string `toml:"protocol"`
 	Type     string `toml:"type"`
 	NodePort int    `toml:"node_port"`
 }
@@ -243,12 +242,12 @@ func (ac *AppConfig) Validate() error {
 				if p.Name == "" {
 					return fmt.Errorf("service %s: ports[%d] name is required", serviceName, i)
 				}
-				if p.Protocol != "" && p.Protocol != "tcp" && p.Protocol != "udp" {
-					return fmt.Errorf("service %s: ports[%d] protocol must be \"tcp\" or \"udp\"", serviceName, i)
+				if p.Type != "" && p.Type != "http" && p.Type != "tcp" && p.Type != "udp" {
+					return fmt.Errorf("service %s: ports[%d] type must be \"http\", \"tcp\", or \"udp\"", serviceName, i)
 				}
-				proto := p.Protocol
-				if proto == "" {
-					proto = "tcp"
+				proto := "tcp"
+				if p.Type == "udp" {
+					proto = "udp"
 				}
 				if p.NodePort < 0 || p.NodePort > 65535 {
 					return fmt.Errorf("service %s: ports[%d] node_port must be between 0 and 65535", serviceName, i)
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -44,12 +44,22 @@ type DiskConfig struct {
 	LeaseTimeout string `toml:"lease_timeout"`
 }
 
+// PortConfig represents a network port for a service
+type PortConfig struct {
+	Port     int    `toml:"port"`
+	Name     string `toml:"name"`
+	Protocol string `toml:"protocol"`
+	Type     string `toml:"type"`
+	NodePort int    `toml:"node_port"`
+}
+
 // ServiceConfig represents configuration for a specific service
 type ServiceConfig struct {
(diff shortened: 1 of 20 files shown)


evanphx (changes requested):
This all looks good, but have it work, a network_v1alpha.Service has to be created but I don't see it here. Something like this: https://github.com/mirendev/runtime/blob/main/pkg/addon/framework.go#L139

phinze:
lol you're right - all this config wiring missed the last mile! fixing

phinze:
Ok this does the thing now. Ready for re-review! (I'll knock out the last few rabbit issues in parallel.)

appconfig/appconfig.go
```diff
@@ -213,6 +223,48 @@ func (ac *AppConfig) Validate() error {
...
+				if p.Name == "" {
+					return fmt.Errorf("service %s: ports[%d] name is required", serviceName, i)
+				}
+				if p.Protocol != "" && p.Protocol != "tcp" && p.Protocol != "udp" {
+					return fmt.Errorf("service %s: ports[%d] protocol must be \"tcp\" or \"udp\"", serviceName, i)
+				}
+				proto := p.Protocol
+				if proto == "" {
+					proto = "tcp"
+				}
```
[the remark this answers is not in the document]
phinze: Good catch gonna make UDP work end-to-end in a follow on commit here stand by.

docs/docs/firewall.md
```diff
@@ -49,11 +49,14 @@ If you're running Miren on a cloud provider, you'll need to configure security g
 | 8443 | UDP | Miren API (QUIC) - CLI and client connections | Yes |
 | 80 | TCP | HTTP traffic to your applications (redirects to HTTPS) | Yes |
 | 443 | TCP | HTTPS traffic to your applications | Yes |
+| NodePorts | TCP/UDP | Direct L4 traffic to non-HTTP services (see [Traffic Routing](/traffic-routing)) | If using TCP/UDP services |
 
 **Miren API (UDP 8443):** The Miren API uses QUIC (HTTP/3) over UDP. This is how the CLI communicates with the server and how remote clients connect to your cluster.
 
 **HTTP Ingress (TCP 80/443):** Application traffic uses standard HTTP/HTTPS. Port 80 handles ACME certificate challenges and redirects to HTTPS. Port 443 serves your applications over TLS.
 
+**NodePorts:** If your app exposes non-HTTP services with `node_port` in the port configuration, those ports must also be open. For example, an IRC server with `node_port = 6667` requires TCP port 6667 to be reachable.
```
evanphx, line 58: Open and reachable in what sense? That the host machine doesn't have firewall rules prevent them?

evanphx (approved):
Looks great! I suspect there will be some "service in app.toml" vs "service entity" confusion until we change the name of the later, but that's fine for now.

docs/docs/firewall.md
phinze: Yeah the section this is in sets the context:

> ## Required External Ports
> 
> If you're running Miren on a cloud provider, you'll need to configure security groups or network ACLs to allow external traffic to reach the Miren server.

phinze:
> I suspect there will be some "service in app.toml" vs "service entity" confusion until we change the name of the later, but that's fine for now.

Yeah we can finalize a name and land #611 soon

PR #642, opened 2026-03-04
Fix saga entity ID type mismatch causing nested execution failures

evanphx:
The saga storage was passing exec.ID as a raw string to entity.New() for the db/id attribute. entity.New wraps strings as KindString, but Entity.Id() only recognizes KindId values. This caused ForceID() to silently generate a new auto-ID, storing the parent saga under a different ID than its execution ID. When a nested child saga tried to reference the parent via parent_execution_id, the ref validation couldn't find it.

Fix: use entity.Id(exec.ID) so the db/id value has the correct type.

Also add a panic guard in ForceID() to catch this class of bug: if a db/id attribute exists but has the wrong value kind, panic immediately instead of silently generating a new ID.

Also fix Get() to handle cond.ErrNotFound from EtcdStore (in addition to entity.ErrEntityNotFound from FileStore).

### DIFF
--- a/controllers/sandbox/volume.go
+++ b/controllers/sandbox/volume.go
@@ -227,7 +227,7 @@ func (c *SandboxController) ensureDisk(ctx context.Context, diskName string, siz
 	name := idgen.GenNS("disk")
 
 	putResp, err := c.EAC.Create(ctx, entity.New(
-		entity.DBId, "disk/"+name,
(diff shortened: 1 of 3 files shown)


phinze (approved):
Good catch, and love the panic guard to catch it going forward. That reminds me - I'll make an issue to wire up some kind of panic alerting for our clusters. ✍️

PR #648, opened 2026-03-05
Parallelize artifact uploads in release workflow

phinze:
The "Upload to Miren API" step in the release workflow was running about 28 sequential curl uploads — 7 artifacts, each with a main file and a checksum file, uploaded to up to 2 paths. Each curl takes around 7 seconds, so we were burning 3-4 minutes just waiting on serial network I/O every release.

The fix splits the upload step into three phases. First, we prepare everything locally — generate checksums, build the artifact manifest, write `version.json`, and queue up all the upload tasks into a simple pipe-delimited file. Then we drain that queue with up to 4 concurrent background curl jobs, using `wait -n` as a semaphore to stay under the cap. Finally, once every artifact upload has succeeded, we upload `version.json` sequentially — it has to go last so consumers never see a version manifest that references artifacts that aren't there yet.

One subtlety worth noting: `set -e` doesn't propagate failures from background processes, so we explicitly track PIDs and check exit codes. The `wait -n 2>/dev/null` handles the edge case where the builtin returns non-zero with no remaining children.

Closes MIR-728

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -320,6 +320,7 @@ jobs:
           set -eo pipefail
 
           REF_NAME="${{ needs.init.outputs.ref }}"
+          TOKEN="${{ steps.get-oidc-token.outputs.token }}"
 
           # Determine if this is a tag build
           IS_TAG_BUILD=false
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -317,6 +317,7 @@ jobs:
           set -eo pipefail
 
           REF_NAME="${{ needs.init.outputs.ref }}"
+          TOKEN="${{ steps.get-oidc-token.outputs.token }}"
```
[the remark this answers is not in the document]
phinze: Good catch on the token TTL, but this isn't a new risk — the previous sequential version also captured the token once and reused it across all ~28 uploads (~3-4 min). With parallelization the upload phase actually completes faster (~1 min), so we're further from the expiry window than before. The worst-case retry scenario would require every upload to hit the 180s max-time cap, which would indicate a much bigger problem than an expired token. Refreshing mid-step would require splitting into multiple workflow steps to call `core.getIDToken()` again — not worth the complexity for a scenario we haven't hit.

.github/workflows/release.yml
```diff
@@ -467,11 +450,61 @@ jobs:
...
+                if kill -0 "$pid" 2>/dev/null; then
+                  new_pids+=("$pid")
+                fi
+              done
+              running_pids=("${new_pids[@]}")
+            done
+
+            upload_one "$file" "$url" &
+            running_pids+=($!)
+          done < upload-queue.txt
```
evanphx, line 496: 😵

evanphx (approved):

.github/workflows/release.yml
phinze: lol nothing to see here just `sync.WaitGroup` in bash :bowtie:

PR #649, opened 2026-03-05
Add macOS .pkg installer for Gatekeeper compatibility

phinze:
The macOS binary is signed and notarized, but bare Mach-O binaries can't have notarization tickets stapled to them. This means when someone downloads `miren` and double-clicks it from Finder, Gatekeeper has to do an online check — and if that check is slow or the system is offline, users get a scary warning dialog even though the binary is perfectly legit.

The fix is to also ship a `.pkg` installer alongside the existing `.zip`. A `.pkg` *can* have the notarization ticket stapled directly to it, so Gatekeeper validates instantly with no network round-trip and no warning. The installer drops the binary into `/usr/local/bin/miren`, which is exactly where you'd put it by hand anyway.

The `.zip` sticks around for programmatic consumers — Homebrew cask and `miren upgrade` both expect it and don't need the `.pkg`. Everything `.pkg`-related is gated on having a `Developer ID Installer` certificate available (via new optional secrets `APPLE_INSTALLER_CERTIFICATE_BASE64` / `APPLE_INSTALLER_CERTIFICATE_PASSWORD`), so the workflow degrades gracefully until that cert is configured. If the existing `.p12` already contains both the Application and Installer identities, no new secrets are needed at all.

Stacked on #648.

Closes MIR-572

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -134,6 +134,8 @@ jobs:
       env:
         APPLE_CERTIFICATE_BASE64: ${{ secrets.APPLE_CERTIFICATE_BASE64 }}
         APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Looks good!

PR #652, opened 2026-03-09
Switch binary release artifacts from .zip to .tar.gz

phinze:
Our per-platform binary artifacts (`miren-linux-amd64.zip`, etc.) required `unzip` on the target system, which plenty of minimal Linux distributions don't include. Meanwhile `tar` is universal on both Linux and macOS, and the base release artifacts were already using tar.gz — so the binary artifacts were the odd ones out.

This switches the Go client code (`pkg/release/artifact.go`, `downloader.go`) to request and extract `.tar.gz` exclusively, removing the `extractZip()` code path entirely. The build scripts and CI pipeline now produce both `.tar.gz` and `.zip` during a transition period, so older clients that still ask for `.zip` can self-upgrade without breaking.

The systemd test Dockerfile also drops `unzip` from its package list since it's no longer needed.

Once old clients have aged out, a follow-up PR will stop producing `.zip` artifacts. There are also companion changes needed in `mirendev/cloud` (installer script) and `mirendev/mirendev` (website download URLs) to switch to the new filenames.

Closes MIR-768

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -199,14 +199,17 @@ jobs:
 
     - name: Package binary
       run: |
-        # Create a zip file with the binary
+        tar -czf miren-${{ matrix.os }}-${{ matrix.arch }}.tar.gz -C bin miren
+        # Transitional: also produce .zip for older clients
         zip -j miren-${{ matrix.os }}-${{ matrix.arch }}.zip bin/miren
 
     - name: Upload build artifact
       uses: actions/upload-artifact@v4
       with:
         name: miren-${{ matrix.os }}-${{ matrix.arch }}
-        path: miren-${{ matrix.os }}-${{ matrix.arch }}.zip
+        path: |
+          miren-${{ matrix.os }}-${{ matrix.arch }}.tar.gz
+          miren-${{ matrix.os }}-${{ matrix.arch }}.zip
         retention-days: 1
 
   build-and-push-docker:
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -199,14 +199,17 @@ jobs:
 
     - name: Package binary
       run: |
-        # Create a zip file with the binary
+        tar -czf miren-${{ matrix.os }}-${{ matrix.arch }}.tar.gz -C bin miren
+        # Transitional: also produce .zip for older clients
```
evanphx, line 203: Do we want to continue to always make both so we can use .zip on macOS, where it's a bit more natural?

evanphx (approved):

.github/workflows/release.yml
phinze: not sure! the macOS "natural" flow is gonna transition to .pkg in #649 so at that point idk if the dual-archive will be that useful

pkg/release/downloader.go
```diff
@@ -98,13 +92,8 @@ func (d *assetDownloader) Download(ctx context.Context, artifact Artifact, opts
...
 
-	// Extract the binary based on file type
-	var binaryPath string
-	if strings.HasSuffix(archivePath, ".zip") {
-		binaryPath, err = d.extractZip(archivePath, opts.TargetDir)
-	} else {
-		binaryPath, err = d.extractTarGz(archivePath, opts.TargetDir)
-	}
+	// Extract the binary
+	binaryPath, err := d.extractTarGz(archivePath, opts.TargetDir)
```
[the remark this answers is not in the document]
phinze: This is pre-existing behavior — the old `extractZip()` did the same thing (only extracted the `miren` binary). Base/release artifacts go through a completely separate code path in `cli/commands/download_release.go` which extracts the full archive. The downloader here is only used for binary upgrades via `Manager.UpgradeArtifact()`.

PR #659, opened 2026-03-11
Add wildcard route support (MIR-781)

evanphx:
## Summary

- Add `LookupWithWildcard` to ingress client: chains exact match → wildcard subdomain (`*.example.com`) → wildcard bare domain → nil, all using existing entity index for O(1) lookups
- Add `ValidateWildcardHost` to reject invalid patterns like `*.com` or `foo.*.com`
- Update HTTP ingress to use wildcard-aware lookup for incoming requests
- Update `RouteSet.HasRoute` in autotls to provision TLS certs for wildcard subdomains
- Add CLI validation when setting wildcard routes (`miren route set *.example.com app`)

## Test plan

- [x] `hack/it ./api/ingress/...` — 27 tests pass (wildcard subdomain, bare domain, exact priority, case insensitivity, validation)
- [x] `hack/it ./components/autotls/...` — 8 tests pass
- [x] `make test` — 3582 tests pass, 20 skipped
- [x] `make lint` — 0 issues


phinze (approved):
Claude and I sat down with this one and traced through the lookup chains, the autotls cert provisioning path, and the validation logic end-to-end. Clean feature, well-structured.

The three-step fallback in `LookupWithWildcard` (exact → wildcard subdomain → wildcard bare domain) is a nice design — deterministic priority with no iteration or regex, just index lookups. And mirroring the same logic in `RouteSet.HasRoute` for cert provisioning means wildcards get per-subdomain certs on demand rather than a blanket wildcard cert, which is the right call.

One thing we noted: on the HTTP ingress hot path, a wildcard-routed subdomain now does 2 serial entity lookups (exact miss → wildcard hit), and a complete miss does 3 before falling through to the default route check. These are local RPC calls so latency is low, but if we ever want to tighten this up, a batched `LookupAny(ctx, ...keys)` that resolves multiple index keys in one RPC round-trip would collapse the chain. Not urgent — exact matches (the common case) are still 1 call — just flagging it as a future optimization opportunity.

evanphx:
@phinze agree on the round trips, we can do that as a follow up. I think we might want to also add a cache on the check side since most of the pain will come from something doing requests over and over, so even a 5 min TTL cache would eliminate most of the pain.

PR #661, opened 2026-03-11
Prefetch TLS certs eagerly when routes are configured

phinze:
When `miren route set` links a subdomain to an app, the very first HTTPS request would time out — autocert only provisions certs lazily on the first TLS handshake, which blocks the connection for 30-90s while the ACME flow completes. Not a great first impression.

The fix is a new `AutocertController` that watches `http_route` entities through the existing controller framework. When a route appears, it eagerly calls `autocert.Manager.GetCertificate()` with a synthetic `ClientHelloInfo` to kick off provisioning before any browser shows up. This is the same reconcile pattern the DNS-01 certificate controller already uses, so both paths now work the same way.

This let us delete the old `RouteWatcher` and `ServeTLS` machinery (~380 lines) and collapse `autotls.go` down to a single `ServeTLSWithController` function that works for both cert paths. The new `HTTPChallengeProvider` interface lets the autocert controller optionally inject ACME challenge handling on port 80, while the DNS-01 path just gets the plain redirect handler.

A few design details worth noting: there's a `HostPolicy` on the autocert manager that checks a controller-maintained set of allowed domains, so rogue TLS handshakes can't trigger cert provisioning for arbitrary domains (they get the self-signed fallback instead). And a ready channel handles the startup ordering problem — the controller manager starts before the port-80 server, so `Reconcile` blocks until `SetReady()` is called after the HTTP server is up.

Stacked on #659.

MIR-771


controllers/certificate/autocert_controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Wait for port-80 ACME challenge server to be ready before attempting provisioning
+	select {
+	case <-c.ready:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	// Eagerly provision the certificate with a synthetic ClientHelloInfo
+	hello := &tls.ClientHelloInfo{ServerName: domain}
```
evanphx, line 102: I don't think this works when domain ends up with, say "*.pds.miren.dev". I think we have to handle wildcards specially in this code path, only getting a eager cert for the base (ie pds.miren.dev) and doing an inline provision when blah.pds.miren.dev arrives.

components/autotls/autotls.go
```diff
@@ -170,99 +71,15 @@ func ServeTLS(ctx context.Context, log *slog.Logger, dataPath string, email stri
...
-				host = hostWithoutPort
-			}
-
-			isLocalhost := host == "localhost" || host == "127.0.0.1" || host == "::1"
-			isIPAddress := net.ParseIP(host) != nil
-
-			if isLocalhost || isIPAddress {
-				h.ServeHTTP(w, r)
-				return
-			}
```
evanphx, line 256: Are we losing this functionality? I don't see where it gets moved to, this allows you to make a request to localhost and have it work, a great property for testing things on the server.

components/autotls/autotls.go
```diff
@@ -170,99 +71,15 @@ func ServeTLS(ctx context.Context, log *slog.Logger, dataPath string, email stri
...
-			if r.Method != "GET" && r.Method != "HEAD" {
-				http.Error(w, "Use HTTPS", http.StatusBadRequest)
-				return
-			}
-
-			target := "https://" + host + r.URL.RequestURI()
-			http.Redirect(w, r, target, http.StatusFound)
-		}),
+		Addr:              ":80",
+		Handler:           port80Handler,
```
evanphx, line 82: won't port80Handler be nil in some cases?

evanphx (changes requested):
Doing the pre-emptive TLS fetch I like, but I think we've ended up losing some functionality and not sure wildcards work right.

controllers/certificate/autocert_controller.go
phinze: Good catch — addressed in f22a6c3a. Wildcard routes now store the `*.example.com` pattern in allowedHosts (for HostPolicy matching of subdomains), but Reconcile eagerly provisions only the base domain cert. Individual subdomains provision inline when they arrive via GetCertificate. Added `isAllowedHost` helper that checks both exact and wildcard matches, plus tests for the matching logic.

components/autotls/autotls.go
phinze: Not lost\! The localhost passthrough is still there — it moved from the old `ServeTLS` into the new `ServeTLSWithController` at lines 57-63 of the updated `autotls.go`. Same logic: localhost/127.0.0.1/::1 and IP addresses get served directly without redirect.

components/autotls/autotls.go
phinze: Nope — `port80Handler` is initialized to `redirectHandler` on line 74 before the conditional wrap. If the certProvider doesn't implement `HTTPChallengeProvider`, it just stays as the redirect handler. So it's always non-nil by the time it's used.

controllers/certificate/autocert_controller.go
evanphx: Ah great!

components/autotls/autotls.go
evanphx: Oh! oh oh oh oh. yes, I *WAS* missing something like I suspected. perfect.

components/autotls/autotls.go
evanphx: ah ah!

evanphx (approved):

controllers/certificate/autocert_controller_test.go
```diff
@@ -0,0 +1,205 @@
...
+func newTestAutocertController(t *testing.T) *AutocertController {
+	t.Helper()
+	log := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
+	dataPath := t.TempDir()
+	// nil eac is fine for unit tests that don't exercise Delete
+	c := NewAutocertController(log, nil, dataPath, "test@example.com")
+	if err := c.Init(context.Background()); err != nil {
+		t.Fatalf("failed to init autocert controller: %v", err)
+	}
+	return c
```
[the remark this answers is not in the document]
phinze: The tests are hermetic — the domains are synthetic and there's no real ACME server, so `GetCertificate` fails at the HostPolicy or cache layer and falls back gracefully. No actual ACME requests are made. We're comfortable with this level of testing for now.

controllers/certificate/autocert_controller.go
```diff
@@ -0,0 +1,185 @@
...
+		log.Info("wildcard route: subdomains will provision certs inline on first request")
+		return nil
+	}
+
+	// Wait for port-80 ACME challenge server to be ready before attempting provisioning
+	select {
+	case <-c.ready:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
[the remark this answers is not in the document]
phinze: This is intentional. If a TLS handshake arrives before port 80 is up, `GetCertificate` tries autocert, it fails (no challenge server to complete HTTP-01), and we return the self-signed fallback. That's strictly better than today's behavior — immediate fallback response vs a 30-90s ACME timeout blocking the handshake.

controllers/certificate/autocert_controller.go
```diff
@@ -0,0 +1,185 @@
...
+func (c *AutocertController) Reconcile(ctx context.Context, route *ingress_v1alpha.HttpRoute, meta *entity.Meta) error {
+	domain := strings.ToLower(strings.TrimSpace(route.Host))
+	routeID := meta.Id()
+	if domain == "" {
+		c.log.Warn("http_route has empty host, skipping certificate provisioning", "route", routeID)
+		return nil
+	}
+
+	c.allowedHosts.Store(domain, struct{}{})
+
```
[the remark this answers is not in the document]
phinze: Fair point — a route changing its host in place would leave the old domain in the allowlist. In practice route host changes are rare, and a stale entry is low-risk (it just means autocert *could* provision for that domain if a handshake arrives). The Delete path handles cleanup when routes are removed. We'll keep it simple for now and revisit if this becomes a real issue.

controllers/certificate/autocert_controller.go
```diff
@@ -0,0 +1,185 @@
...
+// HTTPHandler returns an http.Handler that serves ACME HTTP-01 challenge responses,
+// delegating non-challenge requests to the provided fallback handler.
+func (c *AutocertController) HTTPHandler(fallback http.Handler) http.Handler {
+	return c.mgr.HTTPHandler(fallback)
+}
+
+// isAllowedHost checks whether host is covered by the allowed set, including
+// wildcard entries. For example, if "*.example.com" is in the set, both
+// "foo.example.com" and "example.com" are considered allowed.
```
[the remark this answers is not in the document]
phinze: Good catch — fixed in 579c518b. Comment now correctly states that `*.example.com` matches `foo.example.com` but not `example.com`.

PR #664, opened 2026-03-11
Pre-populate allowed hosts at startup to avoid fallback cert race

phinze:
After a server restart, there's a window where the TLS server is already accepting connections but the autocert controller hasn't reconciled all its routes yet. Any request that arrives during that window for a domain whose `Reconcile` hasn't run gets the self-signed fallback cert — even though the real ACME cert is sitting right there in the DirCache. The `isAllowedHost` guard (from #542) rejects the domain because `allowedHosts` is still empty, so autocert never gets a chance to serve the cached cert.

We hit this in practice when an OIDC discovery request landed a few seconds after a restart. The cert came back with `localhost` in its SANs instead of the actual domain, and the TLS handshake failed. Five seconds later, once reconciliation caught up, everything was fine — but that's a rough experience for the first request.

The fix is to query all existing `http_route` entities during `Init()` and seed `allowedHosts` before the TLS server starts. This way the guard works from the very first handshake, and the controller's reconcile loop still handles new routes and cert provisioning as before.

Closes MIR-786

### DIFF
--- a/controllers/certificate/autocert_controller.go
+++ b/controllers/certificate/autocert_controller.go
@@ -65,7 +65,41 @@ func (c *AutocertController) Init(ctx context.Context) error {
 		},
 	}
 
+	// Pre-populate allowedHosts from existing http_route entities so the
+	// isAllowedHost guard in GetCertificate works immediately — before the
+	// controller manager starts reconciling routes one by one.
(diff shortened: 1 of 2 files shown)


phinze:
lol wrong solution, Claude, fixing now

evanphx (approved):

PR #666, opened 2026-03-12
Fix PDS disk boot failures (MIR-788)

evanphx:
## Summary

A user deployed with `size = 20` instead of `size_gb = 20` in their miren.toml disk config. TOML silently ignored the unknown field, leaving `SizeGB=0`. This caused a cascade of failures:

1. **Sandbox stuck PENDING forever**: `configureVolumes()` failed every boot attempt ("disk does not exist and no size specified"), but the sandbox was never marked DEAD. The pool controller counts PENDING sandboxes as "actual" capacity, so it didn't create replacements — but the stale-PENDING monitor would mark them STOPPED after 5 minutes, then the pool would create a new one. Since STOPPED sandboxes don't trigger the crash counter, the exponential backoff never kicked in.

2. **Lease carousel after fix**: After the user fixed the config, old stuck sandboxes' leases conflicted with new ones from the redeployment, causing ~6 minutes of lease churn before stabilizing.

3. **Orphaned disk on app delete**: `ensureDisk()` never set `CreatedBy` on disk entities, so `DeleteAppTransitive` couldn't find the disk when the app was deleted. The user had to manually clean up the disk and its lease.

## Changes

- **Reject unknown TOML fields** (`appconfig/appconfig.go`): Enable `DisallowUnknownFields()` on all decoder paths so typos like `size` instead of `size_gb` produce a clear parse error at deploy time.

- **Mark sandbox DEAD on any boot failure** (`controllers/sandbox/sandbox.go`): Restructured `createSandbox()` to use a named return with a top-level defer — every error path now transitions the sandbox to DEAD. This is a catch-all that covers `configureVolumes`, `buildSpec`, `NewContainer`, port health checks, and any future failure modes. The pool controller's existing exponential crash-backoff (10s → 20s → 40s → ... → 15min) then kicks in properly.

- **Set `CreatedBy` on disk entities** (`controllers/sandbox/volume.go`): Pass the app ID through to `ensureDisk()` so `DeleteAppTransitive` can find and clean up disks when an app is deleted.

- **Validate disk existence at deploy time** (`servers/build/build.go`): When `size_gb` is 0 (meaning "use existing disk"), query the entity store to verify the disk exists before creating any sandboxes. Fails with: `disk "X" does not exist; set size_gb to auto-create it`.

## Test plan

- [x] `go test ./appconfig/...` — unknown field rejection tests pass
- [x] `go test ./servers/build/...` — disk existence validation tests pass (missing disk, existing disk, auto-create)
- [x] `go vet` and `make lint` clean
- [ ] `make test` full suite in iso

### DIFF
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -98,6 +98,7 @@ func LoadAppConfig() (*AppConfig, error) {
 
 			var ac AppConfig
 			dec := toml.NewDecoder(fi)
+			dec.DisallowUnknownFields()
 			err = dec.Decode(&ac)
 			if err != nil {
 				return nil, err
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -1,6 +1,7 @@
 package appconfig
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -124,6 +125,7 @@ func LoadAppConfigUnder(dir string) (*AppConfig, error) {
 
 		var ac AppConfig
 		dec := toml.NewDecoder(fi)
+		dec.DisallowUnknownFields()
 		err = dec.Decode(&ac)
 		if err != nil {
 			return nil, err
@@ -142,7 +144,9 @@ func LoadAppConfigUnder(dir string) (*AppConfig, error) {
 
 func Parse(data []byte) (*AppConfig, error) {
 	var ac AppConfig
-	err := toml.Unmarshal(data, &ac)
+	dec := toml.NewDecoder(bytes.NewReader(data))
+	dec.DisallowUnknownFields()
+	err := dec.Decode(&ac)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


appconfig/appconfig.go
```diff
@@ -97,6 +98,7 @@ func LoadAppConfig() (*AppConfig, error) {
 
 			var ac AppConfig
 			dec := toml.NewDecoder(fi)
```
phinze, line 100: This decoder path is still permissive — same fix as the other two (`dec.DisallowUnknownFields()`) would close the gap.

phinze (approved):
Claude and I went through this together — traced the full failure cascade from the TOML typo through the PENDING loop and the orphaned disk. Really nice work addressing this at every layer.

One thing: we agree with CodeRabbit that `LoadAppConfig()` at `appconfig.go:100` should also get `DisallowUnknownFields()` to keep all three loader paths consistent. Otherwise this is good to go!

appconfig/appconfig.go
phinze: Oh fixed already! Nice 👍

PR #668, opened 2026-03-12
Fix OIDC provider collision across routes sharing same provider URL

evanphx:
## Summary

- **Provider name collision**: When two routes used the same OIDC provider URL (e.g. `multipass.miren.garden`) with different client credentials, the auto-generated provider name was based only on the URL hostname. The second `route oidc enable` would overwrite the first route's provider entity, scrambling credentials across routes. Provider names now include the route label (e.g. `socials.example.com/multipass.miren.garden`).
- **Stale handler cache**: OIDC handlers were cached by route hostname but never validated against the current provider config. Even after a provider entity was corrected, the old credentials persisted until server restart. The cache now checks provider config on each lookup and replaces stale handlers.

## Test plan

- [x] `TestOidcProviderMatches` — validates the config comparison helper
- [x] `TestGetOrCreateOIDCHandlerCacheInvalidation` — verifies cache returns new handler after provider update
- [x] `TestInjectClaims` — existing tests still pass
- [ ] Manual: enable OIDC on two routes with same provider URL but different client credentials, verify each route redirects with its own client_id

### DIFF
--- a/cli/commands/route_oidc_enable.go
+++ b/cli/commands/route_oidc_enable.go
@@ -68,10 +68,12 @@ func RouteOidcEnable(ctx *Context, opts struct {
 	if providerName == "" {
 		// If no provider name given and provider-url specified, generate a name
 		if opts.ProviderURL != "" {
-			// Use the host from the provider URL as the provider name
-			providerName = strings.ReplaceAll(opts.ProviderURL, "https://", "")
-			providerName = strings.ReplaceAll(providerName, "http://", "")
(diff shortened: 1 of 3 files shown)


phinze (approved):
Claude and I reviewed this together — straightforward bugfix, we read through the full OIDC handler cache path and the CLI naming change end-to-end.

Both fixes look correct and well-scoped. The route-scoped provider naming prevents the entity collision, and the config comparison on cache lookup handles the stale-handler problem cleanly. Tests cover the key scenarios. Ship it!

PR #669, opened 2026-03-12
Enable OIDC authentication in local-only mode

evanphx:
## Summary

- OIDC authentication (e.g., GitHub Actions tokens) was only available when the cluster was registered with Miren Cloud
- Wraps `LocalOnlyAuthenticator` with `CompositeAuthenticator` so OIDC works as a fallback alongside client certificate auth in local-only mode
- Mirrors the existing CloudAuth pattern using the same `CompositeAuthenticator`/`CompositeAuthorizer` types

Fixes MIR-792

## Test plan

- [x] `hack/it ./components/coordinate/...` — 8 tests pass
- [x] `hack/it ./pkg/oidcauth/...` — 52 tests pass


phinze (approved):
Claude and I reviewed this — straightforward change that mirrors the existing cloud auth composite pattern. The `CompositeAuthorizer(nil)` correctly auto-approves cert auth while restricting OIDC to the deploy role. Looks good!

PR #670, opened 2026-03-12
Optimize deploy upload with delta file transfer

evanphx:
## Summary

- Adds manifest-first deploy flow: client sends file hashes, server compares against cached source code image, client uploads only changed files
- New RPC methods `prepareUpload` and `buildFromPrepared` on the Builder service
- Source code cache stored per-app at `{DataPath}/source-code/{app}/` with atomic writes to prevent corruption
- Refactored `MakeTar`/`MakeFilteredTar` to share a common `makeTarWithFilter` implementation with lazy directory emission
- Deploy UI shows cache reuse stats on the "Upload artifacts" phase line
- Falls back gracefully to full upload if server doesn't support the new methods or if manifest computation fails

## Test plan

- [x] Unit tests for `ComputeManifest` and `MakeFilteredTar` in `pkg/tarx/tarx_test.go`
- [x] Unit tests for source cache roundtrip, staging, overwrite in `servers/build/source_cache_test.go`
- [x] Test for directory emission behavior (only dirs containing accepted files are included)
- [x] `make test` passes (3579 tests)
- [x] `make lint` clean
- [x] Manual test: deploy app twice, verify "Reused N/M files" on second deploy

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/moby/buildkit/client"
@@ -70,6 +71,12 @@ func (w *buildLogWriter) write(msg string) {
 	}
 }
 
+type buildSession struct {
+	dir        string
+	appName    string
+	cancelFunc context.CancelFunc
+}
+
 type Builder struct {
 	Log           *slog.Logger
 	EAS           *entityserver_v1alpha.EntityAccessClient
@@ -80,6 +87,7 @@ type Builder struct {
 	TempDir       string
 	Registry      string
 	DNSHostname   string // Cloud-provisioned DNS hostname for default route display
+	DataPath      string
 
 	Resolver  netresolve.Resolver
 	LogWriter observability.LogWriter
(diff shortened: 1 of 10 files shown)


servers/build/source_cache.go
```diff
@@ -0,0 +1,267 @@
...
+	data, err := os.ReadFile(filepath.Join(dir, "manifest.json"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return map[string]cachedFileInfo{}, nil
+		}
+		return nil, err
+	}
+
+	var manifest map[string]cachedFileInfo
+	if err := json.Unmarshal(data, &manifest); err != nil {
```
phinze, line 82: Nit, totally optional: when `manifest.json` is corrupt JSON this silently returns an empty map. Safe behavior (just triggers a full re-upload), but a debug log would save someone a head-scratching session someday.

servers/build/source_cache.go
```diff
@@ -0,0 +1,267 @@
...
+		return fmt.Errorf("marshaling manifest: %w", err)
+	}
+
+	tmpManifestPath := filepath.Join(dir, "manifest.json.tmp")
+	if err := os.WriteFile(tmpManifestPath, manifestData, 0644); err != nil {
+		return err
+	}
+
+	// Atomic rename: both files are complete before becoming visible.
+	layerPath := filepath.Join(dir, "layer.tar.gz")
```
phinze: Another optional one: the two `os.Rename` calls aren't atomic together — a crash between them could leave layer/manifest out of sync. Impact is just a wasted full re-upload next time. If you ever revisit, renaming manifest first is slightly safer (stale layer + accurate manifest = cache misses, not hash/content mismatch).

phinze (approved):
Claude and I dug into this one together — traced the full flow from manifest computation through cache staging and the new build paths. I've already dubbed this "mirsync" in my head.

Really like the design. The fallback-to-full-upload approach means this is pure upside with no risk to existing deploys, and the `buildFromDir` extraction is a clean refactor.

CodeRabbit flagged a few things (auth check on `BuildFromPrepared`, session path tracking, deferred close in Walk) — skipping those here since you've seen them.

Two small optional nits below. Also, food for thought: now that we have per-file sizes in the manifest and the cached `layer.tar.gz` on disk, we've got everything needed to estimate compression ratio and show real upload progress bars with ETAs + rsync-style speedup stats. Going to file a follow-up for that.

servers/build/build.go
```diff
@@ -810,14 +821,181 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
...
+
+	sessionID := idgen.Gen("s")
+
+	cleanupCtx, cancelCleanup := context.WithCancel(context.Background())
+
+	b.sessions.Store(sessionID, &buildSession{
+		dir:        tempDir,
+		appName:    name,
+		cancelFunc: cancelCleanup,
+	})
```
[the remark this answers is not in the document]
evanphx: That's the clients discretion, if it wants to screw up, it's welcome to.

servers/build/source_cache.go
evanphx: Done!

servers/build/source_cache.go
evanphx: yup! improved this with a couple of fixes.

PR #681, opened 2026-03-16
Speed up log queries by dropping server-side sort pipes

phinze:
`m logs system` on garden was taking ~10 seconds to show 100 lines, scaling linearly with the time window (~16s for `--last 1h`). The culprit was `| sort by (_time)` in the LogsQL query pipe — this forces VictoriaLogs to scan, buffer, and sort *every* matching entry in the time window before it can return anything. For system logs on a busy cluster, that's hundreds of thousands of entries getting sorted just to return 100.

Turns out VictoriaLogs already has a built-in `limit` query parameter that returns the N most recent entries without needing a sort pipe at all. We just weren't using it — instead we had `sort by (_time) desc | limit 100` in the query string, which defeats the optimization entirely.

The fix drops the sort pipes from both `executeStreamQuery` and `executeQuery`, uses the native `limit` param, and accepts VictoriaLogs' ingestion order as-is. For a single-writer server the logs come back approximately chronological anyway, so there's no need to sort client-side either. Tested on garden — default query goes from ~10s to under 1s.

### DIFF
--- a/observability/logs.go
+++ b/observability/logs.go
@@ -337,17 +337,17 @@ func (l *LogReader) executeStreamQuery(ctx context.Context, query string, logCh 
 	baseURL := normalizeBaseURL(l.Address)
 	queryURL := baseURL + "/select/logsql/query"
 
-	var sortedQuery string
-	if o.Limit > 0 {
-		// Get the most recent N entries by sorting descending with limit,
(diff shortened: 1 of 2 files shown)


evanphx (approved):
As yes, the classic database blunder: sorting by time when latency is on the line!

PR #686, opened 2026-03-18
Implement cluster-side global router with WebSocket tunneling (MIR-825)

evanphx:
## Summary

- **Global router**: Adds WebSocket-based tunnel between clusters and PoP edge servers for NAT traversal, allowing clusters behind NAT/firewalls to receive inbound traffic via Miren Cloud's PoP infrastructure
- **Netcheck**: Adds public reachability detection via cloud endpoint, caching results and using them to build accurate API address lists in status reports. Hardcodes port 8443 and validates `c.Address` contains a parseable IP before including it
- **HTTP ingress server**: New ingress handler that routes incoming HTTP requests from the global router to the appropriate app sandboxes

## Test plan

- [x] Verify netcheck correctly determines public reachability and caches results
- [x] Verify `apiAddresses()` excludes `c.Address` when it lacks a valid IP host (e.g. `:8443`)
- [x] Verify global router WebSocket tunnel connects to PoP and handles reconnection
- [x] Verify HTTP ingress correctly routes requests through the tunnel
- [x] Run `pkg/cloudauth` and `pkg/globalrouter` unit tests


pkg/globalrouter/client.go
```diff
@@ -0,0 +1,214 @@
...
+		c.log.Warn("websocket disconnected, reconnecting",
+			"error", err, "backoff", backoff)
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(backoff):
+		}
+
+		backoff = min(backoff*2, maxBackoff)
```
phinze, line 106: Backoff never resets after a successful connection — if the connection stays up for hours then drops, the next reconnect still uses the last escalated value. Easy fix: reset `backoff = initialBackoff` at the top of the loop after a successful `runOnce` (or after the connection has been up for some minimum duration).

pkg/globalrouter/pop.go
```diff
@@ -0,0 +1,554 @@
...
+// the request to the correct app.
+func (m *POPManager) newForwardingHandler(popXID string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		hostname := r.Host
+		if fwdHost := r.Header.Get("X-Forwarded-Host"); fwdHost != "" {
+			hostname = fwdHost
+		}
+
+		m.log.Debug("forwarding POP request",
+			"pop_xid", popXID,
```
phinze: `strings.Contains(addr, ":")`  will also match a bare IPv6 address without a port, so it wouldn't get `:8443` appended. Probably doesn't matter today if POP addresses are always hostnames or IPv4, but could bite us later. `net.SplitHostPort` with a fallback might be more robust here.

pkg/globalrouter/client.go
```diff
@@ -0,0 +1,214 @@
...
+
+	err = <-errCh
+	cancel()
+	<-errCh
+	return err
+}
+
+func (c *Client) connect(ctx context.Context) (*websocket.Conn, error) {
+	var token string
+	var err error
```
phinze, line 140: The read and write loops both log at `Info` for every message sent/received. Under steady state with time syncs and keepalives flowing, this could get pretty chatty. Might be worth dropping routine messages to `Debug` and keeping `Info` for connection lifecycle events.

servers/httpingress/httpingress.go
```diff
@@ -1058,6 +1058,136 @@ func (h *Server) DoRequest(ctx context.Context, req *httpingress_v1alpha.Interna
...
+// doesn't execute a request — it gives the caller direct access to the
+// sandbox URL for protocols that need custom connection handling (e.g.,
+// WebSocket tunneling).
+//
+// The path parameter is checked against blocked paths (e.g., admin endpoints).
+// If the route requires OIDC authentication, the tunnel is rejected since
+// OIDC flows cannot be performed over tunneled connections.
+func (h *Server) AcquireTunnel(ctx context.Context, hostname, path string) (*TunnelConn, error) {
+	// Block access to admin endpoints
+	if path == "/.well-known/miren/admin" {
```
phinze: This blocks `/.well-known/miren/admin` specifically — should this be a prefix match on `/.well-known/miren/` to also block the OIDC callback path (`/.well-known/miren/oidc/callback`) for tunnels? OIDC can't work over a WebSocket tunnel anyway, so a request to the callback through this path would just be confusing. (Regular HTTP forwarding through the POP handles OIDC fine — this is just the `AcquireTunnel` path.)

phinze (approved):
Claude and Paul here. Boy are we impressed. Well, Paul was impressed. Claude is a robot. But if Claude could be impressed, this would do it.

We went through the full architecture chunk by chunk, cross-referenced against cloud PR mirendev/cloud#110 and RFD-52 to make sure both sides of the wire protocol agree, and even went down a rabbit hole on QUIC NAT traversal standards to sanity check our approach (tl;dr: we're building exactly the right thing for serving web traffic through edge proxies).

Paul was kind of gushing during the review and Claude didn't stop him. The layering is really clean and the details are 🤌 — sharing the UDP socket across both QUIC connections so source IP matches for phase 2 auth, the `http3.Transport.Dial` override to reuse an already-dialed connection, keepalive-to-ping translation in the tunnel bridge. And sneaking simplified NTP into the initial WebSocket handshake! That's the kind of "oh of course that should be there" foresight that you don't appreciate until you're debugging clock drift across distributed systems at 2am. Mad props.

We filed a few follow-up tickets for things that came up during review: coordinator decomposition, e2e test suite for the full runtime+cloud+POP path, and a reminder to converge other runtime↔cloud communication onto the WebSocket channel once this graduates from labs (we think it's destined to become *the* connection between runtime and cloud).

A few small things inline:

servers/httpingress/httpingress.go
evanphx: Oh good call. Yeah.

pkg/globalrouter/client.go
evanphx: Oh doh! thanks!

pkg/globalrouter/client.go
evanphx: fixed!

PR #688, opened 2026-03-19
Harden PostgreSQL addon for production readiness (MIR-837)

evanphx:
## Summary

- **Compensation on failed provisioning**: If `provider.Provision()` succeeds but a subsequent step fails (attrs, env var injection, version creation, or status update), `provider.Deprovision()` is now called to clean up orphaned resources before setting error status
- **Harden deprovision env var removal**: `removeEnvVars` failure now blocks the deprovision and returns an error so the controller retries, rather than logging a warning and deleting the association with stale `DATABASE_URL` on the app
- **Add `LocalityMode` to provider interface**: New `LocalityMode` type (`on_cluster`/`remote`) added to `AddonProvider` interface, stored on the addon entity at registration time via new `locality_mode` schema field
- **Addon removal reconciliation in app.toml**: `provisionAddons()` now reconciles removals — when the addons section is explicitly present, existing associations not declared in app.toml are deleted. A nil/absent addons section does not trigger removal

## Test plan

- [x] `hack/it controllers/addon` — all 11 tests pass including 2 new tests:
  - `TestProvisionCompensatesOnPostProvisionFailure` — verifies Deprovision is called when post-provision steps fail
  - `TestDeprovisionBlockedByEnvVarRemovalFailure` — verifies env var removal failure prevents association deletion
- [x] `hack/it pkg/addon/...` — all 40 tests pass (registry + postgresql provider)
- [x] `make lint` — 0 issues


pkg/addon/postgresql/shared.go
```diff
@@ -277,23 +316,36 @@ func FindOrCreateSharedServer(ctx context.Context, in FindOrCreateSharedServerIn
 		case "active":
 			serviceHost, err := fw.GetServiceAddress(ctx, server.Service)
 			if err != nil {
-				return FindOrCreateSharedServerOut{}, fmt.Errorf("resolving existing shared service address: %w", err)
+				// Server entity exists but its infrastructure is gone or
+				// unreachable. Remove everything (pool, service, entity) so
+				// the nested saga creates fresh infrastructure with a new
+				// password matching the new data directory.
+				fw.Log.Warn("shared server active but service unreachable, cleaning up stale server",
```
phinze: `GetServiceAddress` can fail for two reasons: the service entity doesn't exist (`cond.ErrNotFound`), or the service has no IP assigned yet (line 193 of `framework.go`). The "no IP" case could be transient — e.g. during initial service setup — but this path tears down the entire shared server, which may be serving other apps.

Could we gate the cleanup on `cond.ErrNotFound` specifically and return the error for anything else so the controller retries?

```go
if errors.Is(err, cond.ErrNotFound{}) {
    fw.Log.Warn("shared server active but service gone, cleaning up", ...)
    if delErr := cleanupStaleSharedServer(fw, ctx, &server); delErr != nil {
        return ..., fmt.Errorf("cleaning up stale shared server: %w", delErr)
    }
    break
}
return ..., fmt.Errorf("resolving shared service address: %w", err)
```

pkg/addon/postgresql/shared.go
```diff
@@ -277,23 +316,36 @@ func FindOrCreateSharedServer(ctx context.Context, in FindOrCreateSharedServerIn
...
 			return FindOrCreateSharedServerOut{
 				ServerID:          server.ID,
 				SuperuserPassword: server.SuperuserPassword,
 				ServiceHost:       serviceHost,
 			}, nil
 		default:
-			return FindOrCreateSharedServerOut{}, fmt.Errorf("shared server exists but has status %q; retry later", server.Status)
+			// Server stuck in a non-terminal state (e.g. "provisioning" from
+			// a previous failed attempt). Remove everything so we start fresh.
+			fw.Log.Warn("shared server in unexpected state, cleaning up stale server",
```
phinze: A concurrent provision request could legitimately find the shared server in "provisioning" state — another request's `EnsureSharedServerSaga` may be mid-flight. Cleaning it up here would tear down that in-progress work.

Would it make sense to return a retryable error for "provisioning" (let the controller try again later when it's hopefully "active") and only clean up states that are genuinely stuck? Even a simple age check on the entity would help distinguish "provisioning right now" from "stuck in provisioning for 30 minutes."

phinze (changes requested):
Claude and I walked through the addon system end-to-end and dug into this PR. Exciting to see this near the finish line! Close-at-hand databases is gonna be a killer feature, and the shared model is a really pragmatic story for the kind of users we're targeting.

**Main question**: This flips the addons feature flag from `default: false` to `default: true` (`features.yaml` + `labs.gen.go`), graduating it out of labs. The PR description doesn't mention this — was that intentional? If we are graduating, two things we'd want:
1. **Docs** — users need to know how to use addons (`app.toml` syntax, CLI commands, what env vars get injected, variants, etc.). "How do I back this up?" should be a key thing we cover.
2. **Shared PG story** — we don't need to solve these for v1, but worth a quick "yep, fine for now" before graduating:
   - **Observability**: The shared PG pool isn't tied to any app, so if one app goes ham on the database, figuring out (a) that it's wheezing and (b) who's being mean to it could be a chore that we're not making any easier yet.
   - **Backups**: What's the story? Users will ask.

If the flag flip was just for testing convenience, we could split it out and graduate separately — and all of the above can wait.

Marking request-changes just so we can discuss the graduation question — that's the only blocker here, the rest is looking solid!

The hardening work itself is great — compensation on failed provisioning, disk cleanup, stale server recovery, and the blackbox tests all hit real production gaps. Inline comments on the stale server cleanup paths are the other things to talk through.

**Nits (not blocking):**

- **Compensation retry**: When compensation deprovision fails during provisioning (`controller.go:115-118`), the association goes to terminal `error`, so leaked infrastructure won't get retried. Future refinement could distinguish "error, cleaned up fine" from "error, stuff leaked" so the controller retries compensation. CodeRabbit flagged this one too.
- **String-based not-found check** in `delete.go:103`: `strings.Contains(err.Error(), "does not exist")` is fragile vs. the `errors.Is(err, cond.ErrNotFound{})` pattern used everywhere else. Root cause is `entity/store.go:1161` using a bare `errors.New`.

pkg/addon/postgresql/shared.go
evanphx: Fixed!

pkg/addon/postgresql/shared.go
evanphx: Fixed!

evanphx:
@phinze A good catch, switch it to on was the expectation! i've beefed up the docs to show how to use addons and the caveats it has today.

phinze (approved):
Oho we're going for it! Docs look good — backup story, shared variant caveats, upgrade path, all covered. Stale server cleanup is properly gated now, compensation is retryable, the nits are all cleaned up. Let's ship this puppy! 🐶📦

PR #693, opened 2026-03-24
Add CLI alias support

evanphx:
## Summary
- Add `[aliases]` section to `.miren/app.toml` for defining custom CLI command shortcuts
- Aliases expand before dispatch, support multi-word names (e.g. `"x console"`), and append extra user arguments
- `miren help alias` shows usage documentation with colored output and lists configured aliases
- Alias names are validated and must not shadow built-in commands

## Example

```toml
# .miren/app.toml
[aliases]
console = "app run bin/rails console"
tail = "logs app -f"
"x console" = "app run bin/rails console"
```

Then `miren console`, `miren tail -n 50`, and `miren x console` all work.

## Test plan
- [x] AppConfig parsing and validation tests for aliases (valid names, invalid names, empty targets)
- [x] Tokenizer tests (quoted strings, escapes, edge cases)
- [x] Alias expansion tests (no config, built-in command priority, no args)
- [x] Linter passes

### DIFF
--- a/cli/alias.go
+++ b/cli/alias.go
@@ -1,0 +1,120 @@
+package cli
+
+import (
+	"fmt"
+	"strings"
+
+	"miren.dev/mflags"
+	"miren.dev/runtime/appconfig"
+)
+
+// expandAlias checks if the given args match a configured alias and expands it.
+// It tries progressively longer prefixes (longest match wins).
+// Returns an error if an alias name conflicts with a built-in command.
+func expandAlias(d *mflags.Dispatcher, args []string) ([]string, error) {
+	if len(args) == 0 {
+		return args, nil
+	}
+
+	ac, err := appconfig.LoadAppConfig()
+	if err != nil || ac == nil || len(ac.Aliases) == 0 {
+		return args, nil
+	}
+
+	// Try progressively longer prefixes, longest match wins
+	var bestMatch string
+	var bestLen int
+
+	for name := range ac.Aliases {
+		words := strings.Fields(name)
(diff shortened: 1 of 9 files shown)


cli/alias.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+
+		if escaped {
+			current = append(current, c)
+			escaped = false
+			continue
+		}
+
+		if c == '\\' && inDouble {
```
phinze, line 85: `\` only escapes inside double quotes — bare `\` outside quotes is literal. Makes sense for these simple command strings, just flagging in case it comes up.

phinze (approved):
Cool stuff! Our hunch is aliases will eventually come from client config too (not just app.toml), and it's nice to have a way to see what's active, so we added `miren alias list`:

```
NAME            COMMAND                        SOURCE
console         app run bin/rails console      .miren/app.toml:4
rails console   app exec -i bin/rails console  .miren/app.toml:6
tail            logs app -f                    .miren/app.toml:5
```

Also noticed the "built-in command not expanded" test was returning early at the no-config check and never hitting the shadow path — went ahead and fixed that up too. Filed MIR-874 for a broader sweep on nice config parse errors (e.g. `help alias` silently swallows parse failures). A couple other small things below, all optional.

<3 P 🤓 & C 🤖

cli/alias.go
evanphx: k!

PR #694, opened 2026-03-24
Add simple disk undelete with soft-delete and background GC

evanphx:
## Summary

- When a disk is deleted, the volume directory is now **moved** to a `deleted-volumes/` holding area instead of being immediately removed
- A background **GC controller** purges soft-deleted volumes after 7 days (configurable), checking hourly
- New **`disk undelete --name <name>`** CLI command restores a deleted disk by recreating entities and moving the volume back
- New **`disk list-deleted`** CLI command shows recoverable disks with expiration times

Closes MIR-847

## Details

**Soft-delete mechanism:**
- `deleteVolume()` calls `softDeleteVolume()` which moves the directory via `os.Rename` (atomic, same filesystem)
- Falls back to hard delete if the move fails — deletion is never blocked
- Writes `metadata.json` alongside the volume with disk name, size, filesystem, volume mode, etc.
- Handles destination collisions by appending a timestamp suffix

**Undelete safety:**
- Verifies `disk.img` exists before recreating entities
- Checks for name conflicts (errors if a disk with the same name already exists)
- Cleans up entities on partial failure at every step

**Files changed:**
- `components/diskio/disk_ops.go` — `MoveVolumeDir` added to interface
- `components/diskio/disk_volume_controller.go` — soft-delete logic
- `components/diskio/deleted_volume.go` — metadata type and helpers
- `components/diskio/deleted_volume_gc.go` — GC controller
- `components/runner/runner.go` — GC wiring
- `cli/commands/disk_undelete.go` — undelete command
- `cli/commands/disk_list_deleted.go` — list-deleted command
- `cli/commands/commands.go` — command registration

## Test plan

- [x] All 3721 existing tests pass (`make test`)
- [x] New unit tests for metadata save/load/list
- [x] New unit tests for GC (purge expired, retain recent, empty dir)
- [x] Updated existing controller tests to verify soft-delete (move) instead of hard delete
- [x] Blackbox end-to-end test: create → delete → list-deleted → undelete → lease → verify mounted
- [x] `golangci-lint` clean

### DIFF
--- a/blackbox/disk_undelete_test.go
+++ b/blackbox/disk_undelete_test.go
@@ -1,0 +1,227 @@
+//go:build blackbox
+
+package blackbox
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"miren.dev/runtime/blackbox/harness"
+)
+
+func TestDiskUndelete(t *testing.T) {
+	c := harness.NewCluster(t)
+	m := harness.NewMiren(t, c)
+
+	diskName := harness.UniqueAppName(t, "undel-disk")
+
+	// Step 1: Create a test disk
+	t.Log("Creating test disk...")
+	r := m.MustRun("debug", "disk", "create", "-n", diskName, "-s", "1")
+	r.RequireContains(t, diskName)
+
+	// Step 2: Wait for disk to be provisioned
+	t.Log("Waiting for disk to be provisioned...")
+	harness.Poll(t, "disk provisioned", 60*time.Second, 2*time.Second,
+		func() (bool, string) {
+			r := m.Run("debug", "disk", "list")
+			if !r.Success() {
+				return false, "debug disk list failed"
+			}
+			if r.OutputContains(diskName) && r.OutputContains("provisioned") {
+				return true, ""
+			}
+			return false, "disk not yet provisioned"
+		},
+	)
+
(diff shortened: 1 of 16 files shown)


phinze (approved):
Claude and I went through this end-to-end — the design is solid. Soft-delete with fallback to hard delete is the right call (deletion never blocked), the GC is clean, and the blackbox test covers the full lifecycle nicely.

We contributed a commit (3bbbd7b7) that tightens up a few edges:

- **Soft-delete atomicity**: metadata is now written before the rename so the move carries both data and metadata together. Previously a metadata write failure would leave an invisible orphaned directory (not listable, not GC-able).
- **Undelete rollback**: uses a deferred rollback pattern — if any step after the directory move fails, the volume gets moved back to `deleted-volumes/` with its metadata intact. Previously partial failures could leave data orphaned in `volumes/` with no entities pointing at it.
- **GC start ordering**: moved `diskGC.Start()` after all controller `Init()` calls so the goroutine can't be orphaned if a later init fails.
- **`disk list-deleted --format json`**: added per CLI conventions.
- **Blackbox test**: added `RequireSuccess` on the post-undelete `list-deleted` call.

CodeRabbit flagged several of the same things — we agreed on the metadata ordering and GC placement findings, and went a bit further on the undelete rollback.

PR #696, opened 2026-03-26
Implement short entity IDs (RFD-69)

phinze:
Entity IDs in Miren are long Base58-encoded UUID v7 values — something like `blog-vCZ1eUgSgNd28ed6vt2DgY`. They push useful columns off-screen in CLI tables and are impossible to type from memory. RFD-69 introduces `db/short-id` to fix this.

Every entity with an `entity/kind` now gets a short (3–8 character) identifier allocated at creation time. The algorithm extracts the base58 suffix from the entity ID and takes the last 3 characters, growing on collision. Named entities without a base58 portion (like `app/blog`) get a random base58 short-id instead. The max length is capped at 8 characters — beyond that it stops being "short."

Uniqueness is enforced via an etcd ref index at `{prefix}/refs/{shortId}`. The create transaction checks both the entity key and the ref key atomically, so a concurrent create can't silently steal another entity's short-id. If a race does occur, the transaction detects it and retries with a freshly allocated short-id (up to 3 times).

The entity server's `Get()` now falls back to ref resolution when a direct ID lookup fails, so `m app inspect DgY` resolves to the right entity automatically. On the display side, `BriefId()` returns the short-id when available, and sandbox list uses it by default (full IDs with `-v`).

A boot-time migration backfills short-ids for all existing entities that don't have one, running after the existing entity format migration and before reindexing. It's idempotent — subsequent boots skip entities that already have a short-id.

MIR-864

### DIFF
--- a/pkg/entity/shortid.go
+++ b/pkg/entity/shortid.go
@@ -1,0 +1,142 @@
+package entity
+
+import (
+	"crypto/rand"
+	"strings"
+)
+
+const base58Alphabet = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+
+// minBase58Len is the minimum length of a base58 segment to be considered
+// high-entropy (i.e., derived from a UUID rather than a human-chosen name).
+const minBase58Len = 10
+
+// defaultShortIdLen is the initial length of a short-id candidate.
+const defaultShortIdLen = 3
+
+// maxShortIdLen caps how long a short-id can grow before we give up.
+// At length 8, there are 58^8 ≈ 128 trillion possibilities.
+const maxShortIdLen = 8
+
+// knownPrefixes are 1-2 character prefixes that precede the base58 portion
+// in entity IDs of Category 2 (name + base58 suffix).
+var knownPrefixes = []string{"ob", "v", "a", "s", "t"}
+
+// isBase58 reports whether every character in s belongs to the Base58 alphabet.
+func isBase58(s string) bool {
+	for i := 0; i < len(s); i++ {
+		if !strings.ContainsRune(base58Alphabet, rune(s[i])) {
+			return false
+		}
+	}
+	return len(s) > 0
+}
+
+// ExtractBase58Suffix extracts the high-entropy base58 portion from an entity ID.
+// Returns the base58 portion and true if found, or empty string and false otherwise.
+//
+// Handles all entity ID categories:
+//   - Category 1 (named): "app/blog" → ("", false)
+//   - Category 2 (name+prefix+base58): "blog-vCZ1eUgSgNd28ed6vt2DgY" → ("CZ1eUgSgNd28ed6vt2DgY", true)
+//   - Category 3 (namespace/prefix-base58): "sandbox/blog-web-CZAtBvhsMNbG38MceikkB" → ("CZAtBvhsMNbG38MceikkB", true)
+//   - Category 4 (kind-base58): "deployment-CZ1eUgSgNd28ed6vt2DgY" → ("CZ1eUgSgNd28ed6vt2DgY", true)
+func ExtractBase58Suffix(entityId string) (string, bool) {
+	id := entityId
+
+	// Strip namespace prefix (everything before and including the last "/")
+	if idx := strings.LastIndex(id, "/"); idx >= 0 {
+		id = id[idx+1:]
+	}
+
+	// Split on "-" and take the last segment
+	if idx := strings.LastIndex(id, "-"); idx >= 0 {
+		id = id[idx+1:]
+	} else {
+		// No "-" at all — this is a purely named entity (Category 1)
+		return "", false
+	}
+
+	// Try stripping known prefixes
+	base58Part := id
+	for _, prefix := range knownPrefixes {
+		if strings.HasPrefix(id, prefix) {
+			candidate := id[len(prefix):]
+			if isBase58(candidate) && len(candidate) >= minBase58Len {
+				return candidate, true
+			}
+		}
+	}
+
+	// Check if the whole segment is base58
+	if isBase58(base58Part) && len(base58Part) >= minBase58Len {
+		return base58Part, true
+	}
+
+	return "", false
+}
+
+// ExistsCheck is a function that reports whether a given short-id candidate
+// is already in use.
+type ExistsCheck func(candidate string) (bool, error)
+
+// AllocateShortId generates a short, globally-unique identifier for an entity.
+// It first tries to derive the short-id from the entity's base58 suffix,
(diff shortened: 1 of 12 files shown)


pkg/entity/shortid.go
```diff
@@ -0,0 +1,142 @@
...
+			taken, err := exists(candidate)
+			if err != nil {
+				return "", err
+			}
+			if !taken {
+				return candidate, nil
+			}
+		}
+	}
+	return "", ErrShortIdExhausted
```
evanphx, line 122: I pity us hitting this.

pkg/entity/shortid_migrate.go
```diff
@@ -0,0 +1,148 @@
...
+
+		// Write the updated entity and ref index entry atomically
+		refKey := refsPrefix + shortId
+		txnResp, txnErr := client.Txn(ctx).
+			If(clientv3.Compare(clientv3.CreateRevision(refKey), "=", 0)).
+			Then(
+				clientv3.OpPut(key, string(newData)),
+				clientv3.OpPut(refKey, entityId),
+			).
+			Commit()
```
evanphx: Any chance we could batch these so that we do like 10 or more in one Txn?

servers/entityserver/entityserver.go
```diff
@@ -59,17 +59,25 @@ func (e *EntityServer) Get(ctx context.Context, req *entityserver_v1alpha.Entity
...
-	entity, err := e.Store.GetEntity(ctx, entity.Id(args.Id()))
+	ent, err := e.Store.GetEntity(ctx, entity.Id(args.Id()))
 	if err != nil {
-		return cond.NotFound("entity", args.Id())
+		// Try resolving as a ref (short-id, name, etc.)
+		if etcdStore, ok := e.Store.(*entity.EtcdStore); ok {
+			if resolvedId, refErr := etcdStore.ResolveRef(ctx, args.Id()); refErr == nil {
+				ent, err = e.Store.GetEntity(ctx, resolvedId)
+			}
+		}
```
evanphx, line 69: Is there a quality of implementing this with the ResolveRef special case that we get rather than querying the db/short-id index directly?

pkg/entity/shortid.go
```diff
@@ -0,0 +1,142 @@
...
+func randomBase58(length int) (string, error) {
+	b := make([]byte, length)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	for i := range b {
+		b[i] = base58Alphabet[int(b[i])%len(base58Alphabet)]
+	}
+	return string(b), nil
+}
```
evanphx, line 138: Maybe leave a comment here that the resulting string is not a "true" base58 encoding of the bytes, such that you can't round-trip back to the original bytes. That's ok since we're just generating a random sequence, but don't want to trip up a future programmer assuming it was.

evanphx (approved):
A couple notes, mostly I'm curious about the ResolveRef rather than querying the db/short-id index directly.

evanphx (approved):
A couple notes, mostly I'm curious about the ResolveRef rather than querying the db/short-id index directly.

servers/entityserver/entityserver.go
phinze: Ok lol sort of came full circle on this one. Reusing the collection indexes seemed great — until we realized they can't enforce uniqueness atomically. The collection key is `{prefix}/collections/{cas}/{entityId}`, so you can enumerate all entities with a value but you can't stick a `CreateRevision = 0` condition on it in an etcd `If` to assert "nobody else has this value."

So the plan is now:

1. **Resolution** uses a new `GetOneIndex` on the `db/short-id` collection index (your suggestion — no special-case ref lookup needed)
2. **Uniqueness enforcement** uses a dedicated key checked atomically in the create transaction — same shape as the ref index, but the motivation is closing the general `UniqueValue` enforcement gap rather than short-id-specific plumbing

Discussing whether to generalize the uniqueness enforcement to all `UniqueValue` attributes in this PR or keep it scoped to short-ids and follow up.

phinze: Prolly just gonna take the swing and fix UniqueValue generally, which is basically this ref thing but for any UniqueValue.

pkg/entity/shortid.go
phinze: Ha — yeah, if we hit this something has gone deeply sideways. 58^8 ≈ 128 trillion possibilities at that point.

pkg/entity/shortid_migrate.go
phinze: Good call — going to leave this as a follow-up. Each entity currently needs its own CAS check (ModRevision guard + unique key creation), so batching means building multi-entity transactions with per-entity conditions. Doable but adds complexity, and the migration is a one-time boot operation that'll be fast enough for the entity counts we're dealing with.

pkg/entity/shortid.go
phinze: Added — good catch, that would definitely trip someone up.

pkg/entity/store.go
```diff
@@ -224,12 +224,26 @@ func (s *EtcdStore) CreateEntity(
...
+	if err != nil {
+		return nil, err
+	}
+
+	var uniqueConditions []clientv3.Cmp
+	for _, attr := range uniqueAttrs {
+		uniqueKey := s.buildUniqueKey(attr)
+		uniqueConditions = append(uniqueConditions, clientv3.Compare(clientv3.CreateRevision(uniqueKey), "=", 0))
+		coltxopt = append(coltxopt, s.addUniqueOp(attr, entity.Id()))
+	}
```
evanphx, line 238: I think we need to something similar over in UpdateEntity, etc in the case that a unique attr is being modified.

evanphx (changes requested):

evanphx (approved):

pkg/entity/store.go
phinze: Yeah misguided "splitting PRs" thing -> ended up going for it

pkg/entity/shortid_migrate.go
phinze: Revisited this with some napkin math. Ran `m debug reindex --dry-run` on our heaviest clusters (neat trick for getting total entity count) — club has 2,645 entities, garden has 2,311, prod has 1,024. At ~2-5ms per etcd transaction that's 5-13 seconds worst case on club, one time on first boot after upgrade (subsequent boots are a no-op).

Batching would bring it under a second, but each entity needs its own ModRevision guard so a batch failure can't tell which entity was concurrently modified — adds retry-one-by-one fallback complexity. The existing `MigrateEntityStore` uses the same one-at-a-time pattern for these same entity counts. We can always revisit if it becomes a concern.

PR #705, opened 2026-03-27
Scope sandbox controller queries to local node in distributed setups

phinze:
We hit a fun one on miren.garden — every app was crash-looping with no errors in app logs or the system journal. Sandboxes would start up perfectly fine (DB init, ports bound, health checks passing) and then just... die. Consistently, about a second after reaching RUNNING.

Turns out a distributed runner (`runner-8a91ac2a` from MIR-890 testing) was the culprit. Three methods in the sandbox controller — `reconcileSandboxesOnBoot`, `CleanupOrphanedContainers`, and `Periodic` — were listing *all* sandboxes system-wide and then checking them against the local containerd. The runner's normal watch loop was already properly scoped to its own node via `compute.Index`, but these three paths bypassed that filter.

So every time the remote runner connected (25 times in the window we checked), it would run boot reconciliation, see the coordinator's sandboxes, fail to find their containers locally, and dutifully patch them all to DEAD. The pool manager would spin up replacements, and the runner would kill those too. A nice little murder loop.

The fix scopes all three methods to use the same `compute.Index(KindSandbox, "node/"+nodeId)` query the watch loop already uses. Each runner now only touches sandboxes that are actually scheduled to it.

Closes MIR-921

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -280,8 +280,11 @@ func (c *SandboxController) reconcileSandboxesOnBoot(ctx context.Context) error 
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
 	defer cancel()
 
-	// List all sandboxes
-	resp, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, compute.KindSandbox))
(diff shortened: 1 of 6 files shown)


evanphx (approved):
Image

PR #707, opened 2026-03-28
Fix overlay IP allocator assigning duplicate IPs after restart

evanphx:
## Summary

Fixes MIR-912: After server restarts, the IP allocator could assign the same IP to two different running sandboxes, causing cross-app traffic routing (security issue).

Three root causes addressed:

- **`SetupOnBridge` ignored provided prefixes**: When `AllocateNetwork` was called for a sandbox with pre-existing IPs (recovered sandbox path), `SetupOnBridge` discarded them and allocated fresh IPs. Now it calls `ReserveSpecificAddr()` to re-reserve the provided addresses.
- **No IP re-registration at boot**: `reconcileSandboxesOnBoot` now calls `ReserveSpecificAddr` for healthy surviving sandbox IPs, ensuring netdb tracks them even if its SQLite state was lost during a crash.
- **Flannel subnet not persisted for `PrevIPv4`**: The coordinator never told flannel which subnet it previously held, risking a different /24 on restart. The leased subnet is now persisted in netdb's metadata table and passed as `PrevIPv4` on startup.

**Upgrade path**: Existing clusters seamlessly pick up the new `metadata` table via `CREATE TABLE IF NOT EXISTS`. The first boot after upgrade behaves identically to today (no `PrevIPv4` yet); the subnet is persisted on that boot and used from the next restart onward.

## Test plan

- [x] New unit tests for `ReserveSpecificAddr` (fresh, idempotent, released, out-of-subnet)
- [x] New unit test for `GetLeasedSubnet`/`SetLeasedSubnet` (persist, update, survive reopen)
- [x] Frozen file hash updated after verifying saga controller delegates to same code paths
- [x] Full test suite passes (3751 tests, 0 failures)
- [ ] Deploy to staging and verify surviving sandboxes retain their IPs across restart
- [ ] Verify new sandbox gets a different IP than surviving ones after restart

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -674,11 +674,16 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		return fmt.Errorf("failed to open netdb: %w", err)
 	}
 
+	prevSubnet, err := ndb.GetLeasedSubnet()
+	if err != nil {
+		return fmt.Errorf("failed to read leased subnet from netdb: %w", err)
+	}
+
 	grungeOpts := grunge.NetworkOptions{
 		EtcdEndpoints: cfg.Etcd.Endpoints,
 		EtcdPrefix:    cfg.Etcd.GetPrefix() + "/sub/flannel",
 		BackendType:   cfg.Server.GetNetworkBackend(),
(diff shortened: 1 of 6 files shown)


phinze (approved):
Nice finds -> nice fixes 😎 👍

PR #711, opened 2026-03-30
Fix slice aliasing bug in buildEntitySaveOps that drops entity attrs on overwrite

phinze:
Garden went down right after the short-id deploy landed. The server was crash-looping with `panic: attribute schema in cache has empty type` — every boot would run `schema.Apply`, overwrite the existing schema attribute entities, and silently lose their `db/type` attr in the process. Then `GetAttributeSchema` would fail and the coordinator couldn't even create the default project.

The root cause is a Go slice aliasing footgun in `buildEntitySaveOps`. It sets `entity.attrs = primary` (sharing the backing array), then `SetUpdatedAt` appends and `SortedAttrs` re-sorts in-place. Any attr that sorts after `db/entity.updated` — like `db/type` — gets shuffled past `primary`'s length. When `CreateEntity`'s retry loop reuses `primary` on the overwrite path, those attrs are gone. This was latent before #696 because `buildEntitySaveOps` was only called once; the retry loop made `primary` reusable across attempts.

The fix is `slices.Clone(primary)` to break the alias. Also improved the panic message in `GetAttributeSchema` to include the schema ID for easier debugging if we ever hit it again.

Regression test included — confirmed it fails without the fix and passes with it.

Closes MIR-931

### DIFF
--- a/pkg/entity/store.go
+++ b/pkg/entity/store.go
@@ -959,7 +959,7 @@ func (s *EtcdStore) buildUniqueUpdateOps(ctx context.Context, entityId Id, oldEn
 func (s *EtcdStore) buildEntitySaveOps(entity *Entity, key string, primary, session []Attr, o *entityOpts) ([]clientv3.Op, error) {
 	var ops []clientv3.Op
 
-	entity.attrs = primary
+	entity.attrs = slices.Clone(primary)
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Image

A non-trivial part of my brain wants to try to write something to detect this most common of footguns.

PR #722, opened 2026-04-02
Fix local disks not mounting for web service sandboxes

phinze:
Apps like uptime-kuma that define `[[services.web.disks]]` with `provider = "local"` were crashing because their disk mounts never showed up in the sandbox. The sandbox was also getting scheduled as stateless since it had no volumes.

The root cause was in `buildSandboxSpec` — there's a guard that skips miren disk attachment when the service isn't using fixed concurrency mode (since miren disks need instance affinity). The problem was that guard used `break` to exit the *entire* disk loop, which meant local disks got thrown out with the bathwater. Web services default to auto concurrency, so any web service with local disks hit this path every time.

The fix scopes the guard to only skip miren disks via a flag + `continue`, letting local disks through regardless of concurrency mode. Two new tests cover the exact MIR-950 scenario (local disks on an auto-scaling web service) and confirm miren disks are still correctly skipped when they should be.

Closes MIR-950

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -633,7 +633,8 @@ func (l *Launcher) buildSandboxSpec(
 	// Add disk volumes and mounts for this service
 	for _, svc := range cfgSpec.Services {
 		if svc.Name == serviceName {
-			// Check concurrency constraints for miren disk volumes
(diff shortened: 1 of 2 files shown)


maryelizbeth (approved):
lgtm

PR #724, opened 2026-04-02
Run blackbox tests in parallel

phinze:
The blackbox test suite has been the critical path of our CI pipeline for a while now — at ~400s wall-clock, it takes about 2.5x longer than any individual test runner group. We dug into the CI data from the last week and confirmed this is consistently the bottleneck, with 274s of serial test execution plus 44s of dev environment setup.

Looking at the test code, every test already does the right thing for parallelism: `UniqueAppName` generates collision-free names using a per-process random ID, all assertions filter by the test's own app name, routes use unique hostnames, and cleanup is registered via `t.Cleanup`. There's no shared mutable state between tests — they were just never marked parallel.

This adds `t.Parallel()` to all 19 blackbox tests. The miren server should be resilient to concurrent interactions, and if we surface any race conditions this way, that's a valuable signal worth investigating.

Expected impact: the serial test time of 274s should collapse to roughly the duration of the slowest individual test (~38s) plus contention overhead. Even if concurrent resource pressure doubles individual test times, we'd be looking at ~80s of test time vs 274s — a solid improvement on the CI critical path.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -245,7 +245,7 @@ jobs:
       run: make dev-server-start
 
     - name: Run blackbox tests
-      run: go test -tags blackbox -timeout 10m -v -count=1 -p 1 ./blackbox/...
+      run: go test -tags blackbox -timeout 15m -v -count=1 -p 1 ./blackbox/...
 
     - name: Server logs on failure
       if: failure()
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -206,7 +206,7 @@ To get started with iso:
 
 ### Testing Notes
 
-- **Unit tests** run with `-p 1` (one package at a time) because runners within an iso container share containerd/buildkit instances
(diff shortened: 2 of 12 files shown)


phinze:
Moving back to draft — we found a real concurrency bug while testing this.

When two tests deploy the same source (`testdata/go-server`) concurrently, they produce the same image digest. Both deploys race to register an artifact entity for that digest, creating duplicates. Subsequent lookups fail with:

```
error locating artifact by digest sha256:...: conflict in entity: more than one entity found
```

This isn't a test isolation issue — it's a server bug where artifact entity registration isn't idempotent for content-addressed digests. Filed as MIR-966.

Once that's fixed, this PR should be ready to go. The test harness isolation is solid (unique app names, dynamic ports, proper cleanup) — it's just the artifact store that can't handle concurrent identical builds.

phinze:
Closing this until we eventually swing around to MIR-966 and/or get mad about CI runtime enough to bang on this.

PR #725, opened 2026-04-02
Drain old pools before starting new ones for services with disks

phinze:
We hit this with victoriametrics on the Garden cluster — deploying a new version would spin up a new sandbox that immediately panicked trying to acquire an flock on a local disk the old sandbox still held. The new sandbox would crash-loop until the old one eventually got cleaned up, at which point it'd finally start. Not great.

The root cause is that `reconcileAppVersion` uses a rolling strategy: create new pool, wait for ready, then clean up old pools. That's the right default for stateless services (avoids 502s during rollover), but for anything with disk mounts it's a non-starter — you can't have two processes holding the same exclusive disk resource.

The fix checks whether a service has disks configured before entering the pool creation loop. If it does, stale pools (ones whose spec doesn't match the new version) get scaled to zero and drained first, so the disk is released before the new sandbox tries to mount it. Services without disks continue using the normal rolling strategy.

This is forward-compatible with the saga-based deployment choreography we're planning (RFD-65) — `drain_old` is just another primitive that'll slot into the step sequence.

Closes MIR-957

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -216,6 +216,20 @@ func (l *Launcher) reconcileAppVersion(ctx context.Context, app *core_v1alpha.Ap
 	// For each service, ensure a pool exists. Collect IDs of newly created pools.
 	var newPoolIDs []entity.Id
 	for _, svc := range spec.Services {
+		// For services with disks, drain old pools before creating new ones.
+		// Disks require exclusive access (local disks use flock, miren disks
+		// use leases), so the old sandbox must release the disk before the
+		// new one can mount it.
+		if serviceHasDisks(spec, svc.Name) {
+			if err := l.drainStaleDiskPools(ctx, app, &ver, spec, svc.Name); err != nil {
+				l.Log.Error("failed to drain old disk pools, skipping service",
+					"app", app.ID,
+					"service", svc.Name,
+					"error", err)
+				continue
+			}
+		}
+
 		poolID, err := l.ensurePoolForService(ctx, app, &ver, spec, svc.Name)
 		if err != nil {
 			l.Log.Error("failed to ensure pool for service",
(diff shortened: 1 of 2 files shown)


controllers/deployment/launcher.go
```diff
@@ -1474,3 +1410,195 @@ func (l *Launcher) cleanupStaleServices(ctx context.Context, app *core_v1alpha.A
...
+			return fmt.Errorf("update pool %s: %w", pool.ID, err)
+		}
+	}
+
+	// Wait for all stale pools to fully drain so disk leases are released.
+	for _, pwe := range stalePools {
+		if err := l.waitForPoolDrained(ctx, pwe.Pool.ID, pwe.Pool.Service, l.PoolReadyTimeout); err != nil {
+			return fmt.Errorf("waiting for pool %s to drain: %w", pwe.Pool.ID, err)
+		}
+	}
```
evanphx, line 1562: we have other places that do similar blocks, I wonder if we shouldn't document them and/or start to come up with mechanisms to allow a workload to wait on a condition but not tie up the worker...

evanphx (approved):

controllers/deployment/launcher.go
phinze: Yeah thinking a lot about this - I think we finish MIR-441, and get working on the stuff in RFD-65 which will give us lattice for that kinda thing.

controllers/deployment/launcher.go
```diff
@@ -1395,3 +1410,195 @@ func (l *Launcher) cleanupStaleServices(ctx context.Context, app *core_v1alpha.A
...
+
+// serviceHasDisks returns true if the named service in the config spec has any
+// disk mounts configured.
+func serviceHasDisks(spec *core_v1alpha.ConfigSpec, serviceName string) bool {
+	for _, svc := range spec.Services {
+		if svc.Name == serviceName {
+			return len(svc.Disks) > 0
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
phinze: Intentional — `serviceHasDisks` only checks explicit disk config. The auto-add path is a migration convenience for apps that predate explicit disk configuration; those apps aren't using flock on their local data. If they are, they should add an explicit `[[disks]]` entry. There's a deprecation warning for the auto-add behavior coming in a separate PR.

PR #732, opened 2026-04-06
Fix Entity.Update to handle multi-valued attrs correctly

phinze:
`Entity.Update` was implemented as a loop over `Entity.Set`, which has "upsert exactly one" semantics. That's fine for single-valued attrs, but `many:true` component fields (like `addon_association.variables`) produce multiple attrs with the same ID from `Encode()`. Each call to `Set` just overwrote the first matching slot, so only the last value survived.

The concrete impact was on addon provisioning: an addon with 8 env vars would only persist the last one in its `variables` field. That meant `removeEnvVars` during deprovision would only clean up 1 of N, leaving stale addon env vars on the app permanently.

The fix changes `Update` to "replace-all-for-key": remove all existing attrs for each incoming ID, then append the new set. This is the same semantic the store-level `UpdateEntity` already implements (via remove + merge), just applied at the in-memory entity layer where controllers operate via `meta.Update()`.

The deployment launcher had a workaround for exactly this bug (with a comment explaining it), so that's unwound here. And the blackbox addon tests get their `WaitForEnvVarRemoved` assertions back (removed in the memcache PR pending this fix).

Closes MIR-974

### DIFF
--- a/blackbox/addon_memcache_test.go
+++ b/blackbox/addon_memcache_test.go
@@ -29,12 +29,10 @@ func TestMemcacheAddonCreateListDestroy(t *testing.T) {
 	harness.WaitForEnvVar(t, m, name, "MEMCACHE_HOST", 30*time.Second)
 	harness.WaitForEnvVar(t, m, name, "MEMCACHE_PORT", 30*time.Second)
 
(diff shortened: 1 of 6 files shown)


evanphx (approved):
Image

PR #744, opened 2026-04-09
Wire metrics and logs through distributed runners

phinze:
Distributed runners were crashing on startup because sandbox metrics were created without CPU/memory writers. The metrics monitor would nil-deref on its first tick, crash-looping the runner. The root cause was that `runner start` created a bare `sandbox.NewMetrics()` without wiring up `CPUUsage` or `MemUsage`, so the coordinator had no way to tell the runner where to ship observability data.

The first commit fixes this by extending the join protocol to carry VictoriaMetrics and VictoriaLogs addresses from coordinator to runner. At startup the runner creates real metrics and log writers pointed at those endpoints over flannel. There's also a `Validate()` check on sandbox metrics so future misconfiguration fails at init instead of panicking in a goroutine. A smaller fix in the same commit handles cluster.local resolution when the coordinator address is a hostname, which was preventing runners from pulling images.

The second commit adds local testing tooling built on iso peers. `hack/dev-distributed up` spins up a coordinator + runner in separate containers on a shared network with flannel, VictoriaMetrics, and VictoriaLogs all wired together. The blackbox test harness gets a new `ModePeers` so `make test-blackbox-distributed` runs the entire suite against the distributed topology. Three new tests verify the distributed-specific plumbing: runner list, metrics pipeline, and log pipeline. The full existing blackbox suite passes against the distributed topology (tested locally, addon tests need a longer timeout due to provisioning saga speed).

Closes MIR-991


evanphx (approved):
Looks great! Makes me wonder if we shouldn't eventually have a way to represent the "services configuration" as a singular value that we can pass around

phinze:
Good idea, filed MIR-998 to track consolidating the service endpoints into a single config blob for the join protocol.

PR #745, opened 2026-04-09
Implement RFD-72: Ephemeral deployments

evanphx:
## Summary

Implements [RFD-72: Ephemeral Deployments](https://github.com/mirendev/rfd/blob/main/rfds/0050/0072-ephemeral-deployments.md) — app versions tagged with a label and TTL that are accessed via subdomain routing without affecting the active version.

- **`miren deploy --ephemeral <label> --ttl <duration>`** creates an ephemeral version accessible at `<label>.app.example.com`
- **`miren app versions`** lists all versions with status (active/ephemeral/replaced), supports `--ephemeral` filter and `--format json`
- Ephemeral subdomains work with normal routes — no wildcard route entity required (just wildcard DNS)
- Deploying with an existing label replaces the old version; per-app limit of 10 with nearest-to-expiry eviction
- GC controller automatically cleans up expired versions and their sandbox pools every 5 minutes
- Activator needs zero changes — ingress resolves the ephemeral AppVersion and passes it directly to `AcquireLease`

## Key design decisions

- Ephemeral deploys skip deployment record creation and locking entirely — they're independent of the normal deployment lifecycle
- Ingress detects ephemeral subdomains by stripping the first DNS label and matching the remainder against known routes (`lookupEphemeralRoute`)
- Server-side label validation ensures only DNS-safe labels are stored, even from direct API callers
- `getAccessInfo` resolves ephemeral hostnames server-side so the CLI displays the correct URL

## Test plan

- [x] Unit tests: label normalization (10 cases), validation, manage functions (lookup, replace, enforce limit, delete expired), GC controller (5 tests), deployment server ephemeral path (3 tests), subdomain extraction (7 cases)
- [x] Blackbox test: full e2e — deploy app, deploy ephemeral, verify active version unchanged, verify `app versions` output, verify HTTP activation via subdomain, verify replace-on-same-label
- [x] Lint clean, all existing tests pass
- [ ] CI


controllers/ephemeral/gc.go
```diff
@@ -0,0 +1,168 @@
...
+	result := &GCResult{}
+
+	gcCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+
+	// List all app versions — filter for ephemeral ones
+	// We use the app_version app index to get all versions across all apps
+	// A more efficient approach would use the ephemeral_expires_at index,
+	// but that requires range queries. For now, scan all versions.
+	resp, err := c.EAC.List(gcCtx, entity.Ref(entity.EntityKind, core_v1alpha.KindAppVersion))
```
phinze, line 107: The exorcism sweep scans every AppVersion in the store to find ghosts among the living. Fine while the spirit population is small, but these entities are never reaped, so the séance gets longer with every deploy the cluster has ever seen. We resurrected MIR-329 to track this.

servers/httpingress/httpingress.go
```diff
@@ -623,25 +676,47 @@ func (h *Server) serveAuthenticatedRequest(w http.ResponseWriter, req *http.Requ
...
+			// Resolve ephemeral version by label
+			ephVer, ephErr := ephemeralx.LookupByLabel(ctx, h.eac, targetAppId, ephemeralLabel)
+			if ephErr != nil {
+				h.Log.Error("error looking up ephemeral version", "error", ephErr, "label", ephemeralLabel)
+				http.Error(w, fmt.Sprintf("error looking up ephemeral version: %s", ephemeralLabel), http.StatusInternalServerError)
+				return
+			}
+			if ephVer == nil {
+				h.Log.Debug("no ephemeral version found", "label", ephemeralLabel, "app", targetAppId)
+				http.Error(w, fmt.Sprintf("ephemeral version %q not found or has expired", ephemeralLabel), http.StatusNotFound)
```
phinze, line 691: A ghost that's past its expiry date can still haunt for up to 5 minutes until the next exorcism sweep. `LookupByLabel` doesn't check `EphemeralExpiresAt`, so the apparition remains visible to visitors even after its time is up. Is that intentional? Totally reasonable if so, just want to make sure it's a deliberate choice.

servers/httpingress/httpingress.go
```diff
@@ -496,19 +498,43 @@ func (h *Server) serveHTTPWithMetrics(w http.ResponseWriter, req *http.Request,
...
 			// Wrap the request handler with OIDC middleware
 			oidcWrapped := h.oidcMiddleware(route, func(w http.ResponseWriter, r *http.Request) {
 				// Continue with normal request handling after auth
-				h.serveAuthenticatedRequest(w, r, targetAppId, routeType, appName)
+				h.serveAuthenticatedRequest(w, r, targetAppId, routeType, ephemeralLabel, appName)
+			})
+			oidcWrapped(w, req)
+			return
+		}
+	} else if label, baseRoute, err := h.lookupEphemeralRoute(ctx, onlyHost); err == nil && baseRoute != nil {
```
phinze, line 520: The `lookupEphemeralRoute` fallback is a nice trick: any route can spawn ghosts without a wildcard route entity. Side effect is that existing wildcard routes (`*.app.example.com`) would conflict, since every subdomain gets interpreted as a ghost sighting. Near-zero chance anyone has wildcard routes today, so this is paranormal trivia, not a blocker.

pkg/ephemeral/manage.go
```diff
@@ -0,0 +1,186 @@
...
+	"sort"
+	"time"
+
+	"miren.dev/runtime/api/compute/compute_v1alpha"
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/entity"
+)
+
+const DefaultMaxEphemeral = 10
```
phinze, line 17: This package is part ghost taxonomy (label normalization, pure strings) and part ghost wrangling (entity CRUD with an injected client). Slightly split personality, but not worth restructuring now. MIR-441 will naturally clarify its identity by pulling the orchestration into saga actions.

phinze (approved):
Clegon and Phenkman here. We spent some quality time with these ephemeral apparitions, tracking their movements from manifestation (deploy) through haunting (ingress routing) to exorcism (GC). The spectral containment system is well-built. A few field notes from our investigation.

CodeRabbit's earlier sightings (pool cleanup ordering, TTL validation, error propagation) were all banished in follow-up commits. Our notes focus on phenomena the automated detectors missed.

👻 Who you gonna call? `miren deploy --ephemeral`

servers/httpingress/httpingress.go
evanphx: Nope! fixed.

components/activator/activator.go
```diff
@@ -896,8 +912,38 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
 			return foundPool, nil
 		}
 
-		// Pool not found after retries - DeploymentLauncher should have created it
+		// Pool not found after retries — try on-demand creation if a PoolCreator
+		// is available (used for ephemeral versions that bypass DeploymentLauncher).
 		a.mu.Unlock()
+
+		if a.poolCreator != nil {
```
phinze, line 920: Two concurrent requests to a cold ephemeral URL can both reach this point (both fell out of the `findPoolInStore` retry loop having found nothing), both release the lock here, and both call `CreatePoolForVersion`. Inside that, `findMatchingPool` is list-then-create, so neither sees the other's work, and we end up with duplicate pools for the same (version, service).

The good news is the sentinel infrastructure already lives in `poolState`. `inProgress`, `done`, `err` were built for exactly this and have been dormant since commit `040329dd` moved pool creation out of the activator. Pragmatic fix: under the same `a.mu` we already hold, insert the sentinel, release the lock, do the creation, then close `done` on completion. Callers who arrive mid-creation hit the existing `state.inProgress` check at line 549 and wait on `done`.

MIR-596 proposes the structural cleanup around this.

components/activator/activator.go
```diff
@@ -896,8 +912,38 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
...
+			poolID, err := a.poolCreator.CreatePoolForVersion(ctx, ver, service)
+			if err != nil {
+				return nil, fmt.Errorf("on-demand pool creation failed for version=%s service=%s: %w",
+					ver.Version, service, err)
+			}
+
+			// Pool was created — retry findPoolInStore to populate our cache
+			foundPool, err := a.findPoolInStore(ctx, ver.ID, service)
+			if err != nil || foundPool == nil {
+				return nil, fmt.Errorf("pool created on-demand (id=%s) but not found in store", poolID)
```
phinze: Minor cosmetic: when `ensurePoolForService` reuses an existing matching pool, it returns `entity.Id("")`, so this error would read `"pool created on-demand (id=) but not found in store"`, which is a little confusing. A non-empty check on `poolID` before formatting would cost nothing.

phinze (approved):
Clegon and Phenkman checking back in. We dug through the delta and ended up doing some git archaeology, because the on-demand pool creation turned out to have a richer backstory than the diff suggests.

Most of the delta is tight. The clone-instead-of-mutate fix (`09b799bb`) and the lease-cache scoping are genuinely important, both were silent correctness failures for `--version <active> --ephemeral`. The shared-pool cleanup (`d8bcb91b`) is the right call. The `LookupByLabel` expiry check fully addresses our earlier "apparition past its expiry" note.

The piece Evan flagged for re-review (pool creation on the activation path, `006f0690`) is where we ended up spending most of our time.

The backstory: `bf97ec7d` (Oct 2025) introduced a sentinel pattern in `poolState` specifically to prevent duplicate pool creation under concurrent requests. Evan reviewed and refined it in `b630ea32`, the commit that added the concurrency doc block still sitting at the top of activator.go. `040329dd` (RFD-39 Phase 4) then moved pool creation out to DeploymentLauncher; the sentinel fields stayed but stopped guarding anything. This PR's on-demand branch reintroduces pool creation in the activator without re-wiring the sentinel, so concurrent requests to a cold ephemeral URL can both fall through to `CreatePoolForVersion`. Since `findMatchingPool` is list-then-create (not atomic), both can succeed.

Suggested pragmatic fix: re-animate the sentinel for the on-demand branch. The fields are still there, just unused. Mark `inProgress: true` and open a `done` channel before releasing the lock, close it after creation finishes. One related cleanup either way: the top-of-file doc block at lines 21-25 still claims `requestPoolCapacity` uses the Sentinel Pattern, which hasn't been true since `040329dd`. Restore the sentinel and the doc is correct again; skip the sentinel and the doc should lose that section.

The bigger frustration, which Paul has been sitting with for a while: the activator's coordination story keeps slowly eroding as features thread through it. MIR-596 tracks the direction already, and we added a comment there pointing at this PR as fresh evidence. That's the refactor pile, not a blocker.

components/activator/activator.go
evanphx: oh yup! should be fixed now.

PR #749, opened 2026-04-10
Split blackbox tests across two CI shards

evanphx:
## Summary
- Add a lightweight `blackbox-tests` job that lists blackbox tests via `go test -tags blackbox -list`, sorts them, and splits into two shards.
- Convert `test-blackbox` to a matrix job that runs each shard in parallel using `-run '^(...)$'`.
- Dynamic split avoids maintaining a static group file like `hack/test-groups.json`.

## Test plan
- [ ] CI runs two `test-blackbox (shard N)` jobs in parallel and both pass
- [ ] Total blackbox wall-clock time roughly halves (~9 min → ~5 min)
- [ ] `test` summary job still gates on all shards succeeding

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,8 +187,48 @@ jobs:
       if: steps.docker-image-cache.outputs.cache-hit != 'true'
       run: docker save runtime-shell -o /tmp/iso-image.tar
 
+  blackbox-tests:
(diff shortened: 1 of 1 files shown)


phinze (approved):
Image

evanphx:
I split the POP tests out in the other PR because they were causing stuff to fail, so we don't need this split for now.

PR #756, opened 2026-04-13
MIR-972: self-healing loop-backed volumes after unclean miren kill

evanphx:
## Summary

Fixes the incident where a SIGKILL to miren during a postgres addon shutdown left the ext4 filesystem on a loop-backed volume corrupted, producing an unrecoverable crash loop that required manual SSH intervention (fsck, losetup -d, force-delete disk leases, entity patches).

Root cause: after the SIGKILL, the new miren re-attached the same `disk.img` to a *second* `/dev/loopN` via `LOOP_CTL_GET_FREE`, producing two loop devices with independent, incoherent kernel page caches. ext4 journal replay on the new mount read a stale view of the file and corrupted the filesystem.

Seven commits, each leaving the tree in a working state:

1. **diskio: adopt existing loop device when backing file is already attached** — new `FindLoopByBacking` op walks `/sys/block/loop*/loop/backing_file`; both `DiskMountController.attachAndMount` and `DiskVolumeController.ensureVolumeMount` now adopt an existing device instead of allocating a second one for the same backing file.
2. **diskio: enable direct I/O on loop devices** — sets `LO_FLAGS_DIRECT_IO` so the loop device no longer keeps an independent page cache layered on the backing file.
3. **diskio: log rollback detach failures in ensureVolumeMount** — replaces silent `LoopDetach` rollback calls with a helper that logs failures.
4. **diskio: sweep orphan loop devices and mounts at boot** — one-shot orphan sweep in `DiskVolumeController.ReconcileWithEntities` that detaches loops / unmounts mounts under miren's volumes dir that don't correspond to any known volume.
5. **diskio: run fsck and retry once on dirty-filesystem mount failures** — new `Fsck` op and `mountWithFsckRetry` helper; a mount failing with EUCLEAN ("Structure needs cleaning") now triggers an automatic fsck and retry.
6. **disk: release orphan leases at boot when owning sandbox is dead** — `DiskLeaseController` sweeps BOUND/PENDING leases whose owning sandbox is DEAD or missing and transitions them to RELEASED, eliminating the "disk has an active lease" deadlock that only cleared via `miren debug disk lease-delete --force`.
7. **sandbox: emit volume-stage failures on the sandbox log stream** — `ConfigureVolumes` errors in the saga path now emit a synthetic lifecycle event on the sandbox's own log stream, so operators see the reason via `miren logs sandbox <id>` without ssh'ing to the node. Shared `sandboxLogMeta` helper keeps the runtime event stream identity/attrs consistent with the container stdio path.

An independent review of the "never double-mount a universal file" invariant walked through every SIGKILL-restart scenario and confirmed the three-layer defense (adoption in `ensureVolumeMount`, direct I/O on new attaches, orphan sweep on boot) closes the corruption path under all realistic production conditions.

**Known follow-up (non-blocking):** `FindLoopByBacking` currently treats a sysfs read error as a warning and falls through to `LoopAttach`. If sysfs is degraded (beyond the normal SIGKILL scenario), the adoption check can be bypassed. Direct I/O softens the blow. Worth a small follow-up to return the error and let the controller retry.

Refs MIR-972.

## Test plan

- [x] `go test ./components/diskio/...` — unit tests including new adoption, orphan sweep, and fsck retry tests
- [x] `go test ./controllers/disk/...` — orphan lease sweep tests
- [x] `go test ./controllers/sandbox/...` — saga volume-config-fails test verifies lifecycle event emission
- [x] `golangci-lint run` clean across touched packages
- [ ] End-to-end reproduction in dev environment: start sandbox with miren volume, `systemctl kill -s KILL miren`, confirm `losetup -a` shows the surviving loop, restart miren, verify the adoption logs and a single loop device remains
- [ ] `losetup -l` shows `DIO=1` for miren-owned loops after the direct-I/O change
- [x] `make test-blackbox` in CI


phinze (approved):
Field Surgeon Claude and Combat Medic Phinze reporting from triage. We examined all seven layers of suture on this one.

Patient presented with a double-loop page cache divergence, complicated by orphan leases blocking the airway and a filesystem too dirty to mount. Prognosis was grim: manual SSH intervention required, crash loop unresponsive to standard treatment.

Operative report: the three-layer defense (adopt surviving loops, direct I/O to prevent future cache splits, orphan sweep at boot) restores hemodynamic stability under all realistic SIGKILL scenarios. The fsck-and-retry is textbook field medicine for dirty filesystems. The lease sweep clears the circulatory blockage that was starving new sandboxes of disk access. And the log stream event emission means next time, the medics on call can read the chart without SSH'ing into the ward.

No complications found. All sutures holding. Patient cleared for deployment.

PR #758, opened 2026-04-13
MIR-970: Unblock addon deprovisioning when provisioning saga is in-flight

evanphx:
## Summary
- Bump addon reconciler worker count from 1 → 4 so a long-running provisioning saga (e.g. valkey waiting on sandbox pool readiness) no longer blocks every other addon event. Same-entity serialization is still guaranteed by `ReconcileController.inFlight`.
- Add a pre-flight re-read at the top of `provision()`: if the association is no longer `pending` (e.g. user ran `addon destroy` during the crash/resync window) skip instead of re-starting a saga that would conflict with leftover resources.

## Why
Observed sequence:
1. Reconciler starts provisioning rabbitmq and valkey.
2. Valkey saga blocks waiting on sandbox pool readiness — occupies the sole worker.
3. User runs `addon destroy`; status flips to `deprovisioning`.
4. Deprovisioning event sits in the work queue forever.

On restart, the resync re-enqueues stale `pending` events and re-triggers provisioning on associations the user had already destroyed, creating conflicting resources.

## Test plan
- [x] `go test ./controllers/addon/` — new `TestProvisionSkipsWhenAssociationNoLongerPending` covers the stale-event path.
- [x] `go vet ./controllers/addon/... ./components/coordinate/...`
- [ ] Smoke in `make dev`: install valkey + rabbitmq, confirm rabbitmq finishes while valkey is still waiting on its sandbox pool.

## Out of scope
Coordinating saga `Recover()` with a concurrent reconciler-driven re-provision is tracked as a follow-up — the fixes here narrow the window significantly but do not fully close it.

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -1033,7 +1033,10 @@ func (c *Coordinator) Start(ctx context.Context) error {
 			eac,
 			controller.AdaptReconcileController[addon_v1alpha.AddonAssociation](addonController),
 			time.Minute,
-			1,
+			// Multiple workers so a long-running provisioning saga for one
+			// association does not block reconciliation of others. Same-entity
+			// concurrency is already prevented by ReconcileController.inFlight.
+			4,
 		)
 		c.cm.AddController(addonReconciler)
 	}
--- a/controllers/addon/controller.go
+++ b/controllers/addon/controller.go
@@ -64,6 +64,20 @@ func (c *Controller) Reconcile(ctx context.Context, assoc *addon_v1alpha.AddonAs
 func (c *Controller) provision(ctx context.Context, assoc *addon_v1alpha.AddonAssociation, meta *entity.Meta) error {
 	c.log.Info("provisioning addon", "association", assoc.ID, "addon", assoc.Addon, "variant", assoc.Variant)
 
+	// Re-read the association to guard against stale events (e.g. on startup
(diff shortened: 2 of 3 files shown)


components/coordinate/coordinate.go
```diff
@@ -1033,7 +1033,10 @@ func (c *Coordinator) Start(ctx context.Context) error {
 			eac,
 			controller.AdaptReconcileController[addon_v1alpha.AddonAssociation](addonController),
 			time.Minute,
-			1,
+			// Multiple workers so a long-running provisioning saga for one
+			// association does not block reconciliation of others. Same-entity
+			// concurrency is already prevented by ReconcileController.inFlight.
+			4,
```
phinze, line 1039: Any reason to go to 4 rather than 2 here? Two workers already solve the stated problem (one slow saga can't starve the queue), and each provisioning saga does a fair bit of entity store work (config versions, app versions, active version patches). Starting conservative and bumping later if someone actually hits the limit seems like the lower-risk move. Don't feel strongly tho, totally shippable as is!

phinze (approved):

components/coordinate/coordinate.go
evanphx: We already done 3 elsewhere and if these are slower, due to startup, felt like it was a good fit to reach for 4 for now.

phinze: SGTM!

PR #762, opened 2026-04-14
Detect dead sandboxes on failed runners

phinze:
When a distributed runner dies (process crash, node failure), its session-scoped `READY` status on the Node entity clears after the ~60s session TTL. The scheduler already uses this to avoid placing *new* sandboxes on dead nodes, but nothing was watching for *existing* sandboxes that were already running there. They'd stay in `RUNNING` status forever, and the coordinator would keep routing traffic to their IPs, causing request timeouts for as long as the node was down.

The fix is a new `nodehealth` controller on the coordinator that watches Node entities. When a node loses READY, the controller starts a 5-minute grace period. If the node doesn't recover within that window, it lists all sandboxes scheduled to the node and marks them DEAD. The activator's existing watch pipeline picks up the status change and immediately invalidates cached leases, so traffic stops routing to the dead IPs. The pool controller then sees the reduced capacity and spins up replacements on healthy nodes.

The grace period is the interesting design choice here. Containers survive miren process restarts (containerd keeps running), and `reconcileSandboxesOnBoot()` re-adopts them when the runner comes back. We verified on Garden that all 13 sandboxes survived today's restart with zero downtime. Marking sandboxes DEAD immediately would break this recovery path and cause unnecessary churn on every upgrade or restart. Five minutes is generous for any normal restart (typically 15-30 seconds) while still bounding the impact of a real failure.

We also discovered during investigation that `reconcileSandboxesOnBoot()` reattaches logs and re-reserves IPs for surviving sandboxes, but doesn't re-register cgroup metrics monitoring. That's tracked separately in MIR-1013.

Validated on Toys with a coordinator + distributed runner topology. Killed the runner process and watched the coordinator detect the node going not-ready after session TTL, tick through the grace period for 5 minutes, then mark 3 sandboxes dead. The pool controller replaced them on surviving nodes within seconds. Also verified the restart path: bounced the runner, it recovered in under a second, grace period cleared, no sandbox disruption.

Closes MIR-1009

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -45,6 +45,7 @@ import (
 	artifactctrl "miren.dev/runtime/controllers/artifact"
 	certctrl "miren.dev/runtime/controllers/certificate"
 	deploymentctrl "miren.dev/runtime/controllers/deployment"
+	nodehealthctrl "miren.dev/runtime/controllers/nodehealth"
 	"miren.dev/runtime/controllers/sandboxpool"
 	schedulerctrl "miren.dev/runtime/controllers/scheduler"
 	"miren.dev/runtime/metrics"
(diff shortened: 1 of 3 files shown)


evanphx (approved):
Approving — ship it. 🚢

Evan + Claude took a look at this together. Clean, well-scoped, good test matrix, and the grace-period / handled-latch design choices all make sense given the reconcile-on-boot recovery path you're protecting.

One thing we spotted that's worth a follow-up (not a blocker):

In `controllers/nodehealth/controller.go:179-185`:

```go
_, err := c.eac.Patch(ctx, entity.New(
    entity.DBId, sb.ID,
    (&compute_v1alpha.Sandbox{
        Status: compute_v1alpha.DEAD,
    }).Encode,
).Attrs(), 0)
```

This reads like "patch just `Status` to DEAD", but `Sandbox.Encode()` in `api/compute/compute_v1alpha/schema.gen.go:1103` doesn't quite play along. Most fields are gated behind `!Empty` / `len() > 0` checks, but `HostNetwork` is emitted unconditionally at line 1107:

```go
attrs = append(attrs, entity.Bool(SandboxHostNetworkId, o.HostNetwork))
```

So the patch actually sends two attrs: `Status=DEAD` and `HostNetwork=false`. Any sandbox on a dead node that was using host networking will silently get `HostNetwork` flipped to `false` on its way to DEAD. Probably harmless in practice since the sandbox is terminal, but it's a latent trap for the next person who tries a targeted `Encode`-based patch.

Two ways to deal with it, whenever you feel like it:

1. Build the attr list by hand here — just `entity.Ref(SandboxStatusId, sandboxstatusToId[DEAD])` — and sidestep `Encode` entirely.
2. Or fix the generator to gate `Bool` fields the same way as the other scalars. That's the real fix and probably affects other kinds too, so it's its own PR.

Also a tiny aside: `Patch(..., 0)` skips optimistic concurrency, where most other callers thread the version through. Low risk here since the runner's already gone, but figured we'd mention it.

Really nice work on this one — the grace-period reasoning in the PR description is great context to have in the commit trail.

— Evan & Claude

PR #764, opened 2026-04-15
MIR-888: Rename route OIDC feature to "route protect oidc"

evanphx:
## Summary

- Present the labs feature as `miren route protect` in the CLI and docs, with `oidc` as a protection-type subcommand (`miren route protect oidc`). This leaves room to add other protection mechanisms under the same verb without a future rename.
- Disable/show stay type-agnostic: `miren route protect disable|show`.
- Labs flag name (`routeoidc`) is kept — it gates an OIDC-specific codepath — but its description now reads "Protect HTTP routes with identity-provider login".
- Docs page `route-oidc.md` → `route-protect.md`; command reference regenerated.
- Schema, API client, ingress runtime, and tests are **not** renamed. OIDC is still the only mechanism under the hood and those names accurately describe what the code does. Only the user-visible UX moved.

## Test plan

- [x] `make bin/miren`
- [x] `make lint` (0 issues)
- [x] `go generate ./pkg/labs/... ./cli/commands/` regenerates cleanly
- [x] `miren route protect --help` shows `oidc`, `disable`, `show` subcommands
- [x] `miren route protect oidc --help` shows the full provider/claim-header flag set
- [x] `cd docs && bun run build` passes (one pre-existing unrelated broken-anchor warning on `/services`)
- [ ] End-to-end smoke in `make dev` with `MIREN_LABS=routeoidc`: attach an OIDC provider to a route via `route protect oidc`, hit the route without a session → redirects to provider, hit again with session → claims forwarded as headers, `route protect disable` removes it


phinze (changes requested):
Big P and little C here (physically at least; from a storage perspective, arguably the reverse).

We spent some time digging into the CLI semantics and the provider model before this goes GA, and wanted to open a conversation about the overall UX shape.

The rename away from `route oidc` is the right instinct. But as we traced through the command structure and the underlying data model, we think there's an opportunity to land on something cleaner before this becomes a public surface.

**Consistent verbs across the CLI.** One thing we're trying to get right as the CLI grows is convergence toward predictable patterns. When the same structural situation uses the same verbs, users build muscle memory faster and AI agents can discover and compose commands more reliably. Right now `route protect` mixes a noun (`oidc`) with verbs (`disable`, `show`) at the same level, and `disable` doesn't appear anywhere else in the CLI (we use `remove`/`delete`/`destroy` everywhere else for teardown operations).

**The provider model is already decoupled from routes.** Looking at the runtime, the `OidcProvider` entity is just IdP credentials (provider URL, client ID, client secret, scopes). The redirect URL is derived at request time from the incoming hostname, not stored on the provider. So a provider is genuinely a reusable, route-independent resource. But the current CLI buries provider lifecycle inside the protect command as a side effect, with an auto-naming scheme (`example.com/accounts.google.com`) that's hard to discover or predict. And `--provider` silently does double duty as both "name this new provider" and "reference an existing one" depending on whether `--provider-url` is also present.

**What we'd suggest considering:** split provider management out as its own CRUD, and make route protection a simple association.

```
# Provider lifecycle (lives under auth, next to the existing auth ci commands)
miren auth provider add google-corp \
  --url https://accounts.google.com \
  --client-id ... --client-secret ...
miren auth provider list
miren auth provider show google-corp
miren auth provider remove google-corp

# Route protection (just the association, no provider creation side effects)
miren route protect example.com --provider google-corp
miren route unprotect example.com
```

This gives you straightforward CRUD for providers following the same `auth ci add/list/remove` pattern that already exists. Route protection becomes a simple verb pair on route (`protect`/`unprotect`, like `set`/`remove`), no subcommand group needed. And the user always knows what providers exist because `auth provider list` is right there.

The IdP-side redirect URL whitelisting (`https://example.com/.well-known/miren/oidc/callback`) is something users have to configure regardless of CLI structure, but this separation makes the mental model clearer: set up the provider once, then attach it to routes as needed.

This shape also extends naturally. Non-OIDC provider types (API keys, basic auth) would just be different flavors of `auth provider add`. And behavior tweaks to protection itself (passthrough mode that forwards identity headers without blocking unauthenticated requests, for example) would be flags on `route protect` rather than new subcommands.

We realize this is a bigger scope than a rename PR. Happy to discuss whether it makes sense to do this now vs. as a follow-up before GA.

Marking as request-changes just to make sure we talk through this before it merges, not because anything here is wrong. The rename is a good idea and we want to help it land in the best shape.

cli/commands/auth_provider.go
```diff
@@ -0,0 +1,226 @@
...
+
+	provider := &ingress_v1alpha.OidcProvider{
+		Name:         opts.Name,
+		ProviderUrl:  opts.ProviderURL,
+		ClientId:     opts.ClientID,
+		ClientSecret: opts.ClientSecret,
+		Scopes:       scopes,
+	}
+
+	_, err = ic.CreateOrUpdateOIDCProvider(ctx, provider)
```
phinze, line 75: One thing worth flagging because it's a small surprise: `CreateOrUpdateOIDCProvider` makes `add` an upsert. Re-running `miren auth provider add my-google ...` silently rotates the client secret without warning. That's actually convenient (no separate `provider rotate-secret` needed), but `auth ci add` (the pattern we cited) is create-only, so users might expect the same here. Probably either worth a docs note that it doubles as an update path, or making the verb explicit (`auth provider set` for upsert, or an `--update` flag). Whatever you decide, even just leaving a comment on the function noting the upsert behavior would help future readers.

cli/commands/route_show.go
```diff
@@ -32,17 +36,86 @@ func RouteShow(ctx *Context, opts struct {
 		return fmt.Errorf("route not found: %s", opts.Host)
 	}
 
+	protected := !entity.Empty(route.OidcProvider)
```
phinze, line 58: Small symmetry gap (anchoring here since the relevant `Host` positional is in unchanged territory): `route protect --default` and `route unprotect --default` both work, but `route show` only takes a required `Host` positional, so there's no way to inspect the protection state of the default route after attaching it. Worth mirroring a `Default bool` flag here so the read side keeps up with the write side.

cli/commands/auth_provider.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	client, err := ctx.RPCClient("entities")
+	if err != nil {
+		return err
+	}
+
+	ic := ingress.NewClient(ctx.Log, client)
+
+	err = ic.DeleteOIDCProvider(ctx, opts.Name)
```
phinze, line 255: `auth provider remove` will happily delete a provider that's still attached to N routes, which is what creates the "missing provider" state that your new `route show` handles so gracefully. The graceful display is honestly the right primary safety net (someone might delete a provider for legitimate teardown reasons and want to keep the route around). But if we want to be a touch kinder, an in-use check that errors out with "provider attached to 3 routes, pass --force" would let people catch typos before things go sideways. Easy to add later if it becomes a real footgun.

cli/commands/auth_provider.go
```diff
@@ -0,0 +1,226 @@
...
+			Scopes      string `json:"scopes"`
+		}
+
+		var items []ProviderJSON
+		for _, p := range providers {
+			items = append(items, ProviderJSON{
+				Name:        p.Name,
+				ProviderURL: p.ProviderUrl,
+				ClientID:    p.ClientId,
+				Scopes:      p.Scopes,
```
phinze: CodeRabbit flagged this one and I think they're right: `scopes` comes out as a single space-delimited string in the JSON output (`"scopes": "openid email profile"`), but `--scope` is a repeated input flag and the project's CLI JSON guidance says to use machine-readable shapes. An array (`"scopes": ["openid", "email", "profile"]`) would let scripts iterate without re-splitting. Same shape applies to the `auth provider show` JSON around line 185. Small fix; `strings.Fields(p.Scopes)` does it.

phinze (approved):
Claude and Paul here, sat down and walked the two follow-on commits together. The shape is exactly what we asked for, and a few of the smaller touches landed especially well — the missing-provider degraded display in `route show` is a really nice safety net for the case where someone deletes a provider out from under an attached route. The redundant labs gating that CodeRabbit flagged on the first review is still in place across the handlers, and that's defensible as defense-in-depth, so we're not pushing on it.

A handful of small things below, none blocking. Approving.

cli/commands/auth_provider.go
evanphx: Adjusted the behavior to make it more obvious.

cli/commands/route_show.go
evanphx: fixed!

cli/commands/auth_provider.go
evanphx: WHOOPS! fixed.

cli/commands/auth_provider.go
evanphx: also fixed!

cli/commands/commands.go
```diff
@@ -503,35 +503,23 @@ miren deploy --analyze
...
+		d.Dispatch("route unprotect", Infer("route unprotect", "Remove identity-provider protection from an HTTP route", RouteUnprotect,
 			WithLabsFeature(labs.FeatureRouteOIDC),
 			WithExample(mflags.Example{
-				Name: "Show OIDC config for a route",
-				Body: "miren route oidc show example.com",
+				Name: "Remove protection from a route",
+				Body: "miren route unprotect example.com",
 			}),
 		))
 	}
```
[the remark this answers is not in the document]
evanphx: It's the provider that determiness the type, so the code is correct.

PR #770, opened 2026-04-21
Gate disk reconciliation on the coordinator node

phinze:
Observed on `toys` during the rust-chat booth deploy (MIR-1022 work): four consecutive addon sandboxes died at "creating disk lease," the addon never came up, and the app served 500s. Root cause was that a `disk_volume` entity had been created on one of the distributed runner nodes, which violates the invariant that disks live only on the primary today. The parent disk stayed stuck at `provisioning` forever because the primary's reconciler kept seeing the foreign volume, correctly refusing to touch it, and then (less correctly) returning early instead of finalizing its own.

The fix adds a `shouldManageDisk` gate on the `DiskController` that today is just `IsCoordinator`. Wiring it as a helper rather than inlining the check gives us a clear extension point for when cross-node disk migration lands — that logic will likely key off a target-node field on the disk entity, and every caller already routes through one function. Distributed runners now return early from `reconcileDisk` entirely, so they can't accidentally manufacture foreign volumes again.

Took the opportunity to harden the coordinator's own path too: `getDiskVolumeForDisk` now prefers a volume it owns when multiples exist (instead of grabbing `values[0]` and hoping), and `handleProvisioning` treats any lingering orphan as noise rather than a reason to bail, so clusters that already have orphaned volumes from the bug will self-heal on the next reconcile.

Closes MIR-1030

### DIFF
--- a/controllers/disk/disk_controller.go
+++ b/controllers/disk/disk_controller.go
@@ -42,6 +42,14 @@ type DiskController struct {
 	// NodeId is the ID of this node, used for creating volume entities
 	NodeId string
 
+	// isCoordinator is true if this controller runs on the coordinator node.
+	// Today all disks are owned by the coordinator, so only the coordinator
+	// may create, update, or delete disk_volume entities. When cross-node
+	// disk migration lands, shouldManageDisk will gain smarter logic (likely
+	// keyed off a target-node field on the disk entity); every caller
+	// already routes through it.
+	isCoordinator bool
+
 	// Base path for disk mounts (e.g., /var/lib/miren/disks)
 	mountBasePath string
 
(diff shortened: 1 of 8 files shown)


controllers/disk/disk_controller.go
```diff
@@ -54,17 +62,35 @@ type DiskController struct {
...
+func (d *DiskController) shouldManageDisk(disk *storage_v1alpha.Disk) bool {
+	_ = disk
+	return d.isCoordinator
+}
+
+// myNodeId returns the entity ID used for disk_volumes owned by this
+// controller's node, normalized so the "node/" prefix is always present
+// exactly once regardless of how NodeId was passed in.
+func (d *DiskController) myNodeId() entity.Id {
+	return entity.Id("node/" + strings.TrimPrefix(d.NodeId, "node/"))
```
evanphx, line 91: We do a similar thing a bunch of places, we should probably figure out a way to do it once up front.

evanphx (approved):
Looks good, one minor thing we should remember to cleanup in other places at some point too

controllers/disk/disk_controller.go
phinze: Good call - filed MIR-1072 for a sweep of these.

PR #773, opened 2026-04-23
Fix miren login cancel being reported as a timeout

bnferguson:
### bnferguson says
Running `miren login` this evening and noticed it taking a little while. Hit Ctrl-C and noticed it said it had timed out at 10m0s. Well, it'd been a while but not _that_ long. So I gave it another shot, immediately canceling via Ctrl-C. Sure enough the same message. Dove into the code to check the behavior and there's the bug.

I did skip the tests on this one — our little bot friend wanted to add a new file and about 3 tests just covering a difference in `context.Canceled` and `context.DeadlineExceeded`. Seemed like there weren't similar tests in the area so did my best to match style, which meant leaving them out.

### Claude says
`pollForToken` returned a hard-coded `"authentication timed out after 10m0s"` whenever its derived timeout context fired, without checking whether the deadline had actually expired or the parent ctx had just been cancelled. It also didn't wrap `ctx.Err()`, so the top-level Ctrl-C handler at `cli/cli.go:53` — which exits 0 silently on `context.Canceled` — had nothing to match against.

## Approach
Check `timeoutCtx.Err()` and only render "timed out after X" on `context.DeadlineExceeded`. Otherwise return the context error directly, wrapped via `%w` so `errors.Is` keeps working up the stack. The top-level handler already eats `context.Canceled`, so Ctrl-C now behaves the same way it does in `deploy`, `server register`, and every other long-running command.

## Reviewer notes
Matches the shape of `PollForApproval` at `pkg/registration/registration.go:161`, which does the same kind of polling against a parent-carried deadline.

## Testing plan
Run `miren login` and hit Ctrl-C immediately — should exit cleanly with no error output. Let it actually time out (or pass a short deadline) — should still print `authentication timed out after <duration>`.


bnferguson:
Gave the first coderabbit bit of feedback another look and from my local testing it seems it works as intended (on MacOS): 

image


No `ERROR: authentication failed: context canceled` to be seen. Seems to behave as other cancellations that I've seen in Miren so far. 

Interestingly, waiting 10 minutes I got `ERROR: authentication failed: server error: invalid_grant - Invalid or expired device code` and not the expected `authentication timed out after 10m0s`. Looks like the server can expire before the client can?

phinze (approved):
👏 🥳 Thanks for your first contribution! Changes looks good and agreed with your interpretation that CodeRabbit is getting mixed up and the code in question catches wrapper errors properly. We'll keep banging on CI and get this in soon.

PR #777, opened 2026-04-29
Switch tarx to go-git's gitignore matcher

phinze:
A user report came in about a vanilla `bridgetown new` app failing to boot after deploy with `Errno::ENOENT - tmp/pids/server.pid`. Bridgetown ships with `tmp/pids/.keep` plus a `.gitignore` that uses `/tmp/*` and `!/tmp/pids/` to keep the pidfile directory tracked. When miren bundles source for deploy, the gitignore walker drops `tmp/` from the tar entirely, so `tmp/pids/` never makes it into the image, and Puma has nowhere to write its pidfile.

The walker was using `shibumi/go-pathspec`, a generic glob library whose `GitIgnore` function takes a single string path with no way to indicate it's a directory. Our `isGitignored` worked around that by checking both `relPath` and `relPath + "/"` and OR-ing the result, but the dual-form check breaks negations: the no-slash form matches `/tmp/*` and short-circuits before the trailing-slash negation `!/tmp/pids/` ever gets a chance. Once the walker `SkipDir`s `tmp/pids/`, lazy directory emission drops `tmp/` from the tar. Pathspec is a generic glob library, not a real gitignore implementation, and we'd been bending it into the wrong shape.

Swapping to `go-git/v5/plumbing/format/gitignore`, the parser go-git itself uses, gets us a proper `Match(path []string, isDir bool) bool` API. All 19 existing tarx tests pass unchanged, which is a strong signal the swap is behavior-preserving for everything except the bug we were after. `TestMakeTarBridgetownTmpPids` is the new regression guard. Binary delta is +20KB after dead-code elimination.

Closes [MIR-1075](https://linear.app/miren/issue/MIR-1075).

### DIFF
--- a/pkg/tarx/tarx.go
+++ b/pkg/tarx/tarx.go
@@ -26,12 +26,17 @@ type FileManifest struct {
 }
 
 // parseGitignoreFile reads a .gitignore file and returns its parsed patterns
-// scoped to the given domain (path components from the walk root). Returns
-// nil if the file doesn't exist or is empty.
-func parseGitignoreFile(path string, domain []string) []gitignore.Pattern {
+// scoped to the given domain (path components from the walk root). A missing
+// file is not an error and returns (nil, nil); other read failures (e.g.
+// permission denied) propagate up so we don't silently treat them as "no
+// patterns" and ship a tar that should have been filtered.
+func parseGitignoreFile(path string, domain []string) ([]gitignore.Pattern, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
-		return nil
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("reading %s: %w", path, err)
 	}
 	var patterns []gitignore.Pattern
 	for line := range strings.SplitSeq(string(data), "\n") {
@@ -41,7 +46,7 @@ func parseGitignoreFile(path string, domain []string) []gitignore.Pattern {
 		}
 		patterns = append(patterns, gitignore.ParsePattern(line, domain))
 	}
-	return patterns
+	return patterns, nil
 }
 
 // parseStringPatterns parses a list of raw gitignore-style pattern strings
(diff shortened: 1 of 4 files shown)


pkg/tarx/tarx.go
```diff
@@ -151,30 +148,12 @@ func ComputeManifest(dir string, includePatterns []string) ([]FileManifest, erro
...
-				}
-				if match {
-					isIncluded = true
-					break
-				}
-			}
-		}
+		segs := pathSegments(rp)
+
+		isIncluded := len(includes) > 0 && gitignore.NewMatcher(includes).Match(segs, info.IsDir())
```
evanphx: Can we move the NewMatcher creation up next to the includes assignment on 132?

pkg/tarx/tarx.go
```diff
@@ -151,30 +148,12 @@ func ComputeManifest(dir string, includePatterns []string) ([]FileManifest, erro
...
 
 		if !isIncluded {
-			ignored, ignoreErr := isGitignored(rp, info.IsDir(), gitignoreMap)
-			if ignoreErr != nil {
-				return ignoreErr
-			}
-			if ignored {
+			if gitignore.NewMatcher(ignorePatterns).Match(segs, info.IsDir()) {
```
evanphx: Ditto on moving this above.

pkg/tarx/tarx.go
```diff
@@ -184,8 +163,12 @@ func ComputeManifest(dir string, includePatterns []string) ([]FileManifest, erro
...
 		if info.IsDir() {
 			nestedGitignore := filepath.Join(path, ".gitignore")
-			if patterns := parseGitignore(nestedGitignore); patterns != nil {
-				gitignoreMap[rp] = patterns
+			more, err := parseGitignoreFile(nestedGitignore, segs)
+			if err != nil {
+				return err
+			}
+			if more != nil {
+				ignorePatterns = append(ignorePatterns, more...)
```
evanphx, line 174: I think this will cause an issue, where a directory that has it's own .gitignore will mutate the ignorePatterns that is then used by later sibling directories incorrectly. We'd need to pass the new patterns into child dirs, but not siblings.

pkg/tarx/tarx.go
```diff
@@ -286,30 +265,12 @@ func makeTarWithFilter(dir string, includePatterns []string, accept func(string)
...
-					}
-					if match {
-						isIncluded = true
-						break
-					}
-				}
-			}
+			segs := pathSegments(rp)
+
+			isIncluded := len(includes) > 0 && gitignore.NewMatcher(includes).Match(segs, info.IsDir())
```
evanphx: Same feedback as the above function.

evanphx (changes requested):
All good, but I think we need to fix the cross-contanimation of the gitignore patterns from sibling directories.

evanphx (approved):
Ah, the pattern domain thing makes sense. I had to go through it, basically it's clamping the pattern to paths that are inside the path to the gitignore.

PR #781, opened 2026-05-05
Fix runner upgrade skipping service restart on already-current binary

phinze:
The repro from MIR-1036: on a runner host, `miren upgrade` happens to update the release-dir binary at /var/lib/miren/release/miren as a side effect. A subsequent `miren runner upgrade --channel main` reads that same binary, decides "already at the target version", and exits — without restarting the service. The daemon PID keeps executing the previous binary inode, which is exactly what you didn't want. `--force` worked around it, but only because force re-downloads a binary you already had just to trigger a restart.

The version check was answering the wrong question. "Is the on-disk binary at the target version?" is fine for deciding whether to download. But "do we need a restart?" is a separate question, and the right answer comes from comparing the running daemon's version against the on-disk binary. So the fix splits those concerns: when the on-disk binary matches the target, we still check whether the running daemon matches the on-disk binary, and if there's drift we just restart (no download). We get the running daemon's version by exec'ing /proc/<MainPID>/exe — the kernel keeps the original inode alive as long as the process holds an exec mapping to it, so this works even after the install path has been atomically replaced.

Same plumbing applies to `server upgrade` since both share the upgrade helper. The `--check` mode surfaces drift directly now, so you can spot the situation without invoking an upgrade.

Picked up an unrelated bonus along the way: `miren logs --format json` was rendering timestamps in the host's local timezone because `time.Unix()` returns a `time.Time` with `Location: Local`. The test for it pinned the expected output to UTC and only passed on UTC hosts. Adding `.UTC()` before formatting makes both the production code and the test TZ-independent. Human text output is untouched.

Closes MIR-1036

### DIFF
--- a/pkg/release/version.go
+++ b/pkg/release/version.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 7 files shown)


pkg/release/version.go
```diff
@@ -81,6 +82,82 @@ func parseVersionText(output string) VersionInfo {
...
+// atomic rename of the install path.
+func GetRunningServiceVersion(serviceName string) (VersionInfo, error) {
+	pid, err := getServiceMainPID(serviceName)
+	if err != nil {
+		return VersionInfo{}, err
+	}
+	if pid == 0 {
+		return VersionInfo{}, fmt.Errorf("service %s is not running", serviceName)
+	}
+	return GetCurrentVersion(fmt.Sprintf("/proc/%d/exe", pid))
```
evanphx, line 144: So tricky! love it.

evanphx (approved):

pkg/release/version.go
phinze: I know right?! I was fairly certain we'd need RPC for this, but lil C came up with a real bankshot.

PR #784, opened 2026-05-05
Add /etc/hosts entry mapping sandbox IP to hostname

evanphx:
## Summary
- Map each sandbox's allocated IP address(es) to its hostname in `/etc/hosts`, so processes that resolve their own hostname (e.g., EPMD) get the real sandbox IP instead of loopback
- Move hostname off the `127.0.0.1`/`::1` localhost lines to dedicated IP→hostname entries
- Update frozen hash for `sandbox.go` (saga path shares `BuildSpec` → `setupHosts`, so both controllers are covered)

Closes MIR-1100

## Test plan
- [x] `make lint` passes
- [x] `make test` passes (including frozen file hash check)
- [ ] Manual: start a sandbox, exec in, verify `cat /etc/hosts` shows `<sandbox-ip> <hostname>` instead of hostname on the localhost line

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1252,6 +1252,10 @@ func (c *SandboxController) AllocateNetwork(
 }
 
 func (c *SandboxController) setupHosts(sb *compute.Sandbox, name string, ep *network.EndpointConfig) error {
+	if ep == nil || len(ep.Addresses) == 0 {
+		return fmt.Errorf("no addresses allocated for sandbox %s", sb.ID)
+	}
+
 	var lines []string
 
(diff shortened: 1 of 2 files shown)


phinze (approved):
Claude and Paul here. Verified live: hostname now points at the bridge IP. Conventional setup, right semantics, ship it.

Paul was sad because he thought he botched this in #728. Claude made him feel better by pointing out #728 has valid and necessary bugfixes, it just didn't go far enough: it got EPMD to *start*, #784 gets EPMD to actually *cluster*. Was Claude being sycophantic or honestly assessing? Very much impossible to know for either of us!

Follow-up worth filing but not blocking: an actual EPMD-clustering blackbox test (RabbitMQ-cluster-forms or similar).

PR #785, opened 2026-05-06
Install NodePort DNAT rules on every node

phinze:
NodePort services were unreachable whenever the sandbox landed on a distributed runner, and `nc -z chat.miren.toys 30669` would refuse the connection while the rule sat correctly installed on the runner hosting the sandbox. Felt like a clear "every-node-should-have-this-rule" gap, but the picture got more interesting once we dug in.

ServiceController already runs on every node (the coordinator runs a runner-style component with `IsCoordinator: true`) and watches every Service entity by Kind, not by node. So the cluster-wide install model was already wired. Why wasn't it working?

Two things, layered. The per-NodePort nft chain was keyed on `(serviceIP, servicePort, proto)`, and `setupNodePort` had an "already created this chain, skipping" shortcut. When a service redeployed with a new NodePort but kept the same service IP and port, the chain name didn't change, so the shortcut fired and the new `service_nodeports { tcp . NPORT : goto … }` map element was never added. The April 21 logs on toys captured exactly this: rust-chat went from `node_port = 6669` to `node_port = 30669`, the coordinator's `Creating service` fired with the new NodePort, but the map element stayed pointing at the old one.

The fix for that part is small: rekey on `(proto, NodePort)`. The map already looks up on those two; the chain name should follow.

While we were in there, we also lifted the early-return-on-empty-Ip in `Create()` (which would otherwise skip NodePort install during the first-reconcile race before the IP allocator runs) and decoupled NodePort from the service chain entirely. The two install paths now share a (port, proto)-keyed endpoint-chain map and run independently. NodePort no longer `goto`s into a service chain; it stands alone with its own counter/masq/vmap. Bonus: it now rebuilds when endpoints change, which the old code silently dropped.

Test coverage at three levels: unit tests for the no-IP and endpoint-rebuild cases, plus a focused `TestDistributedRunnerNodePort` blackbox that deploys tcp-echo (which declares `node_port = 7000`) and verifies the `service_nodeports` map element shows up on both the coordinator and runner1.

Tagging the terraform `node-port` firewall tag work for the runner module as a separate followup against the infra repo. Runners in `miren-development` have no external IP today, so it doesn't block this fix.

Closes MIR-1032

### DIFF
--- a/blackbox/distributed_runner_test.go
+++ b/blackbox/distributed_runner_test.go
@@ -5,6 +5,7 @@ package blackbox
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


phinze:
@coderabbit review

evanphx (approved):
very nice!

PR #786, opened 2026-05-06
Add WAF support to HTTP ingress routes

evanphx:
## Summary

- Integrates Coraza WAF with OWASP Core Rule Set (CRS) into the HTTP ingress layer
- Routes gain a `waf_level` field (0=disabled, 1-4=CRS paranoia level) enabling request-level security filtering (SQL injection, XSS, path traversal, etc.) before traffic reaches user apps
- New CLI commands: `miren route waf <host> --level N` and `miren route unwaf <host>`
- Refactors duplicated OIDC middleware composition into a unified WAF → OIDC → serve chain

## Test plan

- [x] `pkg/waf` unit tests — SQL injection, XSS, POST injection blocked; clean requests pass; paranoia levels 1-4; engine caching
- [x] `servers/httpingress` WAF middleware tests — disabled when level 0, blocks attacks, allows clean, respects levels, handles invalid/negative levels
- [x] `api/ingress/client` tests — set/lookup WAF level, disable (zero-value via Patch), invalid levels, non-existent route
- [x] Blackbox tests — enable/disable WAF, malicious request blocked then allowed after disable, JSON output, default route, invalid level rejection
- [x] `make lint` passes


phinze (changes requested):
Claude and Paul here. The big-picture architectural conversation lives on the 787 review where we walked through both PRs together with notes from yesterday's product standup. Short version: the route schema is growing asymmetric across our three protections, and we're suggesting consolidating the auth side into a single `auth_provider` kind there.

Here's how we're thinking WAF should look: `waf_profile` is its own kind (initially with just `paranoia_level`, leaving headroom for custom rules, allowlists, mode toggles later). The route holds a `waf_profile` ref. The user-facing CLI `m route waf <host> --level N` works as it does today by auto-creating or looking up default profiles (`default-l1`, `default-l2`, etc.) under the covers. Power users get `m waf profile add/list/show/remove` to name and share profiles across routes.

The migration cost asymmetry is what makes doing this now cheap. `waf_level int` to `waf_profile ref` is a route-table backfill on every protected route. `waf_profile ref` to a future middleware ref is a much smaller move on far fewer rows. Doing the entity reshape now sets up the right shape and saves the bigger migration later.

Marking REQUEST_CHANGES on the `waf_profile` reshape specifically. Once that pivot lands we'll come back through for the rest of the review properly.

cli/commands/route_waf.go
```diff
@@ -0,0 +1,102 @@
...
+			return fmt.Errorf("failed to lookup route: %w", err)
+		}
+		if route == nil {
+			return fmt.Errorf("route not found for host: %s", opts.Host)
+		}
+		routeLabel = opts.Host
+	}
+
+	if opts.Disable {
+		if entity.Empty(route.WafProfile) {
```
phinze, line 68: The `--disable` branch still prints human text even with `--format json` (verified live):

```
$ m route waf foo --disable --format json
WAF disabled on route: foo
```

Same on the "already off" branch one line down. Last review's CR comment had a clean fix using a shared `render()` closure — that pattern would close this and stay readable.

docs/docs/waf.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Web Application Firewall (WAF)
+description: Protect your apps from common web attacks with built-in OWASP WAF filtering at the routing layer.
+keywords: [waf, web application firewall, security, owasp, sql injection, xss, coraza, crs]
+---
+
+import CliCommand from '@site/src/components/CliCommand';
+
+# Web Application Firewall (WAF)
```
phinze, line 9: Could we add a "what this doesn't do" callout near the top? CRS catches attack *payloads* (SQL injection, XSS, command injection, path traversal) — not URL pattern scanning. Users hitting this from a "my Rails app logs are full of `/wp-admin/` and `/xmlrpc.php`" angle will turn it on, see the same noise, and be confused.

A line like "the WAF inspects request content for attack patterns; it doesn't fingerprint suspicious URLs, rate-limit, or block bot reconnaissance — for that, consider an upstream proxy like Cloudflare" sets expectations and saves a support thread.

pkg/entity/validation.go
```diff
@@ -248,9 +248,10 @@ func (v *Validator) ValidateAttribute(ctx context.Context, attr *Attr) error {
 		if !ok {
 			return fmt.Errorf("attribute %s must be a string representing an entity ID", name)
 		}
-		// Check if the referenced entity exists
-		if _, err := v.store.GetEntity(ctx, str); err != nil {
-			return fmt.Errorf("attribute %s references a non-existent entity: %w", name, err)
+		if str != "" {
```
phinze, line 251: Quick gut-check: this changes ref-validation semantics globally to "empty refs are valid, skip the existence check." Right call for the WAF detach use case, and probably correct in general — an empty ref is "no reference" not "broken reference." But it's a system-wide loosening; worth confirming nothing else (other detach flows? entity migrations?) was relying on empty refs round-tripping through the existence check.

phinze (approved):
Claude and Paul here. Round two — the `waf_profile` reshape we asked for landed cleanly, and the inline CR findings from last time (nil guards, URI redaction, docs accuracy) are addressed too. Took it for a spin in dev: enabled, blocked SQL injection and XSS, switched levels, disabled, and watched the shared `waf-l1`/`waf-l2`/`waf-l3` profiles get lazily created as singletons in the entity store. The whole shape feels right and leaves room for the future `m waf profile ...` power-user CLI we floated.

One thing the spin surfaced beyond the inlines: we tried this against the real-world "my logs are full of WordPress scans" complaint and the WAF doesn't help with that — CRS catches *payloads* (SQL injection, XSS, command injection), not URL fingerprinting. Worth saying so explicitly in `waf.md` since users will reach for this expecting it to quiet down bot noise and be surprised when it doesn't. Inline comment with details.

cli/commands/route_waf.go
evanphx: Fixed!

pkg/entity/validation.go
evanphx: Fixed!

PR #789, opened 2026-05-07
Add ingress.http_address for plain-HTTP ingress on a custom address

jcasimir:
Standing up a Miren server on a host that already serves another application on 80/443 currently leaves no way to put the Miren ingress somewhere else. Setting `--serve-tls=false` falls back to a hard-coded `:80`, which still collides. The intended deployment shape is the standard one — TLS-terminating proxy out front, plain HTTP to Miren on a loopback port inward — but Miren can't bind anywhere except 0.0.0.0.

This PR introduces a new `[ingress]` config section with one field, `http_address`. When set, Miren binds plain HTTP on that address and skips the standard 80/443 binding entirely. `tls.*` settings have no effect in this mode; a `warnIngressTLSOverride` helper logs one warning per non-default `tls.*` field, naming each so operators can grep and clean up. `tls.standard_tls` is intentionally not in the warn list — it defaults to `true`, so warning whenever it's "set" would fire on every default config.

The listener binds synchronously so a port conflict surfaces as a startup error instead of an orphaned goroutine after the supervisor thinks we're up. The accept loop and graceful shutdown run in background goroutines wired to the parent context.

When `ingress.http_address` is empty (default), all existing paths are unchanged.

## Non-goals

HTTPS on a custom port is a related but distinct change — a follow-up PR adds `ingress.https_address`. A third PR introduces a `selectIngressMode` resolver that rejects setting both ingress addresses simultaneously, once there are two fields to be exclusive with.

## Testing

- `TestWarnIngressTLSOverride` in `cli/commands/server_test.go` table-tests each warn-trigger field, including the silent-on-default `standard_tls` case.
- Verified end-to-end on Linux in a containerized standalone-mode boot: `curl http://127.0.0.1:18080/` against the custom address returns Miren's 404; `ss -tln` confirms no listeners on 80 or 443; the warn line fires as expected when `--self-signed-tls` is set alongside.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -825,7 +825,35 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		}
 	}()
 
-	if cfg.TLS.GetStandardTLS() {
+	if ingressAddr := cfg.Ingress.GetHTTPAddress(); ingressAddr != "" {
+		warnIngressTLSOverride(ctx.Log, &cfg.TLS)
+		// Bind synchronously so a port conflict surfaces as a startup error
+		// instead of an orphaned goroutine after the supervisor thinks we're up.
+		ln, err := net.Listen("tcp", ingressAddr)
+		if err != nil {
+			return fmt.Errorf("listen %s: %w", ingressAddr, err)
+		}
+		ctx.Log.Info("serving plain HTTP ingress", "addr", ln.Addr().String())
+		ingressServer := &http.Server{
+			Handler:           hs,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+		eg.Go(func() error {
+			if err := ingressServer.Serve(ln); err != nil && err != http.ErrServerClosed {
+				return fmt.Errorf("HTTP ingress on %s: %w", ingressAddr, err)
+			}
+			return nil
+		})
+		eg.Go(func() error {
+			<-sub.Done()
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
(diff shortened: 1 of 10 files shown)


jcasimir:
Thanks @coderabbitai — both findings addressed in the latest force-push:

- **Test gap** (`standard_tls=true` exclusion): added a table case `"standard_tls true: silent (excluded by design)"` that asserts no warning fires when `standard_tls` is set. The case is the second entry in the table now, so the design invariant gets explicit coverage.

- **Detached goroutines**: switched both `Serve` and `Shutdown` goroutines from `go func()` to `eg.Go(...)`, returning wrapped errors so the parent errgroup observes them. A fatal serve error now terminates the process via `eg.Wait()` instead of being logged-and-forgotten. The same change is mirrored in the helper extracted by the stacked PR #791 (`serveCustomHTTPIngress` now takes the errgroup and uses `eg.Go(...)` internally) so the fix carries through the consolidation.

Verified end-to-end on the local containerized smoke runner: all five smoke scenarios (validation paths + full standalone-mode boot + listener probe + warn-helper assertion) still pass against the updated branch.

evanphx:
Thanks for the PR @jcasimir! I'm 100% behind the idea that our default http listener port needs to be configurable. Where I think we need to figure out the right balance is that this ALSO disables TLS entirely. It seems like it would be better if it were 2 flags, one to set the http ingress port and another to disable TLS listening.

jcasimir:
Ok, so just to clarify you're saying you might want to enable TLS and non-TLS at the same time? I could see that being interesting.

phinze:
Closing in favor of #799, which folds this work into the broader config-schema reshape described in [RFD-84](https://rfd.miren.garden/rfd/84). Full context in the comment thread on #791.

PR #790, opened 2026-05-07
Add ingress.https_address for HTTPS ingress on a custom address

jcasimir:
Companion to #789. Same root motivation — a Miren host already running another application on 80/443 — but for the operator who wants Miren itself to terminate TLS on a non-default port rather than running plain-HTTP behind an external proxy.

This PR adds `ingress.https_address` to the same `[ingress]` section #789 introduces. When set, Miren binds the HTTPS ingress on that address only — no port-80 redirect, no HTTP-01 challenge server, no \`:443\` listener. The cert source must be either DNS-01 ACME (\`tls.acme_dns_provider\`) or \`tls.self_signed\`, since HTTP-01 cannot complete without port 80; \`validateIngressHTTPSAddressConfig\` enforces this at config-load time.

Two new helpers in \`components/autotls/\` carry the actual serving:

- \`ServeTLSWithControllerOnAddr\` parallels the existing \`ServeTLSWithController\` but takes an address and skips the port-80 server.
- \`ServeTLSSelfSignedOnAddr\` does the same for the self-signed path.

Both drop \`acme.ALPNProto\` from \`NextProtos\` since TLS-ALPN-01 also requires port 443.

Validation runs at config-load — right after \`serverconfig.Load(...)\` returns, before any component bootstrap — so a bad cert configuration fails fast with a clear error rather than looking like a containerd or etcd failure thirty seconds later. We learned the hard way that putting validation inside \`Server()\` means containerd dies first on non-root testing boxes.

## Behavior contract

When \`ingress.https_address\` is empty (default), existing \`tls.standard_tls\` paths are unchanged. When it's set, Miren binds HTTPS on that address only; \`tls.standard_tls\` has no effect; cert issuance must use DNS-01 ACME or self-signed.

## Non-goals

- Plain-HTTP-on-custom-port mode — see #789.
- Mutual-exclusion guard for setting \`ingress.http_address\` and \`ingress.https_address\` together — that lands in a third PR alongside the consolidation refactor.

## Schema collision with #789

Both this PR and #789 independently introduce the \`[ingress]\` section with one field each. Whichever lands second will need a small \`pkg/serverconfig/schema.yml\` rebase to merge the field lists; codegen regenerates cleanly from there.

## Testing

- \`TestValidateIngressHTTPSAddressConfig\` table-tests the cert-source validator (DNS-01 path, self-signed path, neither-set rejection, self-signed-wins-over-DNS).
- \`TestServeTLSWithControllerOnAddr\` exercises the parameterized autotls path end-to-end against \`127.0.0.1:0\` with a fake \`CertificateProvider\` and a real TLS handshake.
- Verified end-to-end on Linux in a containerized standalone-mode boot: with \`--self-signed-tls --ingress-https-address=127.0.0.1:18444\`, \`curl -k https://127.0.0.1:18444/\` returns Miren's 404 over a real TLS handshake; \`ss -tln\` confirms no listeners on 80 or 443.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -68,6 +68,15 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		return fmt.Errorf("failed to load configuration: %w", err)
 	}
 
+	// Validate ingress.https_address preconditions before any I/O so a bad
+	// config fails fast instead of waiting for component startup to surface
+	// an unrelated-looking error.
+	if cfg.Ingress.GetHTTPSAddress() != "" {
+		if err := validateIngressHTTPSAddressConfig(&cfg.TLS); err != nil {
+			return err
+		}
+	}
+
 	// Initialize Miren Labs feature flags
 	labs.Init(ctx.Log, cfg.Labs)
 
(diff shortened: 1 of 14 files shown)


phinze:
Closing in favor of #799, which folds this work into the broader config-schema reshape described in [RFD-84](https://rfd.miren.garden/rfd/84). The `ServeTLSWithControllerOnAddr` / `ServeTLSSelfSignedOnAddr` helpers from this PR land essentially unchanged on #799 as the backbone of the new `behind-proxy-https` mode. Full context in the comment thread on #791.

PR #791, opened 2026-05-07
Reject conflicting ingress addresses and consolidate mode selection

jcasimir:
Stacked on #789 and #790. Once those land, this PR's diff against \`main\` shrinks to just the consolidation work — three commits at the tip of the branch.

#789 and #790 each introduce a new \`[ingress]\` field and add their own arm to the \`if/else if\` ladder in \`Server()\` that picks an ingress mode. Naively merged, the two arms compose by source-line ordering — a contract that's never been a deliberate decision, and one nothing prevents an operator from breaking by setting both \`ingress.http_address\` and \`ingress.https_address\` together.

This PR fixes that in three steps:

1. **\`Extract serveCustomHTTPIngress helper\`** — pure refactor of the plain-HTTP wiring out of \`Server()\` into a self-contained helper, in preparation for the switch.

2. **\`Replace ingress if-ladder with selectIngressMode switch\`** — introduces a typed \`ingressMode\` enum (standard TLS, standard HTTP fallback, custom HTTP, custom HTTPS) and a \`selectIngressMode(cfg) (mode, addr, error)\` resolver that owns the precedence rules. The call site in \`Server()\` becomes a flat switch on the resolved mode. \`selectIngressMode\` is invoked at config-load time, right after \`serverconfig.Load(...)\` returns — so any rejection condition (mutual exclusion below; cert-source from #790) fails fast before component bootstrap. Pure refactor — no new behavior.

3. **\`Reject ingress.http_address and ingress.https_address being set together\`** — adds the mutual-exclusion guard inside \`selectIngressMode\`. Operators get \`ingress.http_address and ingress.https_address are mutually exclusive; set at most one\` at startup rather than a silent winner picked by source order.

## Why a third PR rather than one larger PR per side

Each preceding PR ships a narrowly-scoped feature (\"add this knob\") with clean operator UX in isolation. Adding the consolidation to either #789 or #790 alone would either preempt the other PR's design space or grow that PR's diff with refactor work that's only justified when both fields exist. Keeping it separate gives reviewers the option to merge the building blocks on their own timeline without holding back the consolidation.

## Testing

- \`TestSelectIngressMode\` is a new table test in \`cli/commands/server_test.go\` covering the full mode-selection matrix: each mode resolves to the correct \`(mode, addr)\` pair, the cert-source preconditions for custom HTTPS are enforced, and the new mutual-exclusion rule produces the expected error.
- The existing \`TestWarnIngressTLSOverride\` (from #789) and \`TestValidateIngressHTTPSAddressConfig\` (from #790) carry forward unchanged.
- Verified end-to-end on Linux in a containerized standalone-mode boot: setting both \`--ingress-http-address\` and \`--ingress-https-address\` produces the mutual-exclusion error in <200ms with no component startup attempted; running each mode individually binds correctly on its custom address; the cross-PR warn helper from #789 fires by exact field name when \`tls.self_signed\` is left set alongside \`ingress.http_address\`.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -825,7 +825,35 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		}
 	}()
 
-	if cfg.TLS.GetStandardTLS() {
+	if ingressAddr := cfg.Ingress.GetHTTPAddress(); ingressAddr != "" {
+		warnIngressTLSOverride(ctx.Log, &cfg.TLS)
+		// Bind synchronously so a port conflict surfaces as a startup error
+		// instead of an orphaned goroutine after the supervisor thinks we're up.
+		ln, err := net.Listen("tcp", ingressAddr)
+		if err != nil {
+			return fmt.Errorf("listen %s: %w", ingressAddr, err)
+		}
+		ctx.Log.Info("serving plain HTTP ingress", "addr", ln.Addr().String())
+		ingressServer := &http.Server{
+			Handler:           hs,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+		eg.Go(func() error {
+			if err := ingressServer.Serve(ln); err != nil && err != http.ErrServerClosed {
+				return fmt.Errorf("HTTP ingress on %s: %w", ingressAddr, err)
+			}
+			return nil
+		})
+		eg.Go(func() error {
+			<-sub.Done()
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancel()
+			if err := ingressServer.Shutdown(shutdownCtx); err != nil {
+				return fmt.Errorf("HTTP ingress shutdown on %s: %w", ingressAddr, err)
+			}
+			return nil
+		})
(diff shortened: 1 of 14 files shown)


phinze:
Hi @jcasimir, thanks for this PR (and #789 / #790)! I took your work as a starting point and ran with it a bit, and we ended up with a bigger change that resolves some config schema design tensions your improvements exposed.

The end result is RFD-84 at [rfd.miren.garden/rfd/84](https://rfd.miren.garden/rfd/84): a mode-based redesign of `[ingress]` and `[tls]` that names each supported deployment shape directly (`tls-autoprovision`, `behind-proxy-http`, `behind-proxy-https`) and consolidates your two address fields into a single `ingress.address` override. The implementation is up at #799, and the `ServeTLSWithControllerOnAddr` / `ServeTLSSelfSignedOnAddr` helpers you wrote in #790 are doing the heavy lifting for the `behind-proxy-https` mode.

Your feedback on both the RFD and the PR is very welcome; definitely want to make sure this solves for your use case. Going ahead and closing this PR along with #789 and #790 so we can consolidate the conversation over on #799.

PR #792, opened 2026-05-07
Add per-service port_wait_timeout to app.toml

jcasimir:
## Why

`SandboxSpec.PortWaitTimeout` already exists and is honored by `resolvePortWaitTimeout` in `controllers/sandbox/sandbox.go`, but the only callers are addons (`pkg/addon/postgresql`, `pkg/addon/rabbitmq`). Regular apps are stuck with the 15s default constant.

Concrete failure: LiteLLM with the `miren-postgresql:small` addon. Prisma migrations on first DB connect take >15s, the sandbox dies before the app binds 4000, the pool retries forever. We were running a local-only fork bumping the constant globally to 60s — fine as a stopgap, ugly as a fix.

## What

Surface the existing `SandboxSpec.PortWaitTimeout` knob to user `app.toml` as a per-service field:

```toml
[services.web]
command = "..."
port = 4000
port_wait_timeout = "120s"
```

## Design choices worth flagging

- **Per-service, not app-level.** Web and worker have different startup profiles. A worker doing migrations needs 120s; the web process next to it should still fail fast at 15s.
- **Config, not CLI flag.** The slow-boot characteristic is a property of the service, not the deploy — it would be set the same value every push. Belongs in the manifest.
- **ConfigSpec only, not legacy `Config`.** New versions write `mrv.Config = core_v1alpha.Config{}` (build.go:1507) — the legacy inline path is empty for all live deploys, so plumbing the field through `ConfigSpecFromConfig` would be dead code.
- **Validate at parse time.** A typo like `port_wait_timeout = "120"` (missing the unit) would silently fall back to the 15s default — exactly the failure mode the field exists to prevent. Validation in `AppConfig.Validate()` mirrors `scale_down_delay` / `shutdown_timeout`.

## Touch points

| File | Change |
|------|--------|
| `api/core/schema.yml` | New `port_wait_timeout` attr under `config_spec.services` |
| `api/core/core_v1alpha/schema.gen.go` | Regenerated |
| `appconfig/appconfig.go` | `ServiceConfig.PortWaitTimeout` + parse-time `time.ParseDuration` validation |
| `servers/build/build.go` | Copy through in `buildServicesConfig` |
| `controllers/deployment/launcher.go` | Set `sbSpec.PortWaitTimeout` in the existing service-match loop in `buildSandboxSpec` |

## Tests

- `appconfig`: TOML round-trip; parse-time rejection of `"120"` (no unit); accepts `"2m"`
- `servers/build`: appconfig → `ConfigSpecServices` carries the field; sibling stays empty
- `controllers/deployment`: `buildSandboxSpec` propagates into `SandboxSpec` for the matching service only
- `controllers/sandbox`: positive-path cases (`120s`, `2m`) for `resolvePortWaitTimeout`, complementing the existing fallback cases

All six touched packages green locally (sandbox tests run via `golang:1.25` container since the package is `//go:build linux`).


appconfig/appconfig.go
```diff
@@ -72,6 +72,11 @@ type ServiceConfig struct {
 	EnvVars     []AppEnvVar               `toml:"env"`
 	Concurrency *ServiceConcurrencyConfig `toml:"concurrency"`
 	Disks       []DiskConfig              `toml:"disks"`
+	// PortWaitTimeout overrides the default 15s wait for the service to bind
+	// its port during startup. Accepts a Go duration string (e.g. "60s", "2m").
+	// Empty falls back to the default; invalid duration strings are rejected
+	// at parse time by Validate.
+	PortWaitTimeout string `toml:"port_wait_timeout,omitempty"`
```
evanphx:
```suggestion
	PortWaitTimeout string `toml:"port_timeout,omitempty"`
```

evanphx (dismissed):
Minor change, let's have it just be `port_timeout`. Otherwise looks good!

jcasimir:
Great!

phinze (approved):
This is great. Approving and taking it from here. A couple of small things I'll push on top: `port_timeout` didn't make it into `services.md` or `app-toml.md`, and the schema doc string in `api/core/schema.yml` could use a tweak now that `Validate()` rejects invalid input.

Thanks Jeff!

PR #795, opened 2026-05-08
Fix service controller chain leaks and reconcile drift

phinze:
The service controller had a leak-and-silently-fail interaction that bit us during MIR-965 validation of #785 on Toys, and was set up to bite Garden the same way the moment MIR-954 picked up. `Delete()` was a no-op so chains accumulated forever, and `Create()`'s atomic nft applies would fail silently in a tight loop whenever the kernel state had drifted. Across a version upgrade where the chain hash function shifts, the two compose: new code references state that doesn't match the kernel, the apply aborts, the reconciler retries every minute, and the cluster looks healthy by every other signal.

The fix is a periodic GC pass that diffs the entity store against the kernel and prunes orphan chains and stale map elements. Service deletion happens automatically, chains fall out of the target set on the next tick. Upgrades self-heal for the same reason. nft processes batch deletes sequentially even inside an atomic transaction, so the GC orders carefully: elements first, then service and nodeport chains, endpoint chains last.

While in there, the nft management itself moves onto `sigs.k8s.io/knftables`, the typed Go API kube-proxy uses for its own nftables backend. Same `nft -f -` shell-out underneath, but typed objects for chains, sets, maps, and elements instead of hand-built command strings, and `List*` methods that return parsed structs instead of JSON the controller unmarshals itself. It was already in our dependency closure transitively via the CNI plugins, so promoting it to a direct dep was free. Knock-on effect: a typed `chainKind` enum classifies chains by ownership in one place, replacing scattered prefix-string lists.

Bonus fix: `Init` now does `Add+Flush+rules` on the static dispatcher and masq chains on every startup, which automatically cleans up the duplicate `jump services` rules the original bug report mentioned in passing. That's the third symptom from MIR-1104 closed off without dedicated code.

`UpdateEndpoints`' delete path also gets simplified now that the framework carries pre-delete entities through to delete events (via the etcd `WithPrevKV` watch option). Endpoint deletion now reconciles the one affected service via `eps.Service`, instead of fanning out across every Service entity in the cluster as a workaround for the missing entity value.

Apply-failure observability (entity status fields, metrics) is intentionally not in this PR. With the GC pass keeping kernel state in sync, the silent-fail-loop scenario doesn't have a way to recur, and the reconcile framework already logs `Run()` errors. Worth a separate ticket as observability hygiene rather than tacked on here.

Closes MIR-1104.


controllers/service/gc.go
```diff
@@ -0,0 +1,295 @@
...
+	case chainServices, chainNATPrerouting, chainNATOutput,
+		chainNATPostrouting, chainMarkForMasq, chainMasq:
+		return true
+	}
+	return false
+}
+
+// managedMaps lists the verdict maps the controller writes elements into.
+// Elements outside these maps are not the controller's concern.
+var managedMaps = []string{mapServiceIP4s, mapServiceIP6s, mapServiceNodePort}
```
evanphx, line 60: Is a global the right place for this data?

evanphx (approved):
Good cleanup! Just curious about this random global variable.

controllers/service/gc.go
phinze: Interesting, this felt like a pretty standard pattern for shaking const-esque things out of method bodies (Go won't let me const a slice). Curious what's giving you the smell here. Is it the lone-`var`-among-otherwise-constants-and-functions thing, or something deeper about where the data lives?

If "tucked closer to the controller" would feel better, easy move:

```go
func (s *ServiceController) managedMapNames() []string {
    return []string{mapServiceIP4s, mapServiceIP6s, mapServiceNodePort}
}
```

Same data, just scoped under the struct conceptually. Happy to land that or hear what you'd have in mind.

evanphx: OH OH. you know, I misread this. I thought it was a mutable global. Nevermind!

PR #799, opened 2026-05-11
Reshape ingress config around named modes (RFD-84)

phinze:
This work started because @jcasimir wanted to run Miren behind nginx in production, and Miren didn't quite know how to be run that way yet. He opened #789, #790, and #791 to make it work, extending the existing TLS configuration along the lines its surface seemed to invite. Reading the three PRs as a set, though, surfaced a deeper tension: the ingress and TLS sections together were trying to answer three separate questions (where to listen, whether to terminate TLS, where certs come from) through knobs that each spanned multiple of those answers, and the cross-axis coupling was forcing compensating validation helpers to keep things coherent. We landed on a mode-based redesign that names each deployment shape directly, which RFD-84 walks through in full.

This PR is the implementation. Three modes now: `tls-autoprovision` (default, behavior unchanged from today), `behind-proxy-http` (plain HTTP, defaults to `127.0.0.1:80`), and `behind-proxy-https` (TLS terminated by Miren on a single listener, defaults to `127.0.0.1:443`). An optional `ingress.address` field replaces any mode's default bind, so operators on multi-homed hosts, Tailscale interfaces, IPv6, or non-standard ports can bind wherever they need. Unix sockets are reserved as a future `address` value and rejected today with a clear "not yet supported" error.

One real behavior change comes with this. The autoprovision `:80` handler used to have a Host-based exception that routed `localhost`, `127.0.0.1`, `::1`, and any raw-IP Host directly to the default route app over plain HTTP, bypassing the HTTPS redirect. That was a dev-convenience shortcut that became a small production smell, since operators didn't always realize that raw-IP access could reach apps without TLS. It's gone now. The autoprovision path does only what its name suggests, and operators who want plain-HTTP access for dev workflow pick `behind-proxy-http` explicitly.

One small ergonomic regression worth flagging in dev: `curl http://localhost/` (no scheme, defaults to HTTP) used to hit the default route directly via that side door. Now it 302s to `https://localhost/`. The other dev shortcuts still work, though: `curl -k https://localhost/` and `curl -kL http://localhost/` both reach the default route as before. If the friction proves annoying, RFD-84 leaves room for a dev-only opt-in flag in a follow-up.

`tls.standard_tls` retires entirely. The dev environment already used `--self-signed-tls` rather than that field, so the dev loop is unchanged; I caught and updated a few stranded references in the systemd install path, the embedded systemd entrypoint test script, and the example `server.toml`.

Big thanks to @jcasimir for the use case and the implementation work; both set the stage for this change. His autotls helpers from #790 are the backbone of the new `behind-proxy-https` mode. The three PRs close out once this lands.

Closes [MIR-1115](https://linear.app/miren/issue/MIR-1115).

### DIFF
--- a/pkg/serverconfig/cli.gen.go
+++ b/pkg/serverconfig/cli.gen.go
@@ -22,6 +22,8 @@ type CLIFlags struct {
 	EtcdConfigPeerPort                   *int     `long:"etcd-peer-port" description:"Etcd peer port"`
 	EtcdConfigPrefix                     *string  `long:"etcd-prefix" short:"p" description:"Etcd prefix"`
 	EtcdConfigStartEmbedded              *bool    `long:"start-etcd" description:"Start embedded etcd server"`
+	IngressConfigAddress                 *string  `long:"ingress-address" description:"Optional bind override. Replaces the mode's default bind entirely (interface and port). Ignored for tls-autoprovision. Reserved unix:/path prefix is not yet supported."`
+	IngressConfigMode                    *string  `long:"ingress-mode" description:"Ingress mode: tls-autoprovision (default, :443 + :80 with ACME or self-signed), behind-proxy-http (plain HTTP for use behind a TLS-terminating proxy), behind-proxy-https (TLS terminated by Miren with externally-provided certs, no ACME provisioning)"`
 	ServerConfigAddress                  *string  `long:"address" short:"a" description:"Address to listen on (host:port). For IPv6 use brackets, e.g. \"[::1]:8443\"."`
 	ServerConfigConfigClusterName        *string  `long:"config-cluster-name" short:"C" description:"Name of the cluster in client config"`
 	ServerConfigDataPath                 *string  `long:"data-path" short:"d" description:"Data path"`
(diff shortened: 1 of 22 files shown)


pkg/serverconfig/cli.gen.go
```diff
@@ -38,7 +40,6 @@ type CLIFlags struct {
 	TLSConfigAdditionalIPs               []string `long:"ips" description:"Additional IPs assigned to the server cert"`
 	TLSConfigAdditionalNames             []string `long:"dns-names" description:"Additional DNS names assigned to the server cert"`
 	TLSConfigSelfSigned                  *bool    `long:"self-signed-tls" description:"Use self-signed certificates for TLS (for development/testing only)"`
-	TLSConfigStandardTLS                 *bool    `long:"serve-tls" description:"Expose the http ingress on standard TLS ports"`
```
evanphx, line 41: We should still include this option but ignore it so that upgrades don't break.

evanphx (changes requested):
Looks good, but we need to accept --start-tls as a noop (maybe mark it hidden so it's not normally used), so that upgrades don't break.

phinze:
Yeah fair call I was waffling on the hard-line removal but better to be nicer

phinze:
Addressed in `1145ee7c`. The flag, env var, and TOML field all parse again. The field's never read by any code path so behavior matches the new `tls-autoprovision` default; a startup `WARN` fires when it's explicitly set so operators get nudged toward `ingress.mode`. Tried to mark it hidden in `--help` via a new `cli.hidden` schema knob, but `mflags` doesn't read the `hidden` tag yet so the flag still shows in help with a "Deprecated and ignored" description. Hiding it for real is a small follow-up against `mflags` if we want it.

evanphx (approved):

phinze:
@jcasimir I clicked merge but still very open to any feedback you have at any point! Definitely want to make sure this works for your use case 👍

PR #801, opened 2026-05-13
Reconcile service controller parent chain rules on each GC pass

phinze:
The service controller's periodic GC loop on toys-runner-2 was failing every five minutes with "Device or resource busy" trying to delete an endpoint chain that a live parent chain was still pointing at. Pre-#795 this was silent. After #795 it became an `[ERROR]` log every cycle.

Tracing back: `setEndpoints` early-returned on empty endpoint sets. When an Endpoints entity transitioned from "one endpoint" to "no endpoints" (exactly what happens when a sandbox dies and its backing rotates), the parent's `numgen ... vmap { 0 : goto endpoint_<old> }` stuck around forever. The old endpoint chain was orphan from the entity store's perspective, so GC tried to delete it, but the live parent's stale reference made the kernel refuse, and the reconcile loop kept retrying the same doomed delete.

The fix has two layers. `setEndpoints` now flushes the parent and writes a `counter + drop` body when there are no endpoints, so future zero-endpoint transitions don't leak a stale vmap. `Periodic` also flushes and rebuilds every live service and nodeport chain body before the orphan-delete pass. This is the self-heal path: it catches the existing damage on Toys today plus any future drift class we haven't enumerated, and it means the orphan endpoint chain has no inbound references when the delete fires in the same transaction.

Two ordering subtleties surfaced while deploying to toys-runner-2. First, pre-#795 nodeport chains have bodies containing `goto service_*` rules (the old code never flushed, so rule sets accumulated across redeploys). If both the orphan nodeport and orphan service are queued for deletion in one transaction, the service can't be dropped while the nodeport still references it. So orphan-parent deletes are split into nodeports-first, then services. Second, the chain-hash function bumped at some point (exactly the scenario #795 mentioned), so the entity store wants dispatcher map entries pointing at new-hash chain names while the kernel still has old-hash names. The reconcile path tried to `add element` with the new value, but nft refuses to overwrite an existing key. Reordering so stale dispatcher elements drop before reconcile re-adds takes care of it.

A note on the empty-endpoint choice: `drop` rather than `reject` because the parent chain is reachable from nat-prerouting, where `reject` isn't permitted. Result is a connect timeout instead of `ECONNREFUSED`. Fast-fail would require splitting into a separate filter base chain (kube-proxy-style), which is a bigger refactor than this bug needs.

Verified on toys-runner-2: the stuck `endpoint_DqRxf9quDgDkr2C1RUM5wdhmBNS41LM6gUb5j4i2Bnqp` chain is gone, `service_GAGhE...` rebuilt to `counter + drop`, ~30 orphan endpoint chains in the dead 10.8.54.x subnet cleaned up, no `runner.reconcile.service` errors since the rev2 binary started.

Closes MIR-1129

### DIFF
--- a/controllers/service/gc.go
+++ b/controllers/service/gc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/netip"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


evanphx (approved):
Image

PR #803, opened 2026-05-13
rpc/stream: opt-in batching for bulk ServeReader callers

phinze:
Today's cloud production deploy crawled at 2.4 KB/s against an 80 MB tarball, which works out to about nine hours just for the upload. Past deploys on the same code path ran 300 to 450 KB/s, which is slow but workable. Something cratered, hard.

The miren CLI streams the deploy tarball over a pull-based RPC: the build server calls `Recv(N)` on the client and the client's `serveReader.Recv` returns whatever a single underlying `s.r.Read(buf)` happens to yield. That's one RPC round trip per underlying Read, regardless of how big the request was. For most of the past year `tarx` used `os.Pipe()`, which has a ~64 KB kernel buffer, so the many small writes that `gzip.Writer` produces piled up in the buffer and `serveReader.Recv`'s single Read pulled them all back at once. The bug was real but the kernel was quietly absorbing the damage.

On 2026-04-29 we switched `tarx` from `os.Pipe` to `io.Pipe` so we could surface walk errors via `CloseWithError`. `io.Pipe` is a strict synchronous handoff with no buffer, so each gzip flush became its own Recv response and the latent single-Read bug went from "annoying" to "deploys take all day". Today was the first production deploy in three weeks, so today is the first one to feel it.

The fix lives in `serveReader.Recv` but it needed a knob, because the same function ships interactive stdin for `sandbox exec` and `app run`. Unconditional batching would wedge those streams (CodeRabbit caught this, thank you CodeRabbit). So we get `WithBulkBatching()`, an opt-in `ServeReaderOption` that flips on a 1 MB `io.ReadAtLeast` floor. Bulk callers (`BuildFromTar`, `BuildFromPrepared`, `AnalyzeApp`) pass it; interactive callers stay bare and keep their per-keystroke responsiveness. Throughput on a bulk stream becomes roughly `batch/RTT`, around 10 MB/s on a 100 ms link, regardless of whether the source is `os.Pipe`, `io.Pipe`, a slow disk, or whatever else the future brings.

The Linear writeup at MIR-1135 has the full history (this is the third time we've tangled with this code path, and each prior round was treating symptoms).

Closes MIR-1135

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -601,10 +601,10 @@ func Deploy(ctx *Context, opts struct {
 	}
 	buildCall := func(callCtx context.Context, tarReader io.ReadCloser, cb stream.SendStream[*build_v1alpha.Status]) (buildResults, error) {
 		if useOptimized {
-			tarStream := stream.ServeReader(callCtx, tarReader)
+			tarStream := stream.ServeReader(callCtx, tarReader, stream.WithBulkBatching())
 			return bc.BuildFromPrepared(callCtx, sessionID, tarStream, cb, envVars, ephemeralLabel, ephemeralTTL)
 		}
-		return bc.BuildFromTar(callCtx, name, stream.ServeReader(callCtx, tarReader), cb, envVars, ephemeralLabel, ephemeralTTL)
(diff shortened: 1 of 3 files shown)


evanphx (approved):
Image

PR #804, opened 2026-05-14
Stable upload progress and honest ETA in miren deploy

phinze:
While babysitting a slow deploy I noticed the progress line was lying to me. The "total" in `Uploading artifacts: 0% — 24 KB / ~45.7 MB at 2.4 KB/s` was swinging wildly: it started at ~974 KB, climbed steadily to ~80 MB over a few minutes, dropped back to ~20 MB, climbed again. Useless for figuring out how long the upload had left.

Tracing the math, the underlying bug was that `EstimatedTotalBytes = BytesRead / Fraction` was mixing two counters sampled on different sides of the gzip pipeline. `written` is uncompressed source bytes that have entered tar/gzip, while `BytesRead` is compressed bytes drained by the HTTP uploader on the other side of a gzip buffer plus io.Pipe. Under network back-pressure they advance in bursts at very different ratios, and their quotient is meaningless.

The fix is to stop pretending we know the compressed total. We don't, and projecting it from the runtime ratio doesn't work. The percentage itself was always fine, because both numerator and denominator are uncompressed bytes from the delta-filtered subset of files, so it correctly reports "fraction of the upload work consumed."

Since estimating remaining time is genuinely useful on slow uploads (the calculator-free "this will take 8 hours" use case), I also wove in a time-domain ETA: `elapsed * (1 - frac) / frac`. That formula sidesteps the unit-mixing trap entirely. There's a 5s warmup so early wild numbers don't show, and the display uses a tilde (`eta ~8h 15m`) to be upfront about being approximate.

The new line reads `Uploading artifacts: 1% — 1.0 MB at 2.4 KB/s (eta ~8h 15m)`, in both explain mode and the TUI.

Closes MIR-1136

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -640,11 +640,14 @@ func Deploy(ctx *Context, opts struct {
 			if progress.Fraction > 0 && time.Since(lastPrintTime) >= 500*time.Millisecond {
 				lastPrintTime = time.Now()
 				fmt.Fprintf(os.Stderr, "\r\033[K") // Clear to end of line
-				fmt.Fprintf(os.Stderr, "Uploading artifacts: %d%% — %s / ~%s at %s",
+				line := fmt.Sprintf("Uploading artifacts: %d%% — %s at %s",
 					int(progress.Fraction*100),
 					upload.FormatBytes(progress.BytesRead),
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Image

PR #139, opened 2025-07-22
Improve BuildKit lifecycle logging and retry client connection

phinze:
- **Add comprehensive logging to buildkit connection process**
- **Initialize buildkit cache to avoid warning**

## Summary by CodeRabbit

* **New Features**
  * Enhanced logging throughout the build process, including detailed debug and informational messages during server setup, buildkit client creation, and build phases.
  * Improved error reporting for buildkit cache initialization and client connection issues.

* **Bug Fixes**
  * Increased reliability when establishing connections to the buildkit server by introducing retry logic with exponential backoff.

* **Chores**
  * Added automatic creation of a minimal cache file to prevent first-time warning logs from BuildKit.

PR #806, opened 2026-05-14
coordinate: stop advertising unverified public DiscoveredIPs

phinze:
MIR-1117 / #139 fixed one half of a two-path bug: a non-miren listener on 443 could complete a TLS handshake during netcheck, and the cluster would advertise its WAN IP based on that false positive. PR #139 added cert verification so we only trust responses that prove they're from miren-runtime. That closed Path A.

This is Path B. When netcheck legitimately reports zero reachable (e.g. port 443 isn't forwarded at all), `apiAddresses()` was falling back to every IP in `DiscoveredIPs`, which still included the WAN IP echoed by `ipdiscovery`'s startup netcheck. Same bad outcome: cloud writes `cluster-X.miren.systems A <unreachable WAN IP>`, the POP fleet never gets a turn, external clients hit a closed port. I caught this with a fresh home cluster behind NAT.

The realization on the way to the fix was that public DiscoveredIPs were never authoritative on port reachability in the first place. Interface IPs don't know about firewalls. The startup netcheck only echoes the source IP back; nothing about it proves port 8443 is reachable from outside. The coordinator's port-probing netcheck, surfaced via `publicAddresses()`, is the only signal that actually does. So the fallback wasn't just buggy in the firewalled case, it was structurally wrong as a fallback. There was no scenario where a public DiscoveredIP was correct without netcheck also having said yes.

Fix: drop public DiscoveredIPs from `apiAddresses()` entirely. Private, loopback, and link-local DiscoveredIPs still flow through for in-LAN clients. A public IP in the status report now requires explicit netcheck confirmation, full stop. The bug becomes structurally impossible.

A nice side effect: the existing `AdditionalIPs` mechanism (user-configured, unfiltered) becomes the explicit opt-in path. Users with a Tailscale address or some other non-publicly-reachable interface they want resolved via DNS can just set it there. Default-off-with-opt-in falls out for free.

Closes MIR-1139

### DIFF
--- a/components/coordinate/api_addresses_test.go
+++ b/components/coordinate/api_addresses_test.go
@@ -27,12 +27,20 @@ func TestApiAddresses(t *testing.T) {
 		wantExcludes   []string
 	}{
 		{
-			name:          "no netcheck with discovered public IPs",
+			// Public discovered IPs are never advertised: an interface IP or
+			// startup-netcheck WAN echo doesn't prove port 8443 is reachable
+			// from outside. Only the coordinator's port-probing netcheck can.
+			name:          "no netcheck: public DiscoveredIPs are not advertised, private flows through",
 			discoveredIPs: []net.IP{publicIPv4, privateIP},
-			wantContains:  append(localhost, "203.0.113.10:8443", "10.0.0.5:8443"),
+			wantContains:  append(localhost, "10.0.0.5:8443"),
+			wantExcludes:  []string{"203.0.113.10:8443"},
 		},
 		{
-			name:          "netcheck ran but found nothing reachable",
+			// MIR-1139 repro: cluster is firewalled (port 8443 closed from outside).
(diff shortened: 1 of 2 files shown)


evanphx (approved):

phinze:
Closing as a duplicate of [MIR-1018](https://linear.app/miren/issue/MIR-1018/unusable-addresses-still-present-in-discovery), which @evanphx already has a more thorough fix for on `mir-1018-unusable-addresses-still-present-in-discovery` ([dd03230d](https://github.com/mirendev/runtime/commit/dd03230d2178a2c9aea6595b2664b956e709dab8)). His version trusts negative netcheck per family, factors `ComputeAdvertise` out as a pure function (debug advertise also benefits), and filters CGNAT explicitly. Just never got opened as a PR. We'll land that instead.

PR #808, opened 2026-05-14
coordinate: trust negative netcheck per family, filter CGNAT

phinze:
Opening this on Evan's behalf since the branch has been sitting since April 15. Original commits are his (`dd03230d` "Fix MIR-1018" and `8eb63ab` "Add 'debug advertise' command"); I'm just turning the crank after rediscovering the same bug as MIR-1139.

Two problems addressed.

The advertise-address logic in `coordinate.apiAddresses` was treating "netcheck ran but found zero reachable" as identical to "netcheck didn't run at all" and falling back to discovered IPs in both cases. That fallback would push the WAN IP of a firewalled cluster (or a tailnet address from `tailscale0`) into the status report, and cloud would happily write A records to those unreachable IPs instead of routing through POP. Per-family state fixes this: an IPv4 success no longer silently drops discovered IPv6 public addresses, and a per-family "ran and found nothing" verdict is now trusted rather than papered over.

Addresses in 100.64.0.0/10 (RFC 6598) are also filtered out of the discovered list by default. This range is used by Tailscale and by ISP CGNAT, neither of which is reachable from a generic external client. Users who genuinely want a CGNAT address advertised can set it explicitly via `AdditionalIPs`, which passes through unfiltered.

The logic itself was factored into a pure `ComputeAdvertise` function so the production path and the `debug advertise` command share the same decision tree. That makes the debug command actually useful for diagnosing "why is my cluster advertising X" without needing to start a full coordinator.

Closes MIR-1018
Closes MIR-1139 (duplicate)

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -881,6 +881,7 @@ Warning: These commands are intended for advanced users and developers. They may
 
 `))
 	d.Dispatch("debug connection", Infer("debug connection", "Test connectivity and authentication with a server", DebugConnection))
+	d.Dispatch("debug advertise", Infer("debug advertise", "Show which addresses the server would advertise and why", DebugAdvertise))
 	d.Dispatch("debug reindex", Infer("debug reindex", "Rebuild all entity indexes from scratch", DebugReindex))
 	d.Dispatch("debug test", Section("debug test", "Debug test commands", ""))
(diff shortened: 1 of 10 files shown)


evanphx (approved):
OMG thank you!

PR #809, opened 2026-05-14
coordinate: address CodeRabbit feedback from #808

phinze:
Follow-up to #808 (MIR-1018), which got merged before CodeRabbit's review was addressed. Four findings from CR, all valid; this PR works through them.

**1. `cli/commands/server.go`: parse `AdditionalIPs` before `SetupEtcdTLS`.** The IP set was being handed to `coordinate.SetupEtcdTLS()` at line 299 with only the discovered IPs in it; explicit `cfg.TLS.AdditionalIPs` weren't parsed in until ~220 lines later. The etcd server cert SANs would silently omit user-configured addresses, so a distributed runner reaching the cluster over an explicit `AdditionalIP` would fail TLS verification against etcd. Moved the parse up to right after IP discovery so the set is complete before any cert-issuing code reads it.

**2. `coordinate.PublicIPs()`: route through `ComputeAdvertise`.** This function feeds `AutocertController.dnsPointsToUs`, so getting it wrong means our ACME provisioning trusts addresses we shouldn't. The pre-refactor version had the same two leaks `apiAddresses()` had before MIR-1018: it returned `resp.SourceAddress` even when that family had zero reachable ports, and its fallback admitted `100.64.0.0/10` (CGNAT / Tailscale) through the bare `IsGlobalUnicast`/`IsPrivate` check. Routed it through `ComputeAdvertise` and filtered to candidates classified as `global-unicast` so both gaps close in one move.

**3. `debug_advertise`: warn-and-continue on `ipdiscovery` failure.** The command was aborting where `server.go` just warns. That made `debug advertise` unable to reproduce the very failure mode it exists to diagnose. Mirrored `server.go`'s behavior so it falls through to the explicit-IP and netcheck paths with an empty discovery result.

**4. `debug_advertise`: add `--format json`.** Standard `FormatOptions` pattern for list-style commands. Useful for piping the candidate decisions into tooling. Suppresses the human progress logs in JSON mode so stdout is parseable; emits a `{candidates: [...], advertised: [...]}` document. Regenerated the command doc to pick up the new flag.

Verified locally with `gotestsum ./components/coordinate` (24 tests pass) and `make bin/miren && go generate ./...` produces no diff after the commit.

### DIFF
--- a/cli/commands/debug_advertise.go
+++ b/cli/commands/debug_advertise.go
@@ -16,6 +16,7 @@ import (
 // the final advertised list, so we can debug cases where the server
 // advertises addresses that aren't actually reachable from clients.
 func DebugAdvertise(ctx *Context, opts struct {
+	FormatOptions
 	CloudURL      string   `long:"cloud-url" description:"Cloud URL to use for netcheck (default: https://api.miren.cloud)"`
 	SkipNetcheck  bool     `long:"skip-netcheck" description:"Skip the netcheck call and only report interface scan"`
 	AdditionalIPs []string `long:"additional-ip" description:"Simulate a server-configured AdditionalIP (repeatable)"`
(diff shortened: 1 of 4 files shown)


phinze:
lol I raced a session fixing coderabbit things w/ my own twitchy merge button pushes, and this is what i get

evanphx (approved):

PR #810, opened 2026-05-14
Detect and install npm/bun automatically (MIR-1132)

evanphx:
## Summary

- Adds a stack-augmentation system in `pkg/stackbuild/` that detects when a non-Node/Bun primary stack ships a `package.json` and layers npm (and/or bun) onto its base image — so Rails/Django/Go/Rust apps with JS-driven frontends Just Work without per-app config.
- Each augmentation installs the tool (`apt` for Debian images, `apk` for Alpine/Go) and then runs `npm install` / `bun install` as the unprivileged `app` user (uid 2010) so the resulting `node_modules` is writable at runtime.
- If the app already ships `node_modules`, the tool is still installed (so onBuild commands can shell out) but the install step is skipped — the vendored modules flow through via `copyApp`.

## Implementation notes

- New `Augmentation` type + `DetectAugmentations` in `pkg/stackbuild/augment.go`.
- `Stack` interface gains `BaseDistro()` (`debian` / `alpine`) so the augmentation helpers can dispatch the right package manager.
- `MetaStack` gains `augmentations` + `skipJSInstall` fields populated by `attachAugmentations` after primary stack detection; detection events surface alongside primary events.
- bun augmentation copies the bun binary from the official bun image via multi-stage `llb.Copy`.
- npm install runs without a persistent cache mount (the scratch root is root-owned, which conflicts with running as `app`); per-layer caching still avoids re-runs when package files don't change. A future PR can add a properly-owned cache via `llb.SourcePath`.

## Test plan

- [x] Unit tests in `pkg/stackbuild/augment_test.go` cover: npm-from-package-lock, npm-from-package-json, bun-from-lock/lockb, skip-for-node, skip-for-bun, node_modules skip path, attachment via `DetectStack`, base distros for all six stacks (14 tests, all passing).
- [x] `go build ./pkg/stackbuild/` ✓
- [x] `make lint` ✓
- [ ] End-to-end: build a Rails app with `package.json` + `dartsass-rails` and verify `assets:precompile` succeeds.
- [ ] End-to-end: build a Python app with a vendored `node_modules` and verify the skip path works.

### DIFF
--- a/pkg/stackbuild/augment.go
+++ b/pkg/stackbuild/augment.go
@@ -16,8 +16,9 @@ import (
 type Augmentation string
 
 const (
-	AugNpm Augmentation = "npm"
-	AugBun Augmentation = "bun"
+	AugNpm  Augmentation = "npm"
+	AugYarn Augmentation = "yarn"
+	AugBun  Augmentation = "bun"
 )
 
 // DetectAugmentations returns the set of augmentations needed for the app
@@ -42,15 +43,23 @@ func DetectAugmentations(dir, primaryStack string) (augs []Augmentation, skipIns
 	}
 
 	if hasRegularFile(dir, "package.json") {
-		augs = append(augs, AugNpm)
 		switch {
+		case hasRegularFile(dir, "yarn.lock"):
+			augs = append(augs, AugYarn)
+			events = append(events, DetectionEvent{
+				Kind:    "augmentation",
+				Name:    "yarn",
+				Message: "Found yarn.lock — installing yarn via corepack",
+			})
 		case hasRegularFile(dir, "package-lock.json"):
+			augs = append(augs, AugNpm)
 			events = append(events, DetectionEvent{
 				Kind:    "augmentation",
 				Name:    "npm",
 				Message: "Found package-lock.json — installing npm",
 			})
 		default:
+			augs = append(augs, AugNpm)
 			events = append(events, DetectionEvent{
 				Kind:    "augmentation",
 				Name:    "npm",
@@ -138,6 +147,11 @@ func (h *highlevelBuilder) applyAugmentations(cur, localCtx llb.State, distro st
 			if !skipInstall {
 				cur = h.runNpmInstall(cur, localCtx)
 			}
+		case AugYarn:
+			cur = h.installYarn(cur, distro)
+			if !skipInstall {
+				cur = h.runYarnInstall(cur, localCtx)
+			}
 		case AugBun:
 			cur = h.installBun(cur)
 			if !skipInstall {
@@ -146,6 +160,17 @@ func (h *highlevelBuilder) applyAugmentations(cur, localCtx llb.State, distro st
 		}
 	}
 	return cur
+}
+
+// installYarn layers nodejs+npm and enables corepack so the yarn shim is on
+// PATH. Yarn itself is provisioned lazily by corepack from package.json's
+// "packageManager" field (or falls back to classic yarn 1).
+func (h *highlevelBuilder) installYarn(cur llb.State, distro string) llb.State {
+	cur = h.installNpm(cur, distro)
+	return cur.Run(
+		llb.Shlex("corepack enable"),
+		llb.WithCustomName("[phase] Enabling corepack for yarn"),
+	).State
 }
 
 // ensureAppDir creates /app owned by the app user so the install can run
@@ -169,7 +194,7 @@ func (h *highlevelBuilder) runNpmInstall(cur, mnt llb.State) llb.State {
 	cur = h.ensureAppDir(cur)
 
 	cur = cur.File(llb.Copy(mnt, "/", "/app", &llb.CopyInfo{
-		IncludePatterns:    []string{"package.json", "package-lock.json", "yarn.lock"},
+		IncludePatterns:    []string{"package.json", "package-lock.json"},
 		CreateDestPath:     true,
 		AllowWildcard:      true,
 		AllowEmptyWildcard: true,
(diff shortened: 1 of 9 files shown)


pkg/stackbuild/augment.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func hasDir(dir, name string) bool {
+	st, err := os.Stat(filepath.Join(dir, name))
+	return err == nil && st.IsDir()
+}
+
+// installNpm layers nodejs+npm onto cur using the package manager appropriate
+// for the given base distro.
+func (h *highlevelBuilder) installNpm(cur llb.State, distro string) llb.State {
```
phinze, line 109: Sharpest version of the RFD-85 aesthetic question lands here — `installNpm`'s default branch does `apt-get install nodejs npm`, the bun analog right below does `llb.Copy` from a pinned image we mirror. Resolving the aesthetic question one way or the other will reshape this function. See RFD-85 for the broader frame; the proposed near-term resolution is to mirror the bun pattern (copy from a controlled `oci.miren.cloud/node:20-slim`, with a distro split for Go's Alpine base).

pkg/stackbuild/ruby.go
```diff
@@ -222,6 +226,8 @@ func (s *RubyStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, erro
 	// My kingdom for a pipe operator.
 	base = h.aptInstall(base, "build-essential", "libpq-dev", "nodejs", "libyaml-dev", "postgresql-client", "git", "curl", "ssh")
```
phinze, line 229: Heads up: Ruby's apt list already includes `nodejs`, and with this PR the augmentation also installs `nodejs npm` on top. Bit of wasted layer + build time. If the npm augmentation moves to `llb.Copy` from a controlled image per RFD-85, this resolves naturally — the Ruby base wouldn't need nodejs at all because the augmentation handles it when needed.

pkg/stackbuild/stackbuild.go
```diff
@@ -87,18 +91,66 @@ func DetectStack(dir string, opts BuildOptions) (Stack, error) {
...
+		ms.augmentations = augs
+		ms.skipJSInstall = skipInstall
+		ms.events = append(ms.events, events...)
+	}
+}
+
+// metaStackOf returns the embedded *MetaStack from a Stack implementation.
+// All concrete stacks in this package embed MetaStack, so this should always
+// succeed for stacks created by DetectStack.
+func metaStackOf(stack Stack) (*MetaStack, bool) {
```
phinze: Mild footgun for future-us: adding a new stack means remembering to add a case here, with no compile-time nudge if you forget. An unexported `metaStack() *MetaStack` method on the Stack interface (auto-satisfied by the embedded MetaStack) would make this impossible to miss. Current code works, just less robust to future-stack churn.

phinze (changes requested):
Took our time with this one — initially read as a straightforward
augmentation system, but tracing through the npm and bun halves
surfaced an aesthetic question big enough that we wrote it up as
RFD-85 (Open Auto-Builds).

Short version: the bun half installs via `llb.Copy` from a controlled
image, the npm half installs via `apt-get install nodejs npm` against
whatever Debian Bookworm pinned. Same PR, two aesthetics. The RFD asks
the team to commit to a principle that resolves which one to keep —
and names this PR's npm path as the immediate test case for that
principle.

**The concrete ask, assuming the RFD lands as proposed:**

Swap the npm augmentation install path from `apt-get install nodejs
npm` to `llb.Copy` from a mirrored `node:20-slim` (alpine variant for
Go's base), with a small version-resolver reading `engines.node` and
falling back to a curated default. Mirror `oci.miren.cloud/node:20-slim`
and the alpine variant in image-sync. The bun half is already in this
shape — npm just needs to match.

Holding REQUEST_CHANGES while the principle is being agreed to. If
the team punts on the RFD or decides not to commit to the aesthetic,
this flips to APPROVE with comments — the existing implementation
works fine on its own merits.

**Other items:**

CodeRabbit caught the `yarn.lock` thing and we agree — `yarn.lock` is
in `runNpmInstall`'s IncludePatterns but the command is `npm install`.
Same aesthetic problem actually: detection should respect what the
app says about itself. The cleanest resolution is corepack-aware
install (falls out of "node from a controlled image" naturally — node
16+ ships corepack), but at minimum we shouldn't silently pretend by
copying `yarn.lock` and ignoring it.

Test plan still has two unchecked e2e items — Rails + dartsass-rails
and Python + vendored node_modules. The dartsass one would catch the
node-version concern directly; the vendored-node_modules one
exercises the skipInstall path (only unit-tested today).

Real love for the comments throughout — the cache-mount-why and the
skipInstall rationale especially. "Explain the constraint, not the
code" ages well.

Inline notes below for the smaller things.

---

(P.S. — naming the type `Augmentation` puts a particular image in our
heads. Couldn't unsee it.)

Deus Ex augmented humans

pkg/stackbuild/augment.go
evanphx: The concern is that I don't believe node is single binary at all, so copying from it's image is copying a large, somewhat random chunk of files. The reason is that node links libssl and a bunch of other stuff.

pkg/stackbuild/ruby.go
evanphx: Yeah, I did that on purpose. I didn't want nodejs only included if npm was active because some rubygems use node directly without a package.json.

phinze (approved):
Coming back after we sat down and talked it through — appreciated the patient walk-through. The headline: APPROVING for merge as-is. The principle in our internal stackbuild philosophy doc (RFD-85) still stands (and we're going to rewrite it to be clearer about what it's actually saying), but the specific near-term ask in this review was based on a misread.

Two corrections we owe to Evan, captured in the inline threads above:

- **node isn't a single binary like bun.** `llb.Copy` of node out of the official image isn't the move — it's ~500 files linking libssl and friends. Multi-stage `COPY` from a node image is a real pattern, but it's used to *extract artifacts* (run npm + asset compile, then copy the output out), not to install node into a base image. The "mirror the bun pattern" recommendation was technically wrong.

- **Ruby's apt list including nodejs is load-bearing.** Some rubygems shell out to node directly without a package.json; removing it would break those apps. Worth a one-line comment in the source eventually so the next reader knows it's intentional.

What the principle actually wants for this PR turns out to be: `apt install nodejs npm` is fine here. It's what a careful engineer would do for a Rails+package.json app today. There's no standardized 90%-case pattern for node-as-build-artifact-extractor that we could automate cleanly; users who need that level of control reach for `Dockerfile.miren` or onBuild commands.

The thing the principle *does* still want — and that we'll get to in the RFD rewrite — is making the choices visible: codifying the rules we already follow and communicating them to users (a build philosophy page is one expression of that).

Leftover items from the earlier inline notes (yarn.lock + corepack opportunity, unchecked e2e test-plan items, `metaStackOf` future-stack footgun) stand as follow-ups, none blocking.

Thanks for the patient correction. RFD rewrite coming next.

pkg/stackbuild/stackbuild.go
evanphx: Resolved!

PR #812, opened 2026-05-18
Group top-level CLI commands by user intent

phinze:
A prospect flagged that `miren --help` was hard to navigate as a deployer. The top-level was a flat list of 30+ commands with server-administration sitting right next to deploy and rollback. For someone who only ever deploys an app, that's a lot of noise.

First instinct was to dim operator commands so they faded into the background, but that felt user-hostile, and it didn't solve the navigation problem so much as paper over it. Pivoted to positive grouping: five named buckets organized by what the user is trying to do (Getting started, Monitoring your app, Configuring your app, Client operations, Server operations). Group headers render in bold yellow (matching our table-header convention) with the existing faint sub-command counts. The exercise was itself useful in noticing where the CLI surface had drifted (a couple of follow-ups noted but not landed here, like `download release` and the `debug` outliers).

The rendering side moved into the runtime layer. mflags' role is to parse arguments and own the command tree; this PR makes the runtime render its own top-level help (and intercepts `miren`, `miren -h`, and `miren help` to all go through the same path). The data side landed in the companion mirendev/mflags#14, where we added a `CommandGroupProvider` interface, a `Group` field on `ChildEntry` and `CommandDoc`, and exported `GetDirectChildren` so callers can walk the tree. Same boundary makes future help-rendering polish a runtime concern, not a mflags one.

Along the way: `miren help <space-separated path>` now works the same as the dotted form, the help footer mentions both `miren help <cmd>` and `miren <cmd> --help`, and the `help alias` standalone doc page was retired (aliases content already lives in `docs/docs/app-toml.md#aliases`). A `TestAllTopLevelCommandsHaveKnownGroup` test catches new top-level commands that forget to declare a group, so the grouping stays complete as the CLI grows.

Closes MIR-845


phinze:
New help in action:

2026-05-18 at 17 32 51@2x

evanphx (approved):
Image

my.. it's so... beautiful.

PR #820, opened 2026-05-22
validate: allow tls.additional_names / tls.additional_ips in behind-proxy-http

jcasimir:
## Summary

`pkg/serverconfig/validate.go` rejects `tls.additional_names` and `tls.additional_ips` when `ingress.mode = behind-proxy-http`, on the grounds that "ingress doesn't terminate TLS, so these fields would be ignored". That premise is correct for `tls.self_signed`, `tls.acme_email`, and `tls.acme_dns_provider` — those only affect ingress cert provisioning. But `tls.additional_names` / `tls.additional_ips` also feed:

- **the API server cert** (`components/coordinate/coordinate.go::LoadAPICert` appends `c.AdditionalNames` and `c.IPs.RawIPs()` to the cert SANs before `authority.IssueCertificate`)
- **the etcd cert** (`coordinate.SetupEtcdTLS` is invoked with `cfg.TLS.AdditionalNames` and `ipSet.RawIPs()` from `cli/commands/server.go:315`)

Both of those certs exist regardless of ingress mode. So under `behind-proxy-http`, operators legitimately need to set these fields to make the API/etcd certs cover the cluster's public hostname / IP — but the validation refuses.

In our setup that surfaced as: external `miren deploy` from CI fails Go's TLS hostname check (`leaf cert SAN doesn't match miren.blueyard.app`) because removing `--dns-names=...` from ExecStart — necessary to satisfy this validator — also stripped the SAN from the API cert. No alternative knob exists to set just the API cert SANs.

## Change

Just remove the two over-broad checks. The downstream wiring already handles these fields correctly in every ingress mode.

```diff
 		if c.TLS.GetAcmeDNSProvider() != "" {
 			populated = append(populated, "tls.acme_dns_provider")
 		}
-		if len(c.TLS.AdditionalIPs) > 0 {
-			populated = append(populated, "tls.additional_ips")
-		}
-		if len(c.TLS.AdditionalNames) > 0 {
-			populated = append(populated, "tls.additional_names")
-		}
+		// tls.additional_names / tls.additional_ips are intentionally NOT
+		// gated here: they're applied to the API server cert (always-on
+		// TLS on the API port) and the etcd cert. Both exist regardless
+		// of ingress.mode. The fields gated above (self_signed,
+		// acme_email, acme_dns_provider) really are ingress-cert-only.
 		if len(populated) > 0 {
```

Tests added in `validate_test.go`:
- `behind-proxy-http accepts tls.additional_names`
- `behind-proxy-http accepts tls.additional_ips`
- `behind-proxy-http rejects ingress-only fields even when additional_* are also set`

Verifies the relaxation lands without weakening the ingress-only field check.

## Test plan

- [x] `go test ./pkg/serverconfig/` — passes (including the three new cases)
- [x] `golangci-lint run ./pkg/serverconfig/` — 0 issues
- [x] Manual: built locally, installed on a `behind-proxy-http` cluster with `--dns-names=miren.blueyard.app --ips=160.202.129.155` in ExecStart. miren came up cleanly, regenerated the API cert with the new SANs, and external `miren deploy` from off-box now succeeds where it previously failed Go's hostname check.

### DIFF
--- a/docs/docs/server-config.md
+++ b/docs/docs/server-config.md
@@ -109,7 +109,7 @@ The `behind-proxy-*` modes default to localhost to keep accidental misconfigurat
 
 ## `[tls]` — TLS Settings {#tls}
 
-Controls cert sourcing for ingress modes that terminate TLS (`tls-autoprovision` and `behind-proxy-https`). Ignored under `behind-proxy-http`; populating these fields under that mode is an error. See [TLS](/tls) for setup guides.
+Settings under `[tls]` cover two kinds of certs. `acme_email`, `acme_dns_provider`, and `self_signed` configure the ingress cert and only apply when Miren terminates TLS (`tls-autoprovision` or `behind-proxy-https`); they're rejected at startup under `behind-proxy-http`. `additional_names` and `additional_ips` are different: they extend the SANs on the API server and etcd certs, which exist regardless of ingress mode, so they're valid under any mode. See [TLS](/tls) for setup guides.
 
 | Field | Type | Default | Description | Env Var | CLI Flag |
 |-------|------|---------|-------------|---------|----------|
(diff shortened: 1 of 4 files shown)


phinze (approved):
Nice catch on this! I verified the trace through `LoadAPICert` and `SetupEtcdTLS` and the diagnosis is exactly right. I'll fold in a small docs tweak as I merge.

PR #823, opened 2026-05-26
Improve admin CLI per-method parameter help UX (MIR-746)

evanphx:
## Summary

Smooths out three rough edges in `miren admin <method>`:

- `--help` / `-h` on a method now shows method help (same as `--func-help`) instead of erroring with `unknown parameter(s): help`.
- Calling a method that declares params with no args now renders the help block instead of falling through to a raw `-32602 Invalid params` JSON-RPC error.
- `validateAdminCall` missing-required and unknown-param errors now embed the full method definition (description, tree-formatted params with types and `(required)` markers), matching what `--list` / `--func-help` produce.

Also distinguishes two cases that previously looked identical on the wire:

- `(no parameters)` — method explicitly declares an empty param list.
- `(parameters not advertised by this method)` — method's `$methods` response omits the `params` key entirely.

`servers/admin/admin.go` `fetchMethods` now calls `SetParams` unconditionally when the source JSON includes a `params` key, even if empty. Combined with a tightened client guard (`HasParams()` instead of `HasParams() && len > 0`), declared-zero-params methods now also reject supplied params.

## Test plan

- [x] `go build ./...`
- [x] `go test ./cli/commands/ ./servers/admin/` — 208 passing
- [x] `make lint` — 0 issues
- [ ] Live smoke against a real admin method:
  - `m admin -a <app> <method> --help` matches `--func-help` output
  - `m admin -a <app> <method>` (no params) renders method help, not `-32602`
  - `m admin -a <app> <method> bogus=1` shows rich definition list after "unknown parameter(s)"
  - `m admin -a <app> <method>` (declared zero params) still calls server cleanly
  - `m admin -a <app> <method> --no-validate` still sends empty call
  - `m admin -a <app> --list` and `m admin -a <app> <method> --func-help` unchanged

Refs MIR-746.

### DIFF
--- a/cli/commands/admin.go
+++ b/cli/commands/admin.go
@@ -47,15 +47,16 @@ func Admin(ctx *Context, opts struct {
 		return fmt.Errorf("method name is required (use --list to see available methods)")
 	}
 
-	// Handle --rpc-help flag
-	if opts.FuncHelp {
+	// Handle --func-help flag, or natural --help/-h on a method (treated the same)
+	if opts.FuncHelp || hasHelpFlag(opts.Unknown) {
 		return adminMethodHelp(ctx, adminClient, opts.App, opts.Method)
 	}
 
 	// Fetch method introspection for type-aware parsing and validation
+	var methodInfo *admin_v1alpha.AdminMethod
 	var paramTypes map[string]string
 	if !opts.NoValidate {
-		methodInfo, err := fetchMethodInfo(ctx, adminClient, opts.App, opts.Method)
+		methodInfo, err = fetchMethodInfo(ctx, adminClient, opts.App, opts.Method)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


cli/commands/admin.go
```diff
@@ -129,6 +130,15 @@ func Admin(ctx *Context, opts struct {
 		params[key] = parsed
 	}
 
+	// If the method declares params but none were supplied, render method help
+	// instead of letting an empty call fall through to a raw JSON-RPC error.
+	// Skipped under --no-validate so power users can still issue empty calls.
+	if !opts.NoValidate && len(params) == 0 && methodInfo != nil &&
+		methodInfo.HasParams() && len(methodInfo.Params()) > 0 {
```
phinze, line 140: Asymmetry worth a beat of thought: missing-required-params returns an error with embedded help (non-zero exit), but this branch — only-optional params, none supplied — renders help and exits zero. A script doing `m admin -a foo bar && echo ok` prints `ok` without anything having happened.

Could this match the other branch?

```go
return fmt.Errorf("no parameters supplied\n\n%s", renderMethodHelpString(opts.App, methodInfo))
```

Or if "zero params = query for help" is the intent, fair — but then the missing-required case probably wants to be a help-render too. Either choice works, the split is what feels off.

phinze (approved):
Yer boy and yer bot here. We like this overall — good improvements. The `(no parameters)` vs `(parameters not advertised)` distinction in particular is a small clarity that punches above its weight (which presumes we're gonna be boxing with this API I guess?).

CodeRabbit caught a real one on `hasHelpFlag` — agree with the diagnosis, but we'd lean toward just dropping `case "help"` entirely rather than the preceding-flag heuristic. `--help`/`-h` are the conventions people know; bare `help` was the form that broke things.

One thing inline.

Seeing all this code shift around without any tests is a little itchy for Paul, and for Claude it kinda tickles:

brother may i have some TESTS

cli/commands/admin.go
evanphx: Done!

PR #824, opened 2026-05-27
Restart miren.service after register instead of warning

phinze:
The warning at the end of `miren server register` ("you must now restart the miren server") was the bug, but the fix had a couple of layers. The warning fires at the end of every Register call, including from inside `miren install`, which then turns around and starts the systemd unit itself a few lines later. So during install it's pure noise. During a standalone re-register against a live cluster it's at least correct, but easy to miss in all the other registration output.

I went down the rabbit hole of "could register just be online and not need a restart at all" first. Short version: the cloud authenticator gets wired into RPC server options at construction time with no swap path, plus the build and deployment servers read DNSHostname by value at startup. So it's a real refactor across the auth chain to make any of it dynamic. Maybe worth it someday, but registration is a once-a-quarter operation and the surface area is large.

What we landed on instead: detect a running systemd `miren.service` and restart it ourselves. Fresh installs (systemd or docker) hit the "unit isn't active yet" branch and stay quiet, since install handles the lifecycle a few lines later anyway. Standalone re-register against a live cluster gets an auto-restart plus a "Restarted miren.service" info line. The one case it still doesn't handle is day-2 re-register on a docker install, but docker is the demo path so it pretty much never happens. Left it.

Closes MIR-1155

### DIFF
--- a/cli/commands/commands_linux.go
+++ b/cli/commands/commands_linux.go
@@ -8,7 +8,7 @@ func addCommands(d *mflags.Dispatcher) {
 	// Server command is now defined in commands.go (renamed from dev)
 
 	// Cloud registration commands
-	d.Dispatch("server register", Infer("server register", "Register this cluster with miren.cloud", Register,
+	d.Dispatch("server register", Infer("server register", "Register this cluster with miren.cloud", RegisterStandalone,
 		WithExample(mflags.Example{
 			Name: "Register with cloud",
 			Body: "miren server register --name my-cluster",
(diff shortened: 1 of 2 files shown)


evanphx (changes requested):
We call Register directly from Install as we're doing the install, so I'm worried that a restart here will cause unnecessary churn while the system is trying to come up for the first time.

evanphx (approved):
Oops, missed the check. `is-active` being load berring for a proper install feels... weird but fine.

phinze:
Ya fair point, reworked it so we don't rely on is-active for the install case.

PR #825, opened 2026-05-27
Add Pull Request Environments documentation

evanphx:
## Summary

Closes MIR-995. Adds a dedicated docs page for ephemeral deploys (a.k.a. PR environments), since today the feature is only documented in a one-line changelog entry and the auto-generated `--ephemeral` flag descriptions.

- New `docs/docs/pr-environments.md` covering:
  - How ephemeral deploys work (build + don't activate, subdomain lookup, TTL cleanup)
  - Quick start (DNS-only — no wildcard route needed)
  - **What runs**: `web` service only, configuration is shared with the active version
  - **Using a staging app**: the recommended pattern for keeping previews off production data, with a staging-cluster alternative
  - Label rules (RFC 1123 normalization)
  - TTL & cleanup (default 24h, sweeps every 5 min, 10-version per-app cap)
  - GitHub Actions per-PR preview workflow targeting the staging app, with OIDC + PR-comment posting
  - Limitations and command-flag reference
- Wires the page into the Features sidebar.
- Cross-links from `deployment.md`, `traffic-routing.md`, `ci-deploy.md`, and `intro.md`.

## Test plan

- [x] `bun run build` passes (lint + Docusaurus build, no broken links)
- [ ] `bun run start` and click through every cross-link in the new doc
- [ ] Eyeball the rendered page on https://miren.md/ once merged

### DIFF
--- a/docs/docs/pr-environments.md
+++ b/docs/docs/pr-environments.md
@@ -1,0 +1,253 @@
+---
+title: Pull Request Environments
+description: Deploy a labeled, time-boxed preview of your app on a subdomain — one per pull request, with automatic cleanup.
+keywords: [pr, preview, ephemeral, github, pull request, environment, review app]
+---
+
+import CliCommand from '@site/src/components/CliCommand';
+
+# Pull Request Environments
+
+A pull request environment is a labeled build of your app — called an **ephemeral version** in Miren — that runs alongside the active version on its own subdomain. It runs separetely from your normal deploys and is deleted automatically when its TTL expires. The typical use is one preview per PR, reachable at something like `pr-123.myapp.example.com`.
+
+## How It Works
+
+When you run `miren deploy --ephemeral <label>`:
+
+1. Miren builds the version but doesn't activate it. The active version keeps serving production traffic.
+2. The new version is reachable at `<label>.<your-app-host>`. A request for any subdomain of an existing route is looked up against ephemeral labels for that route — no separate route entity needed.
+3. After the TTL elapses (default 24 hours), a background controller deletes the version.
+
+Ephemeral deploys don't create deployment history records, don't take the deployment lock, and don't block normal deploys.
+
+## Quick Start
+
+**Step 1: Point DNS at your cluster for the subdomains you'll use.** A wildcard CNAME is the usual choice:
+
+```text
+*.myapp.example.com.   CNAME   cluster-jwomf2l0tn8z.miren.systems.
+```
+
+You don't need to configure a wildcard route on your server. Any existing route for `myapp.example.com` will pick up `pr-123.myapp.example.com` as an ephemeral lookup. See [Custom Domains](/traffic-routing#custom-domains) for the full DNS setup.
+
+**Step 2: Deploy with `--ephemeral` and an optional TTL.**
+
+<CliCommand context="client">
+```miren
+miren deploy --ephemeral pr-123 --ttl 48h
+```
+</CliCommand>
+
+Miren builds the version and prints the access URL:
+
+```text
+Ephemeral version myapp-vXYZ created.
+  Label: pr-123
+  TTL:   48h
+  URL:   https://pr-123.myapp.example.com
+```
+
+Open the URL — your preview is live. TLS provisions on first request.
+
+## What Runs in an Ephemeral Version
+
+**Only the `web` service runs.** Workers, background jobs, scheduled tasks, and any other services defined in `.miren/app.toml` aren't started for ephemeral versions — HTTP traffic to the subdomain is the only thing wired up. If reviewing your PR requires a worker too, use a separate staging app instead.
+
+**Configuration is shared with the active version.** Env vars, secrets, addons, and build config all come from the app's current settings. There's no per-ephemeral override:
+
+- Your preview connects to the same database, queues, and external services as production.
+- You can't change `RAILS_ENV`, feature flags, or service URLs just for the preview.
+- Updates to app config (via `miren config set`, addons, etc.) affect both active and ephemeral versions.
+
+If you need isolation from production data, run ephemeral deploys against a separate staging app.
+
+## Using a Staging App
+
+Because ephemeral versions share config with the active version, a preview deployed against your production app talks to your production database, queues, and external services. That's fine for read-mostly UI changes, but risky as soon as a PR writes data, runs migrations, or fires background jobs.
+
+The recommended pattern is to keep a second app — typically `myapp-staging` — that points at a staging database and any other backing services you want isolated. Run all PR previews against that app instead of production.
+
+**Step 1: Create the staging app.** Deploy your main branch to it with whatever staging-specific config you want:
+
+<CliCommand context="client">
+```miren
+miren deploy -a myapp-staging \
+  -e DATABASE_URL=postgres://staging-db.internal/myapp \
+  -e RAILS_ENV=staging
+```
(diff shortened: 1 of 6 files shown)


docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+  URL:   https://pr-123.myapp.example.com
+```
+
+Open the URL — your preview is live. TLS provisions on first request.
+
+## What Runs in an Ephemeral Version
+
+**Only the `web` service runs.** Workers, background jobs, scheduled tasks, and any other services defined in `.miren/app.toml` aren't started for ephemeral versions — HTTP traffic to the subdomain is the only thing wired up. If reviewing your PR requires a worker too, use a separate staging app instead.
+
+**Configuration is shared with the active version.** Env vars, secrets, addons, and build config all come from the app's current settings. There's no per-ephemeral override:
```
phinze, line 56: 💭 We should build this eventually! "Just one env var override" I can imagine being very useful.

Even an auto-set MIREN_EPHEMERAL_LABEL people could switch on would be a good start.

docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+- You can't change `RAILS_ENV`, feature flags, or service URLs just for the preview.
+- Updates to app config (via `miren config set`, addons, etc.) affect both active and ephemeral versions.
+
+If you need isolation from production data, run ephemeral deploys against a separate staging app.
+
+## Using a Staging App
+
+Because ephemeral versions share config with the active version, a preview deployed against your production app talks to your production database, queues, and external services. That's fine for read-mostly UI changes, but risky as soon as a PR writes data, runs migrations, or fires background jobs.
+
+The recommended pattern is to keep a second app — typically `myapp-staging` — that points at a staging database and any other backing services you want isolated. Run all PR previews against that app instead of production.
```
phinze:
```suggestion
For this scenario, we recommend keeping a second app (e.g. `myapp-staging`) that points at a staging database and any other backing services you want isolated. Run all PR previews against that app instead of production.
```

Passive voice swap out 🔄

docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+```
+</CliCommand>
+
+**Redeploying with the same label replaces the prior version.** Push a new commit, redeploy with `--ephemeral pr-123`, and the previous `pr-123` version is deleted before the new one becomes reachable. The TTL resets on each deploy.
+
+## TTL and Cleanup
+
+`--ttl` takes a Go duration string (`30m`, `2h`, `48h`) and defaults to `24h`. Expiration is fixed at deploy time.
+
+A background controller sweeps expired versions every five minutes. Requests to an expired label return a 404 once the next sweep runs. There's no extend command — redeploy with the same label to refresh the TTL.
```
phinze: Small factual nit: this undersells the actual behavior. `LookupByLabel` (`pkg/ephemeral/manage.go:218`) skips expired versions in its own filter, so an expired label 404s on the very next request — no need to wait for the sweep. The sweep is only deleting the entity from storage, not gating the response.

Suggested rewrite:

> Requests to an expired label return 404 immediately — the lookup filters expired versions itself. A background controller sweeps the deleted entities every five minutes.

There's no extend command — redeploy with the same label to refresh the TTL.

docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Deploy preview
+        uses: mirendev/actions/deploy@main
+        with:
+          cluster: ${{ secrets.MIREN_CLUSTER }}
+          app: myapp-staging
+          args: --ephemeral pr-${{ github.event.pull_request.number }} --ttl 48h
```
phinze: 💭 I wish we could lift this pattern up into a proper "miren is doing a preview deploy" action (or first class action args on deploy). Interpolating args feels a little unpolished. Fine for a first cut though.

docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+Each push replaces the previous preview at the same label. When the PR is merged or closed, the version expires on its own — no teardown step needed.
+
+The preview URL follows the pattern `pr-<number>.staging.<your-host>`, so a follow-up step can post it as a PR comment:
+
+```yaml
+      - name: Comment preview URL
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const url = `https://pr-${context.issue.number}.staging.myapp.example.com`;
```
phinze: 💭 Similar polish thing for the future: can our action output the deployed URL so users don't have to construct it?

docs/docs/pr-environments.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Limitations
+
+- **`web` service only** — workers and other services from your app config don't start (see [What Runs in an Ephemeral Version](#what-runs-in-an-ephemeral-version)).
+- **Shared configuration** — env vars, secrets, and addons all come from the app. No per-preview overrides.
+- **No deployment history** — `miren app history`, `miren rollback`, and the deployment lock all ignore ephemeral deploys.
+- **10 per app** — older versions are evicted by expiry as new ones arrive.
+- **DNS must cover the subdomains** — without a wildcard CNAME (or per-label records) pointing at your cluster, the URL won't resolve.
+
+## Command Reference
```
phinze, line 231: 💡 I'm not sure about duplicating a subset of the CLI reference as a pattern... feels like the kinda thing that's asking to drift eventually.

phinze (approved):
Claude and Paul here, with our docs reading glasses on. Paul took a first pass and dropped a few inlines, then we ran through together. Claude cross-checked the technical claims against the implementation — TTL default, sweep interval, the 10-per-app cap, the "web only" mechanism, label normalization, all the flags — and it all holds up.

One small tightening at line 138 worth folding in: the sentence undersells our own behavior, since `LookupByLabel` filters expired versions itself. 404s happen on the next request, not on the next sweep.

The "Using a Staging App" section is the standout. Approving.

docs/docs/pr-environments.md
evanphx: Agreed!

docs/docs/pr-environments.md
evanphx: Oh actually, there is a better way. Claude missed that, fixing.

docs/docs/pr-environments.md
evanphx: I had thoughts about that too. I didn't want to over do it in actions, but we can absolutely add an option to do that.

docs/docs/pr-environments.md
evanphx: Fixed!

docs/docs/pr-environments.md
phinze: Wait… this exists?

docs/docs/pr-environments.md
phinze: lol just noticed this exists and the blog uses it

docs/docs/pr-environments.md
phinze: [https://github.com/mirendev/actions/blob/main/deploy/action.yml#L32-L34](<https://github.com/mirendev/actions/blob/main/deploy/action.yml#L32-L34>) lol

PR #831, opened 2026-05-28
docs: expand admin interface page (MIR-1176)

evanphx:
## Summary

Audits `docs/docs/admin-interface.md` against the live runtime and CLI, and fills in the missing details users need to expose and call admin methods safely.

- **Security**: token format/rotation, `ADMIN_TOKEN` as a reserved env var, and the `X-Miren-Access` trust model.
- **Auditing**: admin calls land in the app's log stream as `source=admin` entries.
- **JSON-RPC shape**: object vs positional params, the 30s call timeout, and HTTP status semantics.
- **Introspection**: `$`-prefixed method names (`$methods`, `$type`) are reserved.
- **CLI**: `--func-help`, `--params-file` (including stdin via `-`), mixed input styles, type-aware parsing, and kebab→snake flag normalization.
- Moves the per-language samples (Python, Node.js, Bun) into one **More Implementation Examples** section below CLI usage; **Other Languages** now links down to it.
- Switches the Go and Node.js auth examples to constant-time token comparison.

> This change drops the "labs feature" callout — intended to land alongside the `adminapi` flag being removed.

Linear: MIR-1176

## Test plan

- [ ] `cd docs && bun run build` — builds cleanly (no MDX/markdown errors).
- [ ] Visual review of `/admin-interface` end-to-end: `<CliCommand>` blocks render, tables render, headings link correctly, and the in-page link from "Other Languages" jumps to the examples at the bottom.


docs/docs/admin-interface.md
```diff
@@ -72,11 +77,24 @@ func authMiddleware(token string, next http.Handler) http.Handler {
...
-- `internal` — Request is from Miren's admin system (trusted)
-- `public` — Request is from an external client (Miren strips any client-provided value)
+#### Network reachability and `X-Miren-Access`
+
+Miren's admin proxy talks to your app over its **internal** httpingress path, never via the public route. The proxy sets one header you can rely on:
+
+- `X-Miren-Access: internal` — request originated from `miren admin` and was routed through Miren's internal admin path.
+- `X-Miren-Access: public` — request arrived through the public ingress. Miren overwrites any client-supplied value here before forwarding, so the header cannot be spoofed by external callers.
+
+For the most defensive setup, be sure to check `X-Miren-Access: internal` as well as the bearer token.
```
phinze: Worth adding that the admin path is hard-blocked on the public route, not just deprioritized: `httpingress.go:488` 404s any public request to `/.well-known/miren/admin` before it reaches your app. So your handler only ever sees `X-Miren-Access: internal` — which makes the header check clean defense-in-depth rather than the only guard. A sentence noting the public route is closed by default would land that point.

phinze (approved):
We reviewed this as an audit: every runtime claim on the page, traced back to the code. They all check out — token generation and env-injection in `launcher.go`, the `X-Miren-Access` strip/set in `httpingress.go`, the audit-log shape (`source=admin`, params size, `status`, `duration_ms`), the 30s timeout, the verbatim status-error string, and the whole CLI surface in `admin.go` (param channels, type coercion, kebab→snake, `--no-validate`). For a doc this dense with specific claims, 100% accuracy is rare. Nice work.

One suggestion inline about the public-route block, and +1 to CodeRabbit's "httpingress" → "internal HTTP ingress" on line 82. Neither blocking.

maryelizbeth (approved):
LGTM 🪄

PR #834, opened 2026-06-01
Add workload identity tokens to sandbox containers

evanphx:
## Summary

- Implements GitHub Actions-style OIDC workload identity for all sandboxes
- Each sandbox gets a signed Ed25519 JWT at `/var/run/miren/identity-token` with claims for org, cluster, app, and sandbox ID
- Ingress serves standard `/.well-known/openid-configuration` and `/.well-known/jwks` endpoints so external systems can verify tokens
- Tokens have 1-hour expiry with automatic background refresh every 45 minutes
- On-demand token request server on the bridge gateway (`<gateway>:7123/v1/token`) lets sandboxes request tokens with custom audience and TTL
- Feature is gracefully disabled when no issuer URL is available (no registration + no TLS names)

## Design

Follows the GitHub Actions OIDC pattern:
- Cluster generates a dedicated Ed25519 signing keypair (persisted at `${dataPath}/server/workload-identity.key`)
- JWT subject format: `org:<org-id>:app:<app-name>:sandbox:<sandbox-id>`
- Standard claims: `iss`, `sub`, `aud`, `iat`, `nbf`, `exp`, `jti`
- Custom claims: `organization_id`, `cluster_id`, `app`, `sandbox_id`
- External verifiers fetch the public key via JWKS endpoint to validate tokens

### Token request server

An HTTP server bound to the bridge gateway IP on port 7123 provides on-demand tokens:
- `GET /v1/token` — returns a token for the calling sandbox (identified by source IP)
- `GET /v1/token?audience=sts.amazonaws.com` — custom audience (multiple values supported)
- `GET /v1/token?ttl=300` — custom TTL in seconds (min 60s, max 24h, default 1h)
- Response: `{"value": "<jwt>"}` (matches GitHub Actions format)
- IP-to-sandbox lookup reuses the existing DNS server's mapping via `ServiceManager.LookupSandboxByIP`
- Sandboxes discover the endpoint via `MIREN_IDENTITY_TOKEN_URL` env var

### Environment variables injected

- `MIREN_IDENTITY_TOKEN_PATH=/var/run/miren/identity-token` — static token file
- `MIREN_OIDC_ISSUER_URL=https://<cluster>` — OIDC issuer for external verification
- `MIREN_IDENTITY_TOKEN_URL=http://<gateway>:7123/v1/token` — on-demand token endpoint

## Test plan

- [x] Unit tests for key generation, token signing/verification, OIDC discovery, JWKS format
- [x] Unit tests for custom audience and TTL (with clamping)
- [x] Unit tests for token request server (default, custom audience, custom TTL, unknown IP, invalid TTL, method rejection)
- [x] `make test` passes (all packages green)
- [x] Manual: deployed app, exec'd into sandbox, verified token file + env vars
- [x] Manual: verified OIDC discovery and JWKS endpoints on ingress
- [x] Manual: verified token signature using JWKS public key (full end-to-end)

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -396,6 +396,15 @@ func (c *SandboxController) reconcileSandboxesOnBoot(ctx context.Context) error 
 					}
 				}
 			}
+
+			// Re-register with token refresher so tokens keep getting renewed
+			if c.tokenRefresher != nil {
+				appName := c.resolveAppName(ctx, &sb)
+				tokenPath := c.sandboxPath(&sb, "identity-token")
+				c.tokenRefresher.register(sb.ID.String(), tokenPath, appName)
+				c.Log.Debug("re-registered sandbox for token refresh",
+					"sandbox_id", sb.ID, "app", appName)
+			}
 		}
 	}
 
@@ -474,6 +483,13 @@ func (c *SandboxController) Init(ctx context.Context) error {
 		return err
 	}
 
+	// Initialize token refresh state before reconcile so surviving sandboxes
+	// can be re-registered during boot reconciliation.
+	if c.WorkloadIssuer != nil {
+		c.tokenRefresher = newTokenRefresher()
+		c.tokenSecrets = newTokenSecretRegistry()
+	}
+
 	// Reconcile sandboxes after containerd restart
 	// This must happen after bridge setup but before starting normal operations
 	err = c.reconcileSandboxesOnBoot(ctx)
@@ -512,9 +528,8 @@ func (c *SandboxController) Init(ctx context.Context) error {
 	c.imageWatchdog.Start(c.topCtx)
 
 	// Start workload identity token refresh loop and token request server
+	// (tokenRefresher and tokenSecrets were created earlier, before reconcile)
 	if c.WorkloadIssuer != nil {
-		c.tokenRefresher = newTokenRefresher()
-		c.tokenSecrets = newTokenSecretRegistry()
 		go c.runTokenRefresh(c.topCtx)
 		go c.startTokenServer(c.topCtx)
 	}
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -37,6 +37,7 @@ import (
 	"miren.dev/runtime/pkg/imagerefs"
 	"miren.dev/runtime/pkg/netdb"
 	"miren.dev/runtime/pkg/netutil"
+	"miren.dev/runtime/pkg/workloadidentity"
 
 	compute "miren.dev/runtime/api/compute/compute_v1alpha"
 	"miren.dev/runtime/api/core/core_v1alpha"
@@ -82,6 +83,7 @@ type SandboxControllerDeps struct {
 	StatusMon      *observability.StatusMonitor
 	Resolver       netresolve.Resolver
 	Metrics        *Metrics
+	WorkloadIssuer *workloadidentity.Issuer
 }
 
 type SandboxController struct {
@@ -106,8 +108,11 @@ type SandboxController struct {
 
 	StatusMon *observability.StatusMonitor
 
-	Resolver netresolve.Resolver
-	Metrics  *Metrics
+	Resolver       netresolve.Resolver
+	Metrics        *Metrics
+	WorkloadIssuer *workloadidentity.Issuer
+
+	tokenRefresher *tokenRefresher
 
 	topCtx context.Context
 	cancel func()
@@ -182,6 +187,7 @@ func NewSandboxController(cfg SandboxControllerDeps) (*SandboxController, error)
 		StatusMon:      cfg.StatusMon,
 		Resolver:       cfg.Resolver,
 		Metrics:        cfg.Metrics,
+		WorkloadIssuer: cfg.WorkloadIssuer,
 	}, nil
 }
 
@@ -503,6 +509,12 @@ func (c *SandboxController) Init(ctx context.Context) error {
 		Config:    DefaultImageGCConfig(),
 	}
 	c.imageWatchdog.Start(c.topCtx)
+
+	// Start workload identity token refresh loop
+	if c.WorkloadIssuer != nil {
+		c.tokenRefresher = newTokenRefresher()
+		go c.runTokenRefresh(c.topCtx)
+	}
 
 	return nil
 }
@@ -2054,6 +2066,30 @@ func (c *SandboxController) buildSubContainerSpec(
 		})
 	}
 
+	// Inject workload identity token
+	if c.WorkloadIssuer != nil {
+		appName := c.resolveAppName(ctx, sb)
+		token, tokenErr := c.WorkloadIssuer.IssueToken(appName, sb.ID.String())
+		if tokenErr != nil {
+			c.Log.Warn("failed to generate workload identity token", "sandbox", sb.ID, "error", tokenErr)
+		} else {
+			tokenPath := c.sandboxPath(sb, "identity-token")
+			if writeErr := os.WriteFile(tokenPath, []byte(token), 0444); writeErr != nil {
+				c.Log.Warn("failed to write workload identity token", "sandbox", sb.ID, "error", writeErr)
+			} else {
+				mounts = append(mounts, specs.Mount{
+					Destination: "/var/run/miren/identity-token",
+					Type:        "bind",
+					Source:      tokenPath,
+					Options:     []string{"rbind", "ro"},
+				})
+				if c.tokenRefresher != nil {
+					c.tokenRefresher.register(sb.ID.String(), tokenPath, appName)
+				}
+			}
+		}
+	}
+
 	// Extract instance number from metadata labels and inject MIREN_INSTANCE_NUM
 	envVars := co.Env
 	var md core_v1alpha.Metadata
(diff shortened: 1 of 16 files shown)


controllers/sandbox/token_refresh.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (tr *tokenRefresher) snapshot() []tokenEntry {
+	tr.mu.Lock()
+	defer tr.mu.Unlock()
+	entries := make([]tokenEntry, 0, len(tr.entries))
+	for _, e := range tr.entries {
+		entries = append(entries, e)
+	}
+	return entries
+}
```
phinze, line 55: The registry this loop walks is in-memory and only gets populated by `register()` at container build time. On a coordinator restart, `reconcileSandboxesOnBoot` re-adopts the surviving sandboxes but never rebuilds their specs, so they never get re-registered here. The refresher comes back up with an empty roster, and those tokens expire within the hour with nothing to renew them. Cleanest fix is to repopulate from the boot reconcile path, since we already have the sandbox list and app resolution there. Longer term, this is the best argument for making refresh reconcile-driven.

controllers/sandbox/sandbox.go
```diff
@@ -2040,6 +2053,30 @@ func (c *SandboxController) buildSubContainerSpec(
...
 	}
 
+	// Inject workload identity token
+	if c.WorkloadIssuer != nil {
+		appName := c.resolveAppName(ctx, sb)
+		token, tokenErr := c.WorkloadIssuer.IssueToken(appName, sb.ID.String())
+		if tokenErr != nil {
+			c.Log.Warn("failed to generate workload identity token", "sandbox", sb.ID, "error", tokenErr)
+		} else {
+			tokenPath := c.sandboxPath(sb, "identity-token")
```
phinze, line 2094: Worth a code comment here: refresh correctness depends on this staying an in-place write. We bind-mount the inode, so rewriting it in place reaches the container. But the usual "safe" atomic pattern (write a temp file, then rename) repoints the directory entry at a new inode, and the container would keep reading the old token forever with no error anywhere. The k8s fix is to bind-mount the directory instead and rename within it (their `..data` trick), which gets you both atomicity and visibility. Not urgent, since the race here is microseconds on a few hundred bytes, but it'd make the safe write and the correct write the same thing.

servers/httpingress/httpingress.go
```diff
@@ -432,6 +437,16 @@ func (h *Server) handleRequest(w http.ResponseWriter, req *http.Request) {
 	defer span.End()
 	req = req.WithContext(ctx)
 
+	// Handle OIDC discovery endpoints for workload identity verification
+	if req.URL.Path == "/.well-known/openid-configuration" {
```
phinze: Two issues. First, these match on every Host at the bare `/.well-known/` level, unlike the namespaced `/.well-known/miren/health` just below, so they shadow any app serving its own OIDC discovery. Second, `AcquireTunnel` blocks `/.well-known/miren/*` as internal endpoints, and since these live outside that prefix they sidestep that guard. `openid-configuration` is spec-locked to the bare path, but we could gate it on `req.Host == issuer hostname` so it only shadows on the cluster's own address. `jwks` isn't path-locked (it's advertised via `jwks_uri`, issuer.go:99), so it could move under `/.well-known/miren/jwks` and rejoin both the convention and the guard. Stepping back, "what counts as a reserved miren path" now lives in three functions that are already drifting, so a shared registry would be a good follow-up.

pkg/workloadidentity/issuer.go
```diff
@@ -0,0 +1,205 @@
...
+func (iss *Issuer) PublicKey() any {
+	return iss.publicKey
+}
+
+const (
+	DefaultTTL = 1 * time.Hour
+	MaxTTL     = 24 * time.Hour
+	MinTTL     = 60 * time.Second
+)
+
```
phinze, line 131: `aud: "miren"` is fine for miren verifying miren, but external federation will want it set per-request rather than as a constant. AWS, Vault, and GCP each check `aud` against their own configured value, and a single workload talking to two of them needs two different audiences (the way GitHub lets the workload request its audience at token time). It's a shape change rather than a value change, worth being aware of for later.

phinze (approved):
Paul here with Claude. We spent real time on this one and rebuilt the workload-identity model from scratch before reviewing, so these are thought-through notes rather than a drive-by. Short version: coherent feature, clean implementation, and it reads clearly as a deliberate first cut. CodeRabbit's `EdDSA` validator catch is already fixed in `2a08e60`, agreed with that one, nothing to add.

The one we'd most want resolved is refresh not surviving a coordinator restart (inline on `token_refresh.go`). A sandbox that outlives a restart ends up holding an expired token within the hour. We're fine with that being a fast-follow ticket rather than a blocker, it just shouldn't slip through the cracks.

A few things worth your read, mostly because the "why" would be great to capture somewhere durable:

- **Per-cluster issuer vs. central.** Today each cluster is its own OIDC issuer, with its own signing key, its own JWKS, and its own canonical hostname. Miren.cloud isn't in the trust path at all; it only contributes the org and cluster ID strings that get baked into the claims as metadata. That's the GitHub/k8s pattern and totally defensible. What's interesting is that those `cluster_id` and `organization_id` claims are exactly what you'd want if cloud later became a single global authority that issued and verified centrally and told clusters apart by claim, so the schema reads as forward-looking toward a central model even though issuance is per-cluster today. The practical fork is how external trust scales: per-cluster means a verifier like AWS configures trust once per cluster (N clusters, N trust configs), while a central issuer would be one config for the whole fleet, scoped by claim. Is per-cluster the intended end state, or a v1 step toward a central cloud issuer?

- **Rotation will need a story (follow-up is totally fine).** Right now there's one persisted key, the JWKS publishes one key, and there's no rotation path, so a key swap is a hard cutover with a verification blackout up to the `max-age=3600` cache window. And since tokens are stateless, there's no per-token revocation either. That's a reasonable thing to punt on for a first cut. We're just flagging that it'll have to land eventually, and a follow-up ticket is the right home for it rather than scope creep here.

- **Food for thought, the canonical cluster address.** This PR quietly promotes the cluster from "a thing that IPs and routes point at" to "a named principal that signs under one canonical address." That's a real and reasonable shift, but it means the cluster now has a cryptographic identity, and how that identity gets chosen matters. Cloud-registered clusters get a clean canonical hostname handed to them. Bare-metal clusters fall back to `cfg.TLS.AdditionalNames[0]`, so their identity anchor is effectively decided by the order of a config list. Reorder the names and the `iss` in every future token (plus every external trust config pinned to it) shifts. We're not asking for a change here, just flagging that the non-cloud identity story is currently implicit, and for something this central it might deserve a more deliberate answer down the road.

Bigger picture: the restart gap, the in-place write coupling below, and the lifecycle hooks spread across build and delete all point at the same missing seam. They'd fold nicely into a reconcile-driven "projected material" helper in the spirit of k8s projected volumes, one component that owns the directory mount, the atomic rotation, and the refresh, and derives its roster from the sandbox list we already walk on boot. That makes refresh self-healing instead of a build-time side effect. Follow-up territory, not this PR.

Really nice work. The flow is tight and stays consistent end to end (the discovery `issuer` can't drift from the token `iss`, the nil-gating is disciplined throughout), and it was genuinely fun to trace.

controllers/sandbox/token_refresh.go
evanphx: Oh yeah, let's fix that. good catch!

controllers/sandbox/sandbox.go
evanphx: OMG a crazy good take away. At a minimum needs a comment.

phinze: ya like i say i think a general projected-volume-flavored primitive is the endgame... so we can dump whatever secrets we like in there safely over time - and eventually yield that facility to users... but for now comment is good!

servers/httpingress/httpingress.go
evanphx: Great call, the, that openid-configuration should only be served when there is NO route matched so that it doesn't collide with all apps. I'll clean it up into a better registry, agree.

pkg/workloadidentity/issuer.go
evanphx: That's why I added the ondemand token server as well, for exactly that case where you want to set the audience.

evanphx:
@phinze On 1 and 3, your intuition is right on that probably that miren.cloud could/should be the mechanism enables this functionality rather than the clusters. But I didn't want to bite that off for the first cut. 

My eventual idea is that when a cluster is join to miren.cloud, the cluster hands it's public keys to miren.cloud and the cluster generates a url like "https://org-xyyz.oidc.miren.cloud" as the issuer, such that miren.cloud can then serve the OIDC and JWKS details directly from miren cloud.

This gives us a surface are to provide more central identities in future. It also solves your concern about cluster reachability being a fundemental identity property.

pkg/workloadidentity/issuer.go
phinze: lol big facepalm, I reviewed this zoomed way in and missed a few files on the first pass, including the whole on-demand token server. Came back for a second look now that you pointed at it. The `?audience=` endpoint is exactly the right answer to my `aud` nitpick, so scratch that one entirely. ✅ 

One thing on `token_server.go` though, and I think it might be the most important thing in the PR: the endpoint's only auth is source IP, no bearer token. That makes it an SSRF magnet. GitHub's equivalent requires `Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN` specifically so an SSRF bug in a workload can't be turned into "go fetch me an identity token" - and with `?audience=sts.amazonaws.com` here, a blind SSRF could mint an AWS-targeted token directly. A blind GET can't set an auth header, which is the whole point of requiring one. Basically the [169.254.169.254 / Capital One](https://blog.appsecco.com/an-ssrf-privileged-aws-keys-and-the-capital-one-breach-4c3c2cded3af) shape.

Not trying to balloon the PR, but prolly a good one to knock out early given it's a path to real cloud creds.

phinze:
> My eventual idea is that when a cluster is join to miren.cloud, the cluster hands it's public keys to miren.cloud and the cluster generates a url like "https://org-xyyz.oidc.miren.cloud/" as the issuer, such that miren.cloud can then serve the OIDC and JWKS details directly from miren cloud.

Oooo splitting discovery from signing there is real nice. Cool things about that:

- clusters keep the private keys and keep signing, cloud just holds the public stuff and serves discovery, so a cloud breach can't mint anything
- you still get one central reachable identity endpoint, kills the whole "cluster has to be reachable to have an identity" problem
- once cloud owns the jwks you can revoke a compromised cluster's key unilaterally, no need for the (maybe pwned) cluster to cooperate, way better than today

👏 good roadmap

pkg/workloadidentity/issuer.go
evanphx: AH! I'll add that token bit, that's easy.

PR #836, opened 2026-06-01
Provision TLS cert for cloud-provisioned cluster hostname

evanphx:
## Summary
- When a cluster has a cloud-provisioned `*.miren.systems` DNS hostname, automatically provision a valid TLS certificate via ACME instead of serving the self-signed fallback
- Both certificate controller paths (HTTP-01 autocert and DNS-01 lego) eagerly provision certs for cluster hostnames in the background on startup
- Cluster hostnames are pinned in the allowed hosts set so they are never removed by route deletions
- Autocert path requires public IPs and DNS-points-to-us checks before attempting provisioning

Closes MIR-1192

## Test plan
- [x] Unit tests for cluster hostnames added to `allowedHosts` during Init
- [x] Unit test confirming cluster hostnames survive `Delete` (route deletion doesn't remove them)
- [x] Unit test for `isAllowedHost` accepting cluster hostnames and ephemeral subdomains
- [x] Unit test verifying `GetCertificate` attempts ACME for cluster hostnames (not immediate fallback)
- [x] Unit test for empty/whitespace cluster hostnames being ignored
- [x] `make lint` passes
- [ ] Manual test: register cluster with cloud, verify HTTPS to `*.miren.systems` hostname uses valid cert


controllers/certificate/controller.go
```diff
@@ -142,13 +144,33 @@ func (c *Controller) initializeLego(ctx context.Context) error {
...
+
+	if len(c.clusterHostnames) > 0 {
+		go c.provisionClusterHostnames(ctx)
 	}
 
 	c.Log.Info("ACME DNS challenge client initialized")
 	return nil
 }
 
+func (c *Controller) provisionClusterHostnames(ctx context.Context) {
```
phinze, line 159: This goroutine runs once at startup, but cluster hostnames have no route to drive the hourly renewal resync, so the cert never renews and expires at ~90 days. Make it a periodic loop (`provisionCertificate` already no-ops when >30 days remain, so this is cheap):

```go
const clusterHostnameRenewInterval = 12 * time.Hour

func (c *Controller) provisionClusterHostnames(ctx context.Context) {
	c.renewClusterHostnames(ctx)
	ticker := time.NewTicker(clusterHostnameRenewInterval)
	defer ticker.Stop()
	for {
		select {
		case <-ctx.Done():
			return
		case <-ticker.C:
			c.renewClusterHostnames(ctx)
		}
	}
}

func (c *Controller) renewClusterHostnames(ctx context.Context) {
	for _, hostname := range c.clusterHostnames {
		if ctx.Err() != nil {
			return
		}
		hostname = strings.TrimSpace(hostname)
		if hostname == "" {
			continue
		}
		if err := c.provisionCertificate(ctx, hostname); err != nil {
			c.Log.Warn("failed to renew cluster hostname certificate (will retry next cycle)", "hostname", hostname, "error", err)
		}
	}
}
```

Also: the existing warning says "will provision inline on first TLS handshake," but this controller never provisions inline (`GetCertificate` only serves cache).

phinze (approved):
Nice change, approving. One gap worth a look: this PR adds a cert with no `http_route` entity, and on the DNS-01/lego path nothing renews it. Route certs renew via the hourly resync; the cluster hostname's goroutine runs once at startup and never again, so the cert expires at ~90 days and (since lego's `GetCertificate` has no fallback) handshakes then hard-fail until a restart. The autocert path is fine here, `autocert.Manager` self-renews. Suggested fix inline. Not a blocker given DNS-01 is the advanced mode, but a cert that dies after one quarter felt worth flagging.

controllers/certificate/controller.go
evanphx: Adopted and commited!

PR #837, opened 2026-06-01
MIR-1198: track ephemeral versions on lease and re-sync activator on watch reconnect

evanphx:
## Problem

Ephemeral preview deploys intermittently fail at the ingress with `error acquiring lease` and **never recover** — the activator logs `tracked: false` / `fallback ticker fired` for a version even though its pool and a healthy RUNNING sandbox both exist. The signature is "works for a while, then stops" (existing ephemeral blackbox tests pass).

## Root cause

`a.versions[key]` (the version→pool mapping every lease lookup keys off) is populated by only three places, and ephemeral versions hit none of the synchronous ones reliably:

1. **On-demand pool creation didn't seed it.** Ephemeral versions skip activation at deploy, so the DeploymentLauncher never pre-creates their pool — they go through `requestPoolCapacity`'s on-demand branch, which set only `a.pools`, not `a.versions`/`a.poolSandboxes`. Only the launcher-found branch (normal deploys) seeded all three. So an ephemeral version was untracked until the async watcher happened to discover its sandbox.

2. **The watcher is the only continuous populator, and it loses events on reconnect.** `WatchIndex` (`pkg/entity/store.go:1538`) is a raw etcd watch from the current revision — no snapshot/replay on (re)connect — and the watch retry loop just re-calls it after a 5s sleep. The only full store reconcile (`recoverPools` + `recoverSandboxes`) runs **once**, at startup. So a watch reset (compaction, leader change, network blip) silently drops events; a sandbox that booted in that window (including after ephemeral scale-to-zero or a re-deploy with a new version ID) is never re-seen, and the version is stranded until a process restart. That's the "works, then stops."

## Fix

- **`registerVersionPoolLocked`** — seed `a.versions` + `a.poolSandboxes` from `requestPoolCapacity`'s on-demand, cached, and launcher-found branches, so the mapping is established when the pool is resolved rather than depending on the watcher.
- **`resyncFromStore`** (`recoverPools` + `recoverSandboxes`) — run on every reconnect in both `watchSandboxes` and `watchPools`, repairing both the version mapping and the sandbox objects after a missed-event window. `recoverPools`' `a.pools` write is now create-only so a live re-sync can't clobber an in-progress creation sentinel (startup behavior unchanged).
- Remove dead `state.pool.Ephemeral` reads (field never set true; ephemeral scaling is enforced by `EphemeralStrategy.MaxInstances()` since MIR-1159).

## Tests

6 new tests in `components/activator/activator_test.go`, including:
- `TestRequestPoolCapacityOnDemandSeedsVersionMapping` / `...Cached...` — direct regression (fail without the fix, proven by revert).
- `TestAcquireLeaseEphemeralOnDemand` — end-to-end lease via the on-demand path, bounded to 2s so a regression fails fast instead of hanging out the 120s cap.
- `TestRecoverPoolsPreservesInProgressSentinel` — live re-sync doesn't clobber a sentinel.
- `TestReSyncRecoversUntrackedEphemeralSandbox` — reproduces the bug-report steady state (pool + RUNNING sandbox in store, activator tracking nothing) and proves the reconnect re-sync recovers it.

42/42 activator tests pass in iso; `make lint` clean.

## Notes / follow-ups

- A residual sub-ms gap remains between the re-sync list and the new watch establishing (vs. the multi-second outage it replaces); a fully gap-free revision-resume watch would be a larger, separate change.

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -292,22 +292,10 @@ func (a *localActivator) AcquireLease(ctx context.Context, ver *core_v1alpha.App
 		return a.waitForSandbox(ctx, ver, service, false)
 	}
 
-	// Ephemeral pools must never scale beyond 1 instance — preview deploys
-	// are short-lived and are meant for inspection, not traffic. If the cached
-	// pool is ephemeral, wait for the existing sandbox instead of incrementing.
-	a.mu.RLock()
-	state, statePresent := a.pools[key]
-	isEphemeral := statePresent && !state.inProgress && state.pool != nil && state.pool.Ephemeral
-	a.mu.RUnlock()
-	if isEphemeral {
-		a.log.Info("ephemeral pool — not scaling, waiting for capacity",
-			"app", ver.App,
-			"version", ver.Version,
-			"service", service)
-		return a.waitForSandbox(ctx, ver, service, false)
-	}
-
-	// No RUNNING or PENDING sandboxes - need to scale up via pool
+	// No RUNNING or PENDING sandboxes - need to scale up via pool.
+	// Ephemeral preview deploys are capped at one instance by EphemeralStrategy
+	// (MaxInstances=1), so requestPoolCapacity won't scale them past their single
+	// sandbox — it returns the at-cap pool and the caller waits for it.
 	a.log.Info("no available sandboxes, requesting capacity from pool",
 		"app", ver.App,
 		"version", ver.Version,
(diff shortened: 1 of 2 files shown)


phinze (approved):
The whole team mob-reviewed this one, so consider this the chorus speaking. We traced the lease lookup through all three `requestPoolCapacity` branches, went into the launcher to work out *why* ephemerals never get a pre-created pool (it's structural — the launcher only reconciles `App.ActiveVersion`, and previews live alongside it, so they never enter its field of view), read the watch and recovery paths until we were satisfied the re-sync is safe to run against live state, and then went grepping through garden's journal to watch the bug actually happen.

Verdict from the group: the diagnosis holds, and we genuinely like that the fix *shrinks* the surface instead of growing it — pulling the version/pool bookkeeping into `registerVersionPoolLocked` means one place now owns "a version resolved to a pool," instead of one branch doing it and two quietly forgetting. The create-only `recoverPools` change with the sentinel-preservation test is exactly the care this function needs now that it runs on every reconnect. The tests landed well with everyone, especially the revert-proven regressions and the 2s-bounded end-to-end lease test.

And garden settled the "is this real?" question for the room: the very ephemeral version from the original report (`mirendev-...ZAm`) shows the `tracked: false`, count-0 stranding signature **18 times** while a healthy RUNNING sandbox exists for it. Not theoretical — recurring in production.

The mob pushed a small follow-up commit (`bd99e3c3`) addressing three nits we turned up: (1) **a fix** — `resyncFromStore` now wakes parked waiters after a reconnect reconcile, because `recoverSandboxes` adopts sandboxes without the watcher's per-sandbox notify, so an in-flight waiter would otherwise sit until the 60s fallback ticker (and log the misleading "channel notification may have failed"); (2) a comment spelling out why create-only `recoverPools` deliberately won't refresh a stale cached pool (the re-sync's store read can be older than an in-flight patched revision, so a refresh could move the cache *backward* — it self-heals via the Patch-conflict path instead); (3) a comment noting the double re-sync on a single blip is intentional and cheaper than cross-goroutine debounce.

One thing the garden dig surfaced that we deliberately kept *out* of this PR: a separate capacity-release notification gap — `ReleaseLease` doesn't wake parked waiters, so capacity-bound requests starve to the 120s timeout, which turns out to be the bulk of garden's ticker/timeout noise. Filed as MIR-1201 (https://linear.app/miren/issue/MIR-1201) rather than smuggling it in here.

The chorus approves. 🎭

PR #838, opened 2026-06-02
Clean up app log output

evanphx:
## Summary
- Parse structured JSON log lines at sandbox log ingress: extract `msg` as body, strip `time`/`level`, promote remaining fields to first-class log attributes, preserve original as `user.orig_msg`
- Namespace internal log attributes under `miren.*` prefix (`miren.sandbox`, `miren.service`, `miren.stage`, `miren.container`, `miren.version`, `miren.short_id`)
- Display the real entity short ID in log brackets (e.g. `[CBZ]` instead of `[clu…jt1uPwog]`)
- Hide `miren.*` attributes from CLI text output — only user/app attributes are shown
- Backwards-compatible queries: sandbox and service filters match both old and new field names

Before:
```
S 2026-05-30 20:36:20: [clu…jt1uPwog] {"time":"...","level":"INFO","msg":"processing step","component":"provisioner"} container=app sandbox=sandbox/... service=web stage=app-run version=...
```

After:
```
S 2026-05-30 20:36:20: [CBZ] processing step component=provisioner cluster_id=ZA8 saga=cluster_teardown
```

Closes MIR-1189

## Test plan
- [x] Unit tests for JSON log parsing at ingress (`TestParseStructuredJSON`, `TestProcessLineJSON`)
- [x] Unit tests for CLI display formatting (`TestPrintLogEntry`, `TestFormatAttributes`)
- [x] Unit tests for backwards-compatible filter construction (`TestBuildFilterWithService`)
- [ ] Manual test with `miren logs` against a running app to verify output format
- [ ] Verify `miren logs sandbox <id>` works for both old and new log entries
- [ ] Verify `miren logs --service <name>` works for both old and new log entries

### DIFF
--- a/cli/commands/logs.go
+++ b/cli/commands/logs.go
@@ -339,8 +339,7 @@ func printLogEntry(ctx *Context, l *app_v1alpha.LogEntry) {
 }
 
 var hiddenAttributes = map[string]bool{
-	"source":        true,
-	"user.orig_msg": true,
+	"source": true,
 }
 
 func formatAttributes(m map[string]string) string {
--- a/cli/commands/logs.go
+++ b/cli/commands/logs.go
@@ -29,7 +29,7 @@ func buildFilterWithService(userFilter, service string) string {
 	if service == "" {
 		return userFilter
 	}
-	serviceFilter := fmt.Sprintf("service:%q", service)
+	serviceFilter := fmt.Sprintf("(service:%q OR miren.service:%q)", service, service)
 	if userFilter == "" {
 		return serviceFilter
 	}
@@ -313,13 +313,19 @@ func printLogEntryJSON(ctx *Context, l *app_v1alpha.LogEntry) {
 
 func printLogEntry(ctx *Context, l *app_v1alpha.LogEntry) {
 	prefix := ""
-	if l.HasSource() && l.Source() != "" {
+	if l.HasAttributes() {
+		if shortID, ok := l.Attributes()["miren.short_id"]; ok && shortID != "" {
+			prefix = "[" + shortID + "] "
+		}
+	}
+	if prefix == "" && l.HasSource() && l.Source() != "" {
 		source := l.Source()
 		if len(source) > 12 {
 			source = source[:3] + "…" + source[len(source)-8:]
 		}
 		prefix = "[" + source + "] "
 	}
+
 	attrs := ""
 	if l.HasAttributes() {
 		attrs = formatAttributes(l.Attributes())
@@ -332,13 +338,24 @@ func printLogEntry(ctx *Context, l *app_v1alpha.LogEntry) {
 		attrs)
 }
 
+var hiddenAttributes = map[string]bool{
+	"source":        true,
+	"user.orig_msg": true,
+}
+
 func formatAttributes(m map[string]string) string {
 	if len(m) == 0 {
 		return ""
 	}
 	keys := make([]string, 0, len(m))
 	for k := range m {
+		if hiddenAttributes[k] || strings.HasPrefix(k, "miren.") {
+			continue
+		}
 		keys = append(keys, k)
+	}
+	if len(keys) == 0 {
+		return ""
 	}
 	slices.Sort(keys)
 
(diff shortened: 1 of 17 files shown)


phinze (approved):
**Paul Bunyan and Claude the Whistlepunk here.** First off, real fine work you're doing in the logyard, this is a good clean haul. We walked the whole length of it, stump to canopy. We like where she's felled, and that `{`-gate keeping plain-text loggers off the clock is a clean cut. We're gonna take the pen for a spell instead of nailing notes to every tree, so here's the lay of the land.

**One genuine widowmaker: user JSON shadows our `miren.*` attribution.** Both `WriteEntry`s merge `Attributes` first, then `Extra`, so `Extra` wins, and `Extra` is raw fields straight off the app's own JSON line. A log line like `{"msg":"hi","miren.sandbox":"sandbox/someone-else"}` overwrites the real sandbox id, and that entry comes out in *somebody else's* `miren logs`. Now, CodeRabbit already marked this tree, and it's logged as felled in `ec13723`, but that commit dragged the merge out of `log.go` and into the writers, and the precedence grew right back. Credit where it's due, but the stump didn't take. We'll flip it so the internal attrs win.

**Pool reuse: we'd like to pull the `specsMatch` change back out of the load.** Comparing `LogAttribute` makes every running pool look stale come upgrade and get rolled, which is a fleet-wide redeploy to relabel some logs. We don't reckon that's worth the diesel. The rename only runs forward, the `(sandbox:X OR miren.sandbox:X)` queries already catch the old logs, reattach flips the computed `miren.*` keys for free when the controller restarts, and any real deploy ships a new image and cuts a fresh pool with clean labels anyhow. Worst that's left behind is a little old `service=`/`stage=` sawdust on historical logs, and it ages out with retention. Paul's fine living with that. Unless you're fixing to tear out the OR-query half soon (which'd want the old records gone first), we'll revert it and the PR comes in lighter.

**Perf, mostly a flag for the next trip.** `user.orig_msg` stuffs the whole original line onto every parsed entry, so JSON loggers near enough double the bytes they store and ship. Worth making opt-in or dropping it outright, since the fields are already split out of the log. And the parser news up a fresh `json.Decoder` per line on the busiest skid road in the system, with nothing benchmarking it: the existing benches feed plain text and only ever hit the fast-path bail. We'll add a JSON bench with `-benchmem`, then see if a hand-rolled flat-object scanner earns its keep.

**What we tipped our hats to:** the streaming tokenizer instead of a full unmarshal, the integration test that round-trips a real `miren.sandbox` through the dotted-field query (that's what convinced us LogsQL handles dotted names, not just a string check), the reused `extra` map being safe since both writers marshal synchronously, and `Stderr()` getting its own map so the two streams don't cross saws. Plus you'd already bucked the real bot findings before we showed up.

Plan: take the pen, push the shadowing fix, likely revert the pool check, add a bench, then circle back on the perf calls. We'll holler down the line as commits land.

phinze (changes requested):
Flipping this to request-changes, and not 'cause anything above changed our minds, the haul's still good. Just chocking the wheels so she doesn't roll out before our commits are on the load. We're still sawing: the attr-shadowing fix and the pool-check revert at minimum. We'll flip her back to approved once the timber's stacked.

evanphx:
@phinze Totally agree, adjusted.

phinze (approved):
Pushed our remaining work on top of your two commits (sorry for the overlap — we'd started in parallel before syncing up). Three commits:

- **Benchmark** for the structured-log ingest path. The existing SandboxLogs benchmarks only fed plain text, so they never measured the JSON path.
- **gjson rewrite of scanJSON**, carrying your `miren.` escape forward unchanged. `json.Decoder.Token()` was allocating ~70 times even for a 2-field line (worse with nested pino req/res objects). On line shapes sampled from real garden/club logs it's 6–9x faster and drops structured lines from 73–194 allocs to 1.
- **One follow-up fix**: extended the shadowing escape to the `source` key. It's the one internal attribution field not under the `miren.` prefix, so an app logging `{"source":"..."}` could still shadow it. Same dash-escape, with a test.

Lint is clean (Go + docs). Flipping back to approve.

PR #839, opened 2026-06-02
Fix cluster not reconnecting to recreated POP VMs

evanphx:
## Summary

- When a POP VM is recreated (same POPXID, new instance), `HandleConnectionRequest` was short-circuiting on the stale map entry without verifying liveness — the cluster sat on the dead QUIC connection for minutes, ignoring all `connection.request` messages from cloud
- Now, a duplicate connection request cancels the old entry and establishes a fresh connection. The `servePOP` defer is guarded (`m.conns[pc.popXID] == pc`) so a replaced goroutine can't delete the new entry

Fixes MIR-1133

## Test plan

- [x] `TestPOPManagerReplaceStaleConnection` — verifies a second request for the same POPXID replaces the entry rather than being ignored
- [x] Existing tests pass (11/11), lint clean
- [ ] Deploy to staging and recreate a POP VM (`terraform taint` + `apply`) to confirm cluster reconnects immediately


phinze (approved):
POP goes the reconnect 💥♻️

PR #840, opened 2026-06-03
Fix NodePort creating duplicate iptables DNAT rules (MIR-1188)

evanphx:
## Summary

- Remove legacy iptables-based NodePort DNAT from the sandbox controller — the service controller already handles this via nftables with proper idempotency (flush+rebuild) and periodic GC
- Loop-delete in `unconfigurePort` so stale duplicate rules from before this change get cleaned up as sandboxes are recycled
- Fix `ensureServiceForPorts` to create a Service entity when any port declares a `node_port`, not only when the port type is non-HTTP — without this the service controller would never see HTTP-typed NodePorts

## Test plan

- [x] `make lint` — 0 issues
- [x] `hack/it controllers/sandbox` — 86 tests pass
- [x] `hack/it controllers/service` — 22 tests pass
- [x] `hack/it controllers/deployment` — 75 tests pass
- [ ] Deploy to staging and verify no duplicate iptables DNAT rules accumulate
- [ ] Verify NodePort connectivity works for both HTTP and non-HTTP services

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -1326,15 +1326,17 @@ func (l *Launcher) ensureServiceForPorts(ctx context.Context, app *core_v1alpha.
 		}
 	}
 
-	// Filter: only act on services with at least one non-HTTP port
-	hasNonHTTP := false
+	// A Service entity is needed when the service has non-HTTP ports (for
+	// cluster-IP load balancing) or when any port declares a NodePort (so the
+	// service controller can install nftables DNAT rules on every node).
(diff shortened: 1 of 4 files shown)


phinze (approved):
Turns out we didn't need NodePort DNAT in *two* places — we don't like NodePorts THAT much. Traced it end to end: the service controller already does everything the sandbox iptables rules did (localhost path included), and the launcher fix makes sure HTTP NodePorts actually reach it. Two implementations enter, the better-behaved one leaves. 👍

PR #841, opened 2026-06-03
Build robust revision-resume abstraction around WatchIndex (MIR-1200)

evanphx:
## Problem

Many processes use `WatchIndex` to react to entity changes, but the watch starts "from now" and can stop at any time (etcd errors, RPC disconnects, server restarts). Several consumers don't re-watch, and even those that do **silently lose every change that happens while the watch is down** — a class of latent timebombs (`components/ipalloc` was a one-shot with no recovery at all).

## What this adds

A self-healing client-side abstraction, **`pkg/entity/indexwatch`**, that watches correctly *and* reconciles the gap between watch requests — built on etcd's native revision-resume so it holds only **O(1) state** (a revision cursor), not a copy of the index.

```
start / post-compaction:  List → EventSync (full set) + cursor = list.revision
steady state:             watch WithRev(cursor+1) + WithProgressNotify
                            put/delete → Added/Updated/Deleted, cursor = ModRev
                            progress watermark → advance cursor
transient disconnect:     re-watch WithRev(cursor+1)   ← gap-free replay, no re-list
compacted (cursor stale): List again → new EventSync    ← only re-snapshot path
```

The initial sync and any post-compaction re-sync are delivered as a **single `EventSync`** carrying the complete current set; consumers reconcile it in their own domain (replace cache, treat absent ids as removed) — the list-reconcile work these consumers already do. Live changes arrive as `EventAdded`/`EventUpdated`/`EventDeleted`.

### Server/store plumbing
- `rpc.yml`: `EntityOp.revision`, `watch_index.from_revision`, `list.revision` (regenerated)
- `store.WatchIndex(fromRev)` uses `clientv3.WithRev` + `WithProgressNotify`; new `ListIndexRevision` surfaces the snapshot revision
- server emits per-event `ModRevision`, progress watermarks, and a distinct compaction signal (`EntityOperationProgress` / `EntityOperationCompacted`)

### Consumer migration
- **`components/ipalloc`** migrated to the watcher, killing its one-shot watch that silently stopped allocating IPs if the watch ever ended (and fixing a latent gap where pre-existing IP-less services never got addresses).
- **`pkg/dns`** sandbox watch migrated; the initial `EventSync` is processed synchronously so DNS is warm before `Watch` returns, and a snapshot sweep removes sandboxes deleted while the watch was down.
- Other direct `WatchIndex` callers pass `from_revision: 0` ("from now"), so existing behavior is preserved.

## Testing
- 7 `indexwatch` unit tests (initial sync, live events, **resume-without-resnapshot**, compaction, progress watermark, blocking backpressure, stop)
- 2 real-etcd store tests proving `ListIndexRevision` returns a real revision and `WithRev` replays only post-cursor events
- ipalloc + `TestServiceController/handles_service_IP_allocation` (real entity-server integration), dns (9 tests)
- Full `make test` green (4389 tests, 0 failures); `make lint` clean

## Deferred follow-ups (work unchanged today via `from_revision: 0`)
- `activator` watches — highest-risk consumer (routing/scaling hot path); its own focused unit
- `ReconcileController` — test-coupling churn
- indexwatch metrics/observability

Closes MIR-1200.

### DIFF
--- a/api/entityserver/entityserver_v1alpha/constants.go
+++ b/api/entityserver/entityserver_v1alpha/constants.go
@@ -10,6 +10,14 @@ const (
 	EntityOperationUpdate EntityOperation = 2
 	// EntityOperationDelete indicates an entity was deleted
 	EntityOperationDelete EntityOperation = 3
+	// EntityOperationProgress is a watch watermark carrying the latest observed
+	// revision with no entity payload. It lets a client advance its resume cursor
+	// during idle periods (etcd progress notifications).
+	EntityOperationProgress EntityOperation = 4
+	// EntityOperationCompacted signals that the requested watch start revision has
+	// been compacted away. The client must re-list to obtain a fresh snapshot and
+	// resume from the new revision. Revision carries the compaction revision.
+	EntityOperationCompacted EntityOperation = 5
 )
 
 // OperationType returns the typed operation for this EntityOp
@@ -30,4 +38,14 @@ func (v *EntityOp) IsUpdate() bool {
 // IsDelete returns true if this is a delete operation
 func (v *EntityOp) IsDelete() bool {
 	return v.Operation() == int64(EntityOperationDelete)
+}
+
+// IsProgress returns true if this is a watch progress watermark
+func (v *EntityOp) IsProgress() bool {
+	return v.Operation() == int64(EntityOperationProgress)
+}
+
+// IsCompacted returns true if this signals the watch start revision was compacted
+func (v *EntityOp) IsCompacted() bool {
+	return v.Operation() == int64(EntityOperationCompacted)
 }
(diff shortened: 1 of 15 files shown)


phinze (approved):
Mob-reviewed live, so this is the recap, not a novel. 🪿

Love the design: one `int64` cursor, let etcd replay the gaps, re-list only on compaction. O(1) state where you'd expect a cache. The DNS migration came out cleaner than the code it deleted.

The one gremlin we hit: the server broadcast the new `Progress`/`Compacted` ops to *every* watcher, including the `from_revision: 0` legacy ones. The activator nil-derefs any non-delete op that carries no entity → panic on the routing hot path, and `len(Events)==0` mistook etcd's `Created` handshake for a watermark — making it immediate *and* advancing the cursor past un-replayed backlog, reopening the exact gap this PR closes. Fixed on the branch: gate `Progress`/`Compacted` on `from_revision > 0`, defer to etcd's own `IsProgressNotify()`, four tests, plus a second commit to back off on clean stream ends.

On the record for future-us, not a blocker: `EventSync` bakes in "pull the whole index into memory and reconcile." Perfect now, but `kind -> sandbox` / `kind -> appversion` get big — optimization fodder later (incremental/diff resync).

Ship it. 🚀

PR #842, opened 2026-06-04
Convert build process to saga framework

phinze:
This converts the build pipeline (`BuildFromTar` / `BuildFromPrepared`) from a ~450-line synchronous function into the saga framework from MIR-439. Eleven actions cover the pipeline, from receiving the source tar through building the image, creating the config and version entities, provisioning addons, and activating the new version, each with a compensating undo so a failure partway through rolls back cleanly instead of leaving half-built state behind.

Two small primitives bridge the saga world (which needs everything serializable for crash recovery) to the non-serializable RPC stream state. `StreamRegistry` stages the inbound tar to disk so a recovered saga can still find the source after the original `io.Reader` is gone, and `StatusRegistry` carries the outbound progress stream, handing back a no-op sender when the saga is recovering and the client is long gone. A `SagaBuilder` wraps the whole thing as a drop-in implementation of the existing Builder RPC interface, and `coordinate.go` selects it behind the existing `labs.Sagas()` flag.

The interesting part is what manual testing turned up. Running a real deploy through the saga path failed at the very end with "execution has status pending, expected completed," even though the build had clearly succeeded: the saga's entire state had frozen at its initial pending write. The cause was in the saga framework itself, not the build code. `EntityStorage.Save` and `EACStorage.Save` persisted through create-if-absent primitives, so every save after the first was silently dropped. The sandbox saga from MIR-439 had been hitting this the whole time, it just never noticed because nothing read its state back. The fix makes Save a real upsert. A second, smaller bug fell out of the same testing: each executor's recovery treated another executor's sagas (build and sandbox share storage) as errors, so we scoped recovery to skip definitions an executor doesn't own.

To keep both from regressing, there's a storage conformance suite that runs the persistence scenarios against all three Storage backends. The deeper question this raised, whether the mock store can be trusted to stand in for etcd in the first place, is addressed separately in the entity.Store conformance suite (#847).

### DIFF
--- a/pkg/saga/storage_conformance_test.go
+++ b/pkg/saga/storage_conformance_test.go
@@ -1,0 +1,169 @@
+package saga
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"miren.dev/runtime/pkg/entity/testutils"
+)
+
+// storageFactory builds a fresh, empty Storage for one test. It registers any
+// teardown via t.Cleanup so callers get a clean backend per subtest.
+type storageFactory struct {
+	name string
+	make func(t *testing.T) Storage
+}
+
+// allStorageBackends returns every production Storage implementation behind a
+// uniform factory. The whole point of this suite is that every backend must
+// satisfy the same contract: the bug in MIR-441 lived in two of these three
+// (EntityStorage and EACStorage both used create-if-absent semantics and
+// silently dropped every save after the first), while MemoryStorage was
+// correct, so memory-backed unit tests stayed green while production froze
+// every saga at its initial pending state. Running the same scenarios against
+// all backends is what closes that gap.
+//
+// This suite tests a different question than pkg/entity's Store conformance
+// suite, and the two are complementary. The entity suite proves MockStore
+// behaves like the real EtcdStore on the underlying write primitives. This
+// suite proves saga storage *uses* those primitives correctly (the MIR-441 bug
+// was a create-only call where an upsert was needed, which no amount of
+// store-level testing would catch). Because the entity suite vouches for the
+// mock, the EntityStorage backend here is mock-backed on purpose and we do not
+// add a real-etcd backend; the EACStorage backend additionally exercises the
+// EntityAccessClient RPC path, which the entity suite does not reach.
+func allStorageBackends() []storageFactory {
+	return []storageFactory{
+		{
+			name: "MemoryStorage",
+			make: func(t *testing.T) Storage {
+				return NewMemoryStorage()
+			},
+		},
+		{
+			name: "EntityStorage",
+			make: func(t *testing.T) Storage {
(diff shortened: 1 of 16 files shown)


pkg/saga/storage_conformance_test.go
```diff
@@ -0,0 +1,169 @@
+package saga
```
evanphx, line 1: We picked this up in another PR already I'm pretty sure.

servers/build/build_saga.go
```diff
@@ -0,0 +1,733 @@
...
+	ProcfileServices map[string]string                    `json:"procfile_services,omitempty" saga:"procfile_services,optional"`
+	ExistingConfig   string                               `json:"existing_config_json" saga:"existing_config_json"`
+	CLIEnvVars       []*build_v1alpha.EnvironmentVariable `json:"cli_env_vars,omitempty" saga:"cli_env_vars,optional"`
+}
+
+type prepareConfigOut struct {
+	ConfigSpec string `json:"config_spec_json" saga:"config_spec_json"`
+}
+
+func prepareConfig(ctx context.Context, in prepareConfigIn) (prepareConfigOut, error) {
```
evanphx, line 238: meta-point: man these are so good for cleaning up the logic. It's immediately obvious what can influence the config.

servers/build/build_saga.go
```diff
@@ -0,0 +1,733 @@
...
+func undoSetActiveVersion(ctx context.Context, in setActiveVersionIn, out setActiveVersionOut) error {
+	if out.Skipped {
+		return nil
+	}
+	deps := saga.Get[*buildSagaDeps](ctx)
+	if err := deps.builder.appClient.SetActiveVersion(ctx, in.AppName, out.PreviousVersionID); err != nil {
+		return fmt.Errorf("restoring previous active version on %s: %w", in.AppName, err)
+	}
+	return nil
+}
```
evanphx, line 548: Image

servers/build/stream_registry.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+// ErrStreamUnavailable is returned by Stage when neither an active reader
+// nor a previously staged path exists for the given stream ID. This is the
+// expected failure when recovering a saga whose tar stream was lost in a
+// crash before the first stage attempt.
+var ErrStreamUnavailable = errors.New("stream unavailable")
+
+// StreamRegistry bridges non-serializable tar streams into the saga world by
+// staging them to durable filesystem paths. The pattern, per RFD-35:
```
evanphx, line 22: This is a good abstraction because I could see us eventually having the ability to consume the tar from other places and this makes that easy.

evanphx (approved):
It's great cleanup!

pkg/saga/storage_conformance_test.go
phinze: Very similar but not the same! Other PR was _entity store_ conformance test, here `pkg/saga` has its own storage indirection layer we wrote the same style of test on.

PR #844, opened 2026-06-04
Add exhaustive switch vetting and enforce it across the tree

evanphx:
## Summary

Enables the `exhaustive` analyzer (bundled in golangci-lint) and makes every enum switch in the tree satisfy it, so adding a new enum member forces a compile-time decision at each relevant switch instead of silently slipping through a `default`.

## What changed

- **Lint config**: enable the `exhaustive` linter in `.golangci.yml` with `default-signifies-exhaustive: false` (a bare `default:` does not by itself make a switch exhaustive). It honors `//exhaustive:ignore` directives.
- **Enforcement (38 switches)**: each switch now enumerates all enum members explicitly. Where the missing members should behave exactly like the existing `default`, they're grouped into a case that `fallthrough`s into the default — so **runtime behavior is unchanged**, but the cases are now documented and checked. Each added case carries a short comment explaining why those members share that path (terminal / transient / unexpected / etc.).

## Intentional exceptions

`//exhaustive:ignore` is kept only on switches over large external/stdlib enums where listing every arm is impractical, with the rationale inline:

| Type | Members | Location |
|------|---------|----------|
| `syscall.Errno` | ~130 | `servers/httpingress/httpingress.go` |
| `tea.KeyType` | ~90 | `pkg/ui/prompt.go`, `cli/commands/deploy_ui.go` |
| `reflect.Kind` | ~27 | `pkg/entity/reader.go`, `controllers/deployment/specs_match_test.go` |

## Coverage note

golangci-lint's path exclusions are global, so the already lint-exempt paths (`lsvd/`, `disk/`, `dataset/`, `pkg/cel`, `pkg/containerdx`, `*.gen.go`) are not exhaustive-checked. The switches in those paths (3 in `lsvd/`) were still made exhaustive in this PR; they're just not re-enforced going forward.

## Verification

- `golangci-lint run ./...` — no issues (exhaustive active; confirmed it flags a removed case)
- `go build ./...` and affected test packages compile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,9 +72,6 @@ jobs:
       with:
         version: v2.6.1
 
-    - name: go vet (exhaustive switch checking)
-      run: make vet
-
     - name: Check go.mod tidiness
       run: |
         # Run go mod tidy and check if there are any changes
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,9 @@ jobs:
       with:
         version: v2.6.1
 
+    - name: go vet (exhaustive switch checking)
+      run: make vet
+
     - name: Check go.mod tidiness
       run: |
         # Run go mod tidy and check if there are any changes
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,10 +1,19 @@
 version: "2"
 
 linters:
+  enable:
+    - exhaustive
+
   disable:
     - errcheck
 
   default: standard
+  settings:
+    exhaustive:
+      # A bare `default:` does not by itself make a switch exhaustive; every
+      # enum member must be listed (use a `//exhaustive:ignore` directive to
+      # opt a specific switch out).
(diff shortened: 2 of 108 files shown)


phinze (changes requested):
Yay, one step closer to Rust ;) We checked the few no-op cases (`SetPortStatus` x2, lsvd `FlushNo`) and they're all safe.

One ask before merging: can we wire it through `golangci-lint` (already in the lint job, bundles the same `exhaustive` analyzer) instead of a standalone `go vet` step? We'd get the shared single load, parallelism, and caching for free, and drop the `vet` target, CI step, and `go.mod`/`go.sum` churn. With golangci-lint already pinned, the bundled version is just as deterministic.

Only difference: golangci-lint still excludes a few dirs (`lsvd/`, `disk/`, etc.), so those won't be covered. Fine to punt, whether those should be linted at all is a separate cleanup TODO.

evanphx:
Sure, let me give it a look.

phinze (approved):

PR #845, opened 2026-06-05
Coalesce repeated log lines in `logs -f` into a live summary

evanphx:
## Summary

When following logs with `m logs -f` on an app that emits a steady stream of near-identical lines (e.g. a 1-second health/DNS ping), the terminal fills with lines that differ **only by their timestamp**, burying genuinely new output:

```
U 2026-06-04 16:58:11: [router] method=POST path="/dns-query" host=clusters.miren.garden ... duration_ms=0 ...
U 2026-06-04 16:58:12: [router] method=POST path="/dns-query" host=clusters.miren.garden ... duration_ms=0 ...
U 2026-06-04 16:58:13: [router] method=POST path="/dns-query" host=clusters.miren.garden ... duration_ms=0 ...
```

Now consecutive lines that are identical except for the timestamp collapse: the first occurrence prints in full, and a single **live-updated** line beneath it summarizes the run, redrawn in place as it continues:

```
U 2026-06-04 16:58:11: [router] method=POST path="/dns-query" host=clusters.miren.garden ... duration_ms=0 ...
[ Repeated 14x over 14s ]
```

## Details

- `renderLogEntry` splits a log entry into its full display string and a timestamp-independent **signature**; consecutive entries with the same signature are treated as a run.
- `logCoalescer` prints the first occurrence in full, then redraws `[ Repeated Nx over <span> ]` in place (`\r\033[K`) for each repeat. A distinct line commits the summary with a newline and prints below it.
- The span is computed from second-truncated timestamps so it matches what the user sees; sub-second bursts (multiple lines in one displayed second) show a bare `[ Repeated Nx ]`.
- **Coalescing engages only for text + `--follow` + TTY.** JSON (`--format json`), piped, and non-follow output emit every line verbatim, so machine consumers and `grep` are unaffected. Wired in via the `logPrinter` factory, which now also returns a `flush` to commit a dangling summary when the stream ends.

## Testing

- New unit tests: `TestRenderLogEntry`, `TestLogCoalescer` (collapse / distinct-line commit / single line / same-second), `TestFormatRunSpan`, `TestRunSpanSuffix`, and `TestLogPrinterNonTTYNeverCoalesces` (guards the machine-output contract).
- Full `cli/commands` package: **234 tests pass**.
- `make lint`: 0 issues.

Closes MIR-1221.

https://github.com/user-attachments/assets/51525b1b-004b-4516-8321-3b71fc431505

### DIFF
--- a/cli/commands/logs.go
+++ b/cli/commands/logs.go
@@ -12,6 +12,7 @@ import (
 	"miren.dev/runtime/pkg/rpc"
 	"miren.dev/runtime/pkg/rpc/standard"
 	"miren.dev/runtime/pkg/rpc/stream"
+	"miren.dev/runtime/pkg/ui"
 )
 
 // normalizeSandboxID ensures the sandbox ID has the "sandbox/" prefix
@@ -171,7 +172,8 @@ func LogsSystem(ctx *Context, opts struct {
 	}
 	// When no --last and no --follow, ts is nil → server returns last 100 lines
 
-	printer := logPrinter(ctx, opts.IsJSON())
+	printer, flush := logPrinter(ctx, opts.IsJSON(), opts.Follow)
+	defer flush()
 
 	ac := app_v1alpha.LogsClient{Client: cl}
 	callback := stream.Callback(func(chunk *app_v1alpha.LogChunk) error {
@@ -200,10 +202,12 @@ type logDispatchArgs struct {
 // dispatchLogs handles protocol negotiation and dispatches to the appropriate
 // log streaming method based on server capabilities.
 func dispatchLogs(ctx *Context, cl *rpc.NetworkClient, args logDispatchArgs) error {
-	printer := logPrinter(ctx, args.json)
+	printer, flush := logPrinter(ctx, args.json, args.follow)
+	defer flush()
 
 	// For app queries, wrap the printer to skip system logs that may have
-	// leaked into app log storage due to entity field collisions.
+	// leaked into app log storage due to entity field collisions. The wrapper
+	// sits in front of the coalescer so filtered-out lines aren't counted.
 	if args.app != "" {
 		inner := printer
 		printer = func(l *app_v1alpha.LogEntry) {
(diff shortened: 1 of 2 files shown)


cli/commands/logs.go
```diff
@@ -330,12 +348,109 @@ func printLogEntry(ctx *Context, l *app_v1alpha.LogEntry) {
...
+
+func newLogCoalescer(ctx *Context) *logCoalescer {
+	return &logCoalescer{ctx: ctx}
+}
+
+func (c *logCoalescer) print(l *app_v1alpha.LogEntry) {
+	display, signature := renderLogEntry(l)
+	ts := standard.FromTimestamp(l.Timestamp())
+
+	if c.count > 0 && signature == c.sig {
```
phinze, line 394: Watched this live during a deploy and noticed: consecutive-only matching means interleaved sandboxes never coalesce. With 3 replicas of a health-pinging app, today you get

```
S 21:54:01: [aaa] ping ok
S 21:54:01: [bbb] ping ok
S 21:54:01: [ccc] ping ok
S 21:54:02: [aaa] ping ok
S 21:54:02: [bbb] ping ok
...forever
```

since the run resets on every line, the feature quietly sits out exactly when the spam is worst. The dream would be per-signature runs with one live line each:

```
S 21:54:01: [aaa] ping ok
[ Repeated 14x over 14s ]
S 21:54:01: [bbb] ping ok
[ Repeated 14x over 14s ]
S 21:54:01: [ccc] ping ok
[ Repeated 13x over 13s ]
```

That's a much hairier rendering problem, and at that point log following starts getting TUI-shaped, with the cursor dancing between N live regions. Consecutive-only seems right for v1; leaving this as food for thought rather than a ticket.

phinze (approved):
Paul here, Claude riding shotgun. We traced the wiring (Ctrl-C → deferred flush included), then deployed a 1s ping app to a dev cluster and watched it live. Works great, and the TTY/follow/JSON gating plus the non-TTY regression test give us a lot of confidence nothing machine-facing changed. One real-world observation inline.

cli/commands/logs.go
evanphx: Totally agree, I defered the more fun and more complex case like this!

PR #846, opened 2026-06-05
Issue workload identities on distributed runners

evanphx:
## Problem (MIR-1224)

The workload identity issuer holds the cluster's Ed25519 signing key and was only wired into the **coordinator's** embedded runner. Distributed runners built their sandbox controller with a `nil` issuer, so their sandboxes never received an identity token, the `MIREN_*` env vars, token refresh, or the on-demand token request server.

## Approach

Runners don't hold the signing key, so they mint tokens by calling the **coordinator over RPC**.

- Add `WorkloadIssuerInfo` and `IssueWorkloadToken` to the `RunnerRegistration` service. The coordinator mints tokens on behalf of runners.
- Introduce a `workloadidentity.TokenIssuer` interface. The concrete `*Issuer` satisfies it on the coordinator; distributed runners use a `remoteIssuer` that proxies minting over RPC. A background loop keeps the issuer URL in sync (re-registration, coordinator restart) and logs only on state transitions.
- Guard against boxing a typed-nil `*Issuer` into the interface field (would defeat the controller's `!= nil` checks and panic).

## Hardening (coordinator side)

- **Per-sandbox authorization:** a runner may only mint tokens for sandboxes scheduled to it, verified via the caller's mTLS client-certificate identity.
- **Full runner ID in the cert CommonName** (not an 8-char prefix), so a runner — which chooses its own `runner_id` at Join — cannot pick a colliding ID and impersonate another runner.
- **App claim derived server-side** from the sandbox entity rather than trusted from the caller, and removed from the RPC arguments entirely (it's part of the federated token subject).
- Startup queries **retry transient failures** so a brief blip doesn't leave a runner permanently without token issuance.

## Testing

`make lint` clean. New/updated unit tests cover: the `TokenIssuer` interface migration, runner-registration Join/ownership/app-derivation/cert-name, and the `remoteIssuer` (URL refresh + enabled-state transitions). All run on host (no containerd).

**Not included:** an end-to-end blackbox test asserting a distributed sandbox actually receives a token + env vars — it wants the live `make dev-distributed` peers topology and is tracked as a follow-up.

### DIFF
--- a/components/runner/remote_issuer.go
+++ b/components/runner/remote_issuer.go
@@ -3,80 +3,43 @@ package runner
 import (
 	"context"
 	"fmt"
-	"log/slog"
-	"sync"
 	"time"
 
 	"miren.dev/runtime/api/runner/runner_v1alpha"
 	"miren.dev/runtime/pkg/workloadidentity"
 )
 
-const (
-	// remoteTokenTimeout bounds a single token-minting RPC to the coordinator.
-	remoteTokenTimeout = 30 * time.Second
-
-	// issuerURLRefreshInterval is how often the cached issuer URL is re-synced
-	// with the coordinator.
-	issuerURLRefreshInterval = 5 * time.Minute
-)
+// remoteTokenTimeout bounds a single token-minting RPC to the coordinator.
+const remoteTokenTimeout = 30 * time.Second
 
 // remoteIssuer satisfies workloadidentity.TokenIssuer by proxying token minting
 // to the coordinator over RPC. Distributed runners do not hold the cluster
 // signing key, so they cannot mint tokens locally and instead ask the
 // coordinator, which holds the key.
 //
-// The issuer URL is cached but kept in sync by a background loop: it can change
-// while the runner is running (e.g. the cluster gains a DNS hostname during
-// re-registration), and the RPC client transparently reconnects after a
-// coordinator restart, so periodic polling re-syncs without a runner restart.
+// The issuer URL is captured once at startup. It is the cluster's OIDC issuer
+// anchor, derived from the coordinator's configuration at boot, so it is stable
+// for a coordinator's process lifetime; a change (e.g. the cluster gaining a DNS
+// hostname via re-registration) restarts the coordinator and is picked up when
+// the runner reconnects/restarts.
 type remoteIssuer struct {
-	ctx    context.Context
-	client *runner_v1alpha.RunnerRegistrationClient
-	log    *slog.Logger
-
-	mu        sync.RWMutex
+	ctx       context.Context
+	client    *runner_v1alpha.RunnerRegistrationClient
 	issuerURL string
-	enabled   bool
 }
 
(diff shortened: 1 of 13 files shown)


components/runner/remote_issuer.go
```diff
@@ -0,0 +1,152 @@
...
+	defer cancel()
+
+	info, err := r.client.WorkloadIssuerInfo(ctx)
+	if err != nil {
+		r.log.Warn("failed to refresh workload issuer info", "error", err)
+		return
+	}
+
+	enabled := info.Enabled()
+	if r.setEnabled(enabled) {
```
phinze: Staring at the transition handling here and wondering if it can be simplified. The coordinator only picks up an issuer at startup, so `enabled` and the URL are basically frozen for the process, and the loop plus `enabled` flag may be more than we need. Not a biggie though.

servers/runner/registration.go
```diff
@@ -604,6 +607,175 @@ func (s *RegistrationServer) RemoveRunner(ctx context.Context, req *runner_v1alp
...
+// runnerCertName is the client-certificate CommonName issued to a runner during
+// Join. It embeds the full runner ID so the coordinator can attribute an mTLS
+// connection back to a specific runner and authorize per-runner actions.
+//
+// The full ID (rather than a prefix) is required for authorization: a runner
+// chooses its own runner ID at Join, so a short prefix would let a malicious
+// runner pick an ID whose prefix collides with a victim's cert name and mint
+// tokens for the victim's sandboxes. A runner ID is a UUID, so "runner-" plus
+// the ID stays within the certificate CommonName length limit.
+func runnerCertName(runnerID string) string {
```
phinze, line 688: Full runner ID is right for the accidental-collision case. One thing worth flagging now that the ownership check leans on this CN: the `runner_id` behind it is whatever the caller handed Join, and Join `Put`s the node by ident, which overwrites rather than rejects. Since runner_ids aren't secret, a reusable-invite holder could Join as a known runner's id, get a cert with its CN, and mint for its sandboxes (and those identities federate outward). Fix belongs in Join, not here, so it's a follow-up, filed as MIR-1225.

phinze (approved):
Really like the shape. "Runners can't sign, so they ask the coordinator to" is the right call.

On the bot notes: +1 on the startup timeout (`runner.go` ~501). Everything else here bounds itself with `remoteTokenTimeout`, but this one rides the raw runner ctx, so a hung coordinator just stalls startup. That's the one worth fixing here. +1 on the `UnscheduledDenied` test nit too. We'd skip the fail-closed-on-empty-app suggestion though, since the local path mints with an empty app and appless sandboxes (system/buildkit) are real, so matching local feels better.

Two inline questions, neither blocking. Filed MIR-1225 for the runner_id one. Nice work!

components/runner/remote_issuer.go
evanphx: yup! done.

servers/runner/registration.go
evanphx: thanks!

PR #847, opened 2026-06-05
Add entity.Store conformance suite (MockStore vs EtcdStore)

phinze:
This came out of the MIR-441 saga work. A storage bug there traced back to something more fundamental: the saga code leaned on `EnsureEntity`'s create-if-absent semantics, `MockStore` and `EtcdStore` happened to agree on that behavior, and the bug surfaced two layers up instead of at the storage boundary where it belonged. The root issue wasn't the saga code, it was that nothing kept the mock honest about the real store.

The entity package has two `Store` implementations, `EtcdStore` for production and `MockStore` for the bulk of unit tests, and their parity was a hand-maintained promise. `mock.go` is sprinkled with comments like "Mirrors EtcdStore.UpdateEntity (store.go:687)," but a comment isn't a test. If the mock drifts, every consumer that trusts it silently loses validity.

So this puts the parity guarantee at the right layer. It runs the same scenarios against both backends across the entire `Store` interface: reads, writes (including the cardinality-aware merge and the ensure-then-replace upsert), indexes and collections, watches, and sessions. The `EtcdStore` backend needs a real etcd, so it runs in the dev environment.

Writing it turned up three places where the mock is a deliberate approximation rather than a faithful model, and rather than hide those, the suite records each with an explicit skip and a reason. `MockStore` numbers each entity's revision from 1 while `EtcdStore` stamps the global etcd revision (a freshly created entity can come back at revision 290), so the suite pins the contract callers can actually rely on, positive and strictly increasing per entity, not absolute values. `GetEntityAtRevision` ignores the revision argument on the mock and always returns the current entity. And sessions are stubs with no scoping. None of these needed a fix, they're reasonable simplifications, but now they're visible instead of folklore.

The upshot: for the core CRUD, index, and watch surface, the mock is now provably faithful to etcd and stays that way, and the three seams where it isn't are documented in the suite itself.

### DIFF
--- a/pkg/entity/store_conformance_test.go
+++ b/pkg/entity/store_conformance_test.go
@@ -1,0 +1,636 @@
+package entity
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"miren.dev/runtime/pkg/cond"
+)
+
+// Store conformance suite.
+//
+// The codebase has two Store implementations: EtcdStore (production) and
+// MockStore (the in-memory fake most unit tests run against). Their behavioral
+// parity has historically been maintained by hand, with comments in mock.go
+// like "Mirrors EtcdStore.UpdateEntity (store.go:687)". Nothing enforced it,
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Image

PR #848, opened 2026-06-05
Refresh runner certificate on start when listen IP changes

evanphx:
## Problem

`miren app run` and `miren sandbox exec` fail against distributed runners with:

```
tls: failed to verify certificate: x509: certificate is valid for
127.0.0.1, ::1, 10.128.0.45, not 10.128.0.47
```

**Root cause:** runner VMs are destroyed and recreated with a persistent disk mounted at `/var/lib/miren`. That disk keeps `runner/config.yaml` — the runner ID **and** the certificate issued at first `join`. On recreate the VM gets a new internal IP, so:

- The runner ID persists → the runner reattaches to its existing `Node` and `setupEntity` rewrites `Node.ApiAddress` to the **new** IP (so `sandbox exec` correctly dials the new IP), but
- The certificate persists → it's still the one minted for the **old** IP and is never re-issued → TLS server verification fails.

## Fix

Add a `RefreshCertificate` RPC. On start, the runner inspects its persisted certificate and, **only if** the discovered listen address isn't covered by the cert's SANs, asks the coordinator to re-issue, writes the new cert back to config, and serves it. This self-heals existing broken runners on their next restart.

A required-but-failed refresh is **fatal** — serving a stale cert would leave the runner silently unreachable.

### Authorization (entirely cert-derived, never from caller input)

1. A client certificate must be presented (mTLS peer cert).
2. It must chain to the cluster CA and be currently valid (`VerifyCert`).
3. It must be a runner certificate (CN `runner-*`, org `miren`).
4. Exactly one registered `Node` must match the CN's runner-ID prefix.

Step 4 ties refresh to live membership. caauth has no revocation, so this prevents a **removed** runner's still-valid certificate from perpetually renewing itself. It **fails closed** on zero or ambiguous matches (two runners sharing an 8-char ID prefix → rejected, not guessed).

### Supporting changes

- `pkg/rpc`: expose the verified peer certificate to handlers via `CurrentConnectionInfo`; add `ContextWithConnectionInfo` (used by the server and tests).
- `pkg/caauth`: add `Authority.VerifyCert` for an already-parsed certificate.
- Factor the runner cert CommonName (`runnerCertName`) and SAN construction (`buildRunnerSANs`) into shared helpers reused by `Join`.

## Tests

- `reissueRunnerCertificate`: happy path (new IP in SANs, CA-signed, CN preserved, via a real `Join`); rejections for nil peer, foreign-CA cert, non-runner CN, wrong org, unregistered cert, **removed runner**, and **ambiguous identity**.
- `RefreshCertificate` RPC rejects a call with no client certificate.
- `buildRunnerSANs` (IP vs DNS host) and `certCoversListenAddr` (covered/changed IP, DNS host, bad PEM).

`go build ./...` and `make lint` are clean.

> Note: also fixed a latent test-harness bug — `newTestServer` created its CA without `ValidFor`, so the root was instantly expired. It went unnoticed because no prior test verified a chain; the new tests do.

## Follow-ups (not in this PR)
- Encoding the **full** runner ID in the certificate at `join` would remove the residual 8-char-prefix ambiguity edge entirely.
- Optional: always-refresh on start / periodic renewal to also pick up CA rotation.

Fixes MIR-1223.

### DIFF
--- a/cli/commands/runner_start.go
+++ b/cli/commands/runner_start.go
@@ -4,6 +4,8 @@ package commands
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"net"
 	"net/netip"
@@ -16,6 +18,7 @@ import (
 
 	containerd "github.com/containerd/containerd/v2/client"
 	"golang.org/x/sync/errgroup"
+	"miren.dev/runtime/api/runner/runner_v1alpha"
 	"miren.dev/runtime/clientconfig"
 	containerdcomp "miren.dev/runtime/components/containerd"
 	"miren.dev/runtime/components/netresolve"
@@ -23,6 +26,7 @@ import (
 	"miren.dev/runtime/controllers/sandbox"
 	"miren.dev/runtime/metrics"
 	"miren.dev/runtime/observability"
+	"miren.dev/runtime/pkg/rpc"
 	"miren.dev/runtime/pkg/runnerconfig"
 )
 
@@ -43,6 +47,30 @@ func RunnerStart(ctx *Context, opts struct {
 		"coordinator", cfg.CoordinatorAddress,
 		"etcd_endpoints", cfg.EtcdEndpoints,
 		"network_backend", cfg.NetworkBackend)
+
+	// Determine listen address. If no explicit address is given, discover the
+	// machine's outbound IP (the one that would route to the coordinator) and
+	// advertise that so the coordinator knows how to reach this runner.
+	listenAddr := opts.ListenAddr
+	if listenAddr == "" {
+		port := "8444"
+		ip, err := discoverOutboundIP(cfg.CoordinatorAddress)
+		if err != nil {
+			return fmt.Errorf("could not discover outbound IP for listen address (use --listen to set manually): %w", err)
+		}
+		listenAddr = net.JoinHostPort(ip.String(), port)
+		ctx.Log.Info("discovered listen address", "addr", listenAddr)
+	}
+
+	// The runner's certificate is persisted in the config (often on a disk that
+	// outlives the VM). If the listen address has changed since the cert was
+	// issued, the persisted cert no longer covers our IP and clients (e.g.
+	// sandbox exec) will fail TLS verification. Re-issue the cert before we use
+	// it to serve. Refreshing is fatal when the address isn't covered: serving a
+	// stale cert would leave the runner silently unreachable.
+	if err := ensureRunnerCertificate(ctx, cfg, opts.ConfigPath, listenAddr); err != nil {
+		return err
+	}
 
 	// Create clientconfig from saved certs for RPC authentication
 	clientCfg := clientconfig.NewConfig()
@@ -136,20 +164,6 @@ func RunnerStart(ctx *Context, opts struct {
 
 	// Create errgroup for background tasks
 	eg, egCtx := errgroup.WithContext(sigCtx)
-
-	// Determine listen address. If no explicit address is given, discover the
-	// machine's outbound IP (the one that would route to the coordinator) and
-	// advertise that so the coordinator knows how to reach this runner.
-	listenAddr := opts.ListenAddr
-	if listenAddr == "" {
-		port := "8444"
-		ip, err := discoverOutboundIP(cfg.CoordinatorAddress)
-		if err != nil {
-			return fmt.Errorf("could not discover outbound IP for listen address (use --listen to set manually): %w", err)
-		}
-		listenAddr = net.JoinHostPort(ip.String(), port)
-		ctx.Log.Info("discovered listen address", "addr", listenAddr)
-	}
 
 	// Build runner configuration
 	runnerCfg := runner.RunnerConfig{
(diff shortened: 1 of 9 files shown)


servers/runner/registration.go
```diff
@@ -360,6 +344,151 @@ func (s *RegistrationServer) Join(ctx context.Context, req *runner_v1alpha.Runne
...
+		return nil, fmt.Errorf("runner is not registered")
+	case 1:
+		// authorized
+	default:
+		s.Log.Warn("RefreshCertificate rejected: runner identity is ambiguous",
+			"subject", peer.Subject.String(), "matches", len(matches))
+		return nil, fmt.Errorf("runner identity is ambiguous")
+	}
+
+	ips, dnsNames := buildRunnerSANs(listenAddr)
```
phinze, line 458: `listenAddr` is the one caller-controlled input in the issued cert, so a runner cert is now a forever-renewable ticket for arbitrary SANs (own CN, so audit trail stays honest). Fine for the compromised-runner threat model, but it composes badly with MIR-1225: a stolen CN via reusable invite becomes *renewable* instead of expiring in a year. Fixing 1225 at Join closes it; just flagging the blast radius until then.

servers/runner/registration.go
```diff
@@ -360,6 +344,151 @@ func (s *RegistrationServer) Join(ctx context.Context, req *runner_v1alpha.Runne
...
+	case 0:
+		s.Log.Warn("RefreshCertificate rejected: runner is not registered",
+			"subject", peer.Subject.String())
+		return nil, fmt.Errorf("runner is not registered")
+	case 1:
+		// authorized
+	default:
+		s.Log.Warn("RefreshCertificate rejected: runner identity is ambiguous",
+			"subject", peer.Subject.String(), "matches", len(matches))
+		return nil, fmt.Errorf("runner identity is ambiguous")
```
phinze: Fail-closed is right. Side effect: a prefix collision = runner fatally can't start after an IP change, re-join only escape. Birthday math says never, so this is fine — your full-runner-ID-in-cert follow-up is the real fix, and we've noted on MIR-1225 that it should ride along with that work.

cli/commands/runner_start.go
```diff
@@ -300,3 +314,107 @@ func RunnerStart(ctx *Context, opts struct {
...
+
+	ctx.Log.Info("runner certificate refreshed", "listen_addr", listenAddr)
+	return nil
+}
+
+// certCoversListenAddr reports whether the leaf certificate in certPEM carries a
+// SAN matching the host of listenAddr: an IP SAN for an IP host, or a DNS SAN
+// for a hostname. This mirrors how the coordinator builds the certificate's SANs
+// from the listen address.
+func certCoversListenAddr(certPEM, listenAddr string) (bool, error) {
```
phinze, line 387: Expiry cliff: once a cert expires, it can't authenticate its own refresh, and it's manual re-join time. Since we're already parsing the cert here, also refreshing when near `NotAfter` (while it can still vouch for itself) is a cheap win ahead of periodic renewal. Filed as MIR-1227.

servers/runner/registration_test.go
```diff
@@ -468,3 +503,253 @@ func TestRemoveRunnerByRunnerId(t *testing.T) {
...
+		// Its certificate is still cryptographically valid, but it is no longer
+		// registered, so refresh must be rejected.
+		_, err = env.server.reissueRunnerCertificate(ctx, peer, "10.0.0.2:8444")
+		if err == nil {
+			t.Fatal("expected error refreshing a removed runner's certificate")
+		}
+	})
+
+	t.Run("ambiguous runner identity is rejected", func(t *testing.T) {
+		// Two runners whose IDs share the same 8-char prefix produce the same
```
phinze: Fun irony: this fixture is built on exactly the non-uniqueness MIR-1225 removes. Whoever fixes 1225 will need to seed the second node via the entity store instead.

servers/runner/registration_test.go
```diff
@@ -28,6 +31,7 @@ func newTestServer(t *testing.T) (*testEnv, func()) {
 	ca, err := caauth.New(caauth.Options{
 		CommonName:   "test-ca",
 		Organization: "test",
+		ValidFor:     24 * time.Hour,
```
phinze, line 34: Nice catch. The footgun that hid this is `caauth.New` happily minting a born-expired CA on zero `ValidFor` — a validation error there would've caught it years ago.

phinze (approved):
Claude and I went through this end to end. The core trick is great: the stale cert is broken for serving but still fine for *authenticating*, so the runner uses its broken credential to ask for a fixed one. Auth chain is solid, fails closed everywhere, and the comments explain the why at every weird spot.

Worth noting the division of labor with the infra side: pinning addresses over there retires the unrecoverable direction (coordinator drift, where nothing can be negotiated with a coordinator you can't find), while this PR self-heals the recoverable one — and repairs the runners that already drifted, which pinning alone freezes in the broken state.

One thing to keep in mind: MIR-1225 means CN uniqueness isn't enforced at Join, and this auth chain leans entirely on CN. Details inline. Filed MIR-1227 for the renewal follow-up. Nothing blocking. 👍

PR #852, opened 2026-06-08
Default workload identity JWKS to RS256, support multiple live keys

evanphx:
## Summary

The cluster's workload-identity OIDC issuer (`pkg/workloadidentity`) signed sandbox tokens with **EdDSA**. EdDSA is exotic and rejected by many OIDC federation verifiers (e.g. AWS STS), so this defaults to **RS256**, which is universally supported — and extends the issuer to hold **multiple live keys** for rotation and mixed key types.

### Phase 1 — RS256 default
- New `pkg/workloadidentity/key.go`: an algorithm-agnostic `signingKey` (RSA or Ed25519 via `crypto.Signer`), **decoupled from `cloudauth`** (which must stay Ed25519 — it backs cluster auth via `Sign`/`VerifySignature`).
- New clusters generate **RSA-2048 / RS256**.
- Clusters with a legacy EdDSA key **auto-migrate**: the EdDSA key is demoted to the `.prev` slot and kept **advertised in JWKS** (so in-flight ≤24h tokens still verify), while new tokens sign with a fresh RS256 key. Ed25519 `kid` computation is preserved so already-issued tokens still match.

### Phase 2 — multiple live keys
- The issuer now holds a `primary` signer plus `keys []*signingKey` — the primary plus any keys dropped into **`server/workload-identity.d/`**. All live keys are **published in JWKS** (deduped by `kid`) alongside verify-only `advertised` keys; the discovery doc lists every algorithm present.
- Supports **key rotation** (publish a new key before cutting over) and **multiple key types** coexisting.
- **Signing remains primary-only** for now — per-request key/type selection (and the runner→coordinator RPC) is intentionally deferred, so `TokenOptions` is unchanged.

## Scope / not changed
- `pkg/cloudauth` (cluster auth — stays Ed25519), `pkg/auth/jwt.go` (Miren Cloud's tokens — separate repo), `pkg/oidcauth/validator.go` (already accepts RS256).

## Notes
- One-time auto-rotation on upgrade for EdDSA clusters; the issuer URL (external trust anchor) is unchanged, so federation configs need no update. The migration renames `key → key.prev`, clobbering any pre-existing `.prev` (an older, almost-certainly-expired EdDSA key).

## Testing
- `go test ./pkg/workloadidentity/` — 19 passed (incl. legacy-migration + directory-key tests)
- `go test ./controllers/sandbox/ -run Token` — 8 passed (token_server end-to-end with RS256)
- `make lint` — 0 issues

### DIFF
--- a/controllers/sandbox/token_server_test.go
+++ b/controllers/sandbox/token_server_test.go
@@ -71,6 +71,7 @@ func TestTokenServer_DefaultToken(t *testing.T) {
 	require.NotEmpty(t, resp.Value)
 
 	token, err := jwt.ParseWithClaims(resp.Value, &workloadidentity.WorkloadClaims{}, func(tok *jwt.Token) (interface{}, error) {
+		assert.Equal(t, "RS256", tok.Method.Alg())
 		return c.WorkloadIssuer.(*workloadidentity.Issuer).PublicKey(), nil
 	})
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


maryelizbeth (approved):
co-signing w/ biscuit!

PR #854, opened 2026-06-10
sandbox: re-register workload-identity token secrets on controller restart

evanphx:
## Summary

Fixes **MIR-1235**: a long-running sandbox authenticating via **workload identity** permanently lost token-server connectivity after the host's sandbox controller / token server restarted. The per-sandbox token server returned `403 {"error":"invalid token"}` for the sandbox's `MIREN_IDENTITY_TOKEN_SECRET` and could **not self-heal** — it required a restart/redeploy. Observed on `linearagent` sandbox `AN2`, looping every 30s with `chitchat connection lost err="token: fetch identity token: token server: 403 Forbidden"`.

## Root cause

The token server validates each request against `tokenSecretRegistry`, an **in-memory** map created empty on controller boot and only ever populated on the sandbox **start** path (where the random secret is generated, registered, and baked into the container env). Boot reconcile re-registered survivors with the token **refresher** but **never with the secret registry**, so on a controller/token-server restart every live sandbox's entry was lost while the running process still held the original env secret — 403 forever until restart.

## Changes (server-side only, no consumer change)

- **Persist + recover the secret.** Each sandbox's token-request secret is now persisted host-side (`0600`, under the sandbox dir, not bind-mounted) on start, and reloaded + re-registered for surviving sandboxes during boot reconciliation. The still-running process self-heals on the next controller restart because the *exact same* secret it holds in its env is re-registered.
- **Key the registry by sandbox identity, not raw IP.** The caller's identity is resolved from the source IP via the authoritative netdb lookup, then the secret is checked against that identity — closing IP-reuse edge cases (a recycled pod IP can't match a stale secret). Also drops the IP plumbing from both register call sites.
- **Cleanup on Delete.** Best-effort removal of the persisted secret so it doesn't linger if `StopSandbox` errors before its dir cleanup.
- **Stop refresh-loop spew.** Boot reconcile now only re-registers a survivor for token refresh when it actually has an `identity-token` file, so the refresh loop no longer logs `failed to write refreshed workload identity token ... no such file or directory` for sandboxes that predate workload identity.

## Testing

- New unit tests in `token_server_test.go`:
  - `TestTokenServer_RecoversSecretAfterRestart` — reproduces the 403→200 recovery after a simulated registry loss.
  - `TestTokenSecretRegistry_KeyedBySandboxIdentity` — a secret can't authenticate a different sandbox.
  - `TestWriteLoadTokenSecret_RoundTrip` / `TestLoadTokenSecret_Missing` — persistence helpers.
- `go test ./controllers/sandbox/ -run 'TestTokenServer|TestTokenSecretRegistry|TestWriteLoadTokenSecret|TestLoadTokenSecret'` — 12 pass.
- `make lint` — 0 issues.

### Manual repro (matches the Linear report)
1. Start a workload-identity sandbox; confirm it connects.
2. Restart the sandbox controller WITHOUT restarting the sandbox.
3. Force a token fetch → now returns `200` instead of `403`; chitchat stays connected.

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -399,13 +399,39 @@ func (c *SandboxController) reconcileSandboxesOnBoot(ctx context.Context) error 
 				}
 			}
 
-			// Re-register with token refresher so tokens keep getting renewed
+			// Re-register with token refresher so tokens keep getting renewed. Only
+			// sandboxes that actually use workload identity have an identity-token file;
+			// skip the rest so the refresh loop doesn't spew write errors for sandboxes
+			// that predate workload identity.
 			if c.tokenRefresher != nil {
-				appName := c.resolveAppName(ctx, &sb)
 				tokenPath := c.sandboxPath(&sb, "identity-token")
-				c.tokenRefresher.register(sb.ID.String(), tokenPath, appName)
-				c.Log.Debug("re-registered sandbox for token refresh",
-					"sandbox_id", sb.ID, "app", appName)
+				if _, err := os.Stat(tokenPath); err == nil {
+					appName := c.resolveAppName(ctx, &sb)
+					c.tokenRefresher.register(sb.ID.String(), tokenPath, appName)
+					c.Log.Debug("re-registered sandbox for token refresh",
+						"sandbox_id", sb.ID, "app", appName)
+				}
+			}
+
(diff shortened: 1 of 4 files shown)


evanphx:
Found this just before release, merging solo (with 🐰 and 🍪 help) to include in 0.10.0

PR #856, opened 2026-06-12
sandbox: prevent duplicate bridge IP assignment (MIR-1238)

evanphx:
## Problem

On garden runner-1 the sandbox IPAM handed `10.8.64.2` to **two live sandboxes** (reviewagent + sportsagent) on the `rt0` bridge, causing ARP conflicts, `martian source` kernel drops, and constant flapping of the apps' long-lived chitchat WebSocket connections (`connection reset by peer` / `i/o timeout`) — surfacing to users as intermittent "empty reply (no responder)".

## Root cause

netdb cannot itself hold two reserved rows for one IP (`ip` is a `PRIMARY KEY`). The bug is a **divergence between netdb's lease bookkeeping and the addresses actually live on the bridge**: a sandbox's `.2` lease was released in netdb while its container kept running on `.2`, and ~19h later the now-"free" `.2` was handed to a second sandbox.

The **container watchdog** caused the premature release. When a labeled container was missing from the node-scoped sandbox list (built from an index that can lag), it:
1. released the IP **before** removing the container, and
2. released it **even when removal failed**, and
3. never re-checked whether the sandbox actually still existed.

A transient/stale index → a live sandbox looks orphaned → its IP returns to the pool → reused.

## Fix (defense in depth)

- **Watchdog** (`controllers/sandbox/watchdog.go`): release an IP only after the container is *confirmed removed* (leave it reserved on failure); before treating a container as orphaned, **re-fetch its sandbox entity directly by ID** (a linearizable key lookup, not an index read) so a sandbox the index transiently omitted is never killed or has its IP released.
- **Allocation-time guard** (`network/config.go`, `sandbox.go`): `AllocateOnBridge` verifies a freshly reserved address isn't already live on the bridge (from running-container labels — no netns entry, no ARP); on conflict it keeps the address quarantined and retries, so a divergence can never surface as a duplicate.
- **IPReconciler** (`controllers/sandbox/ipreconciler.go`): periodically (and on boot) re-reserves any address live on the bridge but free in netdb (repairing the exact divergence), and conservatively reaps reservations with no live owner after several consecutive misses.

## Tests

- netdb: concurrent `Reserve` never duplicates; `ReservedAddrs`.
- network: `reserveConflictFree` skips a live-but-free address and quarantines it.
- reconciler: re-reserves a lost live address; reaps only after the miss threshold; never reaps a live address.
- **iso regression test**: a RUNNING sandbox missing from the node index is kept, not orphaned.

`make lint`: 0 issues. iso: `controllers/sandbox` (120), `watchdogtest` (14), `pkg/netdb` all pass.

## Notes

- `sandbox.go` is a frozen file (saga-transition guard); the changes are in shared code (`AllocateNetwork`, `Init` lifecycle, both reached by the saga path via `inner`), so the frozen hash was updated per the test's documented procedure.
- The live garden duplicate was mitigated separately by recycling the two colliding sandboxes onto distinct IPs.

### DIFF
--- a/controllers/sandbox/ipreconciler.go
+++ b/controllers/sandbox/ipreconciler.go
@@ -1,0 +1,161 @@
+package sandbox
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/netip"
+	"time"
+
+	"miren.dev/runtime/pkg/netdb"
+)
+
+// IPReconciler keeps the netdb IP lease bookkeeping in agreement with the
+// addresses actually live on the bridge. The two can diverge — a release path
+// that ran while a container kept running, a lease lost across a restart — and a
+// divergence where netdb believes a live address is free leads to that address
+// being handed to a second sandbox, a duplicate assignment (MIR-1238).
+//
+// Each cycle it:
+//   - re-reserves any address that is live on the bridge but not reserved in
+//     netdb (the repair direction — always safe, only ever adds a reservation
+//     for an address already in use); and
+//   - releases any address reserved in netdb with no live owner, but only after
+//     it has been absent for several consecutive cycles (the reap direction —
+//     conservative, to avoid racing a sandbox that is mid-create or mid-teardown).
+type IPReconciler struct {
+	Log    *slog.Logger
+	Subnet *netdb.Subnet
+
+	// LiveIPs returns the set of addresses currently assigned to running sandbox
+	// containers, determined independently of netdb.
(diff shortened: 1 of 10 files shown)


phinze (approved):
First review back from a week OOO, so I really went to town: traced all three layers end to end, talked myself through a few theories of how the accounting drifted, collected a respectable pile of observations, went hunting for gaps... and every last one turned out already handled. Distilled all of that effort back down to its true and final form: LGTM.

--p+🤖

PR #861, opened 2026-06-22
Release disk leases promptly when a sandbox dies

phinze:
Debugging `jobsv2` on club, a small port misconfiguration turned into a roughly hour-long wait and a pile of DEAD sandboxes. The app has a single-writer persistent disk, and the situation had an unfortunate way of feeding itself: a sandbox would boot far enough to bind the disk lease, not pass its 15s port health check, and get marked DEAD with the lease still `status.bound`. The next replacement would find that lease still held, fail volume config with `disk … has an active lease`, and land in the same place. The lease only freed once it aged out (~66 minutes observed), at which point the replacement would grab it and the cycle would repeat.

The two ways a sandbox shuts down had drifted apart a little. The graceful `StopSandbox` path releases disk leases on its way out, but the imperative `createSandbox` boot-failure path hadn't picked up that step yet, so a lease could outlive the sandbox that held it. The fix is to release leases in that boot-failure defer too. It's correctly sequenced: `createSandbox`'s defers unwind LIFO, so the inner defer tears the container down before this outer one frees the lease.

That handles the reported case right away, but I didn't want to lean on a single code path remembering to clean up. There was already an orphan-lease sweep that releases leases whose owning sandbox is DEAD or gone, it just only ran once at controller startup, so it never got a chance during a live incident. This promotes it to the existing 5-minute periodic, so any path that leaves a lease behind (a SIGKILL, future code) heals within a bounded window instead of an hour. One wrinkle worth noting: a recurring sweep that reclaims "DEAD or missing sandbox" leases could in principle race a sandbox that's still booting, or one whose own teardown is mid-flight. The storage layer already guards against the worst outcome (a second loop attach gets adopted rather than duplicated), but to be safe the recurring sweep skips any lease younger than two minutes and lets the in-band paths go first. The boot-time sweep keeps a zero grace since it only ever sees leftovers from a crashed process.

Verified end to end in dev with a deliberately-wrong-port app on a single-writer disk: on the old binary the lease stayed `bound` and replacements piled up with `active lease` errors; on the fixed binary the lease releases about 2ms after the sandbox is marked DEAD and replacements boot cleanly, no wedge.

The app still crash-loops on the underlying port misconfig, but that's a separate concern (MIR-1246). While I was in here I also filed MIR-1249 for a related unmount-retry gap I noticed during the investigation.

Closes MIR-1245

### DIFF
--- a/components/runner/runner.go
+++ b/components/runner/runner.go
@@ -993,8 +993,15 @@ func (r *Runner) SetupControllers(
 		workers,
 	)
 
-	// Set up periodic cleanup of old released leases (every 5 minutes)
+	// Set up periodic lease maintenance (every 5 minutes): sweep orphan leases
+	// stranded by sandboxes that died without releasing (SIGKILL, boot failure),
+	// then clean up old released leases. The orphan sweep also runs at Init, but
+	// the periodic tick bounds the worst-case wedge to one interval for sandboxes
+	// that die while the controller is already running.
 	diskLeaseRC.SetPeriodic(5*time.Minute, func(ctx context.Context) error {
+		if err := diskLeaseController.ReconcileOrphanLeases(ctx, disk.OrphanSweepGracePeriod); err != nil {
+			log.Warn("periodic orphan lease sweep failed", "error", err)
+		}
 		return diskLeaseController.CleanupOldReleasedLeases(ctx)
 	})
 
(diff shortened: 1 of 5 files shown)


evanphx (approved):
Man oh man, it really is screaming for a saga.

PR #868, opened 2026-06-26
Add absolute time-window filtering to `miren logs`

phinze:
`miren logs` could only ever look backward from now. The one time control was `--last <duration>`, which set the start to `now - duration` and always pinned the end to the present. That's fine for "what just happened," but useless when you're chasing an incident with a known window and don't want to scroll through everything since 2pm to see what happened between 2:00 and 2:30.

So this adds a `--since`/`--until` pair to all four log subcommands (`app`, `sandbox`, `build`, `system`). Rather than invent our own vocabulary, we leaned into what people already have in their fingers from docker, journalctl, and git: `--since`/`--until`, each overloaded to take either a timestamp or a duration. You can write a full RFC3339 timestamp for a precise boundary, a friendlier naive form like `"2026-06-25 14:00"` (interpreted in your local timezone), or just a duration like `2h` that's read as "that long ago." `--since 2h --until 30m` gives you the window from two hours ago to thirty minutes ago. The old `--last` flag still works, it's just shorthand for `--since` with a duration now, so nobody has to relearn anything.

Most of the backend was already in shape for this. The query layer always called into VictoriaLogs with both a start and an end, but the end was hardcoded to `time.Now()` and there was no way to pass anything else through. The real work was threading an upper bound the rest of the way: `streamLogChunks` gained a `to` parameter, `logReadOpts` grew an `Until` field, and `executeStreamQuery` now honors it. `--since` was almost free since it just produces a `from` timestamp that every protocol version already accepts. A couple of guardrails fall out naturally too: `--until` with `--follow` is rejected (a live tail has no end), `--since` alongside `--last` is rejected (they'd fight over the start), and an inverted window (`--until` before `--since`) is caught up front.

While wiring this up, the obvious gate had a hole: `--until` would be allowed whenever the server had `streamLogChunks`, but that method predates the `to` parameter, so a server in that in-between range would silently ignore the bound. Since the whole fleet released to date is in exactly that range, that's not a rare edge. So the RPC discovery layer now reports parameter names alongside method names, and `--until` checks for real `to` support before sending it. Older servers get a clean "newer server required" error instead of a silently-dropped bound, and the same fix covers the latent `--service`/`--build` case for free. That capability work is the first commit; the logs feature that consumes it is the second.

Closes MIR-1226

### DIFF
--- a/api/admin/admin_v1alpha/rpc.gen.go
+++ b/api/admin/admin_v1alpha/rpc.gen.go
@@ -593,6 +593,7 @@ func AdaptAdmin(t Admin) *rpc.Interface {
 			InterfaceName: "Admin",
 			Index:         0,
 			Public:        false,
+			Params:        []string{"app", "method", "params"},
 			Handler: func(ctx context.Context, call rpc.Call) error {
 				return t.Invoke(ctx, &AdminInvoke{Call: call})
 			},
@@ -602,6 +603,7 @@ func AdaptAdmin(t Admin) *rpc.Interface {
 			InterfaceName: "Admin",
 			Index:         1,
 			Public:        false,
+			Params:        []string{"app"},
 			Handler: func(ctx context.Context, call rpc.Call) error {
 				return t.ListMethods(ctx, &AdminListMethods{Call: call})
 			},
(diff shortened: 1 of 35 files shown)


phinze:
Addressed in commit `ce13c6a`: RPC discovery now reports per-method parameter names, and `--until` gates on `HasMethodParam("streamLogChunks", "to")` in both `dispatchLogs` and `LogsSystem`, so a server too old to know `to` returns a clean "newer server required" error instead of silently dropping the bound. @coderabbitai resolve

evanphx (approved):
🌳 🕐

PR #872, opened 2026-06-29
Page entity-store scans in startup migrations instead of one unbounded read

jcasimir:
## What

`MigrateEntityStore` and `MigrateShortIds` each read the entire entity keyspace with a single `client.Get(ctx, prefix, clientv3.WithPrefix())`. This adds a small `listEntitiesPaged` helper that reads the prefix in bounded pages (`WithLimit` + a key cursor, ascending key order) and uses it in both migrations. Per-entity processing is unchanged — only the read is paged.

## Why

A single unbounded prefix read loads the whole keyspace (keys **and** values) in one RPC. On a large or not-recently-compacted store that request can stall, and because these migrations run early in coordinator startup, a stall blocks the edge listener from coming up. We hit exactly this on a self-hosted node — the migration's full read hung for hours after a restart, leaving every app unreachable (more detail in the companion PR #871). Paging keeps each request bounded and predictable regardless of store size.

As with #871, we may be missing context here, so this is offered as "what worked for us" rather than a prescription.

## Notes

- `go build`, `go vet`, and `gofmt` are clean. The `pkg/entity` integration tests need the dev etcd environment (`make dev`) and weren't run locally.
- The unique-key read in `MigrateShortIds` is keys-only (bounded payload), so it's left as-is — glad to page it too if you'd prefer consistency.
- Companion PR #871 adds a startup-phase timeout as defense-in-depth. The two are independent and complementary; either can merge on its own. If you'd rather take just one, this is the root-cause fix and #871 is the belt-and-suspenders.

### DIFF
--- a/pkg/entity/migrate.go
+++ b/pkg/entity/migrate.go
@@ -129,16 +129,16 @@ func MigrateEntityStore(ctx context.Context, log *slog.Logger, client *clientv3.
 	log.Info("starting entity migration", "prefix", opts.Prefix, "dry_run", opts.DryRun)
 
 	// List all entities
-	resp, err := client.Get(ctx, opts.Prefix, clientv3.WithPrefix())
+	kvs, err := listEntitiesPaged(ctx, client, opts.Prefix)
 	if err != nil {
 		return 0, 0, fmt.Errorf("failed to list entities: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


phinze (approved):
Same as #871 - taking this as-is w/ a few follow ups I'll pick up in MIR-1272. Thanks again!

PR #875, opened 2026-06-30
Cap SandboxLogs per-line buffer to bound control-process memory

jcasimir:
## Problem

A sandbox that emits a large, newline-poor blob to stdout grows
`(*SandboxLogs).buf` without bound. A log line is only flushed on `\n`, so
containerd streaming a multi-GB payload across many `Write` calls accumulates
the whole thing in one `bytes.Buffer`, which is then `gjson`-parsed and
JSON-marshaled into a single log entry.

On our self-hosted cluster this manifested as a recurring control-process memory
balloon (~every 4h) that tripped the host OOM and took down every app in the
shared `miren.service` cgroup. A live heap profile of the control process during
a balloon attributed **97.8% of a 68 GB heap** to this path:

```
containerd cio.copyIO  →  (*SandboxLogs).Write
  bytes.growSlice            20.6 GB   (one unbounded bytes.Buffer)
  encoding/json.Marshal      18.4 GB   (marshaling the giant line)
HeapInuse 53.7 GB, 54M objects
```

The trigger was an app's 4-hourly job occasionally logging a huge raw payload.

## Fix

Cap the partial-line buffer at 1 MiB (`maxLineBytes`): bytes past the cap are
dropped until the next newline flushes the truncated line, bounding memory to
roughly `maxLineBytes` per concurrent sandbox regardless of what an app emits.

Also fixes a latent bug where `buf` was never `Reset` after completing a
buffered line, so a partial line could be re-appended to itself.

Existing `TestSandboxLogs` cases (line handling, partial-line buffering, USER/
ERROR prefixes) are unchanged; a new case asserts a ~12 MiB newline-poor stream
is capped to `maxLineBytes` and flushes truncated on the next newline.

### DIFF
--- a/controllers/sandbox/log.go
+++ b/controllers/sandbox/log.go
@@ -39,27 +39,58 @@ func NewSandboxLogs(
 	}
 }
 
+// maxLineBytes caps a single newline-terminated log line. A sandbox that emits
+// a huge, newline-poor blob to stdout (e.g. a verbose scraper dumping a raw
+// payload) would otherwise grow s.buf — and the control process's Go heap —
+// without bound, since a line is only flushed on '\n'. containerd streams in
+// small chunks, so the unbounded growth happens by accumulation across many
+// Write calls into s.buf; capping the buffer bounds memory to roughly
+// maxLineBytes per concurrent sandbox regardless of what an app spews. Bytes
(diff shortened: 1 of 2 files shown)


evanphx:
@jcasimir great find and sorry about the outage!

evanphx (approved):

PR #877, opened 2026-07-01
Prune old app versions and reclaim their image blobs

phinze:
App version retention is something we always figured we'd need eventually, and this is us building it. Every deploy mints a new AppVersion, and until now they lived forever. That unbounded growth had two costs: it bloated the entity store (and made the ephemeral GC's every-5-minute all-versions scan grow with the cluster's whole deploy history), and it kept disk pinned. Each version references an Artifact, and Artifacts reference registry blobs, so as long as old versions stuck around the image and blob GC could never reclaim old images.

So this adds a periodic version retention GC, modeled on the existing artifact GC and wired in beside it. Per app it keeps the most recent `retention_count` versions or anything newer than `retention_period`, whichever rule keeps it, always keeps the active version, and skips anything a running or pending sandbox still references. Everything else is hard-deleted along with its 1:1 ConfigVersion.

The interesting part is the disk half. Artifacts are shared many-versions-to-one (they're deduped by manifest digest), so you can't just archive a deleted version's artifact, since another live version might share the same image and would start 404ing. Instead, artifact archiving becomes reference-driven: the artifact GC's old blind age/count retention is replaced with "archive any active artifact that zero surviving versions reference," computed across all versions. That gives the existing image and blob GC chain something to actually reclaim, and a shared artifact naturally survives until its last referencing version is pruned.

While in here, the sandbox-pool cleanup that used to live in the ephemeral delete path got pulled into a shared helper. It turned out the version GC doesn't need it at all: the pool manager already reaps drained, unreferenced pools on its own, and the deploy launcher de-refs superseded pools. So there's now a plain `appversion.Delete` (a version plus its ConfigVersion) that the version GC uses, and a `DeleteWithPools` wrapper that only the ephemeral path calls, where prompt teardown actually matters.

Retention is configurable through `app_version.retention_count` and `app_version.retention_period` (defaults 10 and 30d), following the duration-string style of the existing victorialogs/buildkit knobs, and documented in the server config reference.

Closes MIR-329

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -635,30 +635,41 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		}
 	}
 
+	// A malformed retention period parses to 0, which the coordinator treats as
+	// "use the default" rather than "retain nothing". Warn so the operator knows
+	// their configured value was ignored.
+	appVersionRetentionPeriod, err := units.ParseDuration(cfg.AppVersion.GetRetentionPeriod())
+	if err != nil {
+		ctx.Log.Warn("invalid app_version.retention_period, falling back to default",
+			"value", cfg.AppVersion.GetRetentionPeriod(), "error", err)
+	}
+
 	// Build coordinator config
 	coordConfig := coordinate.CoordinatorConfig{
-		Address:                srvaddr,
-		EtcdEndpoints:          cfg.Etcd.Endpoints,
-		Prefix:                 cfg.Etcd.GetPrefix(),
-		NetworkBackend:         cfg.Server.GetNetworkBackend(),
-		DataPath:               cfg.Server.GetDataPath(),
-		AdditionalNames:        cfg.TLS.AdditionalNames,
-		IPs:                    ipSet,
-		AcmeEmail:              cfg.TLS.GetAcmeEmail(),
-		AcmeDNSProvider:        cfg.TLS.GetAcmeDNSProvider(),
-		Resolver:               res,
-		TempDir:                os.TempDir(),
-		CloudAuth:              cloudAuthConfig,
-		Mem:                    mem,
-		Cpu:                    cpu,
-		HTTP:                   httpMetrics,
-		Logs:                   logs,
-		LogWriter:              logWriter,
-		BuildKit:               buildkitComponent,
-		HTTPRequestTimeout:     cfg.Server.HTTPRequestTimeoutDuration(),
-		VictoriametricsAddress: ctx.ServerState.VictoriametricsAddress,
-		VictorialogsAddress:    ctx.ServerState.VictorialogsAddress,
-		WorkloadIssuer:         workloadIssuer,
+		Address:                   srvaddr,
+		EtcdEndpoints:             cfg.Etcd.Endpoints,
+		Prefix:                    cfg.Etcd.GetPrefix(),
+		NetworkBackend:            cfg.Server.GetNetworkBackend(),
+		DataPath:                  cfg.Server.GetDataPath(),
+		AdditionalNames:           cfg.TLS.AdditionalNames,
+		IPs:                       ipSet,
+		AcmeEmail:                 cfg.TLS.GetAcmeEmail(),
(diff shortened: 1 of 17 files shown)


controllers/artifact/gc.go
```diff
@@ -134,94 +133,58 @@ func (c *GCController) RunGC(ctx context.Context) (*GCResult, error) {
...
-			createdAt: time.UnixMilli(e.CreatedAt()),
+		if referenced[art.ID] {
+			result.RetainedArtifacts++
+			continue
 		}
 
-		if art.App == "" {
-			// Orphaned artifact (no app reference)
-			orphanedArtifacts = append(orphanedArtifacts, info)
+		if err := c.archiveArtifact(gcCtx, art, e.Revision()); err != nil {
```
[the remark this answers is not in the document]
phinze: Good catch, this is a real TOCTOU. We are leaving it for now: `app_version.artifact` is not indexed, so a correct pre-archive recheck would mean re-scanning all versions per artifact, and the race is narrow with bounded fallout. It needs an artifact whose referencing version already aged out of retention (so beyond both the count and the period), an identical-digest rebuild of that same app landing inside the hourly window, and then a cold pull before the next redeploy. Running workloads are unaffected (containerd already has the image; image GC keeps running-sandbox images), and blob GC has a 1h recently-modified grace, so the worst case is a deferred cold-pull miss that a redeploy heals. If we ever see it in practice, the clean fix is an `artifact`->versions index plus a just-in-time recheck, which we would rather add deliberately than speculatively.

phinze:
On the `AppVersionConfig.Validate()` nitpick: that empty stub and the "port conflicts" comment are boilerplate the config generator emits for every type, not specific to this change, and the malformed-`retention_period` case already fails safe (parse error falls back to the default). Filed MIR-1284 to fix the template (drop the stale comment, optionally emit real validation) rather than hand-patch a generated file here.

controllers/artifact/gc.go
```diff
@@ -134,94 +133,58 @@ func (c *GCController) RunGC(ctx context.Context) (*GCResult, error) {
...
-					result.FailedArtifacts[info.artifact.ID] = err
-				} else {
-					result.ArchivedArtifacts = append(result.ArchivedArtifacts, info.artifact.ID)
-				}
-			}
-		}
+// referencedArtifacts returns the set of artifact IDs referenced by any
+// AppVersion.
+func (c *GCController) referencedArtifacts(ctx context.Context) (map[entity.Id]bool, error) {
+	resp, err := c.EAC.List(ctx, entity.Ref(entity.EntityKind, core_v1alpha.KindAppVersion))
```
evanphx, line 174: 💡 I'm thinking more about the giant-list-vs-pagination issue. We don't need to fix this here today, but it's something we should do a whole code base pass on.

cli/commands/server.go
```diff
@@ -635,30 +635,36 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		}
 	}
 
+	// A malformed retention period parses to 0, which the coordinator treats as
+	// "use the default" rather than "retain nothing".
+	appVersionRetentionPeriod, _ := units.ParseDuration(cfg.AppVersion.GetRetentionPeriod())
```
evanphx: 🧹  it's worth at least logging if we can't parse this, so that the operator can know that they need to fix their value.

evanphx (approved):
One minor nit about logging when we fallback to the default retention when we can't parse the requested retention.

cli/commands/server.go
phinze: Done, added in the latest push: we now `Warn` with the offending value and fall back to the default, so an operator sees exactly what got ignored instead of silently getting default retention.

controllers/artifact/gc.go
phinze: Agreed, and good call not to scope it here. For this particular list the retention GC now bounds the AppVersion set, so it stops growing with deploy history, but the full-kind-scan-into-memory pattern is all over the GC controllers and the store `List` has no pagination at the API layer. Filed MIR-1302 to track the codebase-wide pass.

PR #878, opened 2026-07-01
Cut no-op reconcile chatter from miren's journal

phinze:
Digging through `journalctl -u miren` on garden, the signal-to-noise ratio was rough: a clean ~1.9h steady-state window held about 80k lines (~700 lines/min, call it 1M lines/day), and roughly 80% of that was reconcile loops heartbeating on every tick even when nothing changed. That's the same theme that bit us in MIR-1279, where garden sat wedged for 14h before a human noticed. A journal that noisy is exactly where a real signal goes to hide.

Three loops drove most of the volume. The sandboxpool `sandbox counts` debug line was the single highest-volume line in the whole journal, firing per-pool every tick even when actual == ready == desired. The deployment launcher narrated its entire reconcile story at INFO on every pass (`reconciling app` → `reconciling app version` → `found pools to check` and onward), seven lines per app per minute regardless of whether anything happened. And the sandbox controller logged `considering sandbox create or update` at the top of every reconcile, plus a handful of debug siblings.

The tempting fix is to just demote everything to debug, but garden already runs at debug, so that buys nothing (the highest-volume line was already debug). So instead of demoting, we took a critical eye to whether these lines earn their place at all and cut the pure heartbeats outright. What's left is gated on actual work: the pool status line already only logs on a count change, so we folded `desired` into it and dropped the redundant counts line. Pool reuse only logs when reuse actually mutates the pool. The deployment loop now emits a single summary line only when it did real work, instead of seven lines of narration. The auto-mount-without-disk-config warning, which was firing for six apps on every tick forever, now warns once per app.

The sandbox `considering` line lives in both the classic and the saga controller, so we removed it from both to keep things quiet regardless of the `labs.Sagas()` flag. That meant touching the frozen legacy `sandbox.go`, which is guarded by a hash test during the saga migration. Per the guard's own instructions I audited the saga path for the parallel change (it was just the twin log line) and bumped the recorded hash.

The two smaller items the issue mentions, the tempo trace-export timeouts and the inbound TLS-handshake scanner noise, are left for separate passes. Those are a different kind of problem than our own loops talking to themselves, and the TLS one in particular wants its own think about `http.Server.ErrorLog` filtering.

Closes MIR-1282

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -47,6 +47,11 @@ type Launcher struct {
 	PoolReadyTimeout time.Duration
 
 	appMu sync.Map // per-app mutexes: app ID -> *sync.Mutex
+
+	// warnedNoDisk tracks apps we've already warned about auto-mounting local
+	// storage without a disk config, so the warning fires once per app instead
+	// of on every reconcile tick. Keyed by app ID.
+	warnedNoDisk sync.Map
 }
 
 // PoolWithEntity wraps a SandboxPool with its entity, allowing updates without re-fetching
@@ -131,7 +136,6 @@ func (l *Launcher) Reconcile(ctx context.Context, app *core_v1alpha.App, meta *e
 
 	span.SetAttributes(attribute.String("miren.app.active_version", current.ActiveVersion.String()))
 
-	l.Log.Info("reconciling app", "app", current.ID, "version", current.ActiveVersion)
 	ready, err := l.addonsReady(ctx, current.ID)
 	if err != nil {
 		l.Log.Error("failed to check addon readiness", "app", current.ID, "error", err)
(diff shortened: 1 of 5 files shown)


phinze:
Thanks for the careful read. Two fixes in the latest push: renamed the summary field `pools_created` → `pools_started` since that count also includes drained pools revived for deploy verification (booting them is real work, so the honest fix is the accurate label), and the `updated pool status` log now caps `desired` at `MaxPoolSize` to match what reconciliation enforces.

On the `warnedNoDisk` unbounded-growth note: leaving it as-is for now. The `appMu` map right above it has the same never-evicted lifetime keyed by every app ID we reconcile, so this stays consistent with the existing pattern, and it's a small transitional set of apps. Happy to add lifecycle eviction for both maps in a follow-up if churn ever becomes a concern.

evanphx (approved):
I have any idea for a slog Handler that captures and holds related logs to see if the operation errors out, and if so, it emits all of the logs. We can look at a mechanism like that as well.

phinze:
> I have any idea for a slog Handler that captures and holds related logs to see if the operation errors out, and if so, it emits all of the logs. We can look at a mechanism like that as well.

Ooo fun idea. Would have to keep an eye on allocs but I like the shape.

PR #879, opened 2026-07-01
Enforce BuildKit's GC byte cap with age-less policy tiers

phinze:
BuildKit's build cache on miren-garden grew to 27GB against a configured 10GB cap and became a primary contributor to the MIR-1279 disk-pressure outage. The frustrating part: when we pruned it by hand it dropped right back under the limit, so the cache was always collectable. Automatic GC just never did it.

Digging in on the box turned up two compounding faults in the `buildkitd.toml` we generate. First, no ceiling field was ever actually set. We configured `keepBytes`, which BuildKit v0.19 deprecates and quietly migrates to `reservedSpace`, and that's a floor ("never prune below this"), not a cap. The real ceiling field, `maxUsedSpace`, was left at zero. `buildctl debug workers` on garden confirmed it: `{reservedSpace:10GB, maxUsedSpace:0, keepDuration:7d}`.

Second, and this is the subtle one, `keepDuration` protects far more than it looks. It shields any record used within the window, but because a parent layer can't be deleted while a child still references it, it transitively pins the entire ancestry of anything touched recently. On a builder that constantly rebuilds the same family of base images, that's nearly the whole cache. We measured it on garden: 73 of 73 records older than 7 days were pinned by a younger descendant, zero were freely deletable. So GC ran faithfully after every build and only ever reclaimed the thin tail that had just aged out, while the cap never bound and the pile climbed.

We proved the mechanism live with a controlled prune: same 2GB target, `--keep-duration 168h` freed 0 bytes while dropping the age gate freed 7.74GB. And we confirmed this isn't something a version bump fixes: the prune logic is byte-identical from v0.19 (what we run) through v0.31.

The fix replaces the single age-gated policy with a graduated set that mirrors BuildKit's own default shape: a 48h/512MB tier for the cheapest-to-rebuild sources, a soft `keepDuration` tier, and then two age-less `maxUsedSpace` tiers that enforce the ceiling regardless of age (the age-less pass drops fresh leaves so their pinned ancestors can finally go). Source filters use the array (OR) form, since a single comma-joined string ANDs the types and matches nothing.

Follow-ups filed while we were in here: MIR-1286 (the `miren-oci` registry has no retention and has ballooned to 236GB) and MIR-1287 (get BuildKit off the release candidate and align the daemon/client/test versions).

Closes MIR-1280

### DIFF
--- a/components/buildkit/buildkit.go
+++ b/components/buildkit/buildkit.go
@@ -269,6 +269,38 @@ func (c *Component) Client(ctx context.Context) (*buildkitclient.Client, error) 
 	)
 }
 
+// generateConfig renders the buildkitd.toml the embedded daemon runs with.
+//
+// The GC policy is the part worth reading. BuildKit prunes the build cache
+// after every build, running each worker.oci.gcpolicy tier in order and
+// trimming down to that tier's limit. We tune the tiers for how miren builds:
+// stackbuild turns an uploaded app into an image using persistent package
+// caches, and a small set of base images gets rebuilt constantly.
+//
+// The rule that matters: a byte cap only holds if some tier enforces it with no
+// keepDuration. A keepDuration gate keeps any record used within the window,
+// and it keeps that record's whole ancestry along with it, since we can't drop
+// a base layer while a fresher layer still sits on top. On a busy builder that
+// covers almost the entire cache, so an age-gated cap never binds and the cache
+// grows forever. That's how garden hit 27GB against a 10GB cap and fed the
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Probably worth a dig into these sorts of time-bomb-esque issues since we continue to fight with buildkit.

phinze:
Good call on the broader dig. Ran an audit for the whole class and it's real, more than just this one. Filed it all under MIR-1296 with children if you want to click through.

PR #880, opened 2026-07-02
Fix build saga crash recovery so resumed builds complete after a restart

phinze:
Saga crash recovery for builds worked in the narrow sense that the executor faithfully recovered an in-flight `build-from-tar` saga from etcd on restart and resumed the `build-image` action. What it couldn't do was finish. The recovered build would try to push its image to `cluster.local:5000` and fail with `dial tcp: lookup cluster.local ... no such host`, burn through buildkit's fast internal retries in about a second, and roll the whole thing back. A fresh deploy minutes later worked fine, which was the tell that this was an ordering problem, not a broken registry.

Mapping the boot sequence made the cause clear, and it's more of a structural inversion than a race. Build saga recovery ran synchronously inside `co.Start()`, but the things a push depends on (the flannel-derived router address, the `cluster.local` host mapping, `SetRegistryIP`, and the registry listener itself) are all established later in the server boot path, after `co.Start()` returns. So on every restart with an in-flight build, recovery resumed the push into a world where the push target didn't resolve yet.

The fix is to run recovery from the boot sequence after those dependencies exist, rather than in the middle of coordinator startup. Recovery is split out of `SagaBuilder.Init` (which now just registers the saga definition) into a `Recover` method the coordinator drives via `RecoverBuildSagas`, called once the registry and name mapping are up.

Verifying it end to end (deploy with an `onbuild = ["sleep 180"]` window, SIGKILL the coordinator mid-build, restart) turned up two more things, which is why the diff is a little bigger than a pure reorder.

First, once recovery actually succeeds it re-runs the whole build to completion, which takes minutes, and doing that synchronously wedged boot: the server never signaled ready and health checks would time out. The old code got away with synchronous recovery only because it always failed in about a second. So recovery now runs in the background, and boot completes promptly while the recovered build finishes on its own.

Second, with recovery finally getting past the push, it started colliding on `create-config-version`. It turned out `saga.ListIncomplete` combined the pending, running, and undoing status-index IDs without deduplicating, so a saga that lingered in more than one index got recovered twice, and the second pass re-ran already-completed actions and conflicted on an entity the first pass had created. This bug predates the change; it was just invisible while recovery always died at the push. Fixed by deduplicating in both saga storage backends.

After all three, the repro is clean: the coordinator boots in a few seconds, recovery runs exactly once, and the resumed build completes and activates the new version, with zero DNS errors and zero conflicts.

The stale deployment lock a crashed deploy leaves behind (the other symptom in the original issue) is intentionally out of scope here and stays with MIR-681, where server-owned deployment lifecycle is the right home for it. The broader "boot ordering is implicit and fragile" problem this exposed is written up separately as an RFD (Boot Dependency Graph).

Closes MIR-1285

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -943,6 +943,15 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		}
 	}
 
+	// The registry is now listening and cluster.local resolves to it, so
+	// it's safe to resume any build sagas interrupted by a previous crash.
+	// Recovery runs here, not inside co.Start, because a resumed image push
+	// needs these dependencies that Start does not yet have (MIR-1285). It
+	// runs in the background on sub (the errgroup/shutdown context) so a
+	// recovered build, which re-runs to completion and can take minutes,
+	// doesn't block boot but is still cancelled on server shutdown.
+	go co.RecoverBuildSagas(sub)
+
 	cert, err := co.IssueCertificate("miren-server")
 	if err != nil {
 		ctx.Log.Error("failed to issue server certificate", "error", err)
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -361,6 +361,12 @@ type Coordinator struct {
 	logAddressesOnce sync.Once
 
 	debugServer *debugsrv.Server
+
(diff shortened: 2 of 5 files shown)


evanphx (approved):
Image

PR #881, opened 2026-07-02
Add runner cordon / drain for distributed runners

phinze:
Taking a distributed runner down for maintenance (a cert rotation, an upgrade, a VM rebuild) had no graceful path. `systemctl stop miren-runner` just drops the node's sandboxes where they sit, and `runner remove --force` is the wrong tool for a temporary maintenance window since it deregisters the node entirely. There was a `Runner.Drain()` that did roughly the right thing, but it was only ever wired to the monolithic server's SIGUSR2 handler, so a distributed runner had no way to invoke it.

This adds three coordinator-driven commands. `runner cordon` and `runner uncordon` flip a node's scheduling eligibility without disturbing anything already running, so the scheduler just stops placing new work there. `runner drain` cordons the node and then evicts its sandboxes so the app controllers reschedule them onto other ready nodes, blocking until the node is empty (or `--timeout` elapses) and reporting how many it moved. This stays simple because distributed runners only ever host stateless workloads (anything stateful pins to the coordinator today), so drained sandboxes float freely to any other ready node, and unlike `remove --force` the runner stays registered and ready to rejoin once maintenance is done.

The interesting design bit is where cordon state lives. The Node already has a `status` (ready/disabled/unhealthy), but that attribute is session-scoped: it's leased to the runner's health session and reset to `ready` on every rejoin. So it can't hold operator intent, since a node "disabled" that way silently becomes schedulable again after a bounce. Instead cordon is a separate, persistent `scheduling` enum. The split is the Kubernetes one: `status` means "the runner process is alive" (reported by the runner), `scheduling` means "the operator wants work here" (set by the coordinator, survives restarts), and the scheduler consults both. Modeling it as an enum rather than a bool leaves room for a future `draining` state if drain ever goes async, without a migration.

Eviction reuses the sandbox-pool cooldown mechanism rather than inventing a new one: drain marks each sandbox STOPPED and lets the existing sandbox controller tear it down and the pool manager recreate it elsewhere. No coordinator-to-runner RPC required, since all of this flows through the entity store the runners already watch.

Getting that working surfaced a real entity-store bug, called out in its own commit and worth a look. `PatchEntity` refused to write to any entity that carried session attributes it wasn't modifying: it re-saved the read-in session attrs and demanded the caller's session, so the coordinator (which holds none) couldn't write a persistent attribute to a Node at all, because a Node always carries that session-scoped `status`. The fix leaves unmodified read-in session attributes in place, which only turns a spurious error into correct behavior. My unit tests missed it because the in-memory store is more lenient than etcd here; it only showed up dogfooding against a live two-peer cluster, which is also where I validated the whole feature end to end (drain a runner, watch a running app migrate to the other node and come back healthy).

Closes MIR-1229. Also filed MIR-1288 for a latent short-form enum write bug in the app restart path spotted along the way.


evanphx (approved):
Image

I needed this a couple weeks back!

PR #882, opened 2026-07-05
Carry non-Go files into the pure-Go runtime image

evanphx:
## Problem

The Go stackbuild's default **pure-Go** path assembles a distroless-static runtime that copied only `/bin/app`, silently dropping every other file in the repo (README, templates, data dirs). This broke apps like `rfd` that read files at runtime relative to `/app`.

Only this one build personality was affected — every other stack (Ruby, Python, Node, Bun, Rust) and even Go's own cgo/JS-augmented path already ship the full `/app` tree.

## Fix

In `GoStack.assembleRuntime`, the distroless-static branch now copies the built `/app` tree onto the runtime image, excluding Go source and the module/vendor build inputs the compiled binary never needs:

```go
ExcludePatterns: []string{"**/*.go", "go.mod", "go.sum", "vendor"}
```

This mirrors the `CopyInfo` the debian-slim branch already uses and removes a footgun from the default Go path. Rather than an explicit opt-in `include` list, non-Go files now travel automatically — consistent with every other stack.

Scope stays on the pure-Go branch; the debian-slim (cgo/augmented) branch is untouched.

## Testing

- New `TestGoRuntimeIncludesNonGoFiles` — builds a real image via buildkit and asserts `app/README.md` and nested `app/data/seed.txt` are present while `app/main.go`, `app/go.mod`, `app/go.sum` are excluded. ✅
- Existing `TestGo`, `TestGoCgo`, `TestGoWithVendor` stay green. ✅
- `make lint` clean (Go + docs). ✅

> Note: these Docker-dependent stackbuild tests skip inside iso (no docker.sock); run on the host. The full `pkg/stackbuild` package (other stacks) was not run.

## Docs

Added a "Runtime Files" note to the Go section of `docs/docs/languages.md`.


phinze (approved):
This inverts a call I made back in #865, where I made distroless strictly binary-only and left go:embed (or flipping on cgo/JS to route to slim) as the escape hatch. And I think you've got it right to undo it. The tell is that we broke our own rfd tool with its config-in-README trick: if *we* were leaning on the "files are just there" invariant without thinking about it, that invariant is worth keeping. Better to match every other stack than make Go the one that quietly drops your files.

One thing worth fixing before this goes in: you updated the user-facing docs, but the `assembleRuntime` header comment (around line 272) still tells the old story, calling this a binary-only image and pointing people at go:embed. That's the misleading bit now, since files come along for free. Whoever edits that function next will read the comment first.

CodeRabbit's point about `vendor` and `go.mod` only matching at the repo root is a fair catch, but it's just bloat on nested-module layouts rather than anything broken, so I'll leave that to you.

--p+🤖

PR #883, opened 2026-07-06
Fix auto-mounted local storage orphaning duplicate pools, and stop the activator stranding ingress on them

phinze:
This came out of a real outage: an unattended reboot took an app with auto-mounted local storage fully offline for ~90 minutes while a healthy sandbox sat idle in a sibling pool. Two separate problems, one the cause and one the amplifier, so this is two revs.

The cause is that auto-mounted storage was patched into the sandbox spec at build time instead of the resolved config. That drifts the pool spec enough to spawn a replacement pool, but leaves `serviceHasDisks` false, which is the gate on the stale-pool drain. So the superseded pool never gets dereferenced and lingers as an immortal duplicate. The first rev registers the auto-mount as a real disk on the resolved config, so the existing drain reaps it like any other disk app.

The amplifier is that the activator binds each `(version,service)` to the first pool it sees at startup and never re-points that binding, so it can get stuck serving from a drained or crash-looping pool while a healthy one sits idle. The second rev makes `watchPools` drop a binding once its pool stops referencing the version, so the next request re-resolves to the pool that actually serves it. It keys off the launcher's own reference, so it guards against a stale binding from any cause, which also makes it safe to roll out alongside the first fix.

Closes MIR-1293

### DIFF
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -2015,6 +2015,49 @@ func (a *localActivator) removePoolFromTracking(poolID entity.Id) {
 	}
 }
 
+// evictStaleVersionBindingsLocked drops version->pool bindings that point at
+// freshPool but whose version freshPool no longer references. When the launcher
+// supersedes a pool it dereferences it (ReferencedByVersions=nil,
+// DesiredInstances=0) without deleting the entity, so the pool still exists and
+// removePoolFromTracking never fires. Without this, a binding established at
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Looks good! 🍪 flagged that `DataPath` might need more enforcement for being correctly set earlier, but not necessary for this change.

PR #885, opened 2026-07-06
buildkit: rebind daemon to the current process on miren restart

evanphx:
## Problem (MIR-1303)

buildkitd runs as a containerd-managed container. On a **clean** miren shutdown, the first SIGTERM cancels the server context (`cli/commands/global.go`), which unwinds `Server()` and runs the deferred `buildkitComponent.Stop()` (`cli/commands/server.go`). `Stop()` SIGTERMs buildkitd and deletes the container, so the next boot's `LoadContainer` returns *NotFound* and creates a fresh daemon. That is the normal, working restart path — the one exercised by `systemctl restart`, deploy-driven bounces, and `make dev-server-restart`.

`KillMode=process` in the systemd unit changes this on an **ungraceful** exit. If miren dies without its `Stop()` defer completing — a crash / panic / OOM-kill, a second SIGTERM (which hits `os.Exit(130)` and bypasses defers), or a graceful stop that exceeds `TimeoutStopSec=90s` and gets SIGKILLed — `KillMode=process` kills only miren's own process and leaves buildkitd (in containerd's cgroup) orphaned and still running.

On the next boot, `restartExistingContainer()` finds that still-running task, sets `c.running = true`, and **reuses it** — only dialing the socket to confirm readiness. It never restarts the buildkitd process. But the reused daemon can't serve the new miren process: buildkit builds use a client-side session (the daemon streams local build context back from the client over a hijacked gRPC stream), and that session state was bound to the miren process that exited. The new process's solves fail with `NotFound: no such job <ref>`. This is **silent** — no error, no log — and **every build fails** until buildkitd is manually killed and miren restarted.

It hit the luanchat cluster on 2026-06-30: all builds failed for ~2.5 hours, and buildkitd had been up ~3 days — i.e. it had survived across the restart via the ungraceful path above.

So the failure is **rare but fatal**: it needs an ungraceful exit to orphan a live daemon, which is why it went unnoticed for so long even though miren has been restarted many times — the vast majority of those restarts took the fresh-daemon path. But once it happens, every build fails until manual intervention.

## Fix

Handle the orphaned daemon at the layer that owns buildkit's lifecycle — the containerd component, on start. When `Start()` finds an existing container, reach the same known-good state a fresh boot produces: a buildkitd task freshly bound to the current process.

- **`restartExistingContainer()`** now always force-stops any existing task and starts a fresh one bound to the current process (the primary fix). It no longer reuses a running task. This evicts an orphaned daemon on the next start regardless of how miren exited.
- **`stopTask()`** no longer returns early when SIGTERM fails, so `task.Delete()` always runs and can't leak a task that would block the next `NewTask`. The delete context keeps the containerd namespace while detaching from the parent's cancellation.
- **Exit-monitor goroutine** (via `task.Wait()`, established before `Start()`) clears `c.running` when buildkitd dies unexpectedly, so `IsRunning()`/`Client()` reflect reality. `Stop()` cancels it and now tears down the container even after such a death.
- **`bkc.Info()` pre-flight** is now fatal — a dead/stale daemon fails the build immediately instead of proceeding into a confusing downstream error.
- **`BuildKitProvider` interface** in `servers/build` lets builds be faked in unit tests (with a typed-nil guard preserving the `BuildKit == nil` check).

## Testing

- `hack/it components/buildkit` — 10 tests pass, including new `TestBuildkitRestart` subtests that prove the buildkitd **PID actually changes** across a simulated restart (old task evicted, new one created and serving), that `IsRunning()` flips false on an out-of-band daemon death, and that `Stop()` cleans up the container even after a death.
- `go test ./servers/build/` — new fake-provider unit test for the connect-error fast-fail; full package green.
- `golangci-lint` / `gofmt` / `go vet` clean.

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -14,6 +14,7 @@ import (
 	"sync"
 	"time"
 
+	buildkitclient "github.com/moby/buildkit/client"
 	"github.com/tonistiigi/fsutil"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -78,6 +79,15 @@ type buildSession struct {
 	cancelFunc context.CancelFunc
 }
 
+// BuildKitProvider is the subset of *buildkit.Component the builder depends on.
+// Abstracting it as an interface lets unit tests inject a fake daemon (e.g. one
+// whose Client fails) instead of standing up a real containerd-managed buildkitd.
+type BuildKitProvider interface {
+	Client(ctx context.Context) (*buildkitclient.Client, error)
+	SocketPath() string
+	IsRunning() bool
+}
+
 type Builder struct {
 	Log           *slog.Logger
 	EAS           *entityserver_v1alpha.EntityAccessClient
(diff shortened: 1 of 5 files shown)


servers/build/build.go
```diff
@@ -95,14 +105,14 @@ type Builder struct {
 
 	// BuildKit is the persistent BuildKit component for container image builds.
 	// When set, uses the shared daemon instead of launching ephemeral sandboxes.
-	BuildKit *buildkit.Component
+	BuildKit BuildKitProvider
```
phinze, line 108: 👏 Love the seam

phinze (approved):
LGTM! Good catch and good fix 🛠️

PR #894, opened 2026-07-07
Runtime-neutral `server container install` (Docker + Podman)

phinze:
Docker is a pain to install, and someone finally asked: can we support Podman instead? (MIR-1311). The clean way there was to stop hard-coding Docker in the containerized install path at all.

So the first commit generalizes it into a container-runtime abstraction. `miren server docker install` becomes `miren server container install`, with the old `server docker *` commands kept as deprecated aliases. A `--runtime docker|podman` flag picks the runtime (auto-detected, preferring Docker when both are present), with a `MIREN_CONTAINER_RUNTIME` env override for scripted installs. The runtime engine itself was never Docker-based (it's containerd + buildkit under the hood), so this only touches the install path and the debug bundle, which now collects from whichever runtime is hosting the container.

We smoke-tested this hard on a Mac with real Podman. Rootful Podman works end to end: install, control plane, real app deploys (including a Postgres app on a persistent LSVD disk), teardown, the whole thing. Rootless Podman is a different story. The server and control plane come up fine, but app sandboxes can't start: Miren's sandbox runtime does privileged, nested containerization (its own containerd, per-sandbox namespaces and devices) that a rootless container can't provide. We root-caused the whole chain (it starts with the pause container's `oom_score_adj`) and filed it as a follow-up (MIR-1313).

That surfaced a UX gap: under rootless, the install would happily "succeed" and then every deploy would crash-loop with an opaque error. So the second commit adds a guard. The install now detects a rootless runtime and stops up front with a clear explanation and the fix (switch Podman to rootful, or use Docker), with `--allow-rootless` to override for control-plane-only testing. While there, it also warns (never blocks) when the runtime's host is short on memory, since the nested stack is hungry.

The third commit cleans up a papercut we kept tripping over along the way (MIR-1166). The install's memory and disk warnings compared reported resources against exact thresholds, but a cloud VM provisioned at a nominal 8GB/100GB reports a bit less (7.76GB, 93.55GB), so people who provisioned exactly the right size still got nagged. A 10% tolerance fixes it, shared across the native and container install paths.

Closes MIR-1311
Closes MIR-1166


evanphx (approved):
Looks great!

PR #896, opened 2026-07-08
Surface unreachable clusters in miren cluster add instead of silently hiding them

phinze:
A user spun up a fresh self-hosted cluster and hit a wall. It registered cleanly, showed up green in the miren.cloud dashboard, and heartbeated every five minutes as `active`. But it never appeared in `miren cluster add`, with no error and no explanation. We spent a good chunk of a debugging session ruling out identity, org scoping, and token caching before finding the real cause: the box had a firewalled inbound port.

miren connects over QUIC (UDP 8443). When the cloud's netcheck can't reach that port, the server correctly decides not to advertise an address no client could dial, so it reports zero API addresses. From there the CLI took over the disappearing act: `miren cluster add` filtered out any cluster with no addresses, and `miren login`'s auto-config path reported "No clusters available for your account." Each individual step is defensible, but the sum is a silent black hole that points users squarely at an auth problem that doesn't exist.

This stops the hiding. Zero-address clusters now show up wherever you'd look for them. The interactive picker lists them disabled (greyed out and unselectable) with "no reachable address" inline and the fix in the footer; the non-interactive listing annotates them the same way; and login now tells the difference between "you have no clusters" and "you have clusters but none are reachable right now." The guidance is specific on purpose: it names UDP 8443 (QUIC is both necessary and sufficient, TCP isn't the thing), mentions the `additional_ips` escape hatch, and points at `miren debug advertise`, which reproduces the exact advertise decision on the host.

This is the first, self-contained slice of a larger fix tracked in MIR-1316. The CLI can only say "no reachable address" generically today because the cloud hands it an empty address list with no reason attached. Later tiers teach the cloud and dashboard to report reachability honestly (the dashboard's current "Connected" indicator is wired to a separate liveness signal, so a firewalled box shows green while being un-addable) and to name the exact blocked port and family, so the message can sharpen to "found 203.0.113.10 but couldn't reach UDP 8443" without leaving your laptop.

Part of MIR-1316.

### DIFF
--- a/cli/commands/config_bind_helpers.go
+++ b/cli/commands/config_bind_helpers.go
@@ -32,6 +32,40 @@ type ClusterResponse struct {
 	OrganizationName  string                 `json:"organization_name"`
 }
 
+// hasReachableAddress reports whether the cloud advertised at least one API
+// address for this cluster. A cluster with none can't be dialed by any client,
+// so it would otherwise be silently hidden from `miren cluster add`. The most
+// common cause is a firewalled inbound port: miren connects over QUIC (UDP
+// 8443), and when the cloud's netcheck can't reach that port it drops the
+// discovered public IP, leaving the cluster advertising nothing. See MIR-1316.
+func (c ClusterResponse) hasReachableAddress() bool {
+	return len(c.APIAddresses) > 0
+}
+
+const (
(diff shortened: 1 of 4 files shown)


evanphx (approved):
I see our new picker UI paradigm has begun!

PR #897, opened 2026-07-08
Remove LSVD completely from runtime (MIR-1194)

evanphx:
## Summary

The runtime moved fully to raw disk images (Universal mode, loop devices) long ago; LSVD only survived as a one-time migration path. This removes LSVD — and the now-dead NBD support — entirely.

- **Delete the `lsvd/` package** (core LSVD, the `lve/` microVM subtree, `cmd/torture`, `pkg/*`).
- **Remove the LSVD→raw-image migration path** from the disk volume controller and its runner startup hook, plus the now-orphaned `stopUserContainers` helper.
- **Remove the `debug disk migrate` CLI command** and its registration.
- **Drop the `lsvd_volume` entity and `disk.lsvd_volume_id` field** from the storage schema (regenerated `schema.go`) and their consumers in the disk controller and `debug_disk`.
- **Remove LSVD-era upgrade shims** that no longer apply: the orphaned `lsvd-server` process cleanup and the `lsvd-state.json` legacy state-file fallback.
- **Remove NBD kernel-module support** (`pkg/nbd` + the `InitializeNBDModule` startup call + the unused legacy `NbdIndex` state field) — it existed only for LSVD's block-device transport.
- **Delete the LSVD tests** and clean up remaining references in build/test tooling.
- `go mod tidy` drops dependencies that only `lsvd/lve` pulled in; regenerated CLI command docs (drops `debug-disk-migrate`).
- **Changelog**: added a Breaking Changes notice (see below).

## Breaking change / upgrade note

A node upgrading **directly** from an old LSVD build will no longer auto-convert LSVD volume data, recover diskio state from `lsvd-state.json`, or stop a leftover `lsvd-server` process.

The automatic LSVD→Universal migration shipped in **v0.6.0** (#639) and runs on first boot of every release since. **Before upgrading to this release, upgrade to at least v0.6.0 and boot it once** so any remaining LSVD volumes are migrated first — otherwise that data can no longer be converted.

## Testing

- Full `make test` suite passed over the LSVD removal (4643 tests, 0 failures, 22 environmental skips).
- Targeted tests + `golangci-lint` + `go vet ./...` pass over the NBD removal; a second full `make test` run was in progress at PR-open time.
- Repo-wide grep confirms zero remaining `lsvd`/`Lsvd`/`LSVD`/`nbd` references in Go source or the storage schema.

Fixes MIR-1194.


docs/docs/changelog.md
```diff
@@ -11,6 +11,9 @@ All notable changes to Miren Runtime will be documented in this file.
 ## Unreleased
 *main*
 
+**Breaking Changes**
+- **LSVD disk volumes are no longer supported** - The legacy LSVD/NBD disk backend has been removed entirely, along with the one-time migration that converted LSVD volumes to the Universal (loop-device) format. That automatic migration shipped in **v0.6.0** and has run on first boot of every release since. If a node still holds un-migrated LSVD volume data when it reaches this release, that data can no longer be converted and the volume will not come back. Before upgrading to this version, upgrade to **at least v0.6.0** and boot it once so every LSVD volume is migrated first.
```
phinze, line 15: Mapped out what actually happens to the unlucky node that skips this gate, just to know the shape of it. On upgrade its disks are still `PROVISIONED` with an empty `VolumeId`, so the controller hits the "provisioned disk, no `disk_volume`" path and `provisionVolume` calls `CreateDiskImage` — a **fresh empty image**. The old interception that would have migrated LSVD data first is gone, so the disk comes back mounted and blank and the app can't tell the difference. Silent, not a crash.

The good news is the bytes aren't shredded: the LSVD segments sit orphaned under `disk-data/` (nothing deletes them anymore), so it's recoverable. Just not by mounting — LSVD is log-structured, so recovery needs the reader this PR deletes. The clean path if it ever happens is a throwaway offline extractor built from the pre-removal `copyLSVDToImage` + `findLSVDVolumeByDirScan`: both reconstruct everything from the filesystem alone (no etcd, no node downgrade, which sidesteps churning a modern entity store with old controllers). So the rescue collapses to "preserve `disk-data/` whole, we build a one-off tool."

Given the install-base size, leaving this as a documented if-this-then-that rather than writing recovery code now is the right call. Recording it here so future-us knows step one is *don't mount the raw bytes, preserve the dir*.

phinze (approved):
Clean removal. Traced every surviving call site: no dangling refs, no orphaned imports, no behavior change in the paths that stay. Confirmed `MountId` still resolves for already-migrated volumes, and the PROVISIONED-with-no-`disk_volume` reconcile still creates a fresh volume in one cycle. `go mod tidy` demoted the still-transitive deps to indirect rather than dropping them, which is correct.

Verified the upgrade-gate claim ourselves — #639 shipped in v0.6.0 with automatic first-boot LSVD migration, so "upgrade to at least v0.6.0 first" names exactly the right release. One inline note recording the failure shape if someone blows past that gate; no action needed on it.

(CodeRabbit's go-billy CVE is a pre-existing indirect-dep thing, unrelated to this change.)

--p+🤖

docs/docs/changelog.md
evanphx: Great, and I think that given our usage pre-0.6.0 i'm not too worried those folks haven't upgraded.

PR #898, opened 2026-07-08
Fix session index-entry leak and patch validation over stale references

phinze:
Baking sagas on garden turned up a persistent error loop that traced back to two independent entity-store bugs, both about the correctness of the index and patch paths. This fixes both. (The larger `pool`-kind index leak from the same investigation is a separate concurrency bug in `DeleteEntity` and gets its own follow-up; this PR is the session-leak and sweeper-loop half.)

**Session-bound entities were leaking index entries.** When an entity's key is leased to a session (`BondToSession`), etcd drops the key when the lease expires. But the plain kind-index entry we write alongside it carried no lease, so it outlived the entity it pointed at. With sessions cycling roughly every 60 seconds, these orphaned entries accumulated fast: on garden, 98% of `session` index entries pointed at entities that no longer existed. The fix leases the plain index entry with the same session lease whenever the entity key itself is leased, so etcd garbage-collects them together. Durable (unleased) entities are untouched and still cleaned up by `DeleteEntity`.

**Patches were choking on pre-existing dangling references.** `PatchEntity` (and `UpdateEntity`/`ReplaceEntity`) re-validated the whole merged entity, including a reference-existence check on attributes the change never touched. So once an entity held a reference to something later deleted out from under it, *every* patch failed, and any controller reconciling that entity looped forever. The disk-lease orphan sweeper is where this surfaced: it couldn't patch an orphaned lease to `RELEASED` because the lease's disk had been deleted, so it retried every five minutes for days. The general fix is that a mutation only owns the attributes it sets, so unchanged references are exempt from the existence check. Creates, and any new or changed reference, are still fully validated. With this in place the sweeper's ordinary release path just works, so no sweeper-specific change is needed.

Both fixes are covered by etcd-backed tests that fail without the change: one asserts a bound entity's index entries share its session lease and vanish when the lease is revoked, the other that a status patch tolerates a pre-existing dangling reference while a brand-new one is still rejected.

Relates to MIR-1320.

### DIFF
--- a/pkg/entity/store.go
+++ b/pkg/entity/store.go
@@ -219,7 +219,7 @@ func (s *EtcdStore) CreateEntity(
 	var coltxopt []clientv3.Op
 	for _, attrs := range indexedAttrs {
 		for _, attr := range attrs {
-			coltxopt = append(coltxopt, s.addToCollectionOp(entity, attr.CAS()))
+			coltxopt = append(coltxopt, s.addToCollectionOp(entity, attr.CAS(), plainIndexLease(o.bind, sid)))
 
 			if sessPart != "" {
 				coltxopt = append(coltxopt, s.addToCollectionSessionOp(entity, attr.CAS(), sessPart, sid))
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Image

Fixin' dem leaks

PR #900, opened 2026-07-08
Auto-tune etcd config to system RAM

evanphx:
## Motivation

As we're working towards further stability, it's important to look at each of our sub-components configuration, this is the first.

## Summary

etcd runs as a small co-tenant inside customer clusters (nodes from ~1 GB to 64 GB RAM) and should hold to roughly **10% of system RAM** rather than grow to fill the node. Until now it launched with a fixed set of options regardless of node size.

This implements the **config-options portion** of the memory-tier tuning analysis: compute a per-node budget and scale etcd's memory-related config off it, auto-detected at startup.

## What it does

Computes a per-node budget `B = max(0.10 × system_RAM, 100 MiB)` from `/proc/meminfo` and derives:

- **`GOMEMLIMIT`** (65% of B) and **`GOGC`** — Go GC pressure toward the budget
- **`--quota-backend-bytes`** (30% of B, capped at 8 GiB) — the actual ceiling on keyspace growth
- Tier-keyed (≤200 MiB / ≤800 MiB / ≤1.6 GiB / else): **auto-compaction retention**, **`--snapshot-count`**, **`--experimental-snapshot-catchup-entries`**, **`--max-concurrent-streams`**, **`--experimental-compaction-batch-limit`**

The computed values are logged at startup (`etcd memory auto-tuning`).

## Design notes

- **Advisory only, no cgroup enforcement.** No `memory.max` is imposed on the container, so miren never OOM-kills etcd. `GOMEMLIMIT` drives the runtime to GC harder as it approaches the budget (soft limit — it will exceed rather than crash if the live set is genuinely too large); `--quota-backend-bytes` is the enforced bound that stops keyspace growth (`NOSPACE`).
- **Auto-detect only** — no new config knobs; operators can still override any value via `ETCD_*` env.
- **etcd v3.5 flag names.** The image is `v3.5.15`, where `snapshot-catchup-entries` and `compaction-batch-limit` are experimental-prefixed (they graduate to stable names in 3.6). Verified against the etcd 3.5 source.
- Bumps `currentConfigVersion` (1→2) so existing containers recreate on upgrade to pick up the new spec.

Out of scope: the fork-level code changes (C.x) and the cgroup enforcement layer from the analysis.

## Testing

- `go test ./components/etcd/` — 13 pass (tier table over 1/2/4/8/16/32/64 GB + sub-1 GB floor + zero-RAM fallback + monotonicity + env/args rendering)
- `go vet`, `make lint` — clean
- `go build ./components/etcd/... ./cli/...` — succeeds

End-to-end verification against a running dev cluster has not been run yet.

### DIFF
--- a/components/etcd/tuning.go
+++ b/components/etcd/tuning.go
@@ -151,6 +151,13 @@ func (t etcdTuning) envVars() []string {
 	}
 }
 
+// signature is a stable fingerprint of the tuning-derived spec (env + args). It is
+// persisted in etcdState so a restart on a node whose RAM has changed regenerates the
+// container instead of reusing a spec built for the old budget.
+func (t etcdTuning) signature() string {
+	return strings.Join(append(t.envVars(), t.args()...), " ")
+}
+
 // args renders the etcd command-line flags for the tuning. Flag names target etcd v3.5,
 // where snapshot-catchup-entries and compaction-batch-limit are experimental-prefixed
 // (both graduated to stable names in 3.6).
--- a/components/etcd/tuning.go
+++ b/components/etcd/tuning.go
@@ -19,9 +19,9 @@ import (
 // GC harder as it approaches the budget.
 //
 // The one exception is --quota-backend-bytes, a hard knob whose breach takes etcd
-// read-only cluster-wide. It is NOT scaled down by RAM: it is floored at etcd's 2 GiB
-// default (see quotaFloorBytes) so the tuning only ever adds backend headroom on large
-// nodes, never removes it on small ones.
+// read-only cluster-wide. It is NOT scaled down by RAM: it is floored well above realistic
+// keyspaces (see quotaFloorBytes) so the tuning only ever adds backend headroom on large
+// nodes, never removes it on small ones. Operators can override it explicitly.
 const (
 	kib = 1024
 	mib = 1024 * 1024
@@ -31,13 +31,14 @@ const (
 	// hold a working keyspace, so the budget never drops under this even on tiny nodes.
 	memoryFloorBytes = 100 * mib
 
-	// quotaFloorBytes is etcd's built-in default backend quota (2 GiB). We never set a
-	// quota below it: --quota-backend-bytes is the one hard knob whose breach raises a
-	// cluster-wide NOSPACE alarm (etcd goes read-only until a manual compact/defrag/
-	// disarm), and keyspace tracks workload rather than the RAM of the node etcd
-	// co-tenants on. Flooring here keeps the tuning purely additive headroom on large
-	// nodes and never a write-outage regression on small ones.
-	quotaFloorBytes = 2 * gib
+	// quotaFloorBytes is the minimum backend quota we ever set. --quota-backend-bytes is the
+	// one hard knob whose breach raises a cluster-wide NOSPACE alarm (etcd goes read-only
+	// until a manual compact/defrag/disarm), and keyspace tracks workload rather than the RAM
+	// of the node etcd co-tenants on. etcd's built-in default (2 GiB) proved too small in
+	// production — a ~711 MB live keyspace bloated to 2.1 GB and hit the quota — so we floor
+	// at 4 GiB (~5.6x that live size, ~2x the bloated file). Flooring keeps the tuning purely
+	// additive headroom on large nodes and never a write-outage regression on small ones.
+	quotaFloorBytes = 4 * gib
 
 	// quotaCapBytes caps the backend quota regardless of how large the node is. This is
 	// etcd's own maximum supported backend size.
@@ -93,19 +94,24 @@ func detectSystemRAMBytes() int64 {
 }
 
 // computeTuning derives etcd config from the node's RAM. A systemRAMBytes of 0 (unknown)
-// yields the smallest tier via the budget floor.
-func computeTuning(systemRAMBytes int64) etcdTuning {
+// yields the smallest tier via the budget floor. quotaOverride, when > 0, sets
+// --quota-backend-bytes explicitly (operator-configured); otherwise the quota is the
+// RAM-scaled value clamped to [quotaFloorBytes, quotaCapBytes].
+func computeTuning(systemRAMBytes int64, quotaOverride int64) etcdTuning {
 	budget := systemRAMBytes / 10
 	if budget < memoryFloorBytes {
 		budget = memoryFloorBytes
 	}
 
-	quota := budget * 30 / 100
-	if quota < quotaFloorBytes {
-		quota = quotaFloorBytes
-	}
-	if quota > quotaCapBytes {
-		quota = quotaCapBytes
+	quota := quotaOverride
+	if quota <= 0 {
+		quota = budget * 30 / 100
+		if quota < quotaFloorBytes {
+			quota = quotaFloorBytes
+		}
+		if quota > quotaCapBytes {
+			quota = quotaCapBytes
+		}
 	}
 
 	streams := budget / (2 * mib)
--- a/components/etcd/tuning.go
+++ b/components/etcd/tuning.go
@@ -13,10 +13,15 @@ import (
 //
 //	B = max(0.10 * system_RAM, 100 MiB)
 //
-// and scale etcd's advisory config (Go GC pressure via GOMEMLIMIT/GOGC, backend quota,
-// compaction/snapshot cadence, concurrent streams) off it. These are advisory only:
-// no cgroup limit is imposed, so etcd is never OOM-killed by us; GOMEMLIMIT simply
-// drives the Go runtime to GC harder as it approaches the budget.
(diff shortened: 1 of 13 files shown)


components/etcd/tuning.go
```diff
@@ -0,0 +1,165 @@
...
+
+// computeTuning derives etcd config from the node's RAM. A systemRAMBytes of 0 (unknown)
+// yields the smallest tier via the budget floor.
+func computeTuning(systemRAMBytes int64) etcdTuning {
+	budget := systemRAMBytes / 10
+	if budget < memoryFloorBytes {
+		budget = memoryFloorBytes
+	}
+
+	quota := budget * 30 / 100
```
phinze: This is the one bound in the PR that fails *hard*: when the backend crosses `quota-backend-bytes`, etcd raises a `NOSPACE` alarm and goes read-only cluster-wide until someone compacts, defrags, and manually disarms it. Worth being conservative here in a way the soft GC knobs don't need.

I measured etcd backend size across the fleet against what this formula (30% of B = 3% of RAM) would set. dbSize is the physical size the quota is actually checked against:

| node | RAM | total keys | etcd dbSize | this PR's quota | headroom |
|---|---|---|---|---|---|
| miren01 | 8 GB | 5.1k | 11 MiB | 238 MiB | 21x |
| miren-club | 16 GB | 23k | 73 MiB | 480 MiB | 6.6x |
| miren-garden | 16 GB | 48k | 81 MiB | 480 MiB | 5.9x |
| selkie | 4 GB | 60k | 51 MiB | 115 MiB | 2.2x |
| **miren-toys** | **16 GB** | **428k** | **470 MiB** | **480 MiB** | **1.02x** |

toys is the tell. It's healthy and busy — ~71k live entities, and only 2,641 of its 285k index entries (0.9%) are orphaned, so this isn't leak bloat that'll drain (#898 won't meaningfully shrink it). It genuinely needs 470 MiB and this caps it at 480. Every node here runs the etcd default 2 GiB quota today, so this only ever *tightens*, most on the small nodes it's aimed at: toys' same 470 MiB keyspace would be 4x over a 4 GB node's 120 MiB quota and 15x over a 1 GB node's 32 MiB. And since `currentConfigVersion` bumps 1→2, every etcd recreates and adopts the new quota on upgrade, so there's no soak period.

The root mismatch is that keyspace tracks workload (entity/app/sandbox count), not the RAM of the node etcd happens to co-tenant on — selkie at 4 GB carries ~10k entities, toys at 16 GB carries ~71k. So a RAM-derived hard ceiling measures the wrong axis with the one knob whose failure mode is a write outage.

Rather than a specific patch, two directions: floor the quota at the current 2 GiB default so the change is purely additive headroom on big nodes and never a regression, or keep the quota independent of RAM and size it off observed keyspace plus margin. The RAM-scaled GOMEMLIMIT is the right tool for "don't hog memory on a small node"; the quota is the wrong one, because being too aggressive here is an outage, not just more GC.

phinze (changes requested):
The memory tiering is clean and I'm happy with most of it. I verified the two experimental flag names against `etcdmain/config.go` at the `v3.5.15` tag (`--experimental-snapshot-catchup-entries` and `--experimental-compaction-batch-limit` are both real there), since e2e hasn't run and a bad flag would fail closed on boot. GOMEMLIMIT/GOGC/compaction/snapshot are all soft knobs, and those I'd ship as-is.

The one thing holding this up is `--quota-backend-bytes`, because it's the only *hard* bound in here and the formula sets it below what we actually run. I measured etcd across the fleet and put the numbers inline — short version: miren-toys is a real 16 GB node sitting at 470 MiB of backend today against the 480 MiB this would give it, and I confirmed that's ~99% live data, not leak bloat, so it won't drain. The same workload on a 4 GB node would land 4x over quota, on 1 GB 15x over. Since keyspace tracks workload and not the RAM of the node etcd happens to sit on, tying a hard, write-stopping ceiling to RAM worries me on exactly the small nodes this targets. Fix options are in the inline note.

(CodeRabbit's retune-on-RAM-change point is real but I'd call it a fine follow-up, not a blocker.)

Requesting changes just to settle the quota question — the rest is a yes.

--p+🤖

evanphx:
@phinze Great analysis, followed your consideration and put a 2GB floor in.

The size of toys vs the usage is really interesting, I want to pull it's data to see why it's got such a big keyspace compared to garden.

phinze (approved):

phinze:
Correcting myself again, and this one's fully on me: I'd been reviewing a stale checkout. My earlier notes here (the inline quota comment and my follow-up) were written against the original commit — I hadn't picked up `565f38257` or `483e79dd5`. With those in, both concerns are already handled:

- **`565f38257` floors `--quota-backend-bytes` at etcd's 2 GiB default.** That's exactly the fix — 3% of RAM is under 2 GiB for every node up to 64 GB, so the quota is always 2 GiB and never scales *down*; the tuning only adds headroom on very large nodes. My whole "this tightens the one hard knob" argument is moot. 
- **`483e79dd5` persists a tuning signature and recreates the container on change**, which covers the retune-on-RAM-change point too.

So scratch the quota concern entirely. For the record, the fleet measurements still stand as useful context and *confirm* the floor is the right call: live keyspace is workload-driven and modest (toys ~200 MiB across 71k entities; garden compacted to ~37 MiB during last night's incident), so 2 GiB is comfortable headroom everywhere, whereas a RAM-scaled value would've been too tight on small nodes. And MIR-1345 is the proof the floor matters — a runaway filled etcd to the 2 GiB default and wedged, so keeping the default rather than dropping below it is precisely what you want.

The real protection against that runaway class isn't a #900 job — it's MIR-1346 (explicit larger quota + NOSPACE self-recovery + runaway monitor). One thing in here worth carrying into that work: the large-node tiers raise auto-compaction retention to 2h/3h (from the current fixed 1h), which retains more churn during a write storm — relevant when MIR-1346 tunes the defrag/self-heal cadence, but not a concern for this PR.

LGTM as it stands. Sorry for the noise off the stale checkout.

--p+🤖

evanphx:
@phinze heads up — the scope of this PR grew after your LGTM (which was at 483e79dd). Following up on the miren-garden NOSPACE incident, I folded the reliability-hardening work you'd scoped to MIR-1346 directly into this PR:

- **80593816** — raise the quota floor 2 → 4 GiB (keep RAM-scaled + 8 GiB cap), make it operator-configurable (`MIREN_ETCD_QUOTA_BACKEND_BYTES` / `--etcd-quota-backend-bytes`); NOSPACE self-recovery in the maintenance loop (AlarmList → compact → defrag → AlarmDisarm), an absolute near-quota compact+defrag trigger, an immediate check at loop startup, loop supervision/retry instead of silent death, and etcd health metrics (db size / in-use / quota / headroom / bloat ratio / alarm).
- **c14dad11** — addresses CodeRabbit's Major note: the near-quota reclaim is now gated on reclaimable bloat (only compact+defrag when it can pull the file back under the high-water; otherwise warn to raise the quota / shed keyspace, rather than thrashing every 5 min). Plus the two nitpicks (schema `min: 0`, hardcoded test golden value).

I know you'd scoped this to MIR-1346 — the call was to keep it here since it's the direct incident fix. Could you take another pass at the added scope, especially the maintenance-loop reclaim/recovery logic in `components/etcd/maintenance.go`? e2e on a live cluster still hasn't run.

evanphx:
@phinze A few revs, including adding better defrag detection and calling, worth a re-review.

phinze (approved):
Our little robo etcd operator gets promoted to senior engineer! Took the second pass on the reliability scope and it holds up.

The part you asked me to look hardest at — the reclaim/recovery logic — is the solid part. I read `recoverFromNoSpace` against the runbook we actually ran by hand to unwedge garden (compact `--physical` → defrag → `alarm disarm`) and it's a faithful automation of it, down to the physical-compact-before-defrag gotcha that bit us live. And the reclaim-vs-warn split does the right thing: compact+defrag only when there's genuinely reclaimable bloat, warn instead of thrash when the backend is just full.

One thing I went back and forth on and want to leave a note on: should `recoverFromNoSpace` gate the same way, and bail with a warning when it can't reclaim? I talked myself out of it, and I think you called it right. Once NOSPACE is armed we're already read-only, so a loop that keeps trying (and heals itself the second space frees up, like when someone bumps the quota) beats one that gives up and stays wedged. The proactive and recovery paths *should* behave differently there, and they do.

So, APPROVE. A heads-up that's got nothing to do with this PR: these new etcd gauges only live in the node's embedded VM right now — we don't ship operational metrics up to central o11y yet, so they won't show in garden Grafana. Filed that as MIR-1390, and the "nobody's watching the ERROR logs" gap it implies is now fleshed out on MIR-1084, so this PR doesn't have to grow a third time.

The one real loose end is the live-cluster e2e you already flagged. I bet you and a bot could run this through its paces in a dev env for smoke testing — probably worth doing before we land all this machinery around our primary data store. Happy to help drive that if useful.

--p+🤖

PR #906, opened 2026-07-09
Add --summary-json to miren deploy for machine-readable results

phinze:
`miren deploy` has only ever spoken to humans. There's a nice TUI while it builds and activates, ending in a "here's where your app lives" message, but nothing structured comes out the other end. So anything downstream that wants the deploy id, the version miren assigned, or the route URLs is stuck either scraping that human output or making a separate `route list` call and reconciling it. Both are fragile: they break the moment we reword a log line.

The thing is, the command already has all of this in hand right at the success point. It knows the deployment record id, it knows the app version it just built, and the server hands it the access info with the route hostnames. It just wasn't handing any of it back.

This adds a `--summary-json <path>` flag that writes a small JSON object with `deploy_id`, `app_version`, `url`, and `urls`. `deploy_id` is empty for ephemeral previews, which deliberately have no deployment record. The URL derivation mirrors exactly what the human output prints (including the ephemeral `<label>.<cluster-host>` form and the default-route fallback), so the summary is authoritative rather than a second guess that might drift.

Writing happens after the deploy has already succeeded, so a failure to write the summary is logged and swallowed. It never turns a good deploy into a failed command.

This is the runtime half of MIR-780; the mirendev/actions deploy action is the first consumer.

Part of MIR-780

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -55,6 +55,7 @@ func Deploy(ctx *Context, opts struct {
 	Sensitive     []string `short:"s" long:"sensitive" description:"Set sensitive environment variable (masked in output)"`
 	Ephemeral     string   `long:"ephemeral" description:"Deploy as ephemeral preview with this label (e.g. feat-login)"`
 	TTL           string   `long:"ttl" description:"TTL for ephemeral version (e.g. 48h)" default:"24h"`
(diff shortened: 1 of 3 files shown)


evanphx (approved):

phinze:
@coderabbitai review

PR #911, opened 2026-07-10
Consolidate env-var display behind one masked surface

phinze:
`miren app status` masked `SECRET_KEY_BASE` and `POSTGRES_PASSWORD` but printed `DATABASE_URL` in full, exposing the Postgres password embedded in the URL. The leak was a symptom of a broader pattern: env-var display had been copy-pasted across six CLI surfaces, each with its own masking behavior. The text path in `app status` matched against a hardcoded list of secret-ish key names (`password`, `secret`, `token`, ...), and `DATABASE_URL` matches none of them, so its value printed verbatim. That same path also ignored the explicit `sensitive` flag from app.toml, so a var deliberately marked sensitive still showed in the clear.

The fix has two halves. First, consolidation: env-var *values* now belong to one noun, `miren env`. Values are gone from `app`, `app status`, `app --json`, and the `deploy --analyze` preview. `app status` keeps the key *names* for glanceability and points at `miren env list`. This also let us drop some dead weight: the `--config-only` flag (a pre-`--json` debug dump that emitted raw proto), an env block in the `app` TUI that was built but never rendered, and the `isSensitiveKey` name matcher.

Second, one masking authority. Every surface that shows a value now goes through a single function, and it encodes a clear strategy. Sensitivity is decided by evidence, in tiers: the declared `sensitive` flag is the authority (set by the user, or stamped by the addon that minted the credential), and on top of that a structural invariant redacts `user:pass@` userinfo in URL-shaped values regardless of key name, since that's provable credential syntax rather than a guess. Name matching still exists, but as a set-time suggestion you can override, not a display-time decision. A name-guess that silently fails at display is how `DATABASE_URL` slipped through, so the display path doesn't guess. Redaction is fixed-width (no first-character or length tell), and values are escaped so a hostile env value can't corrupt a terminal or CI log.

There's also no bulk reveal. `env list` and friends always mask; the only way to see a real value is `miren env get <key> --unmask`, one secret at a time. That keeps the friction on revealing proportional to the exposure.

Smoke-tested end to end against a live deploy with an unmarked `DATABASE_URL`: masked in `env list` (table and json), `env get`, `app status`, and `deploy --analyze`, revealed only through `env get --unmask`.

Closes MIR-1356

### DIFF
--- a/cli/commands/app.go
+++ b/cli/commands/app.go
@@ -2,11 +2,9 @@ package commands
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"time"
 
@@ -164,8 +162,6 @@ func App(ctx *Context, opts struct {
 	FormatOptions
 	Watch bool `short:"w" long:"watch" description:"Watch the app stats"`
 	Graph bool `short:"g" long:"graph" description:"Graph the app stats"`
-
-	ConfigOnly bool `long:"config-only" description:"Only show the configuration"`
 }) error {
 	crudcl, err := ctx.RPCClient("dev.miren.runtime/app")
 	if err != nil {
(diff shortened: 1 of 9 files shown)


evanphx (approved):

phinze:
@coderabbitai review

PR #913, opened 2026-07-13
Advertise containerized state to cloud for Miren Anywhere (MIR-1164)

evanphx:
## What

When the miren server starts **inside a container** (Docker Desktop, `miren server container install`, k8s, etc.) it detects that and advertises it to miren.cloud in the cluster status report, so cloud can keep **Miren Anywhere** (POP routing) forced on for the cluster automatically.

## Why

A containerized server is effectively never reachable directly from the internet — its advertised addresses are container-internal and published ports sit behind NAT, so netcheck can't confirm public reachability. Today an operator in that situation has to flip the per-cluster Miren Anywhere toggle (added in MIR-1163) to **Always** by hand. This closes that gap by reporting the fact and letting cloud apply the policy.

## Changes

- **`pkg/containerenv`** (new) — `InContainer() bool` detects a container environment via `/.dockerenv` (Docker), `/run/.containerenv` (Podman), and container/kube markers in `/proc/{1,self}/cgroup`. Memoized with `sync.Once`; probe paths are injectable for tests.
- **`pkg/cloudauth/status.go`** — new `Containerized bool` field on `StatusReport`, sent without `omitempty` so an explicit `false` keeps cloud in sync if a cluster ever moves off a container.
- **`components/coordinate/coordinate.go`** — populate `Containerized` in both the startup (`ReportStartupStatus`) and periodic (`ReportStatus`) status reports.

## Scope

Runtime-only. The cloud-side consumption of the flag (defaulting the Miren Anywhere toggle to Always) lives in a separate repo. **The JSON key `containerized` is a cross-repo contract** — confirm it matches the cloud consumer before that side ships.

## Testing

- `pkg/containerenv` unit tests: docker/podman/cgroup markers + negative cases.
- `pkg/cloudauth` marshal test: `containerized` is always emitted for both `true` and `false`.
- `components/coordinate` package compiles and its tests pass.
- Verified end-to-end that `InContainer()` returns `true` inside the containerized dev/CI environment (via `/.dockerenv`).
- `make lint` clean.

### DIFF
--- a/pkg/containerenv/containerenv.go
+++ b/pkg/containerenv/containerenv.go
@@ -1,0 +1,75 @@
+// Package containerenv detects whether the current process is running inside a
+// container (Docker, Podman, or a Kubernetes pod). The miren server uses this to
+// advertise to miren.cloud that it's containerized, since a containerized server
+// is effectively never reachable directly from the internet and must keep Miren
+// Anywhere (POP routing) on.
+package containerenv
+
+import (
+	"os"
+	"strings"
+	"sync"
+)
+
+// dockerEnvPath and podmanEnvPath are the marker files each runtime drops into
+// the container filesystem. Their presence is a definitive signal.
+var (
+	dockerEnvPath = "/.dockerenv"
+	podmanEnvPath = "/run/.containerenv"
+)
+
+// cgroupPaths are read as a fallback when the marker files are absent. Container
+// runtimes leave recognizable tokens in the cgroup hierarchy of PID 1 and the
+// current process.
+var cgroupPaths = []string{"/proc/1/cgroup", "/proc/self/cgroup"}
+
+// cgroupMarkers are substrings that indicate a containerized cgroup hierarchy.
+var cgroupMarkers = []string{"docker", "containerd", "podman", "kubepods", "libpod"}
+
+var (
+	once     sync.Once
+	inCached bool
+)
+
+// InContainer reports whether the process is running inside a container. The
+// result is computed once and cached — it can't change over the process
+// lifetime.
+func InContainer() bool {
+	once.Do(func() {
+		inCached = detect()
+	})
+	return inCached
+}
+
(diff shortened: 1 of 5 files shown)


pkg/containerenv/containerenv.go
```diff
@@ -0,0 +1,75 @@
...
+	podmanEnvPath = "/run/.containerenv"
+)
+
+// cgroupPaths are read as a fallback when the marker files are absent. Container
+// runtimes leave recognizable tokens in the cgroup hierarchy of PID 1 and the
+// current process.
+var cgroupPaths = []string{"/proc/1/cgroup", "/proc/self/cgroup"}
+
+// cgroupMarkers are substrings that indicate a containerized cgroup hierarchy.
+var cgroupMarkers = []string{"docker", "containerd", "podman", "kubepods", "libpod"}
```
phinze, line 27: The bots are right to flag `"containerd"`, though the exact false positive they describe doesn't quite hold (the process reading this is the miren server, not the containerd daemon, so its cgroup wouldn't be `containerd.service`). Still worth dropping: it's a loose substring match feeding a policy that sticks, and `/.dockerenv` plus `kubepods`/`cri-containerd-` already cover the real cases. Cheap robustness, not a blocker.

phinze (approved):
Approving. Like we talked through yesterday — I came in skeptical that we needed a `containerized` flag at all, then landed on yeah, this is the right move for demo-path smoothness. Worth being precise on one thing from that convo: the plan is for this signal to nudge cloud's Miren Anywhere default from **off → auto**, not off → on. Auto still lets netcheck do its thing, and anyone running a load-bearing container who wants it truly off can flip to never and have it stick.

Clean package, good tests, and skipping `omitempty` on purpose is a nice touch. Just make sure the cloud consumer reads `containerized` before that side ships — it's a cross-repo contract.

--p+🤖

PR #914, opened 2026-07-13
Fix disk_mount reconcile hot-loop on orphaned mount (MIR-1345)

evanphx:
## Summary

Fixes the MIR-1345 incident where an orphaned `disk_mount` entity spun in a tight, backoff-free reconcile loop, ballooning etcd from ~0.7 GB → the 2 GiB backend quota in ~90 min, arming a persisted `NOSPACE` alarm and wedging the runtime on boot.

This PR covers the **primary root-cause fix only**. The etcd defense-in-depth items from the ticket (explicit `--quota-backend-bytes`, NOSPACE self-recovery, absolute-size defrag triggers/metrics) are intentionally deferred to a separate follow-up.

## Why it looped

The disk-mount `ReconcileController` runs a **watch** (every etcd change to a matching entity is enqueued) and a **worker** (pulls events, calls `reconcileMount`). The framework has a `recentWrites` ring (`pkg/controller/controller.go:207`) whose job is to skip watch events for revisions the controller *itself* wrote — "if I caused this change, I don't need to react to it." It's populated by `RecordWrite`, called from `applyUpdates` (the path where a controller *returns* attrs for the framework to write).

`DiskMountController` doesn't use that path. It writes state **directly** via `c.eac.Patch` inside `updateMountState`, and discarded the result. So its revisions never entered `recentWrites`, and its own writes were never filtered from the watch:

1. Worker reconciles the mount: `DM_ERROR` + `DM_WANT_MOUNTED` → `attachAndMount` → volume missing → `setMountError` → `eac.Patch` writes rev `N`.
2. That write is a change to the entity → the watch delivers an event for rev `N`.
3. The callback enqueues it (it's not in `recentWrites`) → worker reconciles again → writes rev `N+1`.
4. Back to step 2.

**The controller's own write was the input that woke it up again.** Each iteration wrote `DM_ATTACHING` then `DM_ERROR` (two writes, ×7 keys via secondary indexes), with no external driver — self-sustaining churn at ~1000 writes/s.

For the incident mount, both the backing `disk_volume` and the parent `disk_lease` were already gone from etcd (doubly orphaned), so every attempt failed identically and re-fired immediately. #898 (which made previously-failing patches over stale references succeed) didn't cause the loop — it *unmasked* it, converting a silently-failing loop into one that wrote durably at full throughput.

## Fix 1 — Record the controller's direct writes

Capture the revision returned by `eac.Patch` in `updateMountState` and report it through a `WriteTracker`, exactly as the framework intends for "controllers that make manual entity writes outside the reconciliation framework." Now:

- Worker writes rev `N` → `RecordWrite(N)` adds it to the ring.
- The watch event for `N` hits `recentWrites.Contains(N)` → **true** → dropped, never enqueued.
- No self-generated work → worker goes idle. Loop broken.

The mount is still stuck in `DM_ERROR`, but it now sits quietly and is only re-examined on a genuinely external trigger (the 5-min periodic resync, or a real entity change) instead of at full throughput. This is a general fix for any direct-write oscillation in this controller, not just the orphan case.

A minimal one-method `writeTracker` interface is declared locally in `diskio` to avoid a new package dependency on `pkg/controller` (the coupling already lives at the `runner` layer). It's nil-safe, so existing unit tests without a tracker are unaffected.

## Fix 2 — Sweep doubly-orphaned mounts

Fix 1 stops the etcd churn, but a mount whose backing volume is permanently gone would still retry forever on every resync and never settle. `ReconcileOrphanMounts` (mirroring the existing `DiskLeaseController.ReconcileOrphanLeases`) deletes such mounts:

- Lists node-scoped `disk_mount` entities, resolves each `VolumeId` via `eac.Get`.
- Only a definitive `cond.ErrNotFound` counts as orphaned; transient errors are logged and skipped (fail-safe — a blip can't cause mass deletion).
- Best-effort local teardown (unmount/detach/release), then `eac.Delete` on the entity. Deleting through the entity server cleans up the secondary-index keys — a raw etcd `del` of only the `/entity/` key would leave the exact dangling-index state #898 was about.
- Guarded by `OrphanMountSweepGracePeriod` (2m) against a mount created just before its volume is observed.

Wired as a periodic callback on the disk-mount controller (runs once at startup + every 5 min).

## Testing

- 5 new tests in `disk_mount_orphan_sweep_test.go`: self-write recording, nil-tracker safety, sweeper deletes missing-volume mount + tears down local state, sweeper keeps mounts with a live volume, sweeper respects the grace period.
- Full `components/diskio` package (123 tests) passes; `go build ./...` and `golangci-lint` clean.

## Deferred (separate follow-up)

etcd defense-in-depth so a future runaway degrades instead of wedging boot: explicit configurable `--quota-backend-bytes`, NOSPACE self-recovery in the maintenance loop (`AlarmList` → compact → defrag → `AlarmDisarm`), and an absolute-size defrag trigger + DB-size/alarm metrics.

### DIFF
--- a/components/diskio/disk_mount_controller.go
+++ b/components/diskio/disk_mount_controller.go
@@ -2,6 +2,7 @@ package diskio
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"path/filepath"
@@ -10,8 +11,23 @@ import (
 
 	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
 	"miren.dev/runtime/api/storage/storage_v1alpha"
+	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/entity"
 )
+
+// OrphanMountSweepGracePeriod is how old a disk_mount must be before the orphan
+// sweeper will delete it for a missing backing volume. It guards against a race
+// where a mount has just been created but its volume entity hasn't been
+// observed yet.
+const OrphanMountSweepGracePeriod = 2 * time.Minute
+
+// writeTracker records revisions written directly by this controller so the
+// owning ReconcileController's watch can skip its own events. It is satisfied
+// by *controller.ReconcileController; declared locally to keep diskio decoupled
+// from pkg/controller.
+type writeTracker interface {
+	RecordWrite(revision int64)
+}
 
 // DiskMountController watches disk_mount entities and manages loop-device mounts.
 type DiskMountController struct {
@@ -24,6 +40,11 @@ type DiskMountController struct {
 
 	// Cloud client for lease management and segment replay (nil when cloud not configured)
 	cloudClient CloudDiskClient
+
+	// writeTracker records revisions this controller writes directly via eac,
+	// so the reconcile watch skips the resulting self-generated events instead
+	// of re-triggering reconcile in a tight loop. nil in unit tests.
+	writeTracker writeTracker
 
 	// keepMounts, when true, causes Shutdown to skip unmounting.
 	// Set during reload so the new process can pick up existing mounts.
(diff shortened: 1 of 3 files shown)


phinze (approved):
APPROVE — traced both halves; the fix is correct and the tests cover the cases that matter.

biscuit and CodeRabbit already have the non-blocking stuff (teardown-then-delete, missing node-scope test, O(n) `Get`-per-mount), so I won't pile on — agree it all reads as ship-it.

The one thing they didn't catch, and it's a follow-up not a change here: `DiskVolumeController.updateVolumeState` is this bug's identical twin — direct `eac.Patch`, revision discarded, no `SetWriteTracker` wiring, and a non-convergent `DV_ERROR` path that retries `createVolume` and re-writes the error. Same storm, one persistently-failing volume away. Filed as MIR-1388.

--p+🤖

PR #916, opened 2026-07-14
Report a reachability verdict to cloud

phinze:
When a self-hosted cluster is unreachable because inbound QUIC is firewalled, we could *tell* it was unreachable but not say *why*. The dashboard just showed a generic "Not reachable." The annoying part: both ends already did the work and threw it away. `ComputeAdvertise` builds a full per-candidate trace with reasons (including the exact "address family proven unreachable by netcheck" case), and `miren debug advertise` already renders it, but `apiAddresses()` kept only the final `[]string` and dropped the trace on the floor.

This is the runtime half of MIR-1377. It synthesizes a compact `ReachabilityVerdict` from the cached netcheck result: the public address netcheck actually saw, plus which ports failed and their transport (so cloud can say "UDP 8443 (QUIC)"). That per-port detail has to come from the raw `NetcheckResponse.Results`, since the candidate reason is only family-level. The verdict rides along on `StatusReport` as an optional pointer field, populated in both report builders, and degrades gracefully: old agents and reports from before netcheck has run just omit it, and cloud falls back to today's generic copy.

There's also a new Cluster Connectivity docs page explaining the three connectivity checks and how to read a not-reachable verdict, since the dashboard now links into it.

Pairs with the cloud PR that stores and renders the verdict. Part of MIR-1377.

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -1426,6 +1426,18 @@ func (c *Coordinator) apiAddresses() []string {
 	return final
 }
 
+// reachabilityVerdict synthesizes the agent's inbound-reachability verdict from
+// the cached netcheck result, for reporting to cloud. Returns nil when netcheck
+// has produced no usable public source address, so the field is simply omitted
+// from the report and cloud falls back to its generic copy.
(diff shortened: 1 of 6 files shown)


evanphx (approved):
Looking great

PR #917, opened 2026-07-14
Converge stale entity-store index entries with a background GC

phinze:
MIR-1320 fixed the *source* of stale entity-store index entries: deletes became atomic and session-bound index entries now expire with their lease. But nothing removed the entries that had already leaked, and the only cleanup path was a manual `miren debug reindex` that operators had no real reason to know about. The automatic boot reindex isn't a substitute either, since it's gated on the index schema hash and only ever backfills missing entries, never deletes orphans. So a deployed cluster just held onto its bloat indefinitely, and healing was gated behind a maintenance op nobody runs.

This makes clusters self-heal instead. A new background GC controller runs a gentle, bounded sweep that scans collection entries and removes the ones whose backing entity is authoritatively gone under a linearizable read. Because creates are atomic (the entity key and its collection entries land in one transaction), an index entry with no entity is genuinely orphaned rather than a create-in-flight race. Each delete is still CAS-guarded on the entry's mod-revision as an ABA guard, so a slot that gets legitimately recreated between the scan and the delete is left alone.

The sweep is deliberately boring: it runs off the read path so foreground reads stay pure, it's strictly best-effort (errors are logged, never propagated), and its defaults are gentle (hourly, a few hundred deletes per sweep, paced). It streams the keyspace a page at a time and deletes in batches as it goes, so memory stays bounded by a single page and a pass that gets cut short by its deadline still keeps everything it already removed. Removals are reported per-collection, which means that if some future leak ever reappears it shows up as a specific collection that keeps re-accumulating, rather than hiding in an aggregate count. That same streaming, CAS-guarded, batched cleanup core now backs the manual `miren debug reindex` as well, so there's one delete path to reason about and the big-hammer command got safer in the process.

There's no feature flag: we'll let internal soak before release cover the risk, and the gentle bounded defaults are themselves the mitigation since even a worst case drains slowly and visibly. Verified live in a dev cluster by seeding an orphaned collection entry and watching a sweep detect, CAS-delete, and log its removal, alongside unit coverage for the CAS/ABA guard, the batch-conflict fallback, and the bounded-convergence behavior.

Closes MIR-1389

### DIFF
--- a/pkg/entity/cleanup.go
+++ b/pkg/entity/cleanup.go
@@ -1,0 +1,240 @@
+package entity
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+	"time"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"miren.dev/runtime/pkg/cond"
+)
+
+// cleanupDeleteBatchSize is how many stale entries we delete between rate-limit
+// pauses. Deletes are issued one CAS'd transaction at a time (see
+// CleanupStaleCollectionEntries); this only governs how often we sleep for
+// BatchPause, not how many keys share a transaction.
+const cleanupDeleteBatchSize = 100
+
+// CleanupOptions bounds a single stale-cleanup pass so it can run continuously
+// in the background without overwhelming the store.
+type CleanupOptions struct {
+	// DryRun scans and counts stale entries but deletes nothing.
+	DryRun bool
+	// MaxDeletes caps how many stale entries a single pass removes. A large
+	// legacy backlog then drains over several passes rather than one thundering
+	// sweep. Zero means unbounded (the manual reindex "big hammer").
+	MaxDeletes int
+	// BatchPause is slept after every cleanupDeleteBatchSize deletes to
+	// rate-limit write pressure. Zero disables pacing.
+	BatchPause time.Duration
+}
+
+// CleanupStats reports what a stale-cleanup pass observed and did. RemovedByCollection
+// breaks removals down by index/collection so a persistent, post-drain leak shows
+// up as a specific collection that keeps re-accumulating orphans rather than
+// hiding in an aggregate count.
+type CleanupStats struct {
+	CollectionEntriesScanned int64
+	StaleEntriesFound        int64
+	StaleEntriesRemoved      int64
+	// CASConflicts counts entries that changed between scan and delete (the slot
+	// was legitimately recreated/re-leased), so the CAS guard skipped them.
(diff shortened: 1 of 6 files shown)


pkg/entity/cleanup.go
```diff
@@ -0,0 +1,238 @@
...
+	for _, id := range allEntityIDs {
+		validEntityIDs[id] = true
+	}
+
+	collectionPrefix := s.prefix + "/collections/"
+
+	// pending buffers confirmed-stale entries until we have a batch to delete.
+	// We scan and delete incrementally rather than materializing the whole
+	// keyspace: memory stays bounded by one page + one batch, and if the pass is
+	// cut short (deadline/shutdown) everything deleted before that point sticks.
```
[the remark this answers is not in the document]
phinze: Going to keep the valid-set here, but wanted to record why. It's a deliberate read-reduction trade: the set of live entity IDs lets a sweep confirm the common case (an entry whose entity still exists) with an in-memory lookup instead of a linearizable `GetEntity`. Once a store has drained, nearly every entry hits the set, so a steady-state sweep does ~zero point reads. Dropping it for per-batch existence checks would flip that, so with nothing stale, every entry needs an authoritative read every sweep, which is a large permanent read-load increase on an always-on job. The memory it trades for that is modest and bounded here (single-digit MB at our entity counts, freed after each pass).

The heavier memory concern, materializing the entire collection keyspace, is already handled: the sweep streams page-by-page and deletes in batches as it goes rather than loading everything up front. The double entity-listing you flagged only affects the manual `miren debug reindex`, a rare operator op where the extra scan is negligible. Happy to revisit if entity counts ever grow enough that the ID-set memory actually bites, where a partitioned-scan approach would beat both options.

pkg/entity/cleanup.go
```diff
@@ -0,0 +1,238 @@
...
+	// collection entries that look orphaned. After a store has drained, almost
+	// every entry is in this set and the pass does no point reads at all.
+	allEntityIDs, err := s.ListAllEntityIDs(ctx)
+	if err != nil {
+		return stats, fmt.Errorf("cleanup: failed to list entities: %w", err)
+	}
+	validEntityIDs := make(map[Id]bool, len(allEntityIDs))
+	for _, id := range allEntityIDs {
+		validEntityIDs[id] = true
+	}
```
evanphx, line 73: 🧹 Not a problem today, but this is another instance where we should probably have paging rather than bringing the entire set into memory.

evanphx (approved):

pkg/entity/cleanup.go
phinze: Good call, filed as MIR-1400

PR #919, opened 2026-07-14
docs: language guides (39 languages) + Hermes agent example [MIR-1365]

evanphx:
## Summary

Adds a **Language Guides** section to the docs covering **39 languages** — how to get an app in each onto Miren (setup, env vars, whether a Dockerfile is needed) — plus a new **Examples** section with a worked **Hermes agent** deployment. Each guide points at the Miren agent skills so a coding agent can drive the setup.

Every language guide was **validated end-to-end**: a real app generated per language, deployed to a live Miren cluster, routed, and curled for a 200. The bugs that surfaced during validation are baked into the guides.

## Coverage

**Auto-detected** (no Dockerfile): Python, JavaScript (Node & Bun), Go, Ruby, Rust.

**Dockerfile.miren**: Elixir, Gleam, Crystal, Zig, Nim, C, C++, Objective-C, Deno, .NET, F#, Java, Kotlin, Scala, Clojure, Erlang, PHP, Perl, Raku, OCaml, Haskell, Swift, Dart, JRuby, TruffleRuby, Julia, R, Lua, Klong, Common Lisp, COBOL, Bash, plus static sites/SPAs.

**Example**: deploying Nous Research's Hermes agent (prebuilt image + disk + auth dashboard).

## Findings captured from real deploys

- Every `Dockerfile.miren` app must define a service (Procfile or `[services.web]`) — Miren doesn't use the image `CMD`.
- **Elixir**: set `SECRET_KEY_BASE`/`PHX_HOST` before first traffic (scale-to-zero hides boot failures).
- **Gleam**: runtime Erlang OTP must match the builder; `mist.start` (v6 API).
- **Haskell**: Warp needs `-threaded` or it crashes on the first request.
- **Nim**: Jester/httpbeast segfaults on Nim 2.0 ORC — compile `--mm:refc`.
- **OCaml**: `sudo` fails in the build sandbox — use `USER root` + `--no-depexts`.
- **C/C++**: match the runtime glibc to the compiler image (`gcc:14` → trixie).
- **Objective-C**: runs on SOPE (GNUstep web framework); `-WOPort 0.0.0.0:$PORT`.
- **JRuby/TruffleRuby**: Sinatra 4 host authorization + raise `port_timeout` for slow JVM/GraalVM boot.
- **Dart**: `compile exe` needs glibc (not scratch).

## Validation notes

Docs `bun run lint` + `docusaurus build` pass (broken links are fatal, so all cross-links resolve). Throwaway apps and the temporary benchmark/guides clusters used for validation have been cleaned up.


docs/docs/recipes/hermes-agent.md
```diff
@@ -0,0 +1,260 @@
...
+
+The end state: the Hermes **dashboard** served over HTTPS behind basic auth, the
+OpenAI-compatible **API server intentionally disabled**, and all agent state persisted
+on a Miren disk at `/opt/data`.
+
+Along the way it exercises a lot of real Miren behavior — custom Dockerfile builds,
+the `web` service convention, `0.0.0.0` binding, `port_timeout`, and network disks — so
+it doubles as a tour of what matters when you bring your own image.
+
+:::info[This is an application recipe, not a language guide]
```
phinze, line 24: The real split here isn't guide vs example — everything's a guide — it's your own source vs. a prebuilt image you don't own. "Examples" names the format, not the subject, which is why it reads oddly next to "Language Guides." Consider renaming to **Recipes** and making this the first one. Bonus: it gives "where does Rails go?" a home later.

docs/docs/guides/index.md
```diff
@@ -0,0 +1,99 @@
...
+For most languages, Miren detects your stack from your project files and builds a
+container image for you — **no Dockerfile required**. You run `miren init` once and
+`miren deploy`, and Miren figures out the rest. See [Supported Languages](/languages)
+for exactly what's detected and how.
+
+Every other language here — from Elixir and Gleam to Kotlin, Swift, Julia, and even
+COBOL — isn't auto-detected, so its guide shows you a `Dockerfile.miren` you
+can drop into your project. Miren builds from that Dockerfile instead of guessing. This
+is the same escape hatch available to every language when you need full control over the
+build — see [Using Dockerfile.miren](/languages#using-dockerfilemiren).
```
phinze, line 75: Home for the "request native support" callout we landed on this morning. Every Dockerfile guide says "X isn't auto-detected, so use a `Dockerfile.miren`" but never invites people to ask us to build native support. Without that, 33 Dockerfile pages sit in tension with the "we detect your stack" pitch. One admonition here fixes it: native builds for common stacks today, tell us what's next.

docs/docs/services.md
```diff
@@ -33,9 +33,17 @@ Miren detects services in this order:
...
 1. **`.miren/app.toml`** — Services defined in the `[services.*]` sections
 2. **`Procfile`** — Services inferred from Procfile entries
-3. **Dockerfile `CMD`/`ENTRYPOINT`** — If no services are defined above, but your image has a default command, Miren creates a `web` service using that command
+3. **Detected start command** — For an auto-detected language stack (Python, Node, Bun, Go, Ruby, Rust), Miren synthesizes a `web` service from the start command it detects for your framework
 
-If none of these provide a service definition, the deploy will fail with an error.
+If none of these provide a service definition, the deploy fails with
+`no services defined: please define at least one service in a Procfile or .miren/app.toml`.
+
+:::warning[A custom Dockerfile needs an explicit service]
```
phinze, line 41: Correct and worth adding (confirmed in `build.go`). But the config table below still lists the `command` default as "Image's default entrypoint" (line 196) — true for an `image = "postgres:16"` service, false for a `web` service on your own built image, which is exactly this case. Qualify that row with "for services with an explicit `image`" so the two don't contradict.

phinze (approved):
Approving — content's strong, all docs, all cheap to change, so nothing blocks merge.

Three notes inline — two positioning, one nit:
- the "request native support" callout we agreed on this morning isn't in the guides yet (`index.md`)
- "Examples" vs "Language Guides" mixes format and subject; worth a rename (`hermes-agent.md`)
- the new `services.md` warning contradicts a table row below it (`services.md`)

Any of these works as a follow-up.

--p+🤖

docs/docs/recipes/hermes-agent.md
evanphx: Oh yeah, I like making it Recipes.

phinze: You know I got "Cooking with Miren" on the brain 🙃 👨‍🍳

evanphx: 👨🏻‍🍳

PR #920, opened 2026-07-15
Let .miren/app.toml outrank MIREN_APP when resolving the app name

evanphx:
Fixes MIR-1402.

## The bug

Every app sandbox exports `MIREN_APP` naming its own app (`controllers/deployment/launcher.go:573`, `servers/exec_proxy/exec_proxy.go:348`), and the CLI read a variable of that same name to pick its target (`cli/commands/app.go`). Deploying from inside a sandbox therefore picked up the sandbox's app and silently overrode the `.miren/app.toml` in the working directory — the config file was read, parsed, and then ignored.

Repro: `m app shell` into a sandbox for `foo`, `cd` to a checkout whose `app.toml` declares `bar`, run `m deploy` → deploys to `foo`, no error, no warning. Same-app deploys are harmless, which is what kept this hidden until someone did cross-app work from inside a sandbox.

## The fix

Resolution order is now:

    -a/--app flag  >  .miren/app.toml  >  MIREN_APP  >  miren init prompt

The `env:"MIREN_APP"` tag had to come off `AppCentric.App` to make this possible. mflags applies env tags in `FromStruct`, at command registration — before `Parse`, long before `Validate` — and exposes no provenance (`Flag.HasValue` is documented as "set by env var **or** CLI arg"). So while the tag was present, `Validate` had no way to distinguish an explicit `--app` from the environment. `Validate` now reads `MIREN_APP` itself, positioned below the config check.

`MIREN_APP` still works for its documented CI use — deploying from a checkout with no `app.toml`. It just no longer wins against a config file that is present.

## Breaking change worth a look

`docs/docs/pr-environments.md` told readers to set `MIREN_APP=myapp-staging` in CI so PR previews target staging. A repo with a checked-in `app.toml` will now deploy to the app that file names instead. This was a deliberate call: the override is dropped silently, matching the chosen precedence rather than warning or erroring on disagreement. Docs now steer that case to an explicit `-a myapp-staging`, with warning admonitions there and in `ci-deploy.md`.

Docs are the only mitigation — there is no runtime signal for someone who doesn't read them. A one-line stderr warning on env/config disagreement would be easy to add later if this bites anyone.

## Out of scope

The sandbox injection sites and reserved-var skip lists (`launcher.go:1049`, `build.go:873`) are untouched. Renaming the in-container variable would break anything reading it, and the precedence change resolves the bug without it. One residual: deploying from inside a sandbox in a directory with *no* `app.toml` still resolves to the sandbox's own app — that's the documented env fallback working as designed.

## Testing

`TestAppCentric_AppNamePrecedence` covers the ladder, including a subtest driving the full `FromStruct` → `Parse` → `Validate` lifecycle — the path the bug actually lived on. The direct-construction subtests alone would have passed before the fix, since they never run `FromStruct`. `TestAppCentric_NoEnvTagOnApp` pins that the env tag stays off. Both were confirmed to fail with the tag restored.

Verified end-to-end with the built binary:

| Scenario | Resolves to |
|---|---|
| `app.toml` = checkout-app, `MIREN_APP`=sandbox-app | `checkout-app` |
| same, plus `-a flag-app` | `flag-app` |
| no `app.toml`, `MIREN_APP`=ci-app | `ci-app` |
| no `app.toml`, no env, non-TTY | `miren init` error |

`make lint` clean, docs lint passes. Generated command docs regenerated for the flag description change.

### DIFF
--- a/cli/commands/app.go
+++ b/cli/commands/app.go
@@ -25,9 +25,15 @@ import (
 type AppCentric struct {
 	ConfigCentric `group:"Config Options"`
 
-	_   struct{} `group:"App Options"`
-	App string   `short:"a" long:"app" env:"MIREN_APP" description:"Application name"`
-	Dir string   `short:"d" long:"dir" description:"Directory to run from" default:"."`
+	_ struct{} `group:"App Options"`
+
+	// App carries only the -a/--app flag. MIREN_APP is deliberately not an
+	// env tag here: mflags applies env tags in FromStruct, before Validate
+	// runs, which would leave Validate unable to tell an explicit flag from
+	// the environment. Validate reads MIREN_APP itself instead, below the
+	// config file. See Validate for the resolution order.
+	App string `short:"a" long:"app" description:"Application name (defaults to .miren/app.toml, then $MIREN_APP)"`
+	Dir string `short:"d" long:"dir" description:"Directory to run from" default:"."`
 
 	config        *appconfig.AppConfig
 	resolvedDir   string // absolute path to the directory containing .miren/app.toml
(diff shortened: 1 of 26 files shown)


phinze (changes requested):
Like we discussed, dropping this idea and instead moving the runtime env vars so they don't collide w/ our CLI config ones

evanphx:
Switching to a different env var instead!

PR #921, opened 2026-07-15
Add social card image to docs site metadata

evanphx:
## Problem

`https://miren.md/guides/elixir` (and every other docs page) has no `og:image`. But the Docusaurus theme *unconditionally* emits `twitter:card=summary_large_image`:

```js
// @docusaurus/theme-classic/lib/theme/SiteMetadata/index.js
<meta name="twitter:card" content="summary_large_image" />   // always
...
{defaultImage && <PageMetadata image={defaultImage} />}      // only if themeConfig.image
```

So every shared link promises a large-image card and delivers no image, rendering as a bare text stub in social, Slack, and Discord previews. This is the default state of any Docusaurus site that skips `image` — nothing was misconfigured.

## Change

- Set `themeConfig.image` to a 1200×630 card copied from the marketing site (`mirendev/public/miren-og-card.png`), where it is already the site-wide OG default via `BaseLayout.astro`. The two properties now share one card.
- Add `og:type` and `og:site_name` via `themeConfig.metadata`, which the theme does not emit on its own.

No plugin needed — both are built-in `preset-classic` fields. Docusaurus resolves the path with `withBaseUrl(..., {absolute: true})`, so it becomes a fully-qualified URL as scrapers require. Individual pages can still override via frontmatter `image:`.

## Verification

Built the site and inspected the generated HTML. `build/guides/elixir.html`:

```
og:image:      https://miren.md/img/miren-og-card.png
twitter:image: https://miren.md/img/miren-og-card.png
og:type:       website
og:site_name:  Miren Docs
twitter:card:  summary_large_image   ← now backed by a real image
```

Each tag appears exactly once. `build/index.html` matches; all 237 docs pages inherit it. Build is clean.

## Notes for reviewers

- **The card is now duplicated across two repos** with nothing enforcing they match. There's a comment in the config, but if the marketing card is rebranded the docs copy drifts silently. Hotlinking miren.dev was the alternative, but that couples docs previews to marketing deploys.
- **Rendered previews are unverified.** I confirmed the markup is correct, not that Twitter/Slack/Discord render it as intended — that needs a card validator against the deployed URL after merge.
- Possible follow-up: miren.dev generates *per-page* OG cards (`src/pages/og/[...route].ts`). Nicer than one static card, but it's Astro-specific and would need a separate build step here.


phinze (approved):
Reasonable next step, approving. A static default card is exactly what Docusaurus's own docs do, so we're in good company for v1.

Two seeds for way later, nothing to do here: the dream is Stripe-style per-page cards with the page title baked in — understated, branded, one per page. On GitHub Pages that means build-time generation (satori/resvg in a `postBuild` plugin). Bigger dotted line: move docs onto Miren hosting and we can generate at runtime and do whatever we want. Both future PRs — this is a good step toward them.

--p+🤖

evanphx:
@phinze very much agree, i think we're getting closing to wanting to do more with the docs.

PR #925, opened 2026-07-16
Namespace sandbox-injected env vars under MIREN_RUNTIME_ (MIR-1406)

evanphx:
## Problem

`MIREN_APP` meant two contradictory things:

- **Injected into every sandbox** as the running app's name (`controllers/deployment/launcher.go`, `servers/exec_proxy/exec_proxy.go`).
- **Read by the CLI** as the user's target app (`cli/commands/app.go`, via an mflags `env:` tag).

Because mflags applies `env:` as the flag *default*, and a set default outranks the `.miren/app.toml` fallback, running `miren` inside a sandbox silently targeted the sandbox's app and ignored the app config in front of it. That's the ticket's "causes issues with using the CLI inside a sandbox."

## Fix

Move the injected vars to a `MIREN_RUNTIME_` namespace the CLI never reads:

| Old | New |
|-----|-----|
| `MIREN_APP` | `MIREN_RUNTIME_APP` |
| `MIREN_VERSION` | `MIREN_RUNTIME_VERSION` |
| `MIREN_INSTANCE_NUM` | `MIREN_RUNTIME_INSTANCE_NUM` |

The rule is now: **`MIREN_RUNTIME_*` is injected into your app; every other `MIREN_*` is CLI input.** Clean break — no legacy names. The CLI's `MIREN_APP` input (documented for CI) and the `MIREN_IDENTITY_*` workload identity vars are unchanged.

## Notes for reviewers

- **Breaking change, by design.** Apps reading `MIREN_APP`/`MIREN_VERSION` break on next deploy. Documented in the changelog with the old→new mapping. Dual-injecting would leave `MIREN_APP` in the sandbox and therefore leave the bug unfixed.
- **Names defined once** in the new `api/app/runtimeenv.go` instead of the literal being retyped across five files.
- **Dead-code cleanup:** both copies of `isSystemEnvVar` already fell through to a `MIREN_` prefix check, so the explicit `MIREN_*` switch cases did nothing — dropped, leaving `PORT`/`ADMIN_TOKEN` + the prefix.
- **Frozen-file hash updated:** `sandbox.go` is guarded by `sandbox_frozen_test.go` pending the saga cutover. Per its instructions I audited the saga path — `create_saga.go` builds no container env, so there was nothing to mirror.
- **New coverage** for the gap that let this ship: a launcher test asserting the runtime vars land in the sandbox spec (and the old names don't), plus a regression test asserting no injected name collides with a var the CLI reads. Both were mutation-tested — they fail against the old behavior and pass after.

## Verification

- `make lint` (Go + docs) clean; affected package tests pass.
- Verified end-to-end on a deployed app: a real sandbox shows `MIREN_RUNTIME_APP` / `MIREN_RUNTIME_VERSION` / `MIREN_RUNTIME_INSTANCE_NUM=0` with `PORT` intact and no bare `MIREN_APP`. Running the CLI from a dir with no `app.toml` and the old `MIREN_APP` set reproduces the hijack; with `MIREN_RUNTIME_APP` set it correctly errors asking for an app.

## Follow-up (not in this PR)

`MIREN_VERSION` still carries two other unrelated meanings — the image build arg and which miren release CI downloads. Tracked separately.

### DIFF
--- a/api/app/runtimeenv.go
+++ b/api/app/runtimeenv.go
@@ -16,6 +16,11 @@ const (
 // RuntimeEnvNames lists every var Miren injects under MIREN_RUNTIME_. User
 // config must not override these, and none of them may match a var the client
 // reads from its own environment.
+//
+// EnvRuntimeApp and EnvRuntimeVersion are injected into every app sandbox.
+// EnvRuntimeInstanceNum is injected only for instance-backed sandboxes — those
+// carrying an "instance" metadata label; it is set at sandbox-boot time in
+// controllers/sandbox rather than by the deployment launcher.
 var RuntimeEnvNames = []string{
 	EnvRuntimeApp,
 	EnvRuntimeVersion,
--- a/api/app/runtimeenv.go
+++ b/api/app/runtimeenv.go
@@ -1,0 +1,23 @@
+package app
+
+// Environment variables Miren injects into every sandbox, describing the
+// workload to itself.
+//
+// These live under MIREN_RUNTIME_ so they cannot collide with the MIREN_* vars
+// the client reads as input (MIREN_APP, MIREN_CLUSTER, ...). The two directions
+// previously shared the name MIREN_APP, which meant the CLI run inside a sandbox
+// picked up the sandbox's app instead of the one in .miren/app.toml.
+const (
+	EnvRuntimeApp         = "MIREN_RUNTIME_APP"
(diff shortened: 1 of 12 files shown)


phinze (approved):
My brain background process had landed on this exact same prefix 🧠 🤝

phinze:
(Those CR nits look worth qualifying and potentially addressing tho)

PR #927, opened 2026-07-16
Fix `miren app restart` reporting 0 sandboxes and stopping nothing

phinze:
`miren app restart` had been quietly broken: it always printed "stopped 0 sandboxes" and, worse, didn't actually stop anything. This came in as MIR-1410, but digging in turned up that the root cause was already filed as MIR-1288, so this closes that one and folds in the duplicate reports.

The restart handler stops each running sandbox by patching its status to STOPPED. The bug was in how it built the reference: `entity.Id(compute_v1alpha.STOPPED)` produces `"status.stopped"`, the enum's display string, when the actual entity id is the fully-qualified `SandboxStatusStoppedId` (`"dev.miren.compute/status.stopped"`). That ref pointed at an entity that doesn't exist, so the validator rejected every write, and the loop swallowed the rejection with a warn-and-continue. The counter never incremented (hence "0"), and no sandbox ever moved.

The fix itself is one line: use `SandboxStatusStoppedId`. I went with the single-ref form rather than encoding a whole `Sandbox{Status: STOPPED}`, because create_saga.go already documents that the encoder form emits defaults like `hostNetwork=false` and clobbers real values on a patch.

While in there, I stopped the loop from swallowing the stop error. That warn-and-continue is exactly what turned a caught validator rejection into a silent success. A restart that finds a sandbox and can't stop it should fail loudly, not tell you it worked.

There was no server-level test for AppRestart, which is how this shipped, so this adds one. It fails on the old ref (the sandbox status decodes back to empty) and passes on the fix. I also smoke-tested it in the standalone dev env: deployed a two-pool app, ran `m app restart`, and watched it report "stopped 2 sandboxes across 2 pools" while the old sandboxes went dead and fresh ones came up.

Two follow-ups came out of this that are intentionally not here: MIR-1424 (restart should escalate to a hard kill for wedged processes and block until the restart is confirmed) and MIR-1425 (teach the entity validator to enforce schema.Choices on ref writes, so this whole class of bad-enum-ref mistake gets caught centrally).

Closes MIR-1288

### DIFF
--- a/servers/app/app.go
+++ b/servers/app/app.go
@@ -732,10 +732,13 @@ func (r *AppInfo) Restart(ctx context.Context, state *app_v1alpha.CrudRestart) e
 			}
 
 			if err := r.EC.Patch(ctx, sb.ID, 0,
-				entity.Ref(compute_v1alpha.SandboxStatusId, entity.Id(compute_v1alpha.STOPPED)),
+				entity.Ref(compute_v1alpha.SandboxStatusId, compute_v1alpha.SandboxStatusStoppedId),
 			); err != nil {
-				r.Log.Warn("failed to stop sandbox", "sandbox", sb.ID, "error", err)
-				continue
+				// Surface stop failures loudly instead of swallowing them.
+				// The original bug reported "0 sandboxes" success while
+				// silently failing to stop anything; a restart that can't
+				// stop a sandbox it found must not look like it worked.
+				return fmt.Errorf("stopping sandbox %s: %w", sb.ID, err)
 			}
 			stoppedSandboxes++
 		}
--- a/servers/app/restart_test.go
+++ b/servers/app/restart_test.go
@@ -1,0 +1,69 @@
+package app
+
+import (
+	"context"
+	"log/slog"
+	"testing"
+
+	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


servers/app/app.go
```diff
@@ -732,10 +732,13 @@ func (r *AppInfo) Restart(ctx context.Context, state *app_v1alpha.CrudRestart) e
 			}
 
 			if err := r.EC.Patch(ctx, sb.ID, 0,
-				entity.Ref(compute_v1alpha.SandboxStatusId, entity.Id(compute_v1alpha.STOPPED)),
+				entity.Ref(compute_v1alpha.SandboxStatusId, compute_v1alpha.SandboxStatusStoppedId),
```
evanphx, line 735: we should probably audit using of entity.Id!

evanphx (approved):

servers/app/app.go
phinze: Good call. #932 (stacked on this) does exactly that: enforces `schema.Choices` on ref writes in the validator, plus a `guardrails/` static scan that bans `entity.Id(<short enum>)` inside a ref at CI time. I'm expanding it now to widen the sweep beyond refs, so any misuse of `entity.Id` gets caught, not just the in-ref shape. See MIR-1425.

PR #928, opened 2026-07-16
docs: clarify which commands auto-deploy

evanphx:
## Problem

Agents reading the docs tell users to restart their app after changing env vars. That advice is wrong — `miren env set` and `miren env delete` already create a new app version and roll it out automatically. A follow-up `miren app restart` is just a redundant second rollout.

The docs never said this. The behavior was recorded **only** in `docs/docs/changelog.md` — not in `deployment.md`, not in `app-configuration.md`, not in any language guide, and not in the generated `command/env-set.md`. The one page describing env-change redeploy semantics was buried in a recipe (`recipes/hermes-agent.md`). A reader — human or agent — had no way to learn the truth.

## Approach

State the deploy behavior in the docs for **each individual command that causes a deploy**, so it's impossible to read the command's docs and miss it.

`docs/docs/command/*` is generated, so the prose lives in the CLI source via the existing `WithDescription` hook (`cli/commands/infer.go`), which `hack/gen-command-docs` emits directly below the command synopsis.

| Command | Now documents |
|---|---|
| `deploy` | The only command that rebuilds; `--version` activates an existing image with no build |
| `rollback` | Re-activates a previous image, no rebuild |
| `env set` | New version, no rebuild, no restart needed, waits for healthy |
| `env delete` | Same, plus: copies the **current server-side spec**, not local `app.toml` — deploy `app.toml` changes first |
| `addon create` / `addon destroy` | New version rolls out automatically, deferred until provisioning completes |
| `app restart` | Creates **no** new version, and is **not** needed after any of the above |

`app restart` is the direct antidote to the reported bug.

## Verification

- `MIREN_HELP_JSON=1 bin/miren` confirms the descriptions reach the generator
- `go generate ./cli/commands/` touched only the 7 intended pages — no unrelated churn
- `bun run lint` ✓ · `bun run build` ✓ (admonitions parse) · `make lint` → 0 issues

## Notes

Behavior was verified against the code, not assumed. Everything pivots on `App.ActiveVersion` — writing it triggers a rollout (`controllers/deployment/launcher.go`), and config-only changes reuse the existing image rather than rebuilding.

Claims are scoped deliberately: `app restart` is still legitimately needed after `miren disk restore`, so the docs say restart isn't needed after *config/env/addon* changes rather than promising "never".

Out of scope, but found while reading: `api/app/envvar.go:349` activates the new version *before* the deploy-lock check in `servers/deployment/server.go`, so a user can see "env set failed: deployment in progress" while the change is already live. That's a code bug rather than a docs one — worth its own issue.

Fixes MIR-1409.

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -1082,9 +1082,9 @@ Warning: These commands are intended for advanced users and developers. They may
 	addCommands(d)
 }
 
-// Extended markdown descriptions for commands that create a new app version.
-// These render in the generated command docs (docs/docs/command/*.md) and
-// clarify which operations roll out automatically versus require a rebuild.
+// Extended markdown descriptions for lifecycle commands. These render in the
+// generated command docs (docs/docs/command/*.md) and clarify which operations
+// roll out automatically versus require a rebuild.
 
 const deployDescription = `Deploy uploads your source, builds a new container image on the server, and activates the resulting version — replacing the previously running one. This is the only command that rebuilds your image.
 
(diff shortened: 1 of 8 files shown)


cli/commands/commands.go
```diff
@@ -1074,3 +1081,52 @@ Warning: These commands are intended for advanced users and developers. They may
...
+
+:::note[No restart needed]
+Environment variable changes take effect on their own. Running ` + "`" + `miren app restart` + "`" + ` afterward only triggers a redundant second rollout.
+:::`
+
+const envDeleteDescription = `Deleting an environment variable creates a new app version and rolls it out automatically — you do not need to run ` + "`" + `miren deploy` + "`" + ` or ` + "`" + `miren app restart` + "`" + ` afterward. The new version reuses your existing container image (no rebuild), and the command waits for it to become healthy before returning.
+
+:::warning[Deploy app.toml changes first]
+` + "`" + `miren env delete` + "`" + ` builds the new version by copying the *current server-side* spec, not your local ` + "`" + `app.toml` + "`" + `. If you have pending ` + "`" + `app.toml` + "`" + ` changes, deploy them first, then delete the stale variable — otherwise the delete will roll back your unsent changes. Variables declared in ` + "`" + `app.toml` + "`" + ` drop automatically when you remove them from the file and redeploy.
```
phinze: Traced this one: local `app.toml` edits stay inert until `miren deploy`, so they're never in the server-side spec that `env delete` rebuilds from. That means the delete doesn't *roll back* your unsent edits — there's nothing live to revert — it just doesn't include them, same as before you ran it. The "deploy app.toml first" advice is spot on; only "roll back your unsent changes" overstates it, since env delete operates purely on server state. Maybe something like: "otherwise env delete deploys the server-side spec, so your local `app.toml` edits won't be included"?

phinze (approved):
LGTM. Traced all seven commands' behavioral claims end to end against the code and every one holds — including the subtle ones (env delete rebuilding from the server-side spec, rollback excluding the current active version, deploy being the genuinely-only build path). Exactly the kind of docs fix that earns its keep: it kills a real footgun where agents were telling people to restart after env changes. One small wording note inline on the env-delete warning, non-blocking.

--p+🤖

PR #929, opened 2026-07-16
Rewrite Miren Anywhere docs, value-first

phinze:
The Miren Anywhere docs had the right facts in the wrong order. The page opened with tunnel mechanics (WebSockets, QUIC handshakes) before it ever told you why you'd want the feature, and it sat three levels deep under a Miren Cloud "overview" that was really just a list of CLI commands. For a feature that's about to go default-on, that's backwards.

So this rewrites the page problem-first: your cluster is on a home network or behind NAT with no public IP, and Miren Anywhere puts your apps on the internet anyway by having the cluster dial out to Miren Cloud's POP edge fleet. There's a request-flow diagram (which meant enabling Mermaid in the docs site), an honest treatment of the availability modes that leans on "Miren decides by default" and labels the whole thing as still-evolving, and a "Limitations for now" section that scopes what it covers today (app traffic, not deploys; Miren-managed hostnames, not bring-your-own domains yet) without framing those as permanent walls.

While we were in there, a few things got fixed that we found by reading the code instead of trusting the old copy. The Miren Cloud overview got the same problem-first treatment, turning a command runbook into a real hub that frames the offering and maps into the other cloud pages. We were also claiming "automatic cloud backup" in four places (intro, getting-started, terminology, the overview) when the code only ships manual, local `miren disk backup` and files cloud backup under "roadmap," so that overclaim got swept out everywhere it appeared.

Two spots got fact-checked against the actual TLS and networking code. The new encryption note answers the question a reader naturally has once they learn the POP terminates their TLS: every hop across the internet is encrypted, but it's not end-to-end, since the POP decrypts to route like any CDN edge. And the troubleshooting section lost its cargo-cult "restart the server" advice (the connector already auto-reconnects with backoff) and a wrong firewall claim, in favor of the real symptoms, including the most common one the old copy missed entirely: Anywhere is simply switched off for the cluster.

One idea that fell out of the encryption dig got its own issue (MIR-1426): the packet-path code is almost cleanly separable from the closed cloud internals, so we could open-source the whole data plane and say "everything that touches your packets is in the open."

Closes MIR-1341

### DIFF
--- a/docs/docs/miren-cloud/miren-anywhere.md
+++ b/docs/docs/miren-cloud/miren-anywhere.md
@@ -1,66 +1,104 @@
+---
+title: Miren Anywhere
+description: Serve public apps from a cluster with no public IP. No port forwarding, no static IP, no DNS setup. Miren Cloud handles ingress and your cluster dials out.
+keywords: [miren anywhere, pop, point of presence, nat traversal, no public ip, home lab, ingress, global router]
+---
 
 # Miren Anywhere
 
-Miren Anywhere is the connection between your cluster and Miren Cloud. It's how requests for your [subdomain](/miren-cloud/subdomains) reach your cluster, even when the cluster is behind NAT or a firewall.
+Say your cluster runs on a home network, a laptop, or a locked-down VPC. It has no public IP, and you'd rather not punch holes in your router, pay for a static address, or wire up dynamic DNS just to put an app online. So how does the internet reach the apps you deploy?
 
-## How It Works
+That's what Miren Anywhere does. Your apps go public, Miren Cloud takes the inbound traffic at the edge, and your cluster never has to accept a single inbound connection. It reaches out to Miren Cloud, and traffic flows back down that path.
 
-When your cluster is registered with Miren Cloud, the miren server opens a long-lived connection to miren.cloud itself. When a request arrives at a Miren Cloud Point of Presence (POP), the POP determines which cluster the request is for and asks miren.cloud to have that cluster's server connect directly to the POP. Subsequent requests for your hostnames terminate at the POP, which forwards them to your cluster over the new connection.
+## How it reaches your apps
 
-The result: you don't need a public IP, port forwarding, or DNS gymnastics on your cluster — Miren Cloud handles ingress and uses Miren Anywhere as the path back to you.
+Exposing a private server the usual way means changing your network, and Miren Anywhere skips all of it. The only thing your cluster needs is the ability to make outbound connections, which almost every network already allows.
 
-## When It Runs
+It works through a **POP** (Point of Presence): a Miren Cloud edge server that lives on the public internet, terminates TLS for your hostnames, and forwards requests to your cluster. Your cluster opens the connection *out* to the POP, so from your firewall's point of view nothing is listening and nothing is dialing in.
 
-Miren Anywhere starts automatically whenever the miren server has cloud auth configured — i.e. after you've run `miren server install` (without `--without-cloud`) or `miren server register`. There's no flag to enable, no separate process to run.
+```mermaid
+flowchart LR
+    visitor([Visitor on the internet])
+    pop["Miren POP<br/>edge server, holds your TLS cert"]
+    cluster["Your cluster<br/>no public IP, nothing dials in"]
 
-:::info[Standalone clusters]
-If you run a fully standalone cluster (`miren server install --without-cloud`), Miren Anywhere doesn't start; your cluster reaches the network however you've wired it.
+    visitor -->|HTTPS| pop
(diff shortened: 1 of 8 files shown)


docs/docs/miren-cloud/miren-anywhere.md
```diff
@@ -1,66 +1,105 @@
...
+
+Other installs start with Anywhere off. That's a deliberately cautious default: a cluster that's already reachable is working fine, and Miren would rather leave it alone than reroute it through the POP by mistake. When you want Anywhere on one of these clusters, you turn it on yourself.
+
+You control this per cluster from the **Miren Anywhere** setting on the cluster's page in [Miren Cloud](/miren-cloud/overview):
+
+- **Auto** — route through the POP network only when the cluster has no reachable public address. A cluster that can be reached directly keeps serving its apps itself.
+- **Always** — route through the POP network no matter what, even if the cluster has a public IP.
+- **Never** — leave Anywhere off. Apps are reachable only where the cluster's own network reaches.
+
+(There's also a **Default** setting, which just means "let Miren choose," the automatic behavior described above.)
```
evanphx: This parenthetical makes **Default** sound like a synonym for **Auto** — both read as "let Miren choose." But the troubleshooting section needs them to be different: it says a "non-containerized install still on its default" never routes through the POP, whereas **Auto** routes as soon as the cluster has no reachable public address.

So a reader on a non-containerized box with no public IP asks "am I routed?" and gets *yes* from here and *no* from "Your apps aren't reachable at all." And Default is the mode most people are actually sitting on — it's doing the most work of the four and it's the only one in parentheses.

Could we promote it into the list and say what it keys on? If I've got the semantics right from the runtime side — `containerenv` and `Reachability` go up to Cloud as two independent signals — it's something like:

```
- **Default** — what a cluster starts on: Anywhere is on for containerized
  installs (Docker/Podman), which effectively can't accept inbound traffic,
  and off everywhere else.
```

That'd also make the troubleshooting note downstream read as an obvious consequence rather than a surprise. You'd know the intended Cloud-side semantics better than I can infer them from this repo — the distinction is what matters, the wording's yours.

evanphx (approved):
Approving — the problem-first reframe lands, and the page finally answers "why would I want this" before "here's a QUIC handshake." The encryption callout is the standout: it asks the exact question a reader has the moment they learn the POP holds their cert, and answers it honestly instead of hiding behind "encrypted in transit."

One thing worth fixing before merge, inline below: **Default** is defined in a way that makes it read identical to **Auto**, and the troubleshooting section needs them to be different. It's the mode most readers are actually on, so it's the one that can least afford ambiguity.

`--e+🤖`

phinze:
@coderabbitai review

docs/docs/miren-cloud/miren-anywhere.md
```diff
@@ -1,66 +1,105 @@
...
 
-In the future, we'll allow users to bring their own domain names and route them through Miren Anywhere. Today, you can point any DNS record at a cluster's public IP, which works fine for clusters that are publicly reachable.
+**It covers your cluster's Miren-managed hostnames.** Traffic routes through the POP network for your cluster's `cluster-xyz.miren.systems` address and any [Miren Cloud subdomains](/miren-cloud/subdomains) you've claimed. Bringing your own custom domain through Miren Anywhere isn't supported yet. For now, a custom domain needs a cluster that's directly reachable on the public internet.
 
-## Verifying the Connection
+**It carries app traffic, not deploys.** Miren Anywhere routes the HTTP and HTTPS your visitors load, on ports 80 and 443. It doesn't yet carry the control plane, so `miren deploy`, `miren logs`, and other CLI commands still need a direct network path to the cluster they're talking to. In practice that means a cluster with no public address can serve its apps to the whole internet through the POP network while you deploy to it from the same LAN. Carrying that traffic over Miren Anywhere too is on the roadmap.
```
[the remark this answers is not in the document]
phinze: These are bold topic sentences leading full paragraphs, not inline `**NOTE:**`-style callouts standing in for a block. It's the same lead-in pattern used elsewhere in the docs (e.g. `disks.md`). The "Limitations for now" section also already closes with an `:::info` admonition, so converting these two into admonitions would stack three back to back and read heavier, not cleaner. Keeping them as prose intentionally.

docs/docs/miren-cloud/miren-anywhere.md
phinze: Good catch, thanks, and you had the semantics exactly right. Promoted **Default** into the list describing what it keys on (containerized → on, everywhere else → off, matching `ResolveMirenAnywhereMode`), and added "whether or not it's containerized" to **Auto** so the two can't be read as synonyms. The troubleshooting note downstream now lands as a consequence rather than a contradiction. Pushed in d54bb9e.

PR #930, opened 2026-07-16
Fix explicit local disks triggering the legacy auto-mount warning and duplicate mount

phinze:
Declare a local disk the normal way, `provider = "local"` mounted at some path like `/data`, and the deploy would still warn that `app.toml` had no disk declaration. Worse, the resolved sandbox pool ended up with two mounts: the declared disk at its path, plus an injected `local-data` at `/miren/data/local`.

The transitional auto-mount was self-triggering. Both the builder's migration warning and the launcher's auto-mount injection decided "this app hasn't migrated" by looking for a disk mounted at exactly `/miren/data/local`. A local disk mounted anywhere else didn't count. And because all local volumes for an app resolve to the same per-app host directory (keyed on app ID, not disk name or mount path), once the explicit disk had data, that shared directory was non-empty, so every subsequent deploy saw "legacy data with no disk config" and both warned and injected the duplicate.

The fix broadens both checks: any explicit `provider = "local"` disk now satisfies the migration check regardless of its mount path, while the legacy `/miren/data/local` path check stays for disks intentionally placed there. The auto-mount machinery itself is untouched, since registering it as a real disk is deliberate (MIR-1293). Regression coverage lands for the reproducing case, a populated per-app directory with an explicit local disk mounted elsewhere, asserting only the declared disk is present and no warning fires.

Chasing this turned up something worth writing down: local storage is really one shared per-app store, and every local disk an app declares is just another window onto it, even across services. That's a coherent model but it's completely invisible in the config, so I added a docs note spelling it out and filed MIR-1427 to make it more legible rather than risk a migration.

Closes MIR-1423

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -828,12 +828,14 @@ func (l *Launcher) injectAutoMountLocalDisks(spec *core_v1alpha.ConfigSpec, app 
 	for i := range spec.Services {
 		svc := &spec.Services[i]
 
-		// Skip if the service already declares a disk at the auto-mount path or
-		// under the name we would inject; appending a second disk with the same
-		// name would produce a duplicate volume name in the sandbox spec.
+		// Skip if the service already declares an explicit local-provider disk
+		// (all local volumes share the same per-app host directory, so any local
+		// disk already exposes this data regardless of its mount path), a disk at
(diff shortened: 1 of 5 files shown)


evanphx (approved):
I was wondering why I'd get that warning!

PR #932, opened 2026-07-17
Enforce schema.Choices on ref-typed attribute writes

phinze:
The MIR-1288 bug was `app restart` writing a sandbox status ref as the short `entity.Id(STOPPED)` instead of the full `SandboxStatusStoppedId`, so the ref resolved to nothing and the stop was a silent no-op. #927 fixed that one line; this is the class-wide guard so the shape can't recur.

Chasing how the validator was *supposed* to catch it turned up the real story: the entire `schema.Choices` / enum enforcement path was dead code. The `Attr` builder wrote `ab.choices` into its struct but never read it back, so no schema entity ever carried an `EnumValues` attr, the validator's `TypeEnum` branch was unreachable, and `sb.Enum` is called nowhere. Every "enum" is really a ref with `schema.Choices(...)`, and refs were only existence-checked, never membership-checked. A wrong ref that pointed at a real entity would have sailed straight through.

The fix wires it up end to end: the builder surfaces choices as an `EnumValues` attr, and the `TypeRef` validator enforces membership (reusing the enum branch's check). It stays behind the existing guard that exempts clearing-to-empty and unchanged refs on an update, so only newly written out-of-set values get rejected, keeping the MIR-1320 behavior intact.

One trap worth flagging: `db/enumValues` was declared `TypeArray`, but `ArrayValue` yields `[]Value` while the array validator asserts `[]any`. System schemas dodge it via un-validated writes, but domain schemas install through validation, so every real choices schema would've failed to install with "must be an array." Its `ElemType` was already `TypeAny`, so declaring the attribute itself `TypeAny` is the honest fix and leaves the hot array path alone. This is safe on upgrade because we're loosening validation, not tightening it, and the bootstrap blind-overwrites the core entity before `schema.Apply` runs.

Also lands a `guardrails/` package: a static scan banning a short enum value wrapped in `entity.Id()` (bare anywhere, or package-qualified inside a ref), catching the shape at CI time even on paths the validator never runs. This answers Evan's "audit uses of entity.Id" from #927; the tree came back clean.

Verified in iso across `pkg/entity` (etcd and file store paths), schema, sandboxpool, and the coordinator-boot test that installs every real schema on etcd. Closes MIR-1425.

### DIFF
--- a/controllers/sandboxpool/manager_test.go
+++ b/controllers/sandboxpool/manager_test.go
@@ -642,7 +642,7 @@ func TestManagerZeroValuePersistence(t *testing.T) {
 	for _, ent := range results.Values() {
 		_, err := server.EAC.Patch(ctx, []entity.Attr{
 			entity.Ref(entity.DBId, ent.Entity().Id()),
-			entity.String(compute_v1alpha.SandboxStatusId, string(compute_v1alpha.DEAD)),
+			entity.Ref(compute_v1alpha.SandboxStatusId, compute_v1alpha.SandboxStatusDeadId),
 		}, 0)
 		require.NoError(t, err)
 	}
--- a/pkg/entity/system.go
+++ b/pkg/entity/system.go
@@ -122,12 +122,19 @@ func InitSystemEntities(save func(*Entity) error) error {
 		Type, TypeRef,
 	)
 
+	// EnumValues holds a heterogeneous []Value bag of allowed choices, built
+	// with ArrayValue (which yields a KindArray of []Value). The TypeArray
+	// validator asserts []any and would reject that representation, so this is
+	// declared TypeAny — the ElemType was already TypeAny, i.e. never
+	// element-type-checked. This matters because domain schemas that declare
+	// schema.Choices now surface an EnumValues attr and install through the
+	// validating CreateEntity path (see MIR-1425); the base system entities
+	// bypass validation via basicSave, so they were never affected.
 	enumValues := New(
 		Ident, types.Keyword(EnumValues),
 		Doc, "Enum values",
 		Cardinality, CardinalityMany,
-		Type, TypeArray,
-		EntityElemType, TypeAny,
+		Type, TypeAny,
 	)
 
 	enumType := New(
--- a/pkg/entity/validation.go
+++ b/pkg/entity/validation.go
@@ -68,6 +68,25 @@ func asEntityId(val any) (Id, error) {
 	}
 }
 
+// matchesChoice reports whether val equals one of the declared choice values.
+// It backs both the enum check and the ref-typed schema.Choices enforcement.
+func matchesChoice(choices []Value, val Value) bool {
+	return slices.IndexFunc(choices, func(e Value) bool {
+		return e.Equal(val)
+	}) != -1
+}
+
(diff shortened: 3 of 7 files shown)


controllers/sandboxpool/manager_test.go
```diff
@@ -642,7 +642,7 @@ func TestManagerZeroValuePersistence(t *testing.T) {
 	for _, ent := range results.Values() {
 		_, err := server.EAC.Patch(ctx, []entity.Attr{
 			entity.Ref(entity.DBId, ent.Entity().Id()),
-			entity.String(compute_v1alpha.SandboxStatusId, string(compute_v1alpha.DEAD)),
```
evanphx, line 645: More entity.Id misusage!

pkg/entity/system.go
```diff
@@ -122,12 +122,19 @@ func InitSystemEntities(save func(*Entity) error) error {
...
+	// schema.Choices now surface an EnumValues attr and install through the
+	// validating CreateEntity path (see MIR-1425); the base system entities
+	// bypass validation via basicSave, so they were never affected.
 	enumValues := New(
 		Ident, types.Keyword(EnumValues),
 		Doc, "Enum values",
 		Cardinality, CardinalityMany,
-		Type, TypeArray,
-		EntityElemType, TypeAny,
+		Type, TypeAny,
```
evanphx, line 137: This is something we have to be SUPER careful with because we're updating a core entity. In this case, we're changing it from more restrictive to less, so it's ok.

pkg/entity/validation.go
```diff
@@ -292,6 +311,13 @@ func (v *Validator) validateAttribute(ctx context.Context, attr *Attr, exempt re
 			return fmt.Errorf("attribute %s must be a string representing an entity ID", name)
 		}
 		if str != "" && (exempt == nil || !exempt(*attr)) {
+			// When the attribute declares a choice set (schema.Choices),
+			// enforce membership before the existence check: it's cheaper and
+			// gives a more precise error than "references a non-existent
+			// entity" for the common status-ref case (see MIR-1425).
```
evanphx, line 317: ✨

evanphx (approved):
I like it!

pkg/entity/system.go
phinze: Good flag. I traced the upgrade path to be sure, and it is safe precisely because we are loosening rather than tightening, with the bootstrap doing the work for us. On every boot `NewEtcdStore` runs `InitSystemEntities` before `schema.Apply`, and `InitSystemEntities` blind-overwrites `db/enumValues` via `basicSave` (a raw Put, no validation) whenever the stored bytes differ. So an existing cluster's old `TypeArray` definition is replaced with `TypeAny` before any domain schema carrying an `EnumValues` attr gets validated, and the freshly-constructed store's schema cache starts empty so it cannot serve a stale copy. No migration or manual step needed. The one implicit dependency is that ordering (`NewEtcdStore` before `schema.Apply`), so worth not reordering those two.

controllers/sandboxpool/manager_test.go
phinze: Indeed! scoped out #932 a bit to cover a broader set of misusage in its test

PR #935, opened 2026-07-17
Make mounted disks writable by the service's run user by default

phinze:
A Miren disk mounts as `root:root`, but a lot of real-world container images run as a non-root user, so the app can't write to the disk it just asked for. Until now the only way through was to `chown` the mount from inside the image's entrypoint, and since Miren runs the service command in place of the entrypoint, that meant shipping a wrapper script whose whole job was `chown -R appuser /data && exec …`. That's the papercut this closes: it's the disks half of the "bring your own image" story (RFD-91), where running a packaged upstream image should just work.

The behavior now: a mounted miren disk is owned by the user the container runs as. The nice part is we don't derive that uid ourselves. We read it straight off the resolved OCI spec after `WithImageConfig` has run, so containerd's own `/etc/passwd` resolution turns a named `USER app` into a numeric uid for us, and the whole thing composes for free with a future service-level `user` override (which would just be an earlier `WithUser`).

The chown is deliberately cheap. It follows the Kubernetes `fsGroupChangePolicy: OnRootMismatch` pattern: if the disk's top-level directory already has the right owner, we skip the walk entirely, so a correctly-owned disk pays nothing on every boot. Only a first provision or a genuine ownership mismatch triggers the recursive pass, and even then we chown the root directory *last*, so an interrupted walk leaves the root mismatched and retries next boot instead of marking a half-owned disk "done." Ownership rewrites scale with file count, not bytes, and that one-time migration cost is called out in the docs.

For the unusual app that wants something else, an `owner` field on the disk config overrides the default: `"keep"` leaves the raw mount ownership untouched, and `"uid"` or `"uid:gid"` pins a specific owner. Images that run as root, and read-only mounts, are no-ops, so nothing regresses.

Verified live in a dev server across all three modes: default (a `USER 1500` image writing its disk with no shim), `keep` (a fresh disk staying `root:root` and refusing the write), and a `2000:2000` pin (overriding the derived 1500).

Closes MIR-1428


evanphx (approved):
Nice integration at spec time!

Issue #933, opened 2026-07-17
App-version retention period floor defeats the disk-pressure image-GC → unbounded image/registry growth


jcasimir:
### Summary

On a busy single-node cluster, containerd + the internal registry grew to **~450 GB** and the disk hit 86% before we noticed. The growth is legitimate per the documented retention design, but the design has an operational failure mode: **the `retention_period` floor keeps every version younger than 30 days regardless of count, and the disk-pressure image-GC cannot reclaim any of it** because retained versions pin their artifacts and images. The safety net that exists specifically for disk pressure (`pressure_threshold: 80`) runs hourly and frees nothing.

This isn't a bug report — the behavior matches [`server-config.md`](https://github.com/mirendev/runtime/blob/main/docs/docs/server-config.md#app-version) ("the two settings are a floor, not a budget"). It's a proposal to make the disk-pressure path able to override that floor so the cluster self-heals instead of filling up.

### Environment

- Miren `v0.11.1` (commit `6eb97fa`), single-node standalone, embedded containerd + registry + buildkit.
- ~30 apps, mixed deploy frequency. Host disk 878 GB.

### What we observed

- **508 app-version images** in the `miren` containerd namespace (`ctr -n miren image ls`), e.g. one app at **144**, others at 73 / 57 / 49 / 39.
- Disk: `/var/lib/miren` = **558 GB** — containerd overlayfs snapshots **246 GB**, containerd content **104 GB**, registry blobs **94 GB**.
- `version-gc` logged `nothing pruned, retained: 719 total: 719` on every sweep.
- `image-gc` logged `image GC complete, no images deleted … retained: 560 total: 560` on each hourly disk-pressure trigger (`storage_percent: 86.5 threshold: 80`).

### Root cause

Default retention is `retention_count: 10`, `retention_period: 30d` ([`controllers/version/gc.go` `DefaultGCConfig`](https://github.com/mirendev/runtime/blob/main/controllers/version/gc.go#L37-L43)). The retain predicate is an OR with no absolute ceiling ([`gc.go#L190-L198`](https://github.com/mirendev/runtime/blob/main/controllers/version/gc.go#L190-L198)):

```go
if info.version.ID == activeVersion ||
    i < c.Config.RetentionCount ||
    info.createdAt.After(retentionCutoff) {  // younger than retention_period
    result.RetainedVersions++
    continue
}
```

[...]

PR #936, opened 2026-07-17
Proposal: make version retention GC disk-pressure-aware

jcasimir:
## What this is

We run a self-hosted single-node cluster and ran into steady disk-space growth from accumulated app-version images (background in mirendev/runtime#933). This is a **proposed addition** — an implementation we believe would address that — rather than a bug fix. The current retention behavior is intentional and documented, so treat this as a feature concept for discussion; we're very open to a different direction.

## Idea

Today a version is retained if it's among the most recent `retention_count` **or** newer than `retention_period`. On a cluster that deploys frequently, the period clause keeps a lot of recent versions, and since each one pins a container image, an overlayfs snapshot, and a registry blob, disk usage can climb faster than the window ages things out. The image GC runs under disk pressure, but it can only remove images whose version has already been pruned — so it doesn't reclaim much while those versions are still retained.

This adds a disk-pressure signal to the version GC: when storage usage reaches the image GC's threshold (80%), a sweep drops the `retention_period` floor and prunes each app down to its active version plus `retention_count`. Below the threshold, nothing changes.

## Implementation notes

- The active version and `retention_count` are **always** honored — only the age-based surplus is shed, and only while under pressure.
- Live-sandbox and mid-sweep pin re-checks are unchanged, so nothing in use is removed.
- Disk sampling is injected (`storagePercent`) so the behavior is unit-tested without touching real storage; production wires `DataPath` through the coordinator and samples via `sysstats`.

## Validation

- `go build` / `go vet` / `go test ./controllers/version/` pass, including a new test covering below- and above-threshold sweeps. (Full `iso`/`make test` suite should run in CI.)
- On our cluster, applying the equivalent policy by hand reclaimed ~330 GB (86% → 47% disk) — which is what prompted this. The idea here is to make that reclamation happen automatically under pressure instead of needing a manual retention change.

## Open questions — would love your take

- **Default-on vs opt-in:** it's currently on by default at 80% so a stock install benefits, but happy to gate it behind a config flag (default off) if you'd rather.
- **Threshold:** hardcoded to match the image GC's 80%; could be shared config for a single source of truth.
- If this isn't a direction you'd want, no worries at all — the concept (making retention disk-aware) matters more to us than this specific implementation.

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -1154,9 +1154,10 @@ func (c *Coordinator) Start(ctx context.Context) error {
 		versionGCConfig.RetentionPeriod = c.AppVersionRetentionPeriod
 	}
 	c.versionGC = &versionctrl.GCController{
-		Log:    c.Log.With("module", "version-gc"),
-		EAC:    eac,
-		Config: versionGCConfig,
+		Log:      c.Log.With("module", "version-gc"),
+		EAC:      eac,
+		Config:   versionGCConfig,
+		DataPath: c.DataPath,
 	}
 	c.versionGC.Start(ctx)
 
--- a/controllers/version/gc.go
+++ b/controllers/version/gc.go
@@ -19,6 +19,7 @@ import (
 	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
 	"miren.dev/runtime/pkg/appversion"
 	"miren.dev/runtime/pkg/entity"
+	"miren.dev/runtime/pkg/sysstats"
 )
 
 // GCConfig holds configuration for app version retention GC.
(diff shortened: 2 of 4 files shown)


phinze (approved):
LGTM, this is a good change. What sells it is the policy consistency: version GC and image GC now fire on the same signal, same threshold, same data path, so under pressure they act as one release valve instead of the image GC starving while the versions upstream are still pinned.

On your two open questions, both land as fine for us. Default-on at 80% is the better default posture, and a config surface to tune or disable it is a clean enhancement if anyone actually wants it later. Pruning on pressure from any source is also fine, that's exactly what a release valve should do: the subsystem self-stabilizes as hard as it can when storage gets tight, and worst case it unpins images and reclaims *something* even if the pressure originated elsewhere. Nothing in-use is at risk since active + `retention_count` stay honored.

One fast-follow we'll pick up separately: a small docs tweak to make it explicit that `retention_count` is a hard floor and `retention_period` is advisory/best-effort. The current paragraph gets the scenario across; this would just sharpen the mental model.

`--p+🤖`

PR #938, opened 2026-07-18
docs: add recipes for running Amp and OpenHands agents on Miren

evanphx:
Adds two Recipes guides for running headless coding agents on Miren, both validated end-to-end on a live cluster.

## Amp agent runner (`recipes/amp-runner.md`)
Run [Amp](https://ampcode.com) as a headless **runner** that dials out to ampcode.com, registers as a Machine, and executes threads created from the web app. A portless background service — no inbound port, no dashboard. Covers the official installer, `amp --no-tui --runner-id`, enabling remote thread creation, a stable runner id, and clearing Amp's stale per-directory runner lock on redeploy (a persistent-disk gotcha discovered during validation).

## OpenHands agent server (`recipes/openhands-agent-server.md`)
Run OpenHands' **agent server** (the Agent Canvas backend) as a self-contained Miren app that executes the coding agent **in its own container** — no Docker socket, no sibling sandbox — behind a session-key-authenticated HTTP/WebSocket API, with conversation and workspace state on a local disk.

OpenHands' full UI can't run single-container on Miren (it spawns a sandbox container per conversation via the host Docker socket, which Miren doesn't expose). The agent server is the piece that runs the agent in-process, so it fits cleanly and is driven by an Agent Canvas client. The recipe documents the non-obvious bits: absolute `OH_*_PATH` workspace env vars (the relative default crashes with `PermissionError: 'workspace'`), running as the image's non-root user, a `local` disk, session-key auth (`X-Session-API-Key`), CORS for browser clients, and how to get the session key onto the frontend.

## Notes
- Docs-only; `bun run build` (lint + Docusaurus) passes.
- Both recipes were deployed and exercised on a live cluster — Amp registered as a Machine and ran threads; the OpenHands agent server ran a real shell task in-container.

### DIFF
--- a/docs/docs/recipes/amp-runner.md
+++ b/docs/docs/recipes/amp-runner.md
@@ -1,0 +1,321 @@
+---
+title: Run an Amp agent on Miren
+description: A worked example — run the Amp coding agent as a headless runner on Miren that dials out to ampcode.com and shows up as a Machine you drive from the web app.
+keywords: [amp, ampcode, ai agent, runner, machine, headless, remote thread, example, deploy]
+---
+
+import CliCommand from '@site/src/components/CliCommand';
+
+# Run an Amp agent on Miren
+
+Run [Amp](https://ampcode.com), a coding agent, as a headless **runner** on a
+Miren cluster. The container dials *out* to ampcode.com, registers as a Machine, and then
+waits. You create threads from the ampcode.com web app, pick this runner, and the work
+executes on Miren — streaming back to your browser.
+
+The end state: a single portless Miren service running `amp --no-tui`, connected to
+ampcode.com with an API key, with its working directory on a persistent Miren disk. There is
+**no inbound port, no dashboard, and no HTTP ingress** — ampcode.com is the control plane and
+the runner only makes outbound connections.
+
+Along the way it exercises real Miren behavior — custom Dockerfile builds, a **portless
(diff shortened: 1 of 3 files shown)


phinze (approved):
Really nice work, and the end-to-end validation shows. Where I traced the Miren-specific claims they hold up: the portless-`web` warning matches ingress hardcoding the lease on the `web` service, and the OpenHands "disk made writable for the run user" bit matches the diskowner chown path. Mechanics are sound.

One ask before merge, docs-quality not correctness. The app.toml sections stack a lot of admonitions back-to-back (amp-runner has ~12 after its app.toml, 8 of them `warning`), and the genuine gotchas end up drowning next to explanatory "here's why this value" asides in identical yellow boxes. We just hit this across the docs and added a density rule in #939, with the Hermes recipe getting the first destack pass. Could you give these two the same treatment? Rough cut: keep the bite-you gotchas as callouts (portless-`web`, stale lock, disk churn), fold the explanatory ones (keep-alive, stable-id, point-HOME-at-disk on amp) into prose or an app.toml code comment. The roadblock checklist already carries the scannable summary.

Worth a skim of CodeRabbit's notes too. The one with actual teeth is the stale `settings.json` on the persistent disk: it's written only when absent, so an older on-disk copy silently keeps remote-thread-creation off. Same persistent-disk trap you already handle for the pidfile. The rest are optional hardening.

`--p+🤖`

PR #940, opened 2026-07-20
Re-inject old sandbox env-var names as deprecated aliases (MIR-1435)

evanphx:
## Context

[MIR-1406](https://linear.app/miren/issue/MIR-1406) (#925) renamed the env vars Miren injects into every sandbox — `MIREN_APP` → `MIREN_RUNTIME_APP`, `MIREN_VERSION` → `MIREN_RUNTIME_VERSION`, `MIREN_INSTANCE_NUM` → `MIREN_RUNTIME_INSTANCE_NUM` — and shipped it as a **hard cut**: new sandboxes only get `MIREN_RUNTIME_*`, so any app reading an old name gets an empty string on its next recycle, silently. We hit this ourselves within hours (meet's localhost URLs, reviewagent's version logging).

**The rename never reached a tagged release** — it lives only on `main` (latest tag is `v0.11.1`). So rather than walk back a shipped break, this makes the *first public release* alias-and-deprecate from the start.

## What this does

- **Dual-inject** the old names alongside the canonical `MIREN_RUNTIME_*` names for a deprecation window, centralized in `api/app/runtimeenv.go` (`RuntimeEnvWithAlias` helper). Carries the general **policy** as a doc comment: renames of ambient interface surface default to alias-and-deprecate, not a hard cut.
- **CLI guard** in `AppCentric.Validate`: re-injecting `MIREN_APP` would reintroduce the exact collision MIR-1406 fixed (the CLI reads `env:"MIREN_APP"`, so `miren` run inside a sandbox targeted the sandbox's app). The guard detects "inside a sandbox" (`MIREN_RUNTIME_APP` present + matching injected `MIREN_APP`) and defers to `.miren/app.toml`. CI (sets `MIREN_APP`, no `MIREN_RUNTIME_APP`) is unaffected — MIR-1406's fix holds even during the window.
- **Pool-reuse guard**: legacy names added to `filterSystemEnvVars`' skip-list so adding the aliases doesn't invalidate every pool at once — the mass-recreation shape behind the [MIR-1432](https://linear.app/miren/issue/MIR-1432) pool-doubling incident.
- **Docs + changelog** reframed as alias-and-deprecate with removal in a future release (moved out of Breaking Changes → Improvements).

## Decisions
- Alias **all three** old names + add the CLI guard.
- Deprecation is **docs/changelog only** — no runtime warning, since whether an app reads the old vars isn't detectable (reads happen inside app code), so a per-deploy warn would be pure noise.
- **Removal version left unpinned** ("a future release"); the removal is a separate later change.

## Tests
- Inverted the "old names absent" assertion to "present".
- Added a pool-reuse equality test (MIR-1432 guard) and three CLI-guard cases (sandbox alias ignored, explicit `-a` preserved, CI `MIREN_APP` preserved).
- Refreshed the frozen `sandbox.go` hash.

`make lint` (0 issues + docs lint), `go vet`, and targeted unit tests across all six affected packages are green. The end-to-end dev-env check (deploy, `miren exec` in, confirm both name sets + guard, confirm pool reuse) has not been run here.


phinze (approved):
Traced this end to end and it's solid. `RuntimeEnvWithAlias` is the right place to hang the policy, all three injection sites route through it, and I confirmed `MIREN_APP` is the only legacy name any CLI flag actually reads via an `env:` tag — so guarding just that one is complete, not a gap. Pool-reuse filtering and the frozen `sandbox.go` hash both check out.

biscuit's note on the CLI guard is the one worth a beat before merge: an explicit `-a <sandboxapp>` gets cleared too, which only stays harmless while `.miren/app.toml` agrees. Narrow and non-blocking, but your call on whether to tighten it.

Needs a rebase — currently conflicting with main.

--p+🤖

PR #944, opened 2026-07-21
Type the node-id boundary with compute_v1alpha.NodeId

phinze:
Back during the MIR-1030 review, Evan flagged the little `myNodeId` helper we added in #770 as a bandage over a much bigger pattern. A node id in this codebase comes in two shapes, a raw identifier (a UUID, or `"miren"` for the primary) and the prefixed entity id `node/<raw>`, and we had no type discipline about which one any given variable held. Controllers stored `nodeId string` and hand-stitched the `"node/"` prefix on at every point of use, sometimes with a defensive `TrimPrefix` and usually without. Both forms wore the same `string`/`entity.Id` type, so the compiler could never tell anyone they'd mixed them up. That's exactly the fuzziness that let MIR-1030 happen, a runner writing a disk volume with the wrong `NodeId`.

Before touching anything we surveyed the whole tree to see how big the pattern really was: 40-odd non-test construction sites across 13 entity kinds, and for node specifically, exactly one of sixteen sites defended against double-prefixing. So this is the node-shaped first slice of a larger cleanup.

The fix is a real type. `NodeId` is now a defined type with a single constructor, `NewNodeId`, that applies and normalizes the `"node/"` prefix once. After this the literal `"node/"` appears in exactly one place in the runtime. Controller fields and constructors take a `NodeId`, it gets built once at the boundary where the raw id enters (the runner wiring, the test harness), and it converts back to a plain `entity.Id` only at the store edge via `.Id()`, with `.Matches()` for comparisons against untyped entity fields.

One notable placement decision: the type lives in `compute_v1alpha`, right next to where the node kind is actually schema'd, rather than in the generic `entity` package. Baking a node-specific prefix into the domain-agnostic entity framework would have inverted the layering, teaching the generic store about one specific kind.

The typing also shook out a latent bug. `Runner.Drain` was writing node status to the raw `r.Id` instead of the prefixed `node/<id>` the node is actually stored under (the register path a few lines away got it right), so the disable-on-drain was silently hitting a nonexistent id. It's now routed through the typed id like everything else.

The broader cleanup is captured as follow-ups: MIR-1447 to extend the same discipline to the other twelve kinds, and MIR-1448 to teach the schema generator to emit typed ids so the boundary conversions eventually disappear.

Closes MIR-1072

### DIFF
--- a/api/compute/compute_v1alpha/node_id.go
+++ b/api/compute/compute_v1alpha/node_id.go
@@ -1,0 +1,50 @@
+package compute_v1alpha
+
+import (
+	"strings"
+
+	entity "miren.dev/runtime/pkg/entity"
+)
+
+// nodeIdPrefix is the human-id prefix for the compute node kind (KindNode,
+// "dev.miren.compute/kind.node"). This is the one and only place the prefix
+// is written; every node id is built through NewNodeId.
+const nodeIdPrefix = "node/"
+
+// NodeId is the entity ID of a compute node, always in canonical
+// "node/<raw>" form. Construct one only via NewNodeId, which is the single
+// sanctioned way to apply the "node/" prefix; that discipline is what keeps
+// a raw identifier from ever being mistaken for a prefixed one. Convert back
+// to a plain entity.Id with Id() when handing it to the entity store.
+type NodeId entity.Id
+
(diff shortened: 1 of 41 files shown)


evanphx (approved):
I love a good "look at all my one liner changes!" change!

PR #950, opened 2026-07-22
Make release-tag.sh jj-friendly

phinze:
When run from a jj workspace, the working copy has no `.git` of its own (jj keeps a colocated git repo elsewhere), so the script's git commands can't find the repository and the tag has to be created by hand.

Teach it to handle that: when plain git discovery fails, resolve the backing repo with `jj git root` and point git at it. A jj workspace has no checked-out branch to validate, so in that mode we tag `origin/main` directly after confirming the release PR's changelog landed there, which tags the merged commit by construction rather than trusting HEAD. A normal git checkout behaves exactly as before.

Verified: `bash -n` clean; the already-tagged guard and the changelog guard both fire from a jj workspace and stop before tagging; the changelog grep matches present versions. The jj tag/push path itself won't be exercised end-to-end until the next real release.

### DIFF
--- a/hack/release-tag.sh
+++ b/hack/release-tag.sh
@@ -58,15 +58,52 @@ if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
   exit 1
 fi
 
+# Resolve the repository's git dir. In a jj workspace the current directory has
+# no .git of its own (jj drives a colocated git repo elsewhere), so plain git
+# commands fail with "not a git repository". Detect that case and point git at
+# the backing repo via `jj git root`. A normal checkout, including a colocated
+# primary jj repo, keeps git's default discovery untouched.
+JJ_MODE=false
+if git rev-parse --absolute-git-dir >/dev/null 2>&1; then
+  : # plain git checkout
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Image

phinze:
Responding to the bot reviews.

**On `jj git root` returning the working-copy root (biscuit + CodeRabbit):** false positive for the jj we run (0.42). `jj git root` returns the underlying *Git directory*, not the working-copy root — its help text is literally "Show the underlying Git directory of a repository using the Git backend," and I confirmed the returned path is a real git dir (has a `HEAD`) that `git --git-dir` drives correctly. `jj root` / `jj workspace root` are the ones that return the working-copy root, which is likely the mix-up. The dry-run checks (fetch + changelog lookup against `origin/main`) only succeed because `GIT_DIR` resolves correctly, so no change there.

**Adopted the rest** (just pushed):
- `jj git root --ignore-working-copy`, so discovery does not snapshot the working copy.
- The `--force` warning in jj mode now spells out that fetch and `origin/main` targeting still run.
- `TARGET` is pinned to a resolved SHA in both modes (and the jj changelog check reads from `$TARGET`), so nothing drifts between validation and tagging.

PR #951, opened 2026-07-22
Run the image's entrypoint in exec form when no command is set

phinze:
Part of RFD-91's [bring-your-own-image](https://linear.app/miren/issue/MIR-1430) effort, where the goal is to carve out the reasons you need a hand-written `Dockerfile.miren` shim to run an upstream image. One of those reasons is that Miren throws away the image's `ENTRYPOINT`: every service launches as `/bin/sh -c "COMMAND"`, and for a Dockerfile build with no service command we'd synthesize that COMMAND ourselves by flattening the image's `ENTRYPOINT` + `CMD` back into a shell string.

That flatten (`buildImageCommand`) is lossy in a couple of ways worth spelling out. It quotes argv with Go's `%q`, which is *Go* double-quote syntax rather than shell-safe quoting, so an arg like `--flag=$VAR` comes back wrapped in shell double-quotes and then gets expanded by `/bin/sh -c` (a latent injection footgun, not just a cosmetic bug). And wrapping everything in a shell wedges a process between the runtime and the image's real entrypoint, which is where PID 1 and signal forwarding go subtly wrong.

The nice surprise once I dug in: the sandbox seam already does the right thing. `oci.WithImageConfig` loads the image's exec-form argv into the spec, and the `/bin/sh -c` override only fires when a command is actually present. So the whole fix lives upstream. Stop synthesizing a command from the image (drop `buildImageCommand` and its two call sites), and let the image's own `ENTRYPOINT` + `CMD` stand as the process argv in exec form, the way `docker run IMAGE` behaves. The one gap to close alongside it: with nothing synthesizing a web command anymore, a no-command image build needs to still materialize a `web` service (now with an empty command) so that exec path is actually reachable.

Giving a `command` is completely unchanged. You still get `/bin/sh -c "COMMAND"` with `$PORT`, pipes, redirects, and `exec` all intact, which is the full-control escape hatch. Stack (buildpack) builds are unaffected too: their web command still comes from the stack, and their entrypoint still flows through the existing `cfgSpec.Entrypoint` path.

RFD-91 flagged PID 1 and signal forwarding as the load-bearing bet, so I spiked it in the dev environment before writing the change. A no-command image runs its entrypoint as PID 1, receives a `$`-literal `CMD` arg verbatim (proving exec form, since a shell would have expanded it), and catches `SIGTERM` for graceful shutdown. The new blackbox test (`entrypoint-image`) locks that argv fidelity so it can't quietly regress.

The spike also turned up two things worth tracking separately, both filed: [MIR-1452](https://linear.app/miren/issue/MIR-1452) (make `image = "..."` a first-class build source, the step that deletes the shim entirely rather than just shrinking it) and [MIR-1453](https://linear.app/miren/issue/MIR-1453), a pre-existing bug where the coordinator reads the built image's config back through the in-cluster `cluster.local` registry name it can't resolve. This change doesn't depend on that fetch (it reads the image config from the local content store at launch), but the EXPOSE-port sibling will, so it's linked as a blocker there.

Closes MIR-1444.

### DIFF
--- a/blackbox/entrypoint_image_test.go
+++ b/blackbox/entrypoint_image_test.go
@@ -1,0 +1,39 @@
+//go:build blackbox
+
+package blackbox
+
+import (
+	"testing"
+	"time"
+
+	"miren.dev/runtime/blackbox/harness"
+)
+
+// TestDeployImageEntrypointNoCommand verifies that when a service sets no
+// command, Miren runs the image's own ENTRYPOINT+CMD in exec form (argv, no
+// /bin/sh -c). The fixture image declares no service command, so its ENTRYPOINT
+// must run directly. Its CMD carries a $-literal arg; exec form delivers it
+// verbatim, whereas a shell-wrapped command would let /bin/sh expand it to
+// empty (the var is unset). See MIR-1444.
+func TestDeployImageEntrypointNoCommand(t *testing.T) {
+	c := harness.NewCluster(t)
+	m := harness.NewMiren(t, c)
+
+	name := harness.DeployApp(t, m, harness.AppOptions{
+		Testdata: "entrypoint-image",
+	})
+
+	// The app only becomes healthy if the image's ENTRYPOINT ran and bound the
+	// port, so reaching this point already proves the no-command exec path
+	// launches. Now assert argv fidelity from the entrypoint's own logs.
+	harness.Poll(t, "entrypoint argv logged", 30*time.Second, 2*time.Second,
+		func() (bool, string) {
+			r := m.Run("logs", "-a", name)
+			// Verbatim, unexpanded: proves exec form, not /bin/sh -c.
+			if r.OutputContains("arg[1]=<--note=$ENTRYPOINT_IMAGE_SHOULD_NOT_EXPAND>") {
+				return true, ""
+			}
+			return false, "entrypoint argv not yet in logs"
+		},
+	)
+}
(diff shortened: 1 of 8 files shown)


evanphx (approved):
There was a reason I did this way back in the day, but if the tests all pass, we've work around whatever issue it was there to deal with.

phinze:
@evanphx you said "there was a reason I did this way back," so we went off hunting through the history. In the proud tradition of a cat depositing a dead rat on the doorstep, please accept this modest trophy. It's a low-value gift (the reason turns out to be pretty benign), but you asked, so here's what we dragged back:

The command model was a shell string from day one. `ac6e0bfd` ("Add auto building") set up `/bin/sh -c "COMMAND"`, and `61d57920` the next day added "entrypoint" not as argv but as a *string prefix* concatenated onto the Procfile command (`cmd = av.Entrypoint() + " " + cmd`), the whole thing run through the shell. That served the Procfile/`$PORT` ergonomics and, especially, stack builds where a buildpack's wrapper entrypoint has to wrap the web command in one shell invocation.

Image support came much later (`d330f65d`), and `buildImageCommand` was added purely to flatten an image's exec-form `[]string` entrypoint into that existing string slot. So the flatten wasn't guarding a bug, it was the toll for fitting argv into a string-shaped model.

We're not tearing that model out. `/bin/sh -c` still runs for command-given services, and the stack-wrapper entrypoint path (`stack.Entrypoint()` → `cfgSpec.Entrypoint` → launcher prepend) is untouched. We only stop forcing *image* entrypoints through it when there's no command, letting `WithImageConfig` hand containerd the exec-form argv directly.

Bonus rodent: the old flatten would let the outer shell expand `$PORT` inside an exec-form image CMD (`CMD ["srv","--port=$PORT"]` became `--port=8080`), which `docker run` never does. So exec form now actually matches Docker semantics instead of quietly diverging from them.

tl;dr: no landmine. The shell model stays for the cases it was built for.

PR #953, opened 2026-07-22
Add MySQL credential rotation

phinze:
With PostgreSQL rotation landed, MySQL was the natural next engine, and it turned out to lean on the same two pieces of groundwork the Postgres work needed.

MySQL is shaped a little differently, though. Dedicated Postgres had a single role that was both the superuser and the credential the app connected with, so rotation was one live ALTER. MySQL always keeps root as a separate admin and gives the app its own non-root user, so both the shared and dedicated variants actually look like the shared-Postgres case: a per-app "user" credential the app connects with, and an admin "root" credential it never sees. The nice consequence is that every rotation authenticates as root, so we never need the app user's own password to make the change.

Two provisioning gaps had to close first. The shared server derived its data disk name from a hash of the root password, exactly the landmine the Postgres PR defused, so rotating root would have moved the disk out from under the data. The first commit stores an explicit disk_name instead and keeps the old password-derived name only to resolve and back-fill pre-existing servers. It also persists the app's username and database on dedicated associations, which previously recorded only the backing server ref, so rotation knows what to ALTER. Both schema additions are additive, so older associations keep working via a fallback to the app's live config.

The second commit is the rotation itself: user and root for both variants, four flows sharing one set of root-authenticated helpers. The per-app user rotation redeploys the app onto the new secret; root rotation records the new password on the server entity and redeploys nothing. Shared root backfills the decoupled disk name before it touches anything, and every flow is idempotent by connecting with both the old and new root passwords so a crashed retry converges.

Testing follows the pattern we settled on with Postgres: saga-ordering unit tests, plus an integration test that stands up real shared and dedicated MySQL servers and rotates every credential, proving at the wire that the new password authenticates and the old one is rejected, with the shared data disk staying put across a root rotation. All four flows run as subtests under one coordinator so they don't fight over the fixed test port. Beyond the automated tests, all four flows were also smoke-tested end to end against a live dev cluster through the CLI, including the app redeploy and disk-stability checks.

Part of MIR-1355.


pkg/addon/mysql/rotate.go
```diff
@@ -0,0 +1,689 @@
...
+
+type AlterSharedRootPasswordOut struct {
+	// Edge producer: gate the entity update on the engine actually changing.
+	RootAltered bool `saga:"root_altered"`
+}
+
+func AlterSharedRootPassword(ctx context.Context, in AlterSharedRootPasswordIn) (AlterSharedRootPasswordOut, error) {
+	// Try the recorded (old) password and the target; a retry after a crash may
+	// find the engine already on the new one.
+	db, err := connectAsRootTrying(ctx, in.SharedServiceHost, in.SharedRootPassword, in.RootNewPassword)
```
evanphx, line 274: 👏 Trying old-then-new here is the subtle bit that makes root rotation crash-safe — a retry converges whether the engine is still on the old password or already moved to the new one. Easy to miss, would've been easy to get wrong.

pkg/addon/mysql/rotate.go
```diff
@@ -0,0 +1,689 @@
...
+		data.Decode(in.AssocEntity)
+	}
+	user, database := data.Username, data.DatabaseName
+
+	var assoc addon_v1alpha.AddonAssociation
+	if in.AssocEntity != nil {
+		assoc.Decode(in.AssocEntity)
+	}
+
+	// Legacy associations predate these attrs; recover conn info from the app's
```
evanphx, line 455: Non-blocking: this legacy fallback (and the `bestEffortOldUserPassword` rollback restore) is the trickiest new logic, but nothing exercises it — the integration test builds associations that already carry the attrs and have no app ref, so it always takes the fast path, and the unit tests are ordering-only. Everything that *runs* is verified at the wire; these two branches just ride along untested. Might be worth a follow-up to cover a legacy-shaped association (attrs absent, app ref present) end to end.

evanphx (approved):
Approved! This one came out really clean — happy with it.

One non-blocking note inline about test coverage; nothing that should hold it up.

--e+🤖

pkg/addon/mysql/myutil.go
```diff
@@ -55,6 +55,44 @@ func connectAsRoot(ctx context.Context, host string, password string) (*sql.DB,
...
+	}
+	return nil, lastErr
+}
+
+// alterMysqlUserPassword changes a user's password. Works for both a per-app
+// user and root (all accounts here are addressed at the '%' host). Existing
+// connections stay authenticated; only new connections need the new password.
+func alterMysqlUserPassword(ctx context.Context, db *sql.DB, username, password string) error {
+	_, err := db.ExecContext(ctx,
+		fmt.Sprintf("ALTER USER %s@'%%' IDENTIFIED BY '%s'",
```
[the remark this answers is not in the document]
phinze: Tried this — MySQL's `ALTER USER` rejects a bound parameter for the password (`IDENTIFIED BY ?` → `Error 1064`, verified against MySQL 8: the driver sends the `?` server-side and the parser rejects it), so the placeholder isn't available for this statement. Keeping the manual escaping, which is safe in our case: rotation secrets are generated alphanumeric tokens (no backslashes or single-quotes), so `escapeMysqlString` yields a correct literal regardless of `NO_BACKSLASH_ESCAPES`. The escaping is defensive; the real inputs never hit the fragile path.

pkg/addon/mysql/rotate.go
```diff
@@ -0,0 +1,689 @@
...
+		if err := fw.EC.Patch(ctx, in.SharedServerRef, 0,
+			entity.String(addon_v1alpha.MysqlServerDiskNameId, legacy),
+		); err != nil {
+			return BackfillRootDiskNameOut{}, fmt.Errorf("backfilling disk name before root rotation: %w", err)
+		}
+	}
+	return BackfillRootDiskNameOut{RootDiskEnsured: true}, nil
+}
+
+func UndoBackfillRootDiskName(ctx context.Context, in BackfillRootDiskNameIn, out BackfillRootDiskNameOut) error {
```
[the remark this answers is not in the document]
phinze: Added the note — the `disk_name` backfill is a one-way migration ratchet (it decouples the disk identity from the rotating password), so the undo is intentionally a no-op.

pkg/addon/mysql/rotate.go
phinze: Fair catch. You're right that the active-config fallback and the best-effort rollback restore both ride along untested — the integration test always populates the attrs, so it takes the fast path every time. Same gap lives in the dedicated-Postgres path (#952), which shares the pattern. I'll do a follow-up that stands up a legacy-shaped association (attrs absent + a deployed app carrying config) and rotates it end to end, covering both engines. Not gating this on it, per your note.

PR #961, opened 2026-07-23
Cover the legacy active-config fallback in dedicated addon rotation

phinze:
Evan flagged on #953 that our dedicated rotation tests never exercised the legacy fallback. Dedicated Postgres and MySQL read the app's username and database off the association attrs, but associations predating those attrs fall back to the app's active ConfigVersion, and the integration tests always populate the attrs so they take the fast path every time. I promised a follow-up covering both engines, and this is it.

Each engine gets one end-to-end case for the legacy shape: provision a real dedicated server, then build an association with the attrs stripped but pointing at a real app whose active config carries the connection vars. That forces the fallback. The test rotates and checks both halves, that user and database resolve from the active config (only possible via the fallback), and that the rotation still lands at the wire with the new credential authenticating and the old one rejected.

Follows up on #953. Part of MIR-1355.

### DIFF
--- a/pkg/addon/mysql/rotate_integration_test.go
+++ b/pkg/addon/mysql/rotate_integration_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"miren.dev/runtime/api/addon/addon_v1alpha"
+	"miren.dev/runtime/api/core/core_v1alpha"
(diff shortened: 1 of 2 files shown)


phinze:
@coderabbitai review

evanphx (approved):

PR #968, opened 2026-07-24
Make `miren login` ephemeral by default

evanphx:
`miren login` registered a brand new ed25519 key with the cloud on every single login. That persistent key mechanism is genuinely nice for long-lived sessions, but it means every routine login writes a new key server-side, and the vast majority of users never need that. It's a lot of backend load for a feature most people don't use.

The cloud already grew the other half of the answer: the device flow returns a refresh token now, and `/auth/refresh` rotates a pair. So the client just needs to be able to hold tokens instead of a key. Login now stores the access and refresh token it already receives and renews them in the background. `--persistent-key` opts back into the old dance if you want it, and an identity that already has a key keeps it on re-login, so nobody's existing setup changes under them.

The subtle part, and the bit most worth a reviewer's eye, is that the refresh endpoint consumes the token you hand it. Two `miren` processes refreshing at the same moment would leave one holding a spent token and a confusing "access denied" for a user who is perfectly logged in. So refreshes serialize on a file lock and re-read the identity from disk inside it, which turns the loser of a race into a cheap no-op rather than a 401. In the same spirit, only a definitive 401 counts as "session expired". A 5xx or a dropped connection keeps the existing tokens, because mapping a cloud blip to a logout would stampede every CLI user into re-authenticating at once, which is exactly the load we're trying to shed.

Getting there needed a refactor first. Eight call sites had each grown their own copy of "load the keypair, mint a JWT, use it as a bearer token", and several outright rejected any identity that wasn't a keypair. Flipping the default without touching them would have broken `whoami`, `doctor`, `cluster switch` and `cluster add` for anyone logging in fresh, so the first commit routes them all through one seam.

A couple of other things fell out along the way. Logout now revokes the refresh token before deleting the local files (best effort, it never blocks logout), since a copied config would otherwise keep renewing for a week after you thought you'd logged out. A 401 also stopped telling people to go ask their administrator for access when the real problem is usually just an expired session. And two incidental bugs got fixed: `SetLeafConfig` was silently dropping the `Keys` map when building its in-memory view, and leaf configs were written non-atomically, which is a torn read waiting to happen once they're rewritten on every token refresh.

Closes MIR-1385

### DIFF
--- a/cli/commands/login.go
+++ b/cli/commands/login.go
@@ -11,7 +11,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strings"
 	"time"
 
 	"miren.dev/runtime/clientconfig"
@@ -43,8 +42,18 @@ type DeviceFlowExchangeResponse struct {
 	Error            string `json:"error,omitempty"`
 	ErrorDescription string `json:"error_description,omitempty"`
 	AccessToken      string `json:"access_token,omitempty"`
+	RefreshToken     string `json:"refresh_token,omitempty"`
 	TokenType        string `json:"token_type,omitempty"`
 	ExpiresIn        int    `json:"expires_in,omitempty"`
+}
+
+// deviceFlowTokens carries the credentials returned by a successful device-flow
+// exchange. The refresh token is present only when the cloud supports ephemeral
+// login; older clouds return an empty RefreshToken.
+type deviceFlowTokens struct {
+	AccessToken  string
+	RefreshToken string
+	ExpiresIn    int
 }
 
 // BeginKeyRegistrationRequest represents the request to begin key registration
@@ -113,21 +122,22 @@ func getOrCreateKey(ctx *Context, keyName string) (*cloudauth.KeyPair, error) {
 
 // LoginWithDefaults runs the login flow with default settings
 func LoginWithDefaults(ctx *Context) error {
-	return login(ctx, "https://miren.cloud", "cloud", "miren-cli", false, false)
+	return login(ctx, "https://miren.cloud", "cloud", "miren-cli", false, false, false)
 }
 
 // Login authenticates with miren.cloud using device flow
 func Login(ctx *Context, opts struct {
-	CloudURL     string `short:"u" long:"url" description:"Cloud URL" default:"https://miren.cloud"`
-	IdentityName string `short:"i" long:"identity" description:"Name for this identity in config" default:"cloud"`
-	KeyName      string `short:"k" long:"key-name" description:"Name for the authentication key" default:"miren-cli"`
-	NoSave       bool   `long:"no-save" description:"Don't save credentials to config file"`
-	Force        bool   `short:"f" long:"force" description:"Overwrite existing identity without prompting"`
+	CloudURL      string `short:"u" long:"url" description:"Cloud URL" default:"https://miren.cloud"`
+	IdentityName  string `short:"i" long:"identity" description:"Name for this identity in config" default:"cloud"`
+	KeyName       string `short:"k" long:"key-name" description:"Name for the authentication key" default:"miren-cli"`
+	NoSave        bool   `long:"no-save" description:"Don't save credentials to config file"`
+	Force         bool   `short:"f" long:"force" description:"Overwrite existing identity without prompting"`
+	PersistentKey bool   `long:"persistent-key" description:"Register a persistent key instead of the default ephemeral token login"`
 }) error {
-	return login(ctx, opts.CloudURL, opts.IdentityName, opts.KeyName, opts.NoSave, opts.Force)
+	return login(ctx, opts.CloudURL, opts.IdentityName, opts.KeyName, opts.NoSave, opts.Force, opts.PersistentKey)
 }
 
-func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force bool) error {
+func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force, persistentKey bool) error {
 	// Check for existing identity (unless forcing or not saving)
 	if !force && !noSave {
 		config, err := clientconfig.LoadConfig()
@@ -185,6 +195,19 @@ func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force b
 		}
 	}
 
+	// Decide whether to use the persistent-key flow. Ephemeral token login is the
+	// default; --persistent-key opts back in. An existing keypair identity keeps
+	// its key on re-login so we never silently orphan a registered server-side key.
+	usePersistentKey := persistentKey
+	if !usePersistentKey {
+		if existing, err := clientconfig.LoadConfig(); err == nil && existing != nil {
+			if id, err := existing.GetIdentity(identityName); err == nil && id != nil && id.Type == "keypair" {
+				usePersistentKey = true
+				ctx.Info("Identity '%s' already uses a persistent key; keeping it.", identityName)
+			}
+		}
+	}
+
 	// Initialize device flow
 	ctx.Info("Initiating device flow authentication...")
 	initResp, err := initiateDeviceFlow(cloudURL)
@@ -226,7 +249,7 @@ func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force b
 		pollInterval = time.Duration(initResp.PollingInterval) * time.Second
 	}
 
-	token, err := pollForToken(ctx, cloudURL, initResp.DeviceCode, pollInterval, timeout, func(status string) {
+	tokens, err := pollForToken(ctx, cloudURL, initResp.DeviceCode, pollInterval, timeout, func(status string) {
 		if status == "pending" {
 			fmt.Print(".")
 		}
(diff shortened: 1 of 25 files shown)


clientconfig/token.go
```diff
@@ -0,0 +1,355 @@
...
+		pair, err := refreshTokenPair(ctx, issuer, identity.RefreshToken)
+		if err != nil {
+			return "", err
+		}
+		return pair.AccessToken, nil
+	}
+
+	source := c.GetIdentitySource(name)
+	if source == "" {
+		return "", fmt.Errorf("cannot locate config file for identity %q", name)
```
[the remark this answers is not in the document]
evanphx: Good catch on the message being opaque, improved that. One correction on the premise though: `GetIdentitySource` already falls back to `c.sourcePath` for identities in the main `clientconfig.yaml`, so main-config identities do resolve and refresh fine. The `""` case is narrower than it looks: it means the identity has no backing file at all (an in-memory config that was never loaded from disk). Since a refresh rotates the pair and needs somewhere to write the new one, that case genuinely cannot proceed, so it stays an error, but the message now explains why and points at `miren login`.

cli/commands/logout.go
```diff
@@ -112,6 +112,17 @@ func Logout(ctx *Context, opts struct {
 		}
 	}
 
+	// For ephemeral token identities, best-effort revoke the refresh token so a
+	// leaked ~/.config copy can't keep renewing after logout. Never let a failed
+	// revoke block the local logout.
+	if identity.Type == "token" && identity.RefreshToken != "" {
+		if err := clientconfig.RevokeRefreshToken(ctx, identity.Issuer, identity.Token, identity.RefreshToken); err != nil {
```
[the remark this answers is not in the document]
evanphx: Good catch, this was the difference between the feature working and being decorative. Fixed rather than just documented: logout now asks `TokenForIdentity` for a live access token first (which refreshes and persists if the stored one is stale), then re-reads the identity so it revokes the refresh token that is actually current rather than the one the refresh just spent. It also only reports success when a request genuinely succeeded, and `RevokeRefreshToken` now returns an error instead of `nil` when it has no bearer to authorize with, so we can no longer claim a revocation that never happened. Still fully best-effort: every failure path warns and continues with the local logout.

cli/commands/login.go
```diff
@@ -113,21 +122,22 @@ func getOrCreateKey(ctx *Context, keyName string) (*cloudauth.KeyPair, error) {
...
-	IdentityName string `short:"i" long:"identity" description:"Name for this identity in config" default:"cloud"`
-	KeyName      string `short:"k" long:"key-name" description:"Name for the authentication key" default:"miren-cli"`
-	NoSave       bool   `long:"no-save" description:"Don't save credentials to config file"`
-	Force        bool   `short:"f" long:"force" description:"Overwrite existing identity without prompting"`
+	CloudURL      string `short:"u" long:"url" description:"Cloud URL" default:"https://miren.cloud"`
+	IdentityName  string `short:"i" long:"identity" description:"Name for this identity in config" default:"cloud"`
+	KeyName       string `short:"k" long:"key-name" description:"Name for the authentication key" default:"miren-cli"`
+	NoSave        bool   `long:"no-save" description:"Don't save credentials to config file"`
+	Force         bool   `short:"f" long:"force" description:"Overwrite existing identity without prompting"`
+	PersistentKey bool   `long:"persistent-key" description:"Register a persistent key instead of the default ephemeral token login"`
```
phinze: Honestly `ephemeral` doesn't quite land as the *best* word for the concept for me generally — but that's an internal-naming quibble and I don't much care. Where I *do* care is right here, the one place it leaks to a user: it collides with the ephemeral-deploy / ephemeral-version concept, a totally different animal. `renewable` sidesteps it — `"...instead of the default renewable token login"` — then regen docs (the generated `login.md` carries the same string). One-liner.

clientconfig/clientconfig.go
```diff
@@ -22,12 +22,14 @@ const (
 
 // IdentityConfig holds authentication credentials that can be used across clusters
 type IdentityConfig struct {
-	Type       string `yaml:"type"`                  // Type of identity: "keypair", "certificate", etc.
-	Issuer     string `yaml:"issuer,omitempty"`      // The auth server that issued this identity (e.g., "https://miren.cloud")
-	KeyRef     string `yaml:"key_ref,omitempty"`     // Reference to a key in the Keys section (for keypair auth)
-	PrivateKey string `yaml:"private_key,omitempty"` // PEM encoded private key (for keypair auth, deprecated - use KeyRef)
-	ClientCert string `yaml:"client_cert,omitempty"` // PEM encoded client certificate (for cert auth)
-	ClientKey  string `yaml:"client_key,omitempty"`  // PEM encoded client key (for cert auth)
+	Type         string `yaml:"type"`                    // Type of identity: "keypair", "token", "certificate", etc.
```
phinze: Nose-wrinkle, not for this PR: `Type` is a bare `string`, so the (enabled, strict) `exhaustive` linter can't watch these switches. Adding `token` meant hand-auditing ~8 call sites for exactly the missing-arm bug a `type IdentityType string` + consts would catch for free. Parked on MIR-891 with the rest of the identity stuff.

clientconfig/token.go
```diff
@@ -0,0 +1,369 @@
...
+//   - token:   returns the cached access token if still fresh; otherwise refreshes
+//     it via /auth/refresh under a cross-process lock and persists the rotated pair.
+//   - certificate: returns ErrNoBearerToken (the caller supplies a client cert).
+//
+// name identifies which identity leaf to write rotated tokens back to. Pass ""
+// for an anonymous/in-memory identity (e.g. during login before it is named):
+// a refresh is then used for this call only and not persisted.
+//
+// fallbackHost is used as the issuer when the identity carries no Issuer.
+func (c *Config) TokenForIdentity(ctx context.Context, name string, identity *IdentityConfig, fallbackHost string) (string, error) {
```
phinze, line 165: Same smell: a loaded identity comes back half-built, so every caller finishes it against `fallbackHost` — which is really `fallbackIssuer` (everyone hands it an issuer URL, and it only means anything for the legacy cert path). Cleaner world hydrates `Issuer` at the load seam and drops the param. MIR-891 sweep, not now.

cli/commands/login.go
```diff
@@ -113,21 +122,22 @@ func getOrCreateKey(ctx *Context, keyName string) (*cloudauth.KeyPair, error) {
...
 }) error {
-	return login(ctx, opts.CloudURL, opts.IdentityName, opts.KeyName, opts.NoSave, opts.Force)
+	return login(ctx, opts.CloudURL, opts.IdentityName, opts.KeyName, opts.NoSave, opts.Force, opts.PersistentKey)
 }
 
-func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force bool) error {
+func login(ctx *Context, cloudURL, identityName, keyName string, noSave, force, persistentKey bool) error {
```
phinze, line 140: While noticing things: three positional bools makes the call site read `login(ctx, url, name, key, false, false, false)`. Options-struct someday (the flag struct one level up basically already is one). Same bucket, leave it.

cli/commands/logout.go
```diff
@@ -112,6 +112,13 @@ func Logout(ctx *Context, opts struct {
 		}
 	}
 
+	// For ephemeral token identities, best-effort revoke the refresh token so a
+	// leaked ~/.config copy can't keep renewing after logout. Never let a failed
+	// revoke block the local logout.
+	if identity.Type == "token" && identity.RefreshToken != "" {
+		revokeIdentitySession(ctx, cfg, identityName, identity)
```
phinze, line 119: Behavioral heads-up for whenever: this revokes + deletes the identity but leaves clusters still pointing at it, so using one before re-login gives a cryptic `identity not found` instead of a "session expired, log back in" nudge. Pre-existing and self-heals on re-login, but the ephemeral default makes logout/login routine enough that it's more reachable now. Not blocking.

phinze (approved):
LGTM, nice work — the bits that had to be careful (the single-use refresh race, only-a-401-logs-you-out) are, and the client lines up with the real cloud refresh handler on all of it: the 401/5xx split, single-use rotation, the sliding 7-day window. Bot catches look properly fixed too.

Dropped a handful of inline notes as I passed through. Only the `ephemeral` naming one is actually asking for anything — the rest are just spots where the identity code's got a bit of an _ODOUR_ (under-typed, half-hydrated, a row of positional bools). None of it bites today and it's all right in the path of the RBAC / App-Auth pass (MIR-891), so leave it for that sweep — I'm flagging in situ mostly since it's what I'd be mumbling out loud if we were talking through this live.

`--p+🤖`

clientconfig/clientconfig.go
evanphx: Good call, on it.

cli/commands/logout.go
phinze: Filed it for real so it doesn't just live in this thread: MIR-1470. Still not blocking here.

cli/commands/login.go
evanphx: Done — flag help now reads "renewable token login", and regenerated login.md to match. Left the internal naming alone per your note.

clientconfig/clientconfig.go
evanphx: Went ahead and did this one now since it's right where the bug would've bitten. `Type` is an `IdentityType` with consts, and I confirmed exhaustive now fails the build if a switch drops an arm (pulled the cert case out to check). The half-hydrated `Issuer` and the positional bools I'm leaving for the MIR-891 sweep as you suggested.

clientconfig/token.go
evanphx: Agreed on both — it's really `fallbackIssuer`, and hydrating `Issuer` at the load seam would let us drop the param entirely. Leaving it for the MIR-891 sweep.

cli/commands/login.go
evanphx: Yeah. The flag struct one level up is basically that options struct already; collapsing to it belongs with the MIR-891 cleanup. Left as-is here.

cli/commands/logout.go
evanphx: Good flag, and agreed the renewable default makes it more reachable. Tracking in MIR-1470, not touching it here.

PR #970, opened 2026-07-24
fix(disk): undelete a disk through mount-then-ready ordering (MIR-1469)

phinze:
`TestDiskUndelete` was flaking: the lease it puts on a just-undeleted disk would sometimes go straight to `failed` and never recover, timing out after 60s, then pass on re-run.

Turns out `disk undelete` was getting ahead of itself. It wrote the volume `DV_READY` and the disk `PROVISIONED` up front, but those are just entity attributes. The runner still has to register and mount the volume asynchronously, and in that window the disk looks leasable while nothing's actually mounted. A lease created then finds no volume in the runner's state, latches a terminal `DM_ERROR`, and gets stuck `FAILED` forever (failed leases never get re-driven). Reproduced it locally at iteration 44, and the runner log showed the volume mounting ~18s too late.

Fix is to stop lying about readiness. Undelete now creates the volume `DV_PENDING` and leaves the disk `PROVISIONING`, so it flows through the same mount-then-`DV_READY` path a normal create uses, and the disk only reports provisioned once the volume is genuinely mounted. To reuse that path safely, `createVolume` skips reimaging when a `disk.img` is already there instead of truncating it (also closes a latent data-loss hole on the error-recreation path). And the test now fails fast on a `failed` lease so a regression is legible instead of a 60s hang.

Verified: 150 clean iterations (flaked at 44 before), normal disk create/persist still green, lint clean.

Closes MIR-1469

### DIFF
--- a/blackbox/disk_undelete_test.go
+++ b/blackbox/disk_undelete_test.go
@@ -105,7 +105,11 @@ func TestDiskUndelete(t *testing.T) {
 	leaseID := extractLeaseID(t, r)
 	t.Logf("Lease ID: %s", leaseID)
 
-	// Wait for lease to become bound
+	// Wait for lease to become bound. A lease that lands in "failed" is
+	// terminal (the controller never re-drives it), so treat it as fatal and
+	// fail fast with the status output rather than spinning until timeout —
+	// otherwise a regression reads as an opaque 60s hang instead of a lease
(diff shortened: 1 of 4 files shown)


phinze:
@coderabbitai review

evanphx (approved):

PR #980, opened 2026-07-25
stackbuild: detect Next.js and run its production build

evanphx:
> 🤖 Authored autonomously by Claude Code (no human co-author). Opened via the `/bot-friendly-pr` skill; a human reviews before it lands.

## What

Teaches the Node stack to recognize a Next.js app and produce a working image for it.

Before this, the Node stack ran `npm install` and copied the source but never ran a build step, so a Next.js app shipped with no compiled `.next/` and fell over at runtime under `next start`.

## How

Follows the pattern the Ruby stack already uses for Rails — detect the framework, run its build in-image, serve it with the framework's command:

- **Detect**: a `next` dependency in `package.json` (deps or devDeps) sets `hasNext` during `Init`, emitting a `framework: nextjs` event — same shape as `ruby.go` detecting Rails and `python.go` detecting Django/FastAPI.
- **Build**: after the app is copied, run the build (`next build` via the app's `build` script) with user env vars injected, so build-time values like `NEXT_PUBLIC_*` are inlined. This mirrors the Ruby stack's `assets:precompile` step (which likewise injects `opts.EnvVars`).
- **Serve**: `next start -p $PORT`, mirroring `rails server -p $PORT`.

Scoped to Next.js's default output. `output: 'standalone'` and static `export` are follow-ups — same altitude as the Ruby stack not enumerating every Rack server. Plain Node apps are unaffected: no build step runs and the web command is unchanged.

## Verification

`go test ./pkg/stackbuild/` — new `TestNodeNextjs` covers detection (deps + devDeps), the npm/yarn build command, the web command, and that a plain Express app is untouched. The Docker-gated `TestNode`/`TestBun` build tests also ran green, confirming the existing Node/Bun build paths still work. Package builds and vets clean.

Not covered locally: a full end-to-end `next build` image build (needs a real Next fixture through buildkit) — worth adding as a Docker-gated test in a follow-up.

Part of MIR-1462

### DIFF
--- a/pkg/stackbuild/node.go
+++ b/pkg/stackbuild/node.go
@@ -75,6 +75,7 @@ type NodeStack struct {
 	packageManager nodePackageManager
 	scripts        map[string]string
 	entryPoint     string
+	hasNext        bool
 
 	// Parsed dependencies from package.json
 	dependencies    map[string]string
@@ -129,6 +130,13 @@ func (s *NodeStack) Init(opts BuildOptions) {
 		if _, ok := s.scripts["build"]; ok {
 			s.Event("script", "build", "npm build script detected")
 		}
+	}
+
+	// Detect Next.js so we can run its production build and serve it with
+	// `next start`, mirroring how the Ruby stack special-cases Rails.
+	if s.hasDependency("next") {
+		s.hasNext = true
+		s.Event("framework", "nextjs", "Next.js framework detected")
 	}
 
 	// Check for common entry points and store the first one found
@@ -203,9 +211,47 @@ func (s *NodeStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, erro
 
 	state = h.copyApp(state, localCtx)
 
+	// Run the framework build (e.g. `next build`) after the app is copied.
+	// Inject user env vars so build-time values (e.g. NEXT_PUBLIC_* that Next
+	// inlines at build time) are available, mirroring how the Ruby stack
+	// injects env vars for asset precompilation.
+	if buildCmd := s.frameworkBuildCommand(); buildCmd != "" {
+		build := state.Dir("/app")
+		for k, v := range opts.EnvVars {
+			build = build.AddEnv(k, v)
+		}
+		state = build.Run(
+			llb.Shlex(buildCmd),
+			llb.WithCustomName("[phase] Building Next.js application"),
+		).Root()
+	}
+
 	state = s.applyOnBuild(state, opts)
 
 	return &state, nil
+}
+
+// hasDependency reports whether the package.json lists the given package in
+// either dependencies or devDependencies.
+func (s *NodeStack) hasDependency(name string) bool {
+	if _, ok := s.dependencies[name]; ok {
+		return true
+	}
+	_, ok := s.devDependencies[name]
+	return ok
+}
+
+// frameworkBuildCommand returns the in-image build command for a detected
+// framework, or "" when no build step is needed. Next.js must be compiled with
+// `next build` before it can be served; a plain Node app is copied as-is.
+func (s *NodeStack) frameworkBuildCommand() string {
+	if !s.hasNext {
+		return ""
+	}
+	if s.packageManager == nodePkgYarn {
+		return "yarn build"
+	}
+	return "npm run build"
 }
 
 func (s *NodeStack) parsePackageJSON() {
@@ -229,6 +275,11 @@ func (s *NodeStack) parsePackageJSON() {
 }
 
 func (s *NodeStack) WebCommand() string {
+	// Next.js is served with `next start`; bind to the platform-provided port.
+	if s.hasNext {
+		return "npx next start -p $PORT"
+	}
+
 	// Determine the runner based on detected package manager
 	var runner string
 	if s.packageManager == nodePkgYarn {
--- a/pkg/stackbuild/node.go
+++ b/pkg/stackbuild/node.go
@@ -224,6 +224,15 @@ func (s *NodeStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, erro
 			llb.Shlex(buildCmd),
 			llb.WithCustomName("[phase] Building Next.js application"),
 		).Root()
+
+		// The build runs as root, so the generated tree would be root-owned while
+		// the image runs as the app user. Next writes its ISR, image, and prerender
+		// caches beneath .next at runtime, so hand the tree over before export or
+		// those writes fail with EACCES.
+		state = state.Dir("/app").Run(
+			llb.Shlex("chown -R app:app /app/.next"),
+			llb.WithCustomName("[phase] Fixing Next.js build permissions"),
+		).Root()
 	}
 
 	state = s.applyOnBuild(state, opts)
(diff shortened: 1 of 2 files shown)


evanphx:
🤖 Thanks for the review. Addressed **Concern 1** in 04d2d862 — `WebCommand()` now resolves `next` through the detected package manager (`yarn next start -p \$PORT` for yarn, `npx next start -p \$PORT` for npm), matching `frameworkBuildCommand()`. The yarn test case now expects `yarn next start`, so it exercises a distinct path from the npm cases (which also closes the "Minor" point). CI is green.

Leaving two items for human review rather than expanding this autonomous change:
- **Concern 2 (LLB-level integration test for the Next build step):** agreed it's the right next test; it needs a real Next.js fixture through buildkit, which is heavier than this change warrants. Flagged as a follow-up in the PR description.
- **CodeRabbit's nitpick (persistent cache mount for `.next/cache`):** a reasonable perf optimization with a small design choice attached — deferring to the reviewer.

evanphx:
🤖 Addressed the second concern in e37d9d1b: `TestNodeNextjs` now calls `GenerateLLB()` and marshals the graph (with an env var injected) for each case, so the Next.js build-step wiring — the AddEnv loop and `build.Run` — is exercised without needing Docker. CI is green. Re-reviewing now.

evanphx:
🤖 Settled and ready for human review. Summary of the automated round-trip:

- **CI:** green on the latest commit (e37d9d1b) — 21 checks pass, 1 skipped.
- **biscuit (miren-code-agent):** both concerns addressed and confirmed ("happy to close") — package-manager-aware `next start`, and `GenerateLLB`/marshal coverage of the build step.
- **CodeRabbit:** reviewed the first commit; its one item is a trivial perf nitpick (persistent cache mount for `.next/cache`), against install code these fixes didn't touch — **left for the human reviewer** as a conscious call, not addressed.

Still a **draft** — a human decides readiness and merge. This is where the bot's autonomous pass ends.

pkg/stackbuild/node.go
```diff
@@ -229,6 +275,16 @@ func (s *NodeStack) parsePackageJSON() {
 }
 
 func (s *NodeStack) WebCommand() string {
+	// Next.js is served with `next start`; bind to the platform-provided port.
+	// Resolve the local `next` binary through the detected package manager so a
+	// yarn project doesn't shell out to npm's npx (matches frameworkBuildCommand).
+	if s.hasNext {
```
phinze: We need to keep honoring an explicit `start` script here, using a synthesized `next start` only as the fallback. As written, an app whose start script launches a [custom Next server](https://nextjs.org/docs/app/guides/custom-server), runs setup work, or supplies extra flags silently gets the built-in server instead. Next reads `PORT` from the environment, so going through `npm run start` or `yarn start` still binds the platform port.

pkg/stackbuild/node.go
```diff
@@ -203,11 +211,49 @@ func (s *NodeStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, erro
...
+	// Run the framework build (e.g. `next build`) after the app is copied.
+	// Inject user env vars so build-time values (e.g. NEXT_PUBLIC_* that Next
+	// inlines at build time) are available, mirroring how the Ruby stack
+	// injects env vars for asset precompilation.
+	if buildCmd := s.frameworkBuildCommand(); buildCmd != "" {
+		build := state.Dir("/app")
+		for k, v := range opts.EnvVars {
+			build = build.AddEnv(k, v)
+		}
+		state = build.Run(
```
phinze, line 223: This build runs as root, so it leaves `.next` root-owned in the exported filesystem while the image itself runs as UID 2010. Next writes its ISR, image, and prerender caches beneath `.next` at runtime, which can turn those requests into `EACCES`. We need to chown the generated tree to the app user before export, or otherwise give the runtime cache a writable home.

phinze (changes requested):
Requesting changes on two runtime problems inline. The startup-command path breaks custom Next servers and wrappers, and the build can leave Next’s runtime caches unwritable. Both need resolving before this is safe to land.

--p+🤖

pkg/stackbuild/node.go
evanphx: 🤖 Good catch — fixed in 0e08f257. `WebCommand()` no longer short-circuits on Next.js detection: an explicit `start` script now wins for every stack, and the synthesized `next start -p $PORT` is only reached when there's no `start` script. As you note, `next start` reads `PORT` from the environment, so `npm run start` / `yarn start` still binds the platform port. Added test cases for a Next app with a custom `start` script (stays on `npm run start`) and a yarn Next app without one (falls back to `yarn next start -p $PORT`).

pkg/stackbuild/node.go
evanphx: 🤖 Fixed in 0e08f257 — added a `chown -R app:app /app/.next` step right after the build so the tree is handed to the app user before export. Went targeted rather than reusing the `chownApp` helper's `chown -R /app`, since `copyApp` already lands the source as 2010:2011 and a recursive chown over `node_modules` is a lot of churn for no gain.

evanphx:
🤖 On CodeRabbit's nitpick (persistent cache mount for `.next/cache` in `pkg/stackbuild/node.go`) — skipping this one. A persistent cache mount at `/app/.next/cache` isn't exported into the image, and `.next/cache` is also where Next keeps its *runtime* ISR and image caches — so mounting it as a build cache interacts directly with the ownership fix in this same push rather than being an independent perf win. Worth doing deliberately as its own change if repeat-build time becomes a problem.

pkg/stackbuild/node.go
```diff
@@ -203,11 +211,58 @@ func (s *NodeStack) GenerateLLB(dir string, opts BuildOptions) (*llb.State, erro
...
+			llb.Shlex(buildCmd),
+			llb.WithCustomName("[phase] Building Next.js application"),
+		).Root()
+
+		// The build runs as root, so the generated tree would be root-owned while
+		// the image runs as the app user. Next writes its ISR, image, and prerender
+		// caches beneath .next at runtime, so hand the tree over before export or
+		// those writes fail with EACCES.
+		state = state.Dir("/app").Run(
+			llb.Shlex("chown -R app:app /app/.next"),
```
phinze: `chown` exits non-zero on a missing path, and `.next` isn't guaranteed to be there: an app that sets `distDir` in `next.config.js` builds fine and then dies here with a pretty opaque error. That's a build that works on main and breaks on this branch.

Tolerating the failure would keep that case working. `llb.Shlex` doesn't give us a shell, but the `llb.Args([]string{"/bin/sh", "-c", ...})` form at `stackbuild.go:241` and `:255` does. A distDir app then just keeps the ownership situation it already has, which is no worse than today.

pkg/stackbuild/node.go
```diff
@@ -237,9 +292,30 @@ func (s *NodeStack) WebCommand() string {
...
+		if s.packageManager == nodePkgYarn {
+			return "yarn next start -p $PORT"
+		}
+		return "npx next start -p $PORT"
+	}
+
+	// Check for the remaining web server scripts in order of preference
 	if s.scripts != nil {
-		for _, script := range []string{"start", "serve", "server"} {
+		for _, script := range []string{"serve", "server"} {
```
phinze: Same principle as the `start` fix, one step short: a Next app with no `start` but a `serve` script used to get `npm run serve` and now gets `npx next start`. Rare enough that it may not be worth the two lines, but if explicit scripts win, this one arguably wins too.

phinze (approved):
Approving — this clears my earlier block. Both of those were real and both fixes are right, especially the start-script one.

Two non-blocking notes inline. The chown one has slightly more teeth: it can turn a working build into a failed one. The serve/server one is rare enough to skip if you'd rather.

`--p+🤖`

pkg/stackbuild/node.go
evanphx: 🤖 Good call — fixed in 84a51054. Went with a `-d` guard rather than `|| true` so only the distDir case skips and real failures still surface.

pkg/stackbuild/node.go
evanphx: 🤖 Fair — if explicit scripts win, they win. Fixed in 84a51054, back to one loop.

PR #981, opened 2026-07-25
testutils: make GetFreePort return distinct ports per process

evanphx:
> 🤖 Authored autonomously by Claude Code (no human co-author). Opened via the `/bot-friendly-pr` skill; a human reviews before it lands.

## What

Fixes the `TestEtcdComponentIntegration` port-collision flake (MIR-720) at its root: `testutils.GetFreePort`.

`GetFreePort` finds a free port by binding to `:0` and then **releasing** it before returning the number, so two calls in quick succession could hand back the same port — the OS is free to re-offer the just-freed one. In the reported failure both `clientPort` and `peerPort` came back as `36517`; etcd tried to bind its client and peer URLs to one port, failed with `address already in use`, and the test timed out waiting for readiness.

## How

Remember every port handed out in this process (guarded by a mutex) and skip a repeat, rebinding on the next iteration to rotate to a fresh ephemeral port. This resolves the collision for **all** call sites at once — the etcd test takes two/three ports in a row — rather than patching each call, and preserves the existing "free for both TCP and UDP" guarantee. Bumped the attempt budget 10 → 20 to leave headroom for the extra rejections.

Chose this over the ticket's per-test options (assert-and-retry in the test, or `clientPort+1`) because the bug is in the helper's contract — a "get a free port" helper returning the same port twice — so fixing it there is both more correct and narrower in aggregate.

## Verification

`go test ./pkg/testutils/` — new `TestGetFreePortReturnsDistinctPorts` requests 200 ports and asserts every one is unique; run with `-count=5` to confirm the fix isn't itself flaky. Package builds and vets clean.

The flake is probabilistic (two calls *occasionally* collide), so there's no deterministic red-first test to show; the new test instead pins the distinctness *invariant* the old implementation didn't guarantee. I did not run the full `TestEtcdComponentIntegration` (it needs the etcd integration harness), but the fix is entirely within the port helper it depends on.

Part of MIR-720

### DIFF
--- a/pkg/testutils/port.go
+++ b/pkg/testutils/port.go
@@ -24,17 +24,24 @@ var (
 func GetFreePort(t testing.TB) int {
 	t.Helper()
 
-	// Try a few times to find a port free for both TCP and UDP
+	var duplicates int
+
 	for range 20 {
-		// First, get an available TCP port
-		tcpListener, err := net.Listen("tcp", "127.0.0.1:0")
+		// First, get an available TCP port. Bind the wildcard address, not
+		// loopback: callers such as the etcd component run with host
+		// networking and bind 0.0.0.0, and a port already held on a
+		// non-loopback address (an outbound connection's source port, say)
+		// passes a 127.0.0.1 check but fails a 0.0.0.0 bind. Checking the
+		// wildcard is strictly more conservative, so loopback-only callers
+		// are still safe.
+		tcpListener, err := net.Listen("tcp", "0.0.0.0:0")
 		if err != nil {
 			continue
 		}
 		port := tcpListener.Addr().(*net.TCPAddr).Port
 
 		// Try to bind UDP to the same port
-		udpAddr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: port}
+		udpAddr := &net.UDPAddr{IP: net.IPv4zero, Port: port}
 		udpConn, err := net.ListenUDP("udp", udpAddr)
 		if err != nil {
 			tcpListener.Close()
@@ -52,6 +59,7 @@ func GetFreePort(t testing.TB) int {
 		}
 		handedOutMu.Unlock()
 		if seen {
+			duplicates++
 			tcpListener.Close()
 			udpConn.Close()
 			continue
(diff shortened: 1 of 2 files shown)


evanphx:
🤖 Settled and ready for human review. Automated round-trip summary:

- **CI:** green — 21 checks pass, 1 skipped, 0 failures.
- **biscuit (miren-code-agent):** ✅ ready to merge on the first pass, no blockers, no inline comments. Three non-blocking advisories left for the human reviewer to weigh (not addressed here, since each is either a design choice or pre-existing):
  1. the per-process `handedOutPorts` map is never pruned — unbounded over tens of thousands of calls (biscuit: unlikely to matter for typical suites);
  2. the 20-attempt budget now covers both "UDP unavailable" and "already seen" rejections;
  3. the intra-process dedup doesn't close the inter-process TOCTOU window (pre-existing, not a regression).
- **CodeRabbit:** no review posted within the polling window — appears to be skipping this draft. Not blocking.

Still a **draft** — a human decides readiness and merge. End of the bot's autonomous pass.

phinze (approved):
LGTM, and we pushed one commit onto the branch (`481e2a5c`) rather than leaving it as a comment.

It closes a second route to the same `address already in use`. `GetFreePort` was checking `127.0.0.1:P`, but the etcd component runs host networking and binds `0.0.0.0:P`, and those aren't the same question. With a socket held on `10.50.1.35:P`, `bind(127.0.0.1:P)` succeeds while `bind(0.0.0.0:P)` gets EADDRINUSE, so the helper could hand out a port etcd can't actually take. Pre-existing rather than a regression, but it's the thing that would have kept MIR-720 alive after this landed. The wildcard check is strictly more conservative, so the loopback-only callers don't notice. Same commit splits the two rejection reasons in the `t.Fatalf`, since duplicates and TCP/UDP contention now share one attempt budget and a bare "failed after 20 attempts" would misdirect whoever debugs this next.

Then, having apparently not learned when to stop, we sat and watched 20,000 port binds go by to see what the kernel actually does with them. Two things fell out. Your duplicate diagnosis is real: back-to-back identical ports turned up 2 to 3 times per 20,000 draws, about the rate a flake like this surfaces. And the usable pool is nowhere near the nominal 28k, because `bind(0)` clusters hard. Those 20,000 draws only ever yielded ~6,600 distinct ports, and an instrumented copy of the helper hits the 20-attempt wall at ~4,700 handed out. Still ~20x more headroom than the heaviest caller needs, so nothing to do, just a good deal thinner than the number biscuit quoted.

`--p+🤖`

PR #983, opened 2026-07-28
Let only the volume's owner mount a disk lease

phinze:
`TestDiskUndelete` took down the release job on both attempts of run [30291508637](https://github.com/mirendev/runtime/actions/runs/30291508637), an hour after MIR-1469 was closed as fixed. Chasing it turned up something more interesting than a flaky test.

The schema says `disk_lease.node_id` is required, and it means "the node where the disk is mounted." Nothing enforces that on write, and `debug disk lease` never set it. The controller reads a missing `node_id` as "unassigned, everyone have a go", which for a disk lease can't be right: the mount can only happen where the volume physically lives. So both nodes reconciled the same lease and each created its own `disk_mount` pinned to itself, one of which could never work. Then `getDiskMountForLease` took `values[0]` from an unordered index with no node filter, so whether the lease bound or failed terminally came down to which mount the index handed back first. Locally the good one came back first twice; in CI the broken one did, twice. Same commit, opposite outcome.

It needs two nodes, which is why this only ever appeared in `test-blackbox-distributed`, and that job is skipped on `pull_request`. Every PR was green while main was red, which is a rough combination to debug. So this PR also adds a `ci:distributed` label opt-in, and applies it to itself: because `pull_request` runs the workflow from the PR's own head, the fix gets validated by the very job it fixes, before merge rather than after. A path-based auto-trigger for disk and runner changes would be the stronger version, since it doesn't rely on anyone remembering the label, but GitHub has no native per-job path filter and that's more machinery than belongs on a red-main fix.

#970 neither caused this nor was wrong. It fixed a genuine ordering bug in undelete and closed a real data-loss hole, and it changed the test to fail fast on the first `failed` read where the old loop polled for 60s and could still catch a later `bound`. That's why an intermittent flake became a reliable one. The bug was always there; #970 stopped it hiding.

Getting the fix right took two passes. The first version fixed the reported failure but made the pre-existing broken state deterministic instead of 50/50, and would have flapped bound leases between `BOUND` and `PENDING` on non-owner nodes. Both were caught by an external review pass, and both are now regression tests that fail without the fix.

Verified against a local two-node cluster: the full distributed suite passes (45 tests), and the duplicate mount is gone, with only the volume's owner creating one and the other node staying out entirely.

Two follow-ups fall out of this and are not in scope here: `schema.Required` isn't enforced on write, which is why the malformed lease was storable at all, and `disk_lease.node_id` is arguably redundant with the volume's own node and could be derived rather than stored.

Closes MIR-1469

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,10 @@ name: Test
 
 on:
   pull_request:
+    # Explicit list because naming `types` replaces the default set. `labeled`
+    # is what lets the ci:distributed opt-in below take effect on an open PR;
+    # the other three are the defaults we'd otherwise lose.
+    types: [opened, synchronize, reopened, labeled]
   merge_group:
   workflow_call:
     # Called by release.yml for main/tag builds
@@ -410,7 +414,14 @@ jobs:
     # workflow_call, where github.event_name is the caller's event ('push'), so
     # excluding 'pull_request' runs it there while skipping PRs. It would also
     # run in a merge queue if one were enabled ('merge_group').
-    if: github.event_name != 'pull_request'
+    #
+    # Label a PR `ci:distributed` to opt into it before merging. Worth doing for
+    # anything touching disks, leases, or the runner: this job is the only place
+    # a second node exists, so a multi-node bug is invisible until it is already
+    # on main and blocking the release (MIR-1469).
+    if: >-
+      github.event_name != 'pull_request' ||
+      contains(github.event.pull_request.labels.*.name, 'ci:distributed')
     runs-on: depot-ubuntu-latest
     steps:
     - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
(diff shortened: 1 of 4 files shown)


phinze:
Thanks, three good catches. Two were real and are fixed in 222c2fc:

**Missing `disk_volume` treated as terminal** — correct, and worse than it looked. That branch is pre-existing, but moving the volume lookup ahead of the mount state machine made the window much easier to hit. `handleProvisioned` recreates a missing volume for a PROVISIONED disk on the next pass, so it is now a retry rather than a burned lease.

**Unstamped `disk_volume.NodeId`** — also correct. `Matches("")` is false, so a legacy volume with no node stamp read as "owned by another node" and every node would step aside, stranding the lease. `disk_controller.go:184` already guards for exactly this, which confirms it is a live state. The lease's own pin is now the tiebreaker: pinned to this node we proceed, pinned to nobody we wait, because proceeding on a guess is how two nodes end up fighting over one mount, which is the bug this PR exists to fix.

Both have regression tests.

**Dropping `labeled` from the trigger** — pushing back on this one. It is the mechanism that lets an already-open PR opt in; without it you can only opt in at creation time or by pushing an empty commit, which loses most of the value. The cost you are pointing at is real but close to zero here: 1 of the last 60 PRs in this repo carries any label, and it is this PR's own `ci:distributed`. If label usage picks up, the right fix is gating the other jobs on `github.event.action != 'labeled'` rather than removing the trigger, and that needs care around the `test` summary job that branch protection depends on. Happy to revisit then.

evanphx (approved):

PR #988, opened 2026-07-29
Keep tar extraction inside its destination

phinze:
Build contexts reach `TarFS` before BuildKit gets a chance to isolate them, but the extractor trusted every archive entry name. The release downloader did check entry paths, though its prefix comparison had boundary gaps and archive symlinks could still point outside the release directory.

Both extraction paths now share component-aware containment checks and refuse to write through existing symlink components, including a symlinked extraction root. Release artifacts are assembled from copied binaries and do not need links, so the downloader rejects symlink and hard-link entries outright instead of attempting to reproduce attacker-controlled link graphs.

`--force` now builds the replacement in a sibling staging directory and verifies its binary before touching the current release. Installation moves the old directory aside, promotes staging, and rolls back if promotion fails, so a bad download does not destroy the last working release.

Closes MIR-1475

### DIFF
--- a/cli/commands/download_release.go
+++ b/cli/commands/download_release.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	"miren.dev/runtime/pkg/tarx"
 	"miren.dev/runtime/version"
 )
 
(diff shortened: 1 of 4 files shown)


phinze:
@coderabbitai review

phinze:
@coderabbitai review

phinze:
@coderabbitai review

phinze:
@coderabbitai review

evanphx (approved):

PR #989, opened 2026-07-29
Make WireGuard the only overlay backend

phinze:
RFD-51 made encryption part of the distributed runner design, but runtime still exposed the Flannel backend as a choice. That left `vxlan` available as a supported way to send cross-host sandbox traffic in plaintext, even though every supported Miren kernel can run WireGuard.

Distributed runners are still a Labs feature, and we do not know of anyone relying on VXLAN for a real deployment. This is the right point to remove the escape hatch instead of turning it into compatibility baggage.

This removes the choice instead of merely changing its default. Runtime now compiles and selects only WireGuard, ignores the legacy backend field in saved runner configs, and keeps reporting `wireguard` through the join RPC so older runner binaries land on the safe backend too. Startup still removes `flannel.1`, since an old VXLAN device can leave behind more-specific routes that beat the new encrypted path.

The distributed encryption probe remains the backstop for the actual invariant. Its scratch files are now unique per run so concurrent blackbox jobs cannot interfere with each other.

Closes MIR-1482

### DIFF
--- a/.iso/peers.yml
+++ b/.iso/peers.yml
@@ -13,12 +13,6 @@ peers:
     environment:
       ROLE: coordinator
       MIREN_LABS: distributedrunners
-      # Dev runs WireGuard ahead of the shipped default (still vxlan) so the
-      # overlay-encryption blackbox test has something to assert against, and
-      # so the backend the clusters are migrating to is the one exercised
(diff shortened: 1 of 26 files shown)


evanphx (approved):
Nice quick turnaround!

PR #990, opened 2026-07-29
Add system workload identities

phinze:
Miren already has workload identity for customer sandboxes, but the next internal services we need to protect are reached by Miren’s own code. The sandbox controller pulls from the cluster-local registry, and the telemetry writers ship metrics and logs from distributed runners. Neither fits the sandbox identity model.

Rather than introduce another credential system, this extends the existing issuer with system workload identities. They use the same signing keys and short lifetimes, but carry an explicit `identity_type: system`, a closed system-workload name, and subjects such as `org:X:cluster:Y:system:sandboxcontroller`. Audience verification keeps tokens scoped to the service they were minted for.

Distributed runners still do not receive the signing key. They request tokens from the coordinator over the mTLS channel established during Join, where the coordinator checks that the runner remains registered and restricts which workloads it may request. That gives us a useful revocation boundary even though the underlying runner certificates remain valid until they expire.

The initial workloads are `sandboxcontroller` for MIR-1477 and `telemetrywriter` for MIR-1483. System workload identities deliberately receive no RPC privileges by default. Each internal service must verify the token against its own audience instead of inheriting the RPC authorizer’s permissive fallback.

This PR stops at the shared identity primitives so the registry and telemetry work can proceed independently on top of them.

Part of MIR-1477 and MIR-1483.

### DIFF
--- a/api/runner/rpc.yml
+++ b/api/runner/rpc.yml
@@ -303,6 +303,33 @@ interfaces:
             type: string
             doc: Error message if the drain failed
 
+      - name: IssueSystemWorkloadToken
+        index: 12
+        doc: |
+          Mint an identity token for a Miren system workload running on a
+          distributed runner, which does not hold the cluster signing key. The
+          caller is an mTLS-authenticated runner, and the coordinator restricts
+          which system workloads a runner may request; a runner cannot obtain an
+          identity for a workload it does not run.
+        parameters:
+          - name: system_workload
+            type: string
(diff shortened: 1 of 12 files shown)


phinze:
@coderabbitai review

phinze:
@coderabbitai review

phinze:
@coderabbitai review

phinze:
@coderabbitai review

evanphx (approved):
I LOVE how this came together

PR #994, opened 2026-07-30
Give every cluster a workload identity issuer

phinze:
Whether a cluster got a workload identity issuer depended on whether it had a hostname to anchor one to, either a Miren Cloud registration or a `--dns-names` value. A cluster installed with `--without-cloud` and neither got no issuer at all.

That made internal authentication conditional on external addressability, and those two things aren't actually related. Everything consuming these tokens today is the cluster talking to itself: the cluster-local registry (#991) and the telemetry a distributed runner ships (#992). Both verify in-process against the signing key, which needs no DNS, no reachability, and no publicly trusted certificate. Only federating to an outside party needs any of that.

The cost landed in the wrong place. A `--without-cloud` cluster is precisely the one that can't lean on a cloud VPC firewall, and it's the one we gave the weakest story: `authorizeRegistry` passes requests straight through when there's no issuer, so those clusters run an open registry, and runner telemetry has nothing to authenticate with either. The clusters with the least network protection got the least authentication.

So every cluster gets an issuer. One with no hostname anchors at `https://cluster.local` — a name it already owns internally, rather than a synthetic placeholder. Nothing outside can resolve it, which is the honest signal that such a cluster can't federate until it's given a real name, and is exactly why it costs nothing internally. Give it a hostname later and its tokens become federatable with no other change.

Worth being explicit about the two behavior changes, since neither is invisible.

**Registry auth starts being enforced on clusters that previously had none.** That's the point, but it means a pull or push that fails to present a token turns into a broken deploy rather than a quiet passthrough. Both minting sites (`localRegistryHost` and the buildkit push path) already key off the issuer being present, so they'll now mint, and the blackbox suites exercise that path on a server started without `--dns-names`.

**Sandboxes on hostname-less clusters now receive identity tokens where they previously received none.** They're real and verifiable inside the cluster, and they won't federate to AWS or GCP — which is no worse than the nothing they replace. The docs claimed a bare cluster issues no tokens and that the `MIREN_IDENTITY_*` variables are absent, so they're corrected to separate federation from identity rather than conflating them.

One thing I deliberately didn't do: auto-detect a public IP and use that as the anchor. A publicly routable cluster *could* in principle federate on an IP — our discovery handler keys off `url.Parse(issuerURL).Host` and would serve it fine — but `autocert` can't obtain a publicly trusted certificate for a bare IP, and relying parties vary on whether they'll accept one. Anyone who wants to try can already pass `--dns-names 203.0.113.5`, which is used verbatim. Guessing on their behalf would mean an unstable anchor that changes when the IP does.

Prerequisite for #992, which otherwise regresses telemetry on exactly this cohort.

### DIFF
--- a/cli/commands/server.go
+++ b/cli/commands/server.go
@@ -656,27 +656,42 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
 		ctx.Log.Info("no cluster registration found")
 	}
 
-	// Initialize workload identity issuer for sandbox OIDC tokens
+	// Initialize the workload identity issuer.
+	//
+	// Every cluster gets one. It used to be conditional on having a hostname to
+	// anchor it to, which quietly made internal authentication depend on
+	// external addressability: a cluster installed with --without-cloud and no
+	// --dns-names got no issuer, so its own services had no way to authenticate
+	// to each other and fell back to trusting the network. But the services that
+	// consume these tokens (the cluster-local registry, runner telemetry) all
+	// verify in-process against the signing keys, which needs no DNS and no
+	// publicly trusted certificate. Only federating to an outside party needs
+	// those, and a cluster with no hostname simply has an anchor nobody outside
+	// can resolve.
 	var workloadIssuer *workloadidentity.Issuer
 	{
-		var issuerURL string
-		if cloudAuthConfig.DNSHostname != "" {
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Great improvement, makes perfect sense

PR #998, opened 2026-07-31
Bind VictoriaMetrics and VictoriaLogs to loopback

phinze:
#992 routed distributed runner telemetry through the coordinator's authenticated listener, and its whole premise was that VictoriaMetrics and VictoriaLogs could then stop listening anywhere a runner can reach. The premise was right but unimplemented: both components still bound the wildcard, exactly as MIR-1483 describes them doing, so those ports were as open after that PR as before it. What stood in front of them remained a firewall rule, and on the GCP dev clusters that rule turns out to be the auto-created `default-allow-internal`, which permits all TCP and UDP across `10.128.0.0/9` and isn't managed in our infra repo at all.

Neither service authenticates anybody, so reachability is the entire access control. A firewall is a reasonable second layer and a poor only layer, since it can be flushed, reordered, or never applied on a host nobody configured. Binding `127.0.0.1` removes the exposure rather than guarding it, and then there's no rule left to maintain.

This is safe now in a way it wasn't before. The coordinator's own reads and writes go over loopback, and runners ship to `:8443` instead of dialing these directly. The only population that could be stranded is a cluster running distributed runners on a build predating #992, and since distributed runners sit behind `MIREN_LABS`, that's a small and known set. I rolled #992 out to garden, toys and club this morning and confirmed both metrics and logs arriving through the coordinator's ingest before writing this.

Moving the endpoint and readiness probe from `localhost` to the literal address is deliberate rather than cosmetic. On a dual-stack host `localhost` resolves to `::1` first, and against a v4-only listener that spends a failed dial on every new connection, which the readiness probe would pay during startup.

Two things this deliberately doesn't do. It doesn't add authentication to either backend: once they're on loopback the reachable set is host-level access, which already grants the etcd keyspace, the workload identity signing key, and the data files themselves, so a credential on the metrics port defends nothing that isn't already gone. And it leaves the external-instance path alone, where `--victoriametrics-addr` points at something an operator runs and secures themselves.

One residual worth naming: a sandbox with `Spec.HostNetwork` set shares the host netns and can still reach loopback. That needs rights to write a sandbox entity, a much higher bar than "any host on the VPC", but it's why I'd rather see a regression test pinning the bind address than call this airtight.

Closes MIR-1483

### DIFF
--- a/components/victorialogs/victorialogs.go
+++ b/components/victorialogs/victorialogs.go
@@ -135,7 +135,7 @@ func (c *VictoriaLogsComponent) Start(ctx context.Context, config VictoriaLogsCo
 	}
 
 	// Wait for VictoriaLogs to be ready
-	if err := c.WaitForReady(ctx, "localhost", config.HTTPPort); err != nil {
+	if err := c.WaitForReady(ctx, "127.0.0.1", config.HTTPPort); err != nil {
 		task.Delete(ctx)
 		container.Delete(ctx, containerd.WithSnapshotCleanup)
 		return err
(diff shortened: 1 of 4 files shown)


evanphx (approved):
The call is ONLY coming from inside the house!

PR #999, opened 2026-07-31
Re-register sandbox metrics for containers that survive a restart

phinze:
When miren restarts, `reconcileSandboxesOnBoot` adopts the containers that outlived the process. It reattaches log consumers, re-reserves IPs, and re-registers workload identity tokens. It never re-registered cgroup metrics, so CPU and memory for every recovered sandbox stopped at the restart and never came back. Logs kept flowing the whole time, which is what made this expensive to notice: a recovered sandbox looks completely healthy right up until you go looking for its metrics. Every runner and coordinator restart re-triggered it, so the fleet lost metrics for surviving sandboxes on each upgrade.

There's a second way to lose metrics that the boot path can't cover. Boot reconciliation only considers sandboxes the store already has as RUNNING, so one that was still PENDING at restart gets adopted later, when `CheckSandbox` returns `same`. Registering there is trickier than it looks, because that branch runs on every reconcile tick and `Metrics.Add` replaces the entry wholesale, discarding the previous CPU counters. Re-adding on each tick would reset the delta baseline forever and produce no CPU data at all. Hence `Metrics.Has`, and an `ensureMetrics` that does nothing when the sandbox is already monitored.

The log entity and attribute derivation was already copy-pasted between `createSandbox` and the saga's `addMetrics`. A third copy is exactly the kind of drift that let this hide, so all of it now goes through one `sandboxMetricsIdentity` helper.

On the frozen-hash bump for `sandbox.go`: the boot fix is shared between both controllers, since `SagaSandboxController.Init` delegates to the inner controller's `Init`. The adoption half needed mirroring and got it in `saga_controller.go`'s own `same` branch, and `create_saga.go` picks up the shared identity helper.

The test builds a real sandbox, discards the controller that booted it, and `Init`s a fresh one, which is the shape an actual restart takes. I confirmed it fails against the old code.

Closes MIR-1013

### DIFF
--- a/controllers/sandbox/create_saga.go
+++ b/controllers/sandbox/create_saga.go
@@ -398,18 +398,7 @@ func addMetrics(ctx context.Context, in addMetricsIn) (addMetricsOut, error) {
 		return addMetricsOut{}, fmt.Errorf("fetching sandbox: %w", err)
 	}
 
-	le := sb.Spec.LogEntity
-	if le == "" {
-		le = sb.ID.String()
-	}
-
-	attrs := map[string]string{"miren.sandbox": sb.ID.String()}
-	if sb.Spec.Version != "" {
-		attrs["miren.version"] = sb.Spec.Version.String()
-	}
-	for _, lbl := range sb.Spec.LogAttribute {
-		attrs[lbl.Key] = lbl.Value
-	}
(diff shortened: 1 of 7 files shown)


phinze:
@coderabbitai review

evanphx (approved):

PR #1001, opened 2026-08-03
Make tailnet-only clusters reachable

phinze:
Running Miren on a box reachable only over Tailscale didn't really work, and two separate bugs were hiding each other. This fixes the discovery half.

We filtered CGNAT out of advertised addresses deliberately, back when the concern was leaking tailnet addresses to clients who aren't on the tailnet. But Tailscale's IPv6 ULA reads as `IsPrivate()` and went through untouched, so the net effect was advertising the address family that often doesn't work while dropping the one that usually does. Meanwhile `docker0` and our own `rt0` and `flannel.1` were being advertised to everyone, despite only ever routing to workloads on the same host. One cluster I checked advertised five addresses, not one of which a remote client could use.

The original rationale doesn't hold up either. Clients probe every advertised address in parallel and cancel the rest on first success, so a candidate you can't route to costs a goroutine and nothing else. And "don't advertise things that might not work" wasn't a principle we were actually upholding while shipping container bridge addresses to the world.

So the rule is now uniform: keep anything the internet can't route to, drop container bridges, and let netcheck prune the genuinely public ones as before. That's deliberately RFC-based rather than Tailscale-aware, since an overlay network has to live in non-routable space to be an overlay. iroh and Nebula get the same treatment for free. The one place a vendor name survives is the `tailnet` label in `miren debug advertise`, which is display-only and gates nothing.

On the box that reproduced this, the advertised list went from five addresses a remote client couldn't use to four that it can, including the tailnet IPv4 address that was the whole point.

The second rev adds a guide for running on a tailnet, covering setup, the DNS-01 requirement for app certificates, and how app traffic gets to you. It also documents `additional_ips` as an advertisement control, which it has always been, even though both config pages described it purely as a certificate setting.

Needs the cloud side merged first (mirendev/cloud#191). Without it, cloud treats a tailnet address as public and would publish these as A records.

Closes MIR-1509

### DIFF
--- a/cli/commands/debug_advertise.go
+++ b/cli/commands/debug_advertise.go
@@ -74,7 +74,7 @@ func DebugAdvertise(ctx *Context, opts struct {
 			continue
 		}
 		humanInfo("  %-15s %-40s (discovered)", a.Interface, a.IP)
-		ipSet.AddDiscovered(ip)
+		ipSet.AddDiscoveredFrom(ip, a.Interface)
 	}
 
 	for _, s := range opts.AdditionalIPs {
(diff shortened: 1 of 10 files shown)


evanphx (approved):
Image

PR #1002, opened 2026-08-03
Release sandbox token state at teardown, not entity delete

evanphx:
The token refresher was handing back entries for sandboxes that had exited long ago. Every 45 minutes it would mint a fresh JWT for each one and try to write it to a file teardown had already deleted, which is a lot of warn-level noise for no benefit.

The unregister lived in `Delete`, which is the entity-delete callback, and the normal exit path never gets there. A container exits, `monitorTaskExit` flips the sandbox to STOPPED, and reconcile calls `StopSandbox` directly. `StopSandbox` removes the whole sandbox directory but never touched any of the token state, and the entity itself hangs around until `Periodic` sweeps DEAD sandboxes an hour later. So an entry was stale for an hour minimum, and if that delete never landed (delete error, sandbox rescheduled to another node, lease release failing) it leaked for the life of the process. The stale token secret is the part I liked least, since it keeps a dead sandbox's token requests authorized against the token server.

`StopSandbox` is the choke point every teardown path funnels through, so the cleanup moved there, and it now runs up front rather than behind container work that can hang or fail partway. It takes the sandbox as an optional argument, which is the bit that made this awkward before: `Delete` fires after the entity is gone, so the copy it was handed is the only one anyone will ever see, and that's why it had grown a duplicate of the firewall teardown. `Delete` is a two-line wrapper now.

The interesting part was the frozen-file guard on `sandbox.go`, which asks you to audit the saga controller before touching the hash. Doing that turned up the same leak in both boot-failure paths, the non-saga defer in `createSandbox` and `undoBootContainers` in the saga. Both tear down containers and mark the sandbox DEAD without ever reaching `StopSandbox`, and the token file is still sitting on disk at that point, so the self-heal below would never have covered for them either.

Speaking of which, the refresh loop now drops an entry when the write comes back ENOENT, since a missing sandbox directory means a missing sandbox. Teardown should have unregistered it already, so a non-zero drop count logs at info as a hint that some path skipped the cleanup.

Closes MIR-1512

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -2748,6 +2748,38 @@ func (c *SandboxController) Delete(ctx context.Context, id entity.Id, sb *comput
 	return c.StopSandbox(ctx, id, sb)
 }
 
+// entityFallbackIPs returns the addresses recorded on a sandbox entity, for use
+// when the pause container's labels couldn't supply them.
+//
+// A DEAD sandbox yields nothing. DEAD is only reached by way of a teardown that
+// already released the sandbox's addresses, and nothing ever clears Network on
+// the entity, so those addresses are stale by construction. Releasing them again
+// would be a no-op at best; once the netdb cooldown has passed and another
+// sandbox has reserved one of them, ReleaseAddr updates the row by IP with no
+// ownership check and hands a live address back to the pool.
+//
+// The delete callback is the path that makes this reachable, since it carries
+// the prior entity and fires from the periodic sweep an hour after teardown.
+func entityFallbackIPs(sb *compute.Sandbox) map[string]bool {
+	ips := make(map[string]bool)
+	if sb.Status == compute.DEAD {
+		return ips
+	}
+
+	for _, net := range sb.Network {
+		addr := net.Address
+		if strings.Contains(addr, "/") {
+			if prefix, err := netip.ParsePrefix(addr); err == nil {
+				addr = prefix.Addr().String()
+			}
+		}
+		if addr != "" {
+			ips[addr] = true
+		}
+	}
+	return ips
+}
+
 // StopSandbox tears down a sandbox and releases every resource it holds. sb is
 // optional: callers that already have the entity pass it so cleanup still works
 // once the entity has been deleted from the store, and it is fetched here
@@ -2829,19 +2861,7 @@ func (c *SandboxController) StopSandbox(ctx context.Context, id entity.Id, sb *c
 
 		// Use the entity's IPs as fallback if container labels didn't have them
 		if len(sandboxIPs) == 0 {
-			sandboxIPs = make(map[string]bool)
-			for _, net := range sb.Network {
-				addr := net.Address
-				if strings.Contains(addr, "/") {
-					if prefix, err := netip.ParsePrefix(addr); err == nil {
-						addr = prefix.Addr().String()
-					}
-				}
-				if addr != "" {
-					sandboxIPs[addr] = true
-				}
-			}
-
+			sandboxIPs = entityFallbackIPs(sb)
 			if len(sandboxIPs) > 0 {
 				c.Log.Debug("retrieved IPs from entity for cleanup", "id", id, "ips", sandboxIPs)
 			}
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -1048,7 +1048,7 @@ func (c *SandboxController) Create(ctx context.Context, co *compute.Sandbox, met
 		return nil
 	case compute.STOPPED:
 		c.Log.Debug("sandbox is stopped, verifying it is no longer running")
-		return c.StopSandbox(ctx, co.ID)
+		return c.StopSandbox(ctx, co.ID, co)
 	case "", compute.PENDING, compute.RUNNING:
 		searchRes, err := c.CheckSandbox(ctx, co, meta)
 		if err != nil {
@@ -1126,7 +1126,7 @@ func (c *SandboxController) Create(ctx context.Context, co *compute.Sandbox, met
 				}
 
 				// Clean up the unhealthy sandbox
-				err := c.StopSandbox(ctx, co.ID)
+				err := c.StopSandbox(ctx, co.ID, co)
 				if err != nil {
 					c.Log.Error("failed to cleanup unhealthy sandbox", "id", co.ID, "err", err)
 					return fmt.Errorf("failed to cleanup unhealthy sandbox: %w", err)
@@ -1241,6 +1241,12 @@ func (c *SandboxController) createSandbox(ctx context.Context, co *compute.Sandb
 
 			// Clean up the pause container using the common cleanup function
 			c.CleanupContainer(ctx, container)
+
+			// Boot may have gotten far enough to register the sandbox for token
+			// refresh (see buildSubContainerSpec). This path marks the sandbox DEAD
+			// without going through StopSandbox, so release that state here or it
+			// lingers until the entity is swept.
+			c.ReleaseTokenState(co.ID)
 
 			// Update sandbox status to DEAD in entity store
 			co.Status = compute.DEAD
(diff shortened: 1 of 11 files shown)


controllers/sandbox/sandbox.go
```diff
@@ -2691,16 +2698,28 @@ func (c *SandboxController) StopSandbox(ctx context.Context, id entity.Id) error
...
+			fetched.Decode(resp.Entity().Entity())
+			sb = &fetched
+		case !errors.Is(entityErr, cond.ErrNotFound{}):
+			c.Log.Warn("failed to get sandbox entity for cleanup", "id", id, "err", entityErr)
+		default:
+			c.Log.Debug("sandbox entity already deleted and no sandbox supplied, skipping firewall cleanup", "id", id)
+		}
+	}
 
+	if sb != nil {
```
phinze, line 2858: The IP fallback below (`len(sandboxIPs) == 0` → read `sb.Network`) was unreachable from `Delete` before this PR: the entity is gone by then, `EAC.Get` returned `ErrNotFound`, and the whole block was skipped. Delete events *do* carry the prior entity — `entityserver.go:562-564` re-reads it at `ModRevision-1` when the current read misses — so passing `sb` in makes it live.

How it goes wrong:

1. Sandbox A holds `10.x.0.5`. Its container exits, `monitorTaskExit` patches STOPPED.
2. Reconcile → `StopSandbox`. Pause container labels supply the IP, `10.x.0.5` gets released, pause container deleted, entity marked DEAD. The entity still records `10.x.0.5` in `Network` — nothing clears it.
3. Cooldown passes (30 min, `netdb.go:99`). Sandbox B reserves `10.x.0.5` and is running happily.
4. An hour after step 2, `Periodic` deletes A's entity → `Delete` → `StopSandbox(id, sbA)`. The pause container is long gone, so `sandboxIPs` comes back empty and the fallback fires: `ReleaseAddr(10.x.0.5)`.
5. `ReleaseAddr` has no ownership check (`netdb.go:454`, plain `UPDATE ips SET reserved = 0, released_at = now WHERE ip = ?`), so B's live address is back in the pool. `Reserve()` scans from `.2` upward, so sandbox C picks it up a cooldown later. Two sandboxes, one IP.

Step 4 is the steady-state teardown rather than a corner case: `Periodic` only deletes DEAD sandboxes, and DEAD is only reached via `StopSandbox` or the `createSandbox` boot-failure defer, both of which already released the IPs. The entity's addresses are stale by construction every time this path runs.

Worth keeping the passthrough, though — `exec_proxy.go:309` deletes a live RUNNING sandbox and leans on this callback, and there the fallback is a genuine win. A `sb.Status != compute.DEAD` guard on the fallback covers both: on the `unhealthy` path the store is patched to DEAD but the in-memory `co.Status` is still RUNNING at the `StopSandbox` call, and the STOPPED path is STOPPED.

Uncovered either way — every `Delete` in `sandbox_test.go` passes `nil`.

controllers/sandbox/token_refresh.go
```diff
@@ -84,13 +103,29 @@ func (c *SandboxController) refreshTokens() {
...
-		if err := os.WriteFile(e.filePath, []byte(token), 0644); err != nil {
+		err = os.WriteFile(e.filePath, []byte(token), 0644)
+		switch {
+		case err == nil:
+		case errors.Is(err, fs.ErrNotExist):
+			// The sandbox directory is gone, so the sandbox is gone. Teardown
+			// normally unregisters the entry (see ReleaseTokenState); this keeps a
+			// path that skipped it from refreshing a dead sandbox forever.
+			c.Log.Debug("dropping token refresh entry for departed sandbox", "sandbox", e.sandboxID)
+			c.tokenRefresher.unregister(e.sandboxID)
```
phinze: Could this be `c.ReleaseTokenState(entity.Id(e.sandboxID))`? The self-heal drops the refresher entry but leaves `tokenSecrets` registered, which is the stale-secret case the description calls out as the part you liked least. Inert in practice — the token server keys on the sandbox ID resolved from netdb, so a recycled IP resolves to the new sandbox and can't match the old entry — but it'd make the backstop cover the same ground as the path it's backstopping. `entity.Id` is a string type, so the round-trip through `.String()` is exact.

controllers/sandbox/sandbox_frozen_test.go
```diff
@@ -39,7 +39,15 @@ func TestSandboxControllerFrozen(t *testing.T) {
 		//
 		// Registry authentication is shared by both controllers because the saga
 		// controller delegates image pulls to this inner controller.
-		"sandbox.go":  "82a1667dae3227bc67d533b14487187d8a71968c9678f45373996305e2b49897",
+		//
+		// Also updated for the token-refresh leak fix: sandbox resource cleanup moved
```
phinze: FYI, nothing to do here — I'm clearing this whole accumulated block out in a parallel PR. It's been growing a paragraph per touch and git blame tells the story better anyway, so don't feel obliged to keep feeding it.

phinze (changes requested):
Claude turned this up while we were tracing the `sb` passthrough, and it reads real to me. It's not the token work — that all holds up, saga mirror included. It's a side effect: `StopSandbox`'s entity-IP fallback used to be unreachable from `Delete` (entity already gone → `ErrNotFound` → block skipped), and handing `sb` in wakes it up. On the `Periodic` sweep those addresses are stale by construction, so we end up re-releasing IPs a live sandbox may already hold.

The inline on `sandbox.go` lays out the sequence step by step — worth walking it and seeing if you land in the same place. The other two are low-stakes: a small symmetry thing in the self-heal, and an FYI on the frozen-test comment block that needs nothing from you. biscuit's `len(entries)-dropped` nit is optional too.

`--p+🤖`

controllers/sandbox/sandbox.go
evanphx: Good call out, I had worried about that.

evanphx: Walked it and landed in the same place. Confirmed each link: `entityserver.go:564` does re-read at `ModRevision-1`, `controller.go:617-622` decodes that into the `obj` handed to `Delete`, and `ReleaseAddr` at `netdb.go:454` is a bare `UPDATE ... WHERE ip = ?` with nothing checking who holds it. So the fallback was dead code that this PR brought to life on exactly the path where its input is stale by construction.

Took the guard. Checked it against each caller: the periodic sweep is DEAD so it skips, `exec_proxy.go`'s `deleteSandbox` is a live RUNNING sandbox with no prior release so it keeps the fallback, and the STOPPED and unhealthy paths still have unreleased addresses recorded so they stay correct. Left `UnconfigureFirewall` outside the guard since it already ran from `Delete` before this PR.

On coverage: `StopSandbox` is awkward to unit test with containerd in the way, so I pulled the decision into `entityFallbackIPs` and tested that directly. DEAD yields nothing, RUNNING and STOPPED yield the parsed addresses, CIDR forms get stripped.

One thing worth flagging, since it means this guard is narrower than it looks. `DEAD` is really a proxy for "the addresses were already released", and the proxy leaks: `StopSandbox` releases at `sandbox.go:2819` but only patches DEAD at `2838`, using the passed ctx. On shutdown that ctx is cancelled in between, the patch fails, and the error is logged rather than returned, leaving the entity durably STOPPED with `Network` still populated and the addresses already freed. `reconcileSandboxesOnBoot` skips anything not RUNNING, so the restart never looks at it, and the next reconcile calls `StopSandbox` with `Status != DEAD` and walks straight past the guard into the same double-release.

The cooldown does not save it either. `netdb.go:330` loops `[cooldownDur, 0]`, so when the first pass finds nothing the second treats any released address as available, and across a restart there is plenty of time regardless.

That is pre-existing rather than something this PR introduced (the STOPPED path always reached the fallback through the entity fetch), and I do not think another status check is the answer. The root cause is that nothing clears `Network` when the addresses are released, so either clearing it there or making `ReleaseAddr` ownership-aware would close the whole class. Happy to file that as a follow-up and keep this PR scoped to the guard you asked for, unless you would rather it ride along here.

phinze (approved):
Gate's right, and `entityFallbackIPs` is a better shape than the inline guard I'd suggested. The DEAD exclusion doesn't strand anything either — every path to DEAD already releases, via `StopSandbox` itself, the boot-failure defer's `deallocateNetwork`, or `undoAllocNetwork` in the saga.

Ship it.

`--p+🤖`

controllers/sandbox/token_refresh.go
evanphx: Yes, done. A backstop that covers less ground than the path it backs up is worse than no backstop, since it looks like it is handling the case. Also extended `TestRefreshTokens_DropsDepartedSandboxes` to register secrets for both sandboxes and assert the departed one is gone while the live one survives.

controllers/sandbox/sandbox_frozen_test.go
evanphx: Your cleanup landed while I was rebasing, so this resolved itself: I took your cleared form and did not re-add my paragraph. The map is just the three hashes now.

Rebasing onto the new main did surface one thing worth mentioning, since it sits right on this PR's premise. `Delete` had picked up `NetServ.RemoveSandboxMapping` on trunk, so it conflicted. I moved it into `StopSandbox` with the rest of the cleanup, which I think actually serves its own comment better: it wants the address claim dropped the moment the sandbox holding it goes, and teardown is that moment, well ahead of the entity delete.

controllers/sandbox/token_refresh.go
```diff
@@ -84,13 +103,29 @@ func (c *SandboxController) refreshTokens() {
...
+			dropped++
+		default:
 			c.Log.Warn("failed to write refreshed workload identity token", "sandbox", e.sandboxID, "error", err)
 		}
 	}
 
 	if len(entries) > 0 {
-		c.Log.Debug("refreshed workload identity tokens", "count", len(entries))
+		c.Log.Debug("refreshed workload identity tokens", "count", len(entries)-dropped)
```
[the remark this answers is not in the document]
evanphx: Tidied, though the reasoning needs one correction: this cannot go negative. `dropped` increments at most once per entry in the loop over `entries`, so it is bounded above by `len(entries)`. Zero was reachable, which is the real point, and logging `count=0` when every entry got dropped was silly. It is now computed once and only logged when there is something to report.

controllers/sandbox/sandbox_test.go
```diff
@@ -440,7 +440,7 @@ func TestSandbox(t *testing.T) {
 		r.NoError(err)
 
 		defer func() {
-			r.NoError(co.StopSandbox(context.WithoutCancel(ctx), id))
+			r.NoError(co.StopSandbox(context.WithoutCancel(ctx), id, nil))
 		}()
```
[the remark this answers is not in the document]
evanphx: Skipping this one as out of scope. This PR's only change to this file is a one-line signature fixup on 443, adding the new `sb` argument to a `StopSandbox` call; the restart-simulation structure you are describing came from main and is untouched here.

The observation itself looks fair, so worth raising against the test on its own rather than folding it into an unrelated change.

PR #1003, opened 2026-08-03
Make the sandbox IP ownership map authoritative

phinze:
A sandbox that landed on a recently-recycled pod IP could never get a workload identity token: `403 {"error":"invalid token"}` forever, and restarting the app didn't help. It took biscuit down for ~7 hours on Aug 3, and stayed invisible because it only bites newly created sandboxes. Everything already running holds a valid token and keeps working, so the cluster looks healthy right up until something deploys.

The token server identifies a caller by looking its source address up in the DNS server's map, so that map decides which app an issued token claims to be. Nothing ever fixed a wrong answer: deletes bailed when another entity still claimed the address without re-pointing at the survivor, a tracked live sandbox never re-asserted its own address, and the fallback took the first match regardless of status (so a stopped sandbox, whose entity lingers for up to an hour with its address, could outrank the live one). Ownership is now derived from per-sandbox claims, and the controller registers the mapping itself in BootContainers rather than waiting on the watcher.

There's also a recovery path: a failed secret verification is the signature of a poisoned mapping, so the token server re-derives the owner once and retries. A stale entry costs one rejected request instead of the sandbox's whole life.

Five of the eleven new tests fail against the old code. `miren debug dns` from the issue's remediation notes isn't here: that table is per-runner in-memory state, and reaching it needs a runner-side RPC plus coordinator node routing, which deserves its own change.

Closes MIR-1511

### DIFF
--- a/controllers/sandbox/sandbox.go
+++ b/controllers/sandbox/sandbox.go
@@ -116,6 +116,7 @@ type SandboxController struct {
 
 	tokenRefresher *tokenRefresher
 	tokenSecrets   *tokenSecretRegistry
+	lookupRefresh  refreshLimiter
 
 	topCtx context.Context
 	cancel func()
@@ -2053,6 +2054,8 @@ func (c *SandboxController) BootContainers(
 		}
 	}()
 
+	c.registerSandboxDNS(ctx, sb, ep, meta)
+
 	for _, container := range sb.Spec.Container {
 		opts, err := c.buildSubContainerSpec(ctx, sb, &container, ep, sbPid, meta, volumeMounts)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


phinze:
Good catch on the unbounded map. `allow` now sweeps entries older than the cooldown once the map passes a modest threshold, with a test covering it.

On the `ignoreID` guard in `releaseIPLocked`: it's redundant on the delete path as you say, but it's load-bearing for the other caller. `addSandboxMappingLocked` calls it for a sandbox that has moved addresses, and that sandbox is still in `s.sandboxes` under its old address at that point, so without the guard it would be selected as its own heir. Added a comment saying so.

/biscuit review

phinze:
@coderabbitai review

evanphx (approved):
Looks good! Nice robustnesses added!

PR #1007, opened 2026-08-03
Nag in Slack while code scanning alerts are open

phinze:
Five alerts sat open on main for months, two of them since May, and nothing
told anyone. The natural fix would be to react to the alert itself, but
GitHub Actions has no code_scanning_alert trigger, so a scheduled poll is
the only option that does not mean standing up a webhook receiver.

It queries the alerts API daily and posts only when the count is non-zero,
which makes silence meaningful: an empty channel is a clean board, not a
broken job. It also means the nagging stops on its own once the work is
done, rather than becoming a green message everyone learns to skim past.

The Slack rail already exists. SLACK_BOT_TOKEN is an org secret and
SLACK_CHANNEL_ID an org variable, both already used by the docs deploy, so
this adds no new configuration.

Two details worth knowing. The message is assembled inside jq rather than
by interpolating alert text into a JSON string in the shell, since rule
descriptions and paths are not ours to trust as JSON. And the workflow only
ever speaks when something is wrong, which means the posting path would
otherwise go untested until the first real alert, exactly when a silent
failure costs the most; the force_post dispatch input posts regardless of
count so the wiring can be checked on demand.

### DIFF
--- a/.github/workflows/code-scanning-watch.yml
+++ b/.github/workflows/code-scanning-watch.yml
@@ -1,0 +1,119 @@
+name: Code Scanning Watch
+
+# GitHub Actions cannot trigger on code_scanning_alert, so there is no way to
+# react the moment an alert opens. This polls instead, and stays quiet unless
+# there is something to say: at zero open alerts it posts nothing, so silence
+# in Slack genuinely means the board is clean.
+
+on:
+  schedule:
+    # 15:00 UTC daily, which is 10am CT in summer and 9am CT in winter.
+    - cron: '0 15 * * *'
+  workflow_dispatch:
+    inputs:
+      force_post:
+        description: 'Post to Slack even when there are no open alerts, to prove the wiring still works'
+        type: boolean
+        default: false
+
+permissions:
+  security-events: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch open alerts
+        id: alerts
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+        run: |
+          gh api "repos/$REPO/code-scanning/alerts?state=open&per_page=100" \
+            --paginate --slurp | jq 'add // []' > alerts.json
+
+          COUNT=$(jq 'length' alerts.json)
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Very nice

phinze:
Taken, though the specific vector doesn't hold: `github.repository` can't contain a `"` or `\`. GitHub restricts owner names to alphanumerics and hyphens, and repository names to alphanumerics, hyphen, underscore and period, so there's no input that breaks out of that string literal today.

The conclusion is still right for the reason in your last paragraph. Having exactly one field in the payload built by shell interpolation while everything else goes through jq is an inconsistency that only survives as long as nobody puts alert-derived text in it, and this file is a plausible thing to copy.

So `text` now comes out of jq as a JSON string, quotes included, and the payload splices it in unquoted the way `blocks` already was. Verified it stays valid JSON with a repo name of `evil"}],"x":[{"y` — jq escapes it rather than producing invalid JSON. It also incidentally fixes the pluralization, which was `alert(s)` in the fallback text while the blocks did proper singular/plural.

Your second point retracted itself, so nothing to do there.

phinze:
Thanks — both checked against `d64a8ef`, which landed after this review was generated.

**`COUNT` in `env:`** — already done. That commit moved `COUNT` into the step's `env:` block alongside `REPO`/`ALERTS_URL`/`RUN_URL`, so there's no longer a `${{ }}` expression inside the shell script. It came along with the `text` fix for the same reason you give.

**Random heredoc delimiter** — skipping, because the collision it protects against can't occur. `BLOCKS` is built with `jq -c`, which emits the whole array as exactly one line, and jq escapes any newline inside a JSON string as `\n` rather than emitting a real one.

I tested that rather than assuming it. Feeding an alert whose file path is literally `evil\nSLACK_EOF\nblocks=x` (real newlines, the exact shape needed to close the heredoc early and inject an output key) still produces one line of output:

```
$ jq '[.[0] | .most_recent_instance.location.path = "evil\nSLACK_EOF\nblocks=x"]' alerts.json \
    | jq -c -f blocks.jq | wc -l
1
```

Worth stating the invariant out loud since it isn't local to the heredoc: this holds because of the `-c` on line 62. Drop that flag and the concern becomes real.

PR #1009, opened 2026-08-03
Make daemon logging deliberate, and stop the audit trail repeating itself

phinze:
`miren server install` defaulted `--verbosity` to `-vv`, so every systemd-installed coordinator ran at Debug indefinitely, while `miren runner install` had no verbosity option at all and inherited the CLI default of Warn. Neither was chosen for the job it was doing.

Before changing anything I measured what a running cluster actually logs, which was worth doing because the fix as filed would not have accomplished much. On a busy coordinator Info already accounts for around 90% of lines, so moving the default from Debug to Info barely registers. The real cost is the audit trail, roughly half of all bytes, and within it a single message. `audit cert auth` fires per-RPC rather than per-connection and re-serializes the full subject DN, issuer DN, serial and fingerprint every time, about 330 bytes a line, thousands of times an hour, to say that the same two runner certs are still the same two runner certs. Because the audit stream is pinned at an Info floor by design it is immune to `-v`, so no choice of default was ever going to reach it.

Hence three revs, in the order the problem revealed itself. Daemons now start their `-v` ladder at Info instead of inheriting the CLI's Warn, with the marker at the command registration site so the level is right however the process was launched rather than depending on every launch path remembering a flag. Cert-auth records are deduplicated per certificate and peer, emitted on first sight and once per interval afterwards carrying a count of what they absorbed. And the Info tier gets a pass over the measured top offenders, driven by emitted volume rather than by reading all 1,315 Info call sites.

Together those cut roughly two thirds of the bytes a coordinator writes.

One deliberate omission. `rpc access` is the largest remaining source, almost entirely `EntityAccess.get/list/put` from runners, and demoting it was the obvious next lever. It is the wrong one: runners hold a general-purpose entity API, so those records are the only visibility into an unconstrained surface, and quieting them would hide exactly the thing worth watching. That is MIR-1520 instead.

The rubric added to CLAUDE.md is the part meant to outlast this. 1,315 Info call sites against 325 Debug says Info is not a considered tier but the level everyone reaches for, and without a written rule it drifts back.

Measuring also surfaced real failures the noise had been hiding, filed as MIR-1518 and MIR-1519.

Closes MIR-1503

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -279,3 +279,37 @@ See `cli/commands/route_list.go` for a reference implementation.
   - Avoid redundant comments that restate what the code does (e.g., `// Start server` above `server.Start()`)
   - Good comments explain complex logic, document assumptions, or clarify non-obvious behavior
   - Function/method comments should explain the purpose and any important side effects, not just restate the name
+
+### Log Levels
+
+Daemons (`miren server`, `miren runner start`) default to **Info**; one-shot CLI
+commands default to **Warn**. `-v` moves one step louder from there, and
+`SIGTTIN`/`SIGTTOU` retune a running process without a restart. Because Info is
+what an operator actually reads on a live cluster, it has to stay worth reading.
+
+- **Error** — the system failed at something it is responsible for. A user's app
+  failing to bind its own declared port is not this; the platform handled that
+  correctly.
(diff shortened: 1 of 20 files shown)


evanphx (approved):
Good call on this.

PR #1018, opened 2026-08-04
Connect to the cluster API from inside a sandbox, with per-app roles

evanphx:
Closes MIR-1344.

## What this does

Lets code running inside a sandbox call the cluster's own API as itself — the Kubernetes service-account model — authenticating with the workload identity token already mounted into every sandbox, scoped by a **role chosen per app**. Before this, the token existed only for *outbound* federation (AWS STS etc.); nothing validated it inbound and there was no way to reach the API from the sandbox network.

## In-cluster access

- **Verify our own tokens inbound.** New `pkg/workloadidentity` validator + authenticator, wired into the coordinator's auth chain (coordinator-only, ahead of OIDC). Validates issuer, audience, signature, expiry against the cluster's own keys, in-process.
- **Reach the API from the sandbox network.** Open the API port on the bridge (tcp **and** udp — the transport is QUIC), inject `MIREN_API_ADDRESS` / `MIREN_IN_CLUSTER` / `MIREN_CA_CERT_PATH`, and verify against an `api.miren` cert SAN rather than skipping verification (sandboxes share a bridge). On a distributed runner the address points at the coordinator, resolved to a literal IP host-side.
- **Client path.** `clientconfig.InCluster()` / `InClusterState()` (the `rest.InClusterConfig()` analogue), a per-request bearer-token func (the token refreshes on a loop), and a TLS server-name override. `m` just works inside a sandbox.

## Roles

A catalog of canned roles in `pkg/workloadroles`:

| Role | Scope | Grants |
|---|---|---|
| `none` | — | nothing |
| `app-readonly` *(default)* | own app | status, logs, deployment reads |
| `app-deploy` | own app | + build & (re)deploy |
| `app-debugger` | own app | + exec/run in own sandboxes |
| `app-admin` | own app | + config/env writes, exec |
| `cluster-readonly` | cluster | reads across all apps + infra |
| `cluster-deployer` | cluster | + deploy/manage any app |
| `cluster-debugger` | cluster | + exec/inspect any app |
| `cluster-admin` | cluster | broad, minus the carve-outs below |

App-scoped roles are confined to their own app; cluster roles are not. **No role — not even `cluster-admin` — can mint identity tokens, do raw entity-store writes, or reconfigure the network fabric.** Those stay on the internal cert plane; a role is a bearer token in a sandbox, and these limits bound the blast radius.

The role is a token claim, resolved **server-side from the app entity** in every mint path (coordinator, refresher, distributed-runner proxy), never trusted from the caller.

**Assignment:** `workload_role` in `.miren/app.toml` (app-scoped roles only — cluster values are rejected at deploy), or `m app set-workload-role` (operator path, the only way to grant a cluster role).

## Authorization guards (the load-bearing correctness work)

The authorizer only sees `(resource, action)`, never the target app — so per-app confinement lives in handlers via `rpc.AllowApp`. An app-scoped role can therefore only safely grant methods whose handler calls `AllowApp`. This change:

- Makes `rpc.BoundApp`/`AllowApp` method-driven and adds `AllowApp` guards to the config-write handlers (`crud.set*`, `deployment.setEnvVars`/`deleteEnvVars`) and **both** exec entry points, so an app-scoped role can only ever reach its own app. A tripwire test enforces *"app-scoped role ⇒ only guarded methods."*
- Closes the `CompositeAuthorizer` default-allow: an unknown auth method now fails closed.
- Verifies a deployed `app_version` belongs to the target app, so a role that can deploy can't pull another app's built image (and its secrets) into its own runtime. Enforced leniently at `CreateDeployment`/`UpdateDeploymentAppVersion` (a real cross-app version is rejected; a not-yet-built placeholder is left alone) and strictly at `DeployVersion`.
- `SetWorkloadRole` uses a single-attribute entity patch, so an operator setting a role can't clobber a concurrent deploy's write to the same app record.

## Docs

New **In-Cluster API** page (connection + roles), `workload_role` in the app.toml reference, and the generated `app set-workload-role` command doc. The dev server gains `--dns-names` so a default `make dev` has an issuer and can exercise the feature.

## Follow-ups (filed separately)

- MIR-1528 — `crud.list` leaks all apps to any OIDC caller (pre-existing; deliberately kept out of every workload role).
- MIR-1529 — rate-limit auth failures now that the API port is reachable from the bridge.

## Testing

- Unit tests throughout: the role catalog + guarded-method tripwire, the validator (expired / wrong-issuer / wrong-audience / forged / legacy-key / rotation), the authenticator (foreign-issuer declined, app vs cluster scoping, unknown role stays confined), the `AllowApp`/`BoundApp` table incl. a regression guard for the old default-allow, per-handler own-app-vs-other-app guard tests, the cross-app version-deploy rejection, and the in-cluster client config.
- The feature was driven end-to-end in a dev cluster (before this branch was rebased): default `app-readonly` (own logs work, exec/write denied), `app-admin` via app.toml (own-app exec + env-write work, other app denied with `bound to app …`), `cluster-admin` in app.toml rejected at deploy, operator-granted `cluster-readonly` reading across apps, and wrong-CA TLS rejected.
- etcd/containerd-backed integration tests run in CI (they can't run on a bare host).


pkg/workloadroles/roles.go
```diff
@@ -0,0 +1,189 @@
...
+// The package has no internal dependencies so both the authorizer (pkg/oidcauth)
+// and the authenticator (pkg/workloadidentity) can import it without a cycle.
+package workloadroles
+
+// Role names. These are the values stored on an app and carried in the token's
+// role claim.
+const (
+	RoleNone            = "none"
+	RoleAppReadonly     = "app-readonly"
+	RoleAppDeploy       = "app-deploy"
```
phinze: Small thing that gets expensive once it's on an entity attribute and a published docs page. The catalog is a clean 4×2 grid with one break:

```
readonly  app-readonly   cluster-readonly
deploy    app-deploy     cluster-deployer   <- verb vs agent noun
debugger  app-debugger   cluster-debugger
admin     app-admin      cluster-admin
```

`app-deployer` regularizes it (or `cluster-deploy`, either direction works). Renaming after the fact means migrating app.toml files, the `workload_role` attribute, and the docs.

pkg/workloadroles/roles.go
```diff
@@ -0,0 +1,189 @@
...
+var Roles = map[string]Role{
+	RoleNone: {Perms: perms{}},
+
+	RoleAppReadonly: {Perms: appRead},
+	RoleAppDeploy:   {Perms: merge(appRead, appDeploy)},
+	RoleAppDebugger: {Perms: merge(appRead, execBlock)},
+	RoleAppAdmin:    {Perms: merge(appRead, appDeploy, appConfig, execBlock)},
+
+	RoleClusterReadonly: {Perms: merge(appRead, clusterRead), ClusterScoped: true},
+	RoleClusterDeployer: {
```
phinze, line 128: `cluster-deployer` picking up `crud.new`/`crud.destroy` is asserted in the tests, so I read it as intentional. What surprised me is the asymmetry with the app tier: `appDeploy` has no destroy, and `RoleAppAdmin` deliberately doesn't either (the spot-check asserts that too). So the ladder reads as "the admin of my app can't delete my app, but a cluster deployer can delete every app."

Nobody predicts that from the names. My lean is moving destroy up to `cluster-admin` and leaving deployers deploying, since "deployer deletes your app" is the surprising direction to be wrong in. If the grant is right as-is, the docs row should say create and destroy out loud instead of "manage any app."

Related: the `app-admin` docs row says "Full control of its own app" when it deliberately stops short of destroy. Worth a more precise word.

pkg/workloadroles/roles.go
```diff
@@ -0,0 +1,189 @@
+// Package workloadroles is the single source of truth for the canned roles a
```
phinze, line 1: The package doc is good on the mechanics. Worth one more line on the horizon: RFD-67 (MIR-891) moves authorization to `{Resource, Action, Instance}` with `Scope` on the permission, at which point this catalog becomes policy presets rather than a Go map and every `rpc.AllowApp` call site deletes.

Nothing here blocks that, and the one decision that would have (carrying the resolved permission set in the token instead of the role name) went the right way, so swapping the resolution backend needs no token format change and no reissue. Saying so here keeps the next person from reading the map as the permanent model.

docs/docs/in-cluster-api.md
```diff
@@ -0,0 +1,127 @@
...
+| Environment variable | Value | Use |
+| --- | --- | --- |
+| `MIREN_IN_CLUSTER` | `1` | Marks the process as running inside a sandbox |
+| `MIREN_API_ADDRESS` | e.g. `10.x.x.1:8443` | The cluster API address, reachable from the sandbox |
+| `MIREN_CA_CERT_PATH` | `/var/run/miren/ca.crt` | The cluster CA, used to verify the API's certificate |
+| `MIREN_IDENTITY_TOKEN_PATH` | `/var/run/miren/identity-token` | The token presented on each request |
+
+The connection is verified against the cluster CA — it is *not* insecure. Because sandboxes share a network bridge, skipping verification would let a neighbor impersonate the API, so the CA is mounted and checked.
+
+### From Go
```
phinze: I'd cut this section, and `clientconfig.InClusterState` with it, rather than fix it.

The example doesn't work as written: `State.Client(name)` resolves through an exact-match lookup at `/_rpc/lookup/<name>`, and the capability is exposed as `dev.miren.runtime/app`, so `state.Client("app")` fails. Every caller in the tree uses the full string.

But fixing it publishes a Go client we're not set up to stand behind. `go list -deps ./clientconfig` is 375 packages, including `dave/jennifer` (a code generator), all of cel-go, lipgloss, and `pkg/oidcauth`, so user code importing the client links the server's authorization tables. That's all accidental coupling and all predates this PR, but merging is the moment we start promising it.

With #978 landing a REST/JSON surface, that's the natural in-cluster call anyway, and advertising the heavyweight door right before the light one opens leaves us maintaining two clients with users defaulting to the wrong one. `InClusterState` has no callers anywhere in the tree, so deleting costs nothing and keeps `clientconfig` rearrangeable.

I'll file the boundary cleanup separately. The moves are small: GitHub Actions helpers out of `oidcauth`, keypair helpers out from under `rbac`, then the `pkg/rpc` client/server/generator split, then `internal/`.

docs/docs/in-cluster-api.md
```diff
@@ -0,0 +1,127 @@
...
+:::info[Two directions, one token]
+- **[Workload Identity](/workload-identity)** — present the token to *external* services to prove which workload you are.
+- **In-Cluster API Access** (this page) — present the same token to the *Miren* API, scoped by a role.
+:::
+
+Typical uses: a CI job running inside a sandbox that redeploys its own app, a sidecar that tails its app's logs, an operator tool that reads status across the cluster, or an app that manages its own configuration.
+
+## Connecting
+
+From inside a sandbox, the `miren` CLI just works — it detects that it's running in-cluster and authenticates with the mounted token:
```
phinze, line 23: Worth a clause here: nothing mounts the `miren` binary into a sandbox, so this works once you've installed the CLI in your image. "Just works" plus "No `miren login`, no config file" reads as "it's already there," and for a distroless or slim image it isn't.

docs/docs/in-cluster-api.md
```diff
@@ -0,0 +1,127 @@
...
+
+:::warning[Cluster-scoped roles are operator-only]
+`set-workload-role` is not reachable by in-sandbox workloads or app-scoped deploy identities — only a cluster operator can call it. That's what stops an app owner from self-granting cluster-wide access.
+:::
+
+You can see an app's current role in `miren app status`.
+
+## Sharp edges
+
+:::warning[A role change takes effect on the next sandbox build]
```
phinze, line 93: Adjacent sharp edge worth its own line: deleting `workload_role` from app.toml doesn't revoke the role. `applyWorkloadRole` no-ops on empty, so the last value sticks. `workload_role = "none"` is the revoke path and it works correctly today, it just isn't written down, and deleting the line is what everyone tries first.

servers/app/app.go
```diff
@@ -598,6 +613,33 @@ func (r *AppInfo) SetHost(ctx context.Context, state *app_v1alpha.CrudSetHost) e
 	return nil
 }
 
+// SetWorkloadRole sets the app's workload identity role. It has no rpc.AllowApp
+// guard on purpose: this is the operator path, and it is absent from every token
+// role map, so only cert (internal/operator) and JWT-with-RBAC callers reach it
+// — never an app-scoped OIDC or workload identity. That is what lets it grant
+// cluster-scoped roles, which the app.toml path must not.
+func (r *AppInfo) SetWorkloadRole(ctx context.Context, state *app_v1alpha.CrudSetWorkloadRole) error {
```
phinze, line 625: `setworkloadrole` staying operator-only currently rests on a negative invariant: it's absent from every role map, backed by the denylist in `TestCarveOutsAbsentFromAllRoles`. That works and the tripwire is worth having, but a one-line `if rpc.BoundApp(ctx) != "" { return rpc.AppAccessError(ctx, name) }` says it affirmatively and doesn't depend on nobody ever adding a `crud` method to `appConfig`. It's also the shape the eventual `cluster:admin` gate takes.

servers/build/build.go
```diff
@@ -727,6 +728,57 @@ func mergeCliEnvVars(existingVars []core_v1alpha.ConfigSpecVariables, cliVars []
 	return result
 }
 
+// applyWorkloadRole persists a workload_role declared in app.toml onto the app
```
phinze: Leftover from splitting validate out of apply: this paragraph documents `applyWorkloadRole` but now sits on top of `validateWorkloadRole`, which has its own comment right below. The split itself was the right call.

phinze (changes requested):
Requesting changes on two things, neither of them a bug: the role names and the Go surface. Both are cheap to change now and a migration once they're published.

Also worth carrying into #978: the authenticate/authorize gate lives in the QUIC request path (`pkg/rpc/server.go:958`), not in `Method.Handler`. If `restCall` reaches handlers without an equivalent gate, every role here is enforced on QUIC and bypassed on REST, and REST is the path we're about to point sandboxes at.

One thing that might deserve its own changelog line: `verifyVersionOwnedByApp` closes a live hole on the shipping CI path, not just a hypothetical workload one. `oidcDeployRole` already grants `deployversion`/`createdeployment`/`updatedeploymentappversion`, so an OIDC token bound to app X could name app Y's built version and pull Y's image and baked secrets into X.

Inline below. The role naming and the Go cut are the two with teeth, the rest are notes.

`--p+🤖`

pkg/workloadroles/roles.go
evanphx: 🤖 Agreed — renamed `app-deploy` → `app-deployer` so the grid reads consistently (deployer/debugger/admin as agent nouns). Done before merge, so nothing to migrate.

pkg/workloadroles/roles.go
evanphx: 🤖 Good catch on the surprising direction. Moved `crud.new`/`crud.destroy` out of `cluster-deployer` — app lifecycle is `cluster-admin`'s now; deployers just deploy and configure. Updated the spot-check and the docs rows, and tightened the `app-admin` row, which deliberately stops short of destroy.

pkg/workloadroles/roles.go
evanphx: 🤖 Added a paragraph noting RFD-67/MIR-891 turns this catalog into policy presets, and that the token carries the role *name* so the backend can swap without a token change or reissue. Thanks for the context — worth keeping the next person from reading the map as permanent.

docs/docs/in-cluster-api.md
evanphx: 🤖 Agreed on all counts — dropped `InClusterState` (no callers) and the "From Go" section rather than fix the broken example and start promising a client we're not set up to stand behind. REST (#978) is the right in-cluster door. Thanks for taking the `pkg` boundary cleanup separately.

docs/docs/in-cluster-api.md
evanphx: 🤖 Added a clause: the `miren` binary has to be in your image — nothing mounts it — so "just works" is once it's installed. Matters for distroless/slim.

docs/docs/in-cluster-api.md
evanphx: 🤖 Documented it — deleting the line keeps the last value; `workload_role = "none"` is the revoke path. Good one to call out, since deleting is what everyone tries first.

servers/app/app.go
evanphx: 🤖 Added it: `SetWorkloadRole` now returns `AppAccessError` when `rpc.BoundApp(ctx) != ""`, so operator-only is stated directly instead of resting on the method staying absent from every role map (kept the tripwire too). Added a test for the app-scoped-rejected case.

servers/build/build.go
evanphx: 🤖 Fixed — removed the orphaned `applyWorkloadRole` comment and folded its escalation rationale into `validateWorkloadRole`'s. Thanks for catching it.

phinze (approved):
All addressed, and the role table came back sharper than what I asked for.

Still open from last round, not a blocker here: whether #978's `restCall` runs the same authenticate/authorize gate as the QUIC path. That's what decides whether these roles hold once REST is the in-cluster call.

`--p+🤖`

evanphx:
@phinze yeah, i'm going to decide that over in #978

PR #1019, opened 2026-08-05
app.toml: [tasks] blocks and the web opt-out (RFD-97, 1/3)

evanphx:
Implements the config surface for [RFD-97 "App Tasks"](https://github.com/mirendev/rfd/blob/main/rfds/0075/0097-app-tasks.md), the approved design RFD for MIR-853. First of three stacked PRs (config → execution → triggers).

This PR is inert on merge: `[tasks]` parses, validates, and reaches `ConfigSpec`, but nothing runs a task yet.

## `[tasks.<name>]`

A task is a command the platform knows how to run; a service is a process it keeps up.

```toml
[tasks.migrate]
command = "bin/rails db:migrate"
trigger = "deploy"
timeout = "10m"

[tasks.cleanup]
command = "bin/cleanup-sessions"
trigger = "schedule"
every = "6h"
```

`[services]` is untouched — no new fields, no new values, nothing to relearn. Tasks deliberately have no `ports`, `concurrency`, `image`, or `disks`: they're absent from the schema rather than present-and-rejected, so the grammar never admits a configuration the validator has to refuse. (Per-task `image` and `disks` are explicit v1 cuts in the RFD, tracked there as open questions.)

## Scheduling

`schedule` takes systemd `OnCalendar` syntax, which reads left to right — `Mon *-*-* 09:00:00` is legible as "Monday at 9am" without knowing which position means what. New `pkg/oncalendar` parses the subset we accept; there was no existing dependency for this.

`every` is pure sugar over `schedule`, desugared during config assembly so **only the calendar form is ever stored** and every downstream consumer has exactly one scheduling mechanism to understand.

Intervals anchor to midnight UTC rather than to the deploy. An interval measured from "whenever this was deployed" would move every time the app ships — a daily job on an app that deploys ten times a day would fire ten times a day — and tick-based dedup couldn't catch it, because a new anchor produces a new tick. The cost is that `every` only accepts durations tiling a day evenly; `every = "7h"` is rejected at build time with an error pointing at `schedule`.

## `web = false`

`ensureWeb` gains an opt-out and **nothing else**. An app declaring only `[services.worker]` still gets a synthesized `web`, because narrowing that would drain a running pool on the next deploy of every app relying on it. Opting out is opt-in.

That preserved default is only safe if the ambiguous case is caught, so declaring tasks with no services and no explicit `web` is now a build error carrying its own fix:

```
this app declares tasks but no services, so it's unclear whether it
should have a web service.
  If it has no long-running process:  set web = false
  If it should serve traffic:         declare it, e.g. [services.web]
```

Without it, a task-only app would quietly acquire a web service synthesized from its image entrypoint and keep it running forever — the accidental long-running process tasks exist to eliminate, and hard to notice because it costs money quietly rather than failing. It matters most for the readers who aren't reading: an agent generating an `app.toml` has no way to know the platform will invent a web service it didn't ask for.

A service declared in a Procfile answers the question just as well as one in `app.toml`.

## Also

- `errNoServices` becomes "no services or tasks" — an app made entirely of tasks is valid, it just has nothing running between invocations.
- `miren app analyze` reports declared tasks with their resolved schedule, so it doesn't silently omit half of what the config now says.
- Stored task order is sorted; map iteration would otherwise make identical input produce a different `ConfigSpec` on every build.
- `services.*` gained `port_timeout` in the "did you mean?" field list, which it was missing.

## Testing

`make lint` clean, `make test` clean. `pkg/addon/postgresql` and `pkg/addon/rabbitmq` integration tests fail intermittently in my environment, but they fail identically on `main` with this branch absent.

Unit coverage for the OnCalendar grammar and `Next` (including leap-day, month/year rollover, and timezone-independence), every rejected `every` value, the full task validation table, `web` in all three states, and the ambiguous-web error.


servers/build/build.go
```diff
@@ -201,18 +201,117 @@ func mergeServiceEnvVars(existingEnvs []core_v1alpha.ConfigSpecServicesEnv, newE
...
+// generating an app.toml has no way to know the platform will invent a web
+// service it didn't ask for, and this error is the only moment it can act on
+// that. Nothing here is new-user-hostile, since the config shape it fires on
+// cannot exist before tasks.
+func validateWebIntent(ac *appconfig.AppConfig, procfileServices map[string]string) error {
+	if ac == nil || len(ac.Tasks) == 0 {
+		return nil
+	}
+	// A service declared anywhere — app.toml or Procfile — answers the question.
+	if len(ac.Services) > 0 || len(procfileServices) > 0 {
```
phinze, line 251: This short-circuit is looking at a map that the thing it's guarding against just populated.

`buildServicesConfig` calls `ac.ResolveDefaults(allServiceNames)`, which mutates `ac.Services` **in place** on the caller's config — so once the synthesized `web` is staged, `ac.Services` has a `web` entry. Both call sites assemble the config before validating (`buildFromDir:1406-1418`, `prepareConfig:252-264`), so by the time we get here the answer is always "a service exists," and the error can't fire.

Ran the two in the order `buildFromDir` calls them:

```
services in spec: 1, ac.Services: map[web:0x...]
Error: Expected error with "this app declares tasks but no services..." but got nil
--- FAIL: current order: assemble then validate
--- PASS: validate before assembling
```

So a task-only `app.toml` with no `web` key deploys clean and quietly acquires the forever-running web service — the exact outcome the check exists to prevent, and the one thing in this PR that isn't inert on merge.

`TestValidateWebIntent` passes because it calls this with a fresh struct, which is precisely where the aliasing is invisible. Whatever the fix, a test that goes through `buildVersionConfig` first would pin it.

servers/build/build.go
```diff
@@ -1298,11 +1409,19 @@ func (b *Builder) buildFromDir(ctx context.Context, name string, path string,
...
+		Log:              b.Log,
 	})
 
-	// Fail the deploy if no services are defined - this prevents deploying an app
-	// that can't serve any traffic
-	if err := validateServicesExist(configSpec); err != nil {
+	// Ask about web intent before checking for workloads: an ambiguous app
+	// would pass the workload check on a synthesized web service, which is
+	// exactly the outcome the question exists to prevent.
+	if err := validateWebIntent(ac, procfileServices); err != nil {
```
phinze: Moving this above `buildVersionConfig` fixes the aliasing (see the note on `validateWebIntent` itself) and also gets us a fail-fast: right now someone learns they need `web = false` only after sitting through a full image build, and nothing this check reads comes from build output — the Procfile is read from the source FS.

In the saga, `loadSource` looks like the natural home: it already produces both `AppConfig` and `ProcfileServices`, and its doc comment already claims fail-fast-before-buildkit as its job.

docs/docs/app-toml.md
```diff
@@ -265,6 +292,112 @@ containers that run as root are left untouched. See
...
+(`09..17`), and a repetition (`00/6`). Everything is UTC. Cron syntax is not
+accepted; `man systemd.time` documents the full grammar.
+
+:::note[Validation]
+- `command` is required.
+- `trigger` must be `deploy`, `schedule`, or `manual`.
+- `every` and `schedule` are mutually exclusive, and exactly one is required when `trigger = "schedule"`. Setting either without that trigger is an error rather than a silent no-op.
+- `every` must divide a day evenly.
+- `timeout` must be a valid Go duration and must not be negative.
+- `retries` must be non-negative, and has no effect on a `manual` task — the caller decides whether to retry.
```
phinze: Small contradiction with the bullet two lines up: `retries` on a manual task isn't ignored, it's a build error. "Setting either without that trigger is an error rather than a silent no-op" promises exactly the opposite of "has no effect" here.

Maybe: "`retries` must be non-negative, and is rejected on a `manual` task — the caller decides whether to retry."

phinze (changes requested):
One blocker: `validateWebIntent` never fires on a real deploy, so the ambiguous task-only app still gets its silent web service. Details inline — the fix is a reorder, not a redesign, and it buys a fail-fast as a bonus.

Everything else holds up. `oncalendar` is a nice little package and the day-alignment argument for `every` is the right call. Of CodeRabbit's four, the `validateRequiredVars` gap on task env is worth picking up; the rest are optional.

`--p+🤖`

servers/build/build.go
evanphx: Confirmed and fixed. You're right that the guard was reading a map its own guardee had just populated — `ResolveDefaults` mutates `ac.Services` in place, so after assembly every app claims a service.

Moved the call ahead of `buildVersionConfig` in the legacy path and into `loadSource` in the saga, per your suggestion — it already produces both the `AppConfig` and the Procfile services, and the fail-fast is a real gain: previously you learned you needed `web = false` only after a full image build.

Added `TestValidateWebIntentSurvivesConfigAssembly`, which goes through `buildVersionConfig` first and asserts both halves: the check catches the ambiguity before assembly, and provably cannot after. That second assertion is the one that would have caught this.

servers/build/build.go
evanphx: Done — `loadSource` it is. It takes `StreamID` now so a config error still reaches the deploying client, which is the point of failing early.

docs/docs/app-toml.md
evanphx: Fixed, using your wording: "is rejected on a `manual` task". The prose contradicted the bullet two lines above it.

pkg/oncalendar/oncalendar.go
```diff
@@ -0,0 +1,439 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// searchHorizon bounds Next's forward scan. An expression that matches nothing
+// within five years is treated as unsatisfiable rather than scanned forever.
+const searchHorizon = 5
```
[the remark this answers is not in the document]
evanphx: Good catch, and a real bug. The horizon was sized for the four-year leap cycle, but intersecting a leap day with a weekday stretches much further — the Gregorian century rule drops leap days that would otherwise land mid-cycle, so `Mon *-02-29` can go 40 years between hits.

Widened to 50 years with the reasoning in the comment. Added tests for weekday-on-leap-day parsing and for `Mon *-02-29` resolving to 2044-02-29 from now, plus a test that impossible dates (`*-02-30`, `*-04-31`) are still rejected — so the wider horizon isn't just trading a wrong answer for a slower one.

servers/build/build.go
```diff
@@ -201,18 +201,117 @@ func mergeServiceEnvVars(existingEnvs []core_v1alpha.ConfigSpecServicesEnv, newE
...
+
+		for _, ev := range tc.EnvVars {
+			task.Env = append(task.Env, core_v1alpha.ConfigSpecTasksEnv{
+				Key:         ev.Key,
+				Value:       ev.Value,
+				Required:    ev.Required,
+				Sensitive:   ev.Sensitive,
+				Description: ev.Description,
+				Source:      "config",
+			})
```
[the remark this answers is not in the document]
evanphx: Taking this one — phinze flagged it as worth picking up too. `validateRequiredVars` now walks `spec.Tasks[*].Env`. A required-but-empty task variable fails the deploy for the same reason a service's does: otherwise the run starts and falls over on a value the platform already knew was missing. The message names the task rather than saying "service".

docs/docs/app-toml.md
```diff
@@ -265,6 +292,112 @@ containers that run as root are left untouched. See
...
+### Triggers {#triggers}
+
+| `trigger` | Starts when |
+|-----------|-------------|
+| `deploy` | A new version is deployed, before it takes traffic. The deploy fails if the task does. |
+| `schedule` | A calendar expression comes due |
+| `manual` | Only when someone asks |
+
+Any task can be invoked by hand whatever its trigger — that's how you re-run a
+migration without redeploying.
```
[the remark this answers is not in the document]
evanphx: Fair, though it resolves itself by merge order: this is 1/3 of a stack, and #1043 (which implements execution) sits directly on top and merges after. Adding a not-yet-available admonition here would mean landing documentation that's already stale by the time the next PR merges. Leaving as-is deliberately.

cli/commands/deploy.go
```diff
@@ -1771,6 +1771,27 @@ func analyzeApp(ctx *Context, bc *build_v1alpha.BuilderClient, dir string) error
...
+					detail = fmt.Sprintf("%s: %s", task.Trigger(), task.Schedule())
+				}
+
+				ctx.Printf("  %s: %s%s\n",
+					analyzeValueStyle.Render(task.Name()),
+					task.Command(),
+					lipgloss.NewStyle().Foreground(theme.Muted).Render(fmt.Sprintf(" (%s)", detail)))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
evanphx: Leaving this one. Task commands come from the app's own `app.toml`, which is already fully trusted input — it specifies the commands the platform executes in that app's containers. A control character in `miren app analyze` output is not a privilege boundary anyone crosses by writing their own config. If terminal-escape sanitizing is wanted it belongs in the CLI's rendering layer for every field it prints, not on this one.

evanphx:
Pushed fixes for the blocker and the review items, rebased onto `main` (conflict cleared).

**Blocker** — `validateWebIntent` moved ahead of config assembly: `buildVersionConfig` in the legacy path, `loadSource` in the saga. The new `TestValidateWebIntentSurvivesConfigAssembly` goes through assembly first and pins both directions, since a fresh-struct test is exactly where the aliasing hides.

**Also fixed:** `validateRequiredVars` now covers task env; `oncalendar`'s horizon widened for weekday-on-leap-day expressions (real bug — `Mon *-02-29` was rejected as never firing); dropped-task log moved to `Warn`; the `retries` doc contradiction.

Two CodeRabbit items declined with reasoning inline (the docs admonition, since #1043 implements execution and merges right after this; and the terminal-control-character one, since `app.toml` is already trusted input).

While rebasing, main's exec app-scoping guard landed. `Attach` in #1043 needed the same guard and didn't have it — fixed there, since attaching hands the caller a container's stdio and would otherwise have been a bypass.

evanphx:

- Documented the task-only exception where web synthesis is described, not only in the validation note further down.

The three earlier blockers (validateWebIntent ordering, task env in validateRequiredVars, the oncalendar leap-day horizon) were fixed and replied to in the previous round.

docs/docs/app-toml.md
```diff
@@ -73,6 +83,27 @@ tail = "logs app -f"
...
+If neither `app.toml` nor your `Procfile` declares a `web` service, Miren
+synthesizes one from the image's entrypoint. That is the historical default and
+it stays the default: leaving `web` unset never changes what your app does.
+
+There is one exception, and it's the case where guessing would be expensive: an
+app that declares [tasks](#tasks) and no services has to say which it meant. See
+the validation note below.
+
+`web = false` opts out. It's how an app that only declares tasks says it has no
+long-running process at all — no web service, no route, and nothing running (or
```
phinze, line 99: This overpromises slightly. `web = false` vetoes the *synthesis*, but a Procfile `web:` line has already populated `srvMap["web"]` by the time the `ensureWeb && wantWeb` guard runs, so that service survives and nothing warns. Preferring the explicit declaration over the top-level bool seems right, but someone migrating an app with `web: bin/server` who adds `web = false` will get a web service anyway and no signal about it. Probably just wants a clause here — "unless a service named `web` is declared in `app.toml` or your Procfile, which wins" — rather than code.

servers/build/build.go
```diff
@@ -1887,6 +2022,25 @@ func (b *Builder) AnalyzeApp(ctx context.Context, state *build_v1alpha.BuilderAn
 		result.SetServices(&services)
 	}
 
+	// Tasks come straight from spec, which already resolved defaults and
```
phinze, line 2105: Worth noting for 2/3: analyze reaches `buildVersionConfig` without going through `validateWebIntent`, so for a task-only app with no `web` key it prints `web: image default` under Services and then the deploy hard-fails with `errAmbiguousWeb`. Analyze skips the other deploy validations too, so it's consistent in spirit, but this one shows something affirmatively wrong rather than just staying quiet — and analyze is the command you'd reach for to preview exactly this. Fine to leave until execution lands.

phinze (approved):
Blocker's cleared, and the regression test that pins *why* the ordering matters is the better half of the fix — if the `ResolveDefaults` aliasing ever goes away, the next person gets told instead of rediscovering it.

Two non-blocking notes inline, neither with teeth: one is a docs sentence, the other probably belongs in 2/3.

`--p+🤖`

docs/docs/app-toml.md
evanphx: 🤖 You are right, and the precedence is worth keeping — an explicit declaration should beat a bool. Documented rather than changed: the section now says `web = false` opts out of the *synthesized* service, not one you asked for, and that a `web` in app.toml or a Procfile `web:` line survives it, with the fix being to delete the declaration.

servers/build/build.go
evanphx: 🤖 Execution has landed in #1043, so I took this now rather than deferring. analyze runs the same `validateWebIntent` check and reports it as a config finding, and stops listing services when the intent is ambiguous — the only entry there is the synthesized web the app never asked for, so listing it was the wrong claim rather than a missing one. Staying quiet about the others is consistent with how analyze treats the rest of the deploy validations.

docs/docs/app-toml.md
```diff
@@ -73,6 +83,27 @@ tail = "logs app -f"
...
+billed for compute) between invocations.
+
+:::note[Validation]
+If an app declares tasks, declares no services, and doesn't set `web`, the build
+fails rather than guessing. Without that, a task-only app would quietly acquire
+a web service built from its image entrypoint and keep it running forever.
+:::
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, thanks — the table now says "task-only app with no services", matching the validation note below it.

PR #1020, opened 2026-08-05
Adopt Go 1.26 idioms across the runtime

phinze:
With Go 1.26 established as the runtime baseline, we can retire compatibility scaffolding around several now-native patterns: pointer helpers, untyped error targets, indexed reflection walks, and manual benchmark loops.

This is the deliberately curated follow-up. Ordinary `errors.As` calls now use the type-safe `errors.AsType`; the one dynamic target in `MultiError.As` stays as-is. Config defaults use `new(expression)` from the generator source, struct-field walks use the new reflection iterator, and the remaining manual benchmark loops use `B.Loop`.

I kept the broader `go fix` suggestions out of this pass. Integer ranges, JSON tag cleanups, and general `reflect.TypeFor` conversions have a wider review surface and can stand on their own.

CI passes across lint, builds, five unit-test runners, and four blackbox shards. CodeRabbit and biscuit are both clean.


phinze:
Checked both caveats against the current tree. `github.com/pkg/errors` is still a direct dependency: it is imported by eight runtime files, including `controllers/sandbox/sandbox.go`, and `go mod why -m github.com/pkg/errors` confirms that direct path. `go mod tidy -diff` is clean.

The commented snippet in `pkg/rpc/client.go` was updated deliberately. This file now uses the standard-library `errors` package for `AsType`, so leaving `errors.Errorf` in the preserved block would make the snippet stale and uncompilable if it were restored. Deciding whether to delete or reinstate that older block is separate from this mechanical migration.

phinze:
The latest head tightens `pkg/rpc/client.go` to the mechanical scope: it keeps `github.com/pkg/errors` for the file’s existing `New`, `Is`, and `Errorf` behavior, and imports the standard library as `stderrors` only for `AsType`. That also removes the incidental edit inside the commented block. Targeted lint, `go test ./pkg/rpc`, and `go mod tidy -diff` are clean.

phinze:
Removed the obsolete commented-out status check. The live branch immediately above already handles both successful and non-OK responses, so the block was redundant. Targeted lint and `go test ./pkg/rpc` pass.

pkg/oidcauth/authenticator_test.go
```diff
@@ -330,6 +330,7 @@ func TestOIDCAuthenticator_RepositoryClaimBinding(t *testing.T) {
 			}
 			if identity == nil {
 				t.Fatal("expected identity, got nil")
+				return
```
evanphx: Non-blocking: this `return` is unreachable, and it's the one hunk in the PR that isn't one of the four idioms. I figured a nil-analyzer wanted it, so I pulled it out and ran both `go vet ./pkg/oidcauth/` and `golangci-lint run ./pkg/oidcauth/...` — clean either way. Looks like a leftover we can drop.

evanphx (approved):
Approved — feels great to get all that scaffolding out at once.

One non-blocking note inline about a stray `return` that seems to have snuck in with the sweep.

--e+🤖

pkg/oidcauth/authenticator_test.go
phinze: Plot twist: CI’s pinned staticcheck does require this guard. Removing it reports SA5011 on `identity.Subject`, since the analyzer doesn’t treat `t.Fatal` as proving `identity` non-nil. I restored the `return` so lint stays green.

PR #1022, opened 2026-08-05
Give saga executions a retention window

phinze:
Saga executions are durable by design but nothing ever deleted them, so every execution ever started was still in the entity store with a JSON blob of every action's output. Garden reached ~4,900 in a month of unremarkable use. MIR-1519 shows the untame version: one app that fails to bind its port retries on a loop, and each retry is a permanent entity. MIR-953 turns this on everywhere.

First rev is a prerequisite that's a bug on its own: `Execution` has always carried `CreatedAt`/`UpdatedAt` and the executor stamps them on every transition, but the schema had no fields for them and `Save` dropped them silently. Nothing could ask how old a saga was.

Second rev is a small cleanup that fell out of the third. `pkg/saga` had two near-identical in-memory `Storage` implementations, the exported `MemoryStorage` and an unexported double in `saga_test.go`. Adding interface methods meant writing them twice, identically, which is how it surfaced.

Third rev is the sweeper. RFD-35 wanted two mechanisms, a TTL plus compacting failed sagas so they could be kept longer. I backed off that: compaction only shrinks bytes within an entity while every number here is about entity count, and the long failure window was the only thing making it worthwhile. So it's one rule with one knob, seven days for terminal executions either way, exposed as `saga.retention_period` so it's tunable without a rebuild.

Retention goes through `Storage` rather than reaching into the entity store from the controller. The deciding detail is the legacy-timestamp fallback: executions written before the first rev read back with a zero `updated_at`, so the entity-backed stores fall back to the store's own timestamp instead of treating that as infinitely old and eating a saga an old-binary runner just finished. `MemoryStorage` has no second timestamp to fall back to. That's backend-specific behavior, which is what the interface is for. `ListTerminal` returns summaries rather than executions so a large backlog doesn't have to be materialized with its blobs just to read timestamps, which leaves the sweep trivial and `controllers/sagagc` a ticker. In-flight executions are never touched at any age, including undoing ones.

Verified three ways: unit tests across all three backends, a real-etcd pass asserting that deletion clears the status index entry (a stale one there would have every future sweep rediscovering a phantom), and a live standalone cluster where a real deploy's sagas were swept and the app kept serving.

Only open item is the club measurement the issue asked for, which still didn't happen: `miren debug entity list -k saga -C club` returned nothing in ten minutes before I killed it. That's MIR-1525, and it realistically wants MIR-1523 first.

Closes MIR-1526


cli/commands/server.go
```diff
@@ -741,6 +741,21 @@ func Server(ctx *Context, opts serverconfig.CLIFlags) error {
...
+	// back to the default instead, and say so.
+	sagaRetentionPeriod, err := units.ParseDuration(cfg.Saga.GetRetentionPeriod())
+	if err != nil {
+		defaultSaga := serverconfig.DefaultSagaConfig()
+		sagaRetentionPeriod, _ = units.ParseDuration(defaultSaga.GetRetentionPeriod())
+		ctx.Log.Warn("invalid saga.retention_period, falling back to default",
+			"value", cfg.Saga.GetRetentionPeriod(),
+			"default", sagaRetentionPeriod,
+			"error", err)
+	}
```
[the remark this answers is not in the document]
phinze: Fixed, though the failure mode isn't quite what's described here. `coordinate.go` guards with `if c.SagaRetentionPeriod >= 0`, so a negative never reaches `GCController.Start` and can't trip the `<= 0` disable path. It falls through to the 7d default and GC keeps running.

The real problem is that it does that *silently*: someone who typo'd `-7d` gets no signal their setting went nowhere. Negatives now take the same path as unparseable values, warn and fall back to the default. Zero stays the only way to disable deletion.

controllers/sagagc/gc.go
```diff
@@ -0,0 +1,153 @@
...
+	result, err := saga.RunRetention(sweepCtx, c.Storage, saga.RetentionConfig{
+		Retention:  c.Config.Retention,
+		MaxDeletes: c.Config.MaxDeletesPerSweep,
+	})
+	if err != nil {
+		// A truncated sweep (deadline, shutdown) is expected and not alarming;
+		// the next one resumes. Warn so a persistent hard error stays visible.
+		c.Log.Warn("saga retention sweep ended early", "error", err,
+			"scanned", result.Scanned, "deleted", result.Deleted)
+		return
```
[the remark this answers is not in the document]
phinze: Good catch, taken. I'd followed `indexgc`'s uniform Warn, but you're right that our own log-level guidance splits these: a store that won't answer is the platform failing at something it owns, whereas hitting our own sweep deadline or shutting down is expected and resumes next tick. Now `context.Canceled`/`DeadlineExceeded` stay at Warn and everything else logs at Error.

pkg/saga/retention.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	terminal, err := storage.ListTerminal(ctx)
+	if err != nil {
+		return result, err
+	}
+
+	cutoff := time.Now().Add(-cfg.Retention)
+
+	for _, exec := range terminal {
```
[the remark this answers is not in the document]
phinze: Good catch, taken. Spending the whole budget on the final execution is a complete sweep, and reporting it as capped would tell an operator a backlog remains on a store that just converged.

Also tightened the field's doc comment while I was in there. It claimed `Capped` meant "there is known remaining work," which overstates it even with the index fix: the executions we never reached might all be inside the retention window. It now says what it actually means, that the sweep stopped before it finished looking. Added the exact-boundary test too.

pkg/saga/storage.go
```diff
@@ -178,6 +188,107 @@ func (s *EntityStorage) ListIncomplete(ctx context.Context) ([]*Execution, error
...
+// The store's index lookups are equality-only, so there is no range query over
+// a timestamp that would let us ask for expired executions directly. We list
+// the terminal set (IDs only, which stays cheap) and read each one's finish
+// time, fetching in batches and keeping only the summary.
+func (s *EntityStorage) ListTerminal(ctx context.Context) ([]TerminalExecution, error) {
+	var ids []entity.Id
+	seen := make(map[entity.Id]struct{})
+
+	for _, status := range []entity.Id{
+		saga_v1alpha.SagaStatusCompletedId,
```
evanphx, line 212: Non-blocking, but worth writing down somewhere: this selects on status alone, so a completed *child* saga is eligible while its still-running parent is protected. `Recover` deliberately doesn't recover children (`executor.go:473`) — it leans on the parent re-finding them by deterministic ID in `createChildExecution`. Delete the child and that lookup misses, so the parent re-runs the nested saga instead of reusing it.

Traced it and it isn't reachable today: resume skips already-recorded actions, so you'd need a parent that crashed in the gap between the child's last save and the parent persisting its own action result, then went unrecovered past the window. And both production `RunNested` callers do their own entity lookup before ever reaching `RunNested`, so they short-circuit regardless.

So nothing to fix — just that the next nested saga written without that pre-check inherits a hole. A `ParentExecutionId == ""` filter here would close it cheaply if you want the belt-and-suspenders.

controllers/sagagc/gc.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// DefaultGCConfig returns the default configuration. The seven-day retention is
+// what RFD-35 committed to. The per-sweep cap and interval together drain about
+// 48,000 executions a day, comfortably ahead of the worst observed write rate.
+func DefaultGCConfig() GCConfig {
+	return GCConfig{
+		Retention:          7 * 24 * time.Hour,
+		CheckInterval:      30 * time.Minute,
+		MaxDeletesPerSweep: 1000,
```
evanphx, line 57: Non-blocking. This caps the writes, but every sweep still lists and materializes the whole terminal set to find those 1000 — the cap doesn't shorten the scan. I sized it and it's fine (~500 `GetEntities` round trips per 100k entities, well inside the 10m `SweepTimeout`), so I'm not raising it as a problem.

The bit worth knowing is the shape if it ever did blow the timeout: `ListTerminal` returns the ctx error before a single delete runs, so a sweep that can't finish listing makes zero progress rather than partial — and would stay that way every tick. The delete loop preserves partial work; the listing phase is all-or-nothing.

pkg/saga/retention_etcd_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"an ancient running saga must survive: recovery still needs it")
+
+	// The index must not still be advertising the deleted executions. A stale
+	// entry here is invisible to Get but would make every future sweep rescan
+	// entities that no longer exist.
+	completedIds, err := store.ListIndex(ctx, entity.Ref(
+		saga_v1alpha.SagaStatusId, saga_v1alpha.SagaStatusCompletedId))
+	require.NoError(t, err)
+	assert.NotContains(t, completedIds, entity.Id("etcd-old-completed"),
+		"deleting an execution must clean up its status index entry")
```
evanphx, line 70: 👏 This is the assertion I'd have been most likely to skip and most likely to regret. A delete that left the index entry behind is invisible to `Get`, so every later sweep would keep rediscovering a phantom — and nothing else in the suite would have caught it.

pkg/saga/storage.go
```diff
@@ -178,6 +188,107 @@ func (s *EntityStorage) ListIncomplete(ctx context.Context) ([]*Execution, error
 	return executions, nil
 }
 
+// terminalFetchBatch is how many entities ListTerminal materializes at a time.
+// Every entity carries its action-output blobs, so a store holding a large
+// backlog must not be pulled into memory all at once just to read timestamps
+// off it.
+const terminalFetchBatch = 200
```
evanphx, line 199: ✨ Worth flagging that this isn't redundant with the batching `GetEntities` already does internally — that one only chunks the etcd txn ops, then still allocates and returns the full `[]*Entity`. This outer batch is what actually bounds memory. Easy to misread as belt-and-suspenders and delete later.

evanphx (approved):
Approved! Good to have these stop piling up forever.

Four inline notes, all non-blocking — the `ListTerminal` one about nested children is the only one worth a real think, and even that's covered by both callers today.

--e+🤖

pkg/saga/storage.go
phinze: Good catch, and thanks for tracing reachability instead of just flagging the shape. Fixed it here rather than filing it, since once I priced the fix it was smaller than the follow-up would have been worth.

One amendment to the suggested filter though: `ParentExecutionId == ""` would mean child executions are never collected at all. Both `RunNested` callers are addon provisioning (`ensure-shared-mysql-server`, `ensure-shared-server`), so children do get written in production, and we'd be trading a rare correctness hole for a permanent leak of exactly the kind this PR closes. Smaller volume than create-sandbox, but still unbounded.

The rule that holds is narrower: a child is safe once its parent can no longer re-find it, so parent terminal or parent gone. That turned out cheap because `ListIncomplete` already returns precisely the in-flight set, making it a set membership test rather than a lookup per child. The extra listing only happens when the terminal set actually contains a child, so clusters running no nested sagas pay nothing. Held-back children are counted in a new `Skipped` field so the sweep log says so instead of staying quiet.

controllers/sagagc/gc.go
phinze: Agreed, and it's the asymmetry I'd stopped thinking about. Worth adding that the failure mode isn't just "no progress this tick," it's self-perpetuating: the terminal set only grows, so a sweep that can't finish listing won't finish the next one either. If we ever approach that, the fix is making the listing resumable rather than raising the timeout.

pkg/saga/retention_etcd_test.go
phinze: It nearly didn't get written. I only reached for it after realizing the mock couldn't vouch for index cleanup, which is exactly the property MIR-1320 was about.

pkg/saga/storage.go
phinze: Good distinction, and exactly what gets tidied away in two years by someone who assumes `GetEntities` already handles it. Put it in the comment so it's defended in place: that batching chunks the etcd txn ops and still returns the full slice, so it bounds request size, not memory.

PR #1023, opened 2026-08-05
Add an etcdctl debug helper

phinze:
Getting at embedded etcd currently means either dropping down to `miren debug ctr`, remembering the container and task incantation, then reconstructing the active endpoint and mTLS flags, or installing `etcdctl` separately and wiring up those same connection details by hand. Both work, but both are fiddly little rituals for a common debugging job.

This adds `miren debug etcdctl ...` as the direct path. The command reads the live etcd container spec, supplies the advertised endpoint and mounted client credentials when TLS is enabled, and passes the remaining arguments through unchanged. It still delegates the actual exec to `ctr`, so its I/O, signal, and exit behavior stays familiar.

The command is deliberately local to coordinators running embedded etcd. Missing containers, inaccessible sockets, and connection failures now explain that requirement rather than leaking an opaque containerd error.

The `make dev` smoke test exercised the mTLS path against etcd 3.5.15: endpoint status and health succeeded, a limited prefix query returned live Miren keys, and etcdctl flags worked both before and after the subcommand. The CLI package tests pass and lint is clean.

### DIFF
--- a/cli/commands/debug_etcdctl.go
+++ b/cli/commands/debug_etcdctl.go
@@ -1,0 +1,198 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/pkg/namespaces"
+	"github.com/containerd/errdefs"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"miren.dev/runtime/components/etcd"
+)
+
+const debugEtcdctlProbeTimeout = 5 * time.Second
+
+const debugEtcdctlDescription = `Place Miren's ` + "`" + `-n` + "`" + `/` + "`" + `--namespace` + "`" + ` and ` + "`" + `--socket` + "`" + ` flags before the etcdctl arguments. Once the etcdctl arguments begin, all remaining flags are passed through to etcdctl unchanged.`
+
+type debugEtcdctlOptions struct {
+	Namespace   string   `short:"n" long:"namespace" description:"containerd namespace" default:"miren"`
+	Socket      string   `long:"socket" description:"path to containerd socket"`
+	Args        []string `rest:"true"`
+	Passthrough []string `unknown:"true"`
+}
+
+func (o debugEtcdctlOptions) etcdctlArgs() []string {
+	args := make([]string, 0, len(o.Args)+len(o.Passthrough))
+	args = append(args, o.Args...)
+	return append(args, o.Passthrough...)
+}
+
+func DebugEtcdctl(ctx *Context, opts debugEtcdctlOptions) error {
+	socket := opts.Socket
+	if socket == "" {
+		socket = defaultContainerdSocket()
+	}
+
+	client, err := containerd.New(socket)
+	if err != nil {
+		return localEtcdUnavailableError(socket, opts.Namespace, err)
+	}
+	defer func() {
+		if client != nil {
+			client.Close()
+		}
+	}()
+
+	containerCtx, cancel := context.WithTimeout(
+		namespaces.WithNamespace(ctx, opts.Namespace),
+		debugEtcdctlProbeTimeout,
+	)
+	defer cancel()
+
+	container, err := client.LoadContainer(containerCtx, etcd.ContainerName)
+	if err != nil {
+		return localEtcdUnavailableError(socket, opts.Namespace, err)
+	}
+	if _, err := container.Task(containerCtx, nil); err != nil {
+		return localEtcdTaskUnavailableError(err)
+	}
+
+	spec, err := container.Spec(containerCtx)
+	if err != nil {
+		return fmt.Errorf("reading etcd container spec: %w", err)
+	}
+
+	args, err := buildEtcdctlExecArgs(spec, opts.etcdctlArgs(), newEtcdctlExecID())
+	if err != nil {
+		return err
+	}
+	cancel()
+
+	_ = client.Close()
+	client = nil
+
+	return execCtr(socket, opts.Namespace, args)
+}
+
+func localEtcdUnavailableError(socket, namespace string, err error) error {
+	if errdefs.IsNotFound(err) {
+		return fmt.Errorf(
+			"embedded etcd container %q was not found in containerd namespace %q; "+
+				"this command must be run on a local Miren coordinator with embedded etcd",
+			etcd.ContainerName, namespace,
+		)
+	}
+
+	if errdefs.IsPermissionDenied(err) || strings.Contains(strings.ToLower(err.Error()), "permission denied") {
+		return fmt.Errorf(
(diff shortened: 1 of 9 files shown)


cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+const etcdContainerName = "miren-etcd"
+const debugEtcdctlProbeTimeout = 5 * time.Second
+
+type debugEtcdctlOptions struct {
+	Namespace   string   `short:"n" long:"namespace" description:"containerd namespace" default:"miren"`
+	Socket      string   `long:"socket" description:"path to containerd socket"`
+	Args        []string `rest:"true"`
+	Passthrough []string `unknown:"true"`
```
evanphx, line 26: Once any etcdctl flag shows up, mflags swallows everything from there to the end into `Passthrough` — including our own `-n` and `--socket`. So `miren debug etcdctl get / --prefix --socket /custom.sock` quietly hands `--socket` to etcdctl, which rejects it.

That's the right trade, since flags-working-after-the-subcommand is the more valuable property. But nobody's going to guess it. `WithDescription` on the registration flows into an extended-description section in the generated docs — a line there saying miren's own flags go first would cover it. Non-blocking.

cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}()
+
+	containerCtx, cancel := context.WithTimeout(
+		namespaces.WithNamespace(ctx, opts.Namespace),
+		debugEtcdctlProbeTimeout,
+	)
+	defer cancel()
+
+	container, err := client.LoadContainer(containerCtx, etcdContainerName)
```
evanphx: `LoadContainer` succeeds when the container exists but has no running task, so a stopped etcd falls straight through to `ctr: no running task found: not found` — the one case that still leaks containerd's wording after all the care taken above it.

Worth a `container.Task(ctx, nil)` probe here with an error in the same voice as the others: the container's there but not running, go look at `miren server` / the logs. Non-blocking, but it's the gap I'd close first.

cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+	"strings"
+	"time"
+
+	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/containerd/v2/pkg/namespaces"
+	"github.com/containerd/errdefs"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const etcdContainerName = "miren-etcd"
```
evanphx: 💡 Third copy of `"miren-etcd"` now — `components/etcd/etcd.go:27`, `cli/commands/server.go:355`, and here — and `/certs/*` plus `/usr/local/bin/*` are mirrored out of that package too. Already duplicated before this PR so not a regression, but a rename would break this command silently at runtime with nothing standing in the way. Exporting the name from `components/etcd` would pin at least the one that matters. Non-blocking.

cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	args, err := buildEtcdctlExecArgs(spec, opts.etcdctlArgs(), newEtcdctlExecID())
+	if err != nil {
+		return err
+	}
+	cancel()
+
+	if err := client.Close(); err != nil {
+		return fmt.Errorf("closing containerd client: %w", err)
```
evanphx: Small shape thing: a failed close aborts before we ever exec. I checked whether it can fire and it can't today — the first close on a grpc conn returns nil — so this isn't a bug report. But a close error here isn't something anyone could act on, and ignoring it keeps a working command working. Non-blocking.

cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+		etcdContainerName,
+		"/usr/local/bin/etcdctl",
+	}
+	args = append(args, connectionArgs...)
+	args = append(args, etcdctlArgs...)
+	return args, nil
+}
+
+func etcdClientEndpoint(processArgs []string) (string, bool, error) {
+	endpoint := processFlagValue(processArgs, "--advertise-client-urls")
```
evanphx, line 152: 👏 Reaching for the advertised URL first instead of the listen URL. Under TLS the listen side binds `0.0.0.0` while the cert only carries `localhost`/`127.0.0.1`, so the other ordering would fail verification and send someone hunting for a cert problem that isn't there. Easy to get backwards.

cli/commands/debug_etcdctl.go
```diff
@@ -0,0 +1,183 @@
...
+	args, err := buildEtcdctlExecArgs(spec, opts.etcdctlArgs(), newEtcdctlExecID())
+	if err != nil {
+		return err
+	}
+	cancel()
+
+	if err := client.Close(); err != nil {
+		return fmt.Errorf("closing containerd client: %w", err)
+	}
+	client = nil
```
evanphx, line 77: ✨ Closing and nil-ing the client before the exec. `syscall.Exec` never comes back so the defer would never run — that's the kind of thing that normally gets noticed six months later, if ever.

evanphx (approved):
Approved! Glad to never type the `ctr tasks exec` incantation again.

Four non-blocking notes inline. The flag-ordering one is the only one a user will actually hit; the stopped-etcd one is worth a real error message instead of letting `ctr` do the talking.

--e+🤖

cli/commands/debug_etcdctl.go
phinze: Good catch. We documented that `-n`/`--namespace` and `--socket` need to come before the etcdctl arguments. Preserving arbitrary etcdctl flag order is the more useful side of that parser tradeoff.

cli/commands/debug_etcdctl.go
phinze: Yep, this was the missing edge. We now probe the task before exec and return a Miren-owned “embedded etcd is not running; check the server and coordinator logs” error.

cli/commands/debug_etcdctl.go
phinze: Good call. We exported the container name from `components/etcd` and consume it here. One wrinkle: the identical string in `server.go` is the etcd member name rather than the container ID, so that one stays separate.

cli/commands/debug_etcdctl.go
phinze: Agreed. A close failure is not actionable here and should not prevent an otherwise valid exec, so we now ignore it after explicitly closing before `syscall.Exec`.

cli/commands/debug_etcdctl.go
phinze: Yep, that TLS bind/advertise split is exactly why the live spec is the source of truth here.

cli/commands/debug_etcdctl.go
phinze: Thanks! The `syscall.Exec` lifetime makes this worth being explicit about.

PR #1026, opened 2026-08-05
Add `miren debug saga` for listing and inspecting executions

phinze:
Inspecting a stuck saga meant running `debug entity list -k saga` and reading base64 out of the output by hand. The fields you actually want (initial inputs, per-action results, execution order) are stored as JSON blobs, so the raw entity view shows you bytes rather than what happened. RFD-35 left saga observability as an open question and it stayed open.

`debug saga list` defaults to active executions only, meaning pending, running, and undoing. Completed sagas dominate the store and are never what you're looking for when something is wedged, so `--all` widens and `--status` narrows. Both filters the issue asked for, definition name and status, were already indexed on the schema, so there's no migration and no reindex. `debug saga show` walks the execution order with per-action timing, undo state, errors, and decoded output, truncated by default and whole with `--full`. It also lists child sagas, since `parent_execution_id` is indexed too. Everything supports `--format json`.

Then building it made the ID format impossible to ignore, and prerelease is the window for that. Sagas were minting `idgen.Gen("saga")`, the generator meant for short opaque tokens that get embedded in longer names, rather than the `GenNS` every other entity uses. The result was `sagaCdjkwhnngwnS8AZYPSTcK`: the word running straight into lowercase base58 with no separator your eye can find, and no kind namespace at all, sitting next to things like `sandbox/sb-…` and `pool/…`. They're `saga/sg-…` now, with `deriveChildID` moved in lockstep so nested-saga recovery stays deterministic, plus a test pinning both shapes. Listings strip the namespace the way sandbox and app listings already do, and lookups accept the bare name, the qualified ID, or the legacy unnamespaced form, so the ID you copy off the table pastes straight back and existing sagas keep resolving without a migration.

Two things worth flagging. `show` renders the actions that ran, not the full set the definition declares, because the saga registry is built at server startup and isn't exposed over the API. So "where it stopped" is the last executed action plus the error, and a registry-listing RPC would unlock real DAG visualization if we want it.

And the CREATED column is blank on real sagas today. Saga storage saves by full replace, and `ReplaceEntity` carries the revision forward but not `db/entity.created`, so an execution loses its creation timestamp on the first save after the initial one. The MIR-1526 retention work is fixing that, so the column stays and starts working when it lands. Worth knowing how it surfaced: an in-memory integration test that saves through the real `EntityStorage` asserted the timestamp was present and passed, because MockStore's replace preserves the attr where EtcdStore's doesn't. Anything using `NewInMemEntityServer` to reason about persistence has that blind spot.

One known risk, taken deliberately. `deriveChildID` produces the idempotency key that `createChildExecution` uses to find an already-created nested child during recovery, so changing the ID format means a child created before this lands won't be found after it, and a parent re-running that action would spawn a second child and re-execute its actions. Top-level sagas are unaffected, since they recover by reading their stored ID rather than deriving it. We're prerelease and that's precisely why the format is moving now, so we're taking the narrow upgrade window rather than carrying the old derivation in the saga hot path indefinitely.

Closes MIR-1523

### DIFF
--- a/cli/commands/debug_saga.go
+++ b/cli/commands/debug_saga.go
@@ -1,0 +1,601 @@
+package commands
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"sort"
+	"strings"
+	"time"
+
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	saga_v1alpha "miren.dev/runtime/api/saga/saga_v1alpha"
+	"miren.dev/runtime/pkg/cond"
+	"miren.dev/runtime/pkg/entity"
+	"miren.dev/runtime/pkg/saga"
+	"miren.dev/runtime/pkg/ui"
+)
+
+// sagaOutputPreview is how much of an action's output we show before
+// truncating. Wide enough to recognize what came back, short enough that a
+// twenty-action saga still fits on a screen. Use --full to see all of it.
+const sagaOutputPreview = 100
+
+// sagaIDNamespace is the kind prefix execution IDs carry. Listings print IDs
+// without it, the way sandbox and app IDs are shown, and lookups put it back.
+const sagaIDNamespace = "saga/"
+
+// activeSagaStatuses are the non-terminal statuses. These are what `debug saga
+// list` shows by default: a completed saga is history, but a saga sitting in
+// one of these is either working or wedged, which is the thing you opened a
+// debug command to find out.
+var activeSagaStatuses = []saga.Status{
+	saga.StatusPending,
+	saga.StatusRunning,
+	saga.StatusUndoing,
+}
+
+// sagaRecord pairs a decoded execution with the entity store's own timestamps.
+// The saga schema does not persist Execution.CreatedAt/UpdatedAt, but the
+// entity store records when it created and last wrote each entity, and for a
+// saga those are the same events: it is written once per action. A running
+// saga whose entity has not been updated in an hour is the stuck one.
+type sagaRecord struct {
+	exec      *saga.Execution
+	createdAt time.Time
+	updatedAt time.Time
+}
+
+// lastAction returns the most recently executed action name, or "" if the saga
+// has not run anything yet.
+func (r *sagaRecord) lastAction() string {
+	if len(r.exec.ExecutionOrder) == 0 {
+		return ""
+	}
+	return r.exec.ExecutionOrder[len(r.exec.ExecutionOrder)-1]
+}
+
+func DebugSagaList(ctx *Context, opts struct {
+	FormatOptions
+	ConfigCentric
+	Status     string `short:"s" long:"status" description:"Filter by status (pending, running, undoing, completed, failed)"`
+	Definition string `short:"d" long:"definition" description:"Filter by saga definition name"`
+	All        bool   `short:"A" long:"all" description:"Include completed and failed sagas"`
+}) error {
+	var wantStatus saga.Status
+	if opts.Status != "" {
+		var err error
+		wantStatus, err = parseSagaStatus(opts.Status)
+		if err != nil {
+			return err
+		}
+	}
+
+	eac, err := sagaClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Pick the narrowest index the server can answer directly. definition_name
+	// and status are both indexed; when both filters are given we query by
+	// definition and narrow by status below, since a definition name is far
+	// more selective than a status.
+	var indexes []entity.Attr
+	switch {
+	case opts.Definition != "":
+		indexes = append(indexes, entity.String(saga_v1alpha.SagaDefinitionNameId, opts.Definition))
+	case wantStatus != "":
+		attr, err := sagaStatusIndex(wantStatus)
+		if err != nil {
+			return err
+		}
+		indexes = append(indexes, attr)
+	case opts.All:
+		res, err := eac.LookupKind(ctx, "saga")
+		if err != nil {
+			return fmt.Errorf("looking up saga kind: %w", err)
+		}
+		indexes = append(indexes, res.Attr())
+	default:
+		for _, s := range activeSagaStatuses {
+			attr, err := sagaStatusIndex(s)
+			if err != nil {
+				return err
+			}
+			indexes = append(indexes, attr)
+		}
+	}
+
+	records, err := listSagas(ctx, eac, indexes)
+	if err != nil {
+		return err
+	}
+
+	// Filtering to active statuses applies when the user asked for neither a
+	// specific status nor --all. Note that --definition does not opt out: a
+	// definition whose sagas have all completed still lists empty by default.
+	// The empty-state message below keys off this same variable rather than
+	// restating the condition, since the two drifting apart is what silently
+	// suppressed the --all hint in exactly the case that needed it.
+	filterActive := !opts.All && wantStatus == ""
+
+	// Apply whatever the index could not.
+	records = slices.DeleteFunc(records, func(r *sagaRecord) bool {
+		switch {
+		case wantStatus != "":
+			return r.exec.Status != wantStatus
+		case filterActive:
+			return !slices.Contains(activeSagaStatuses, r.exec.Status)
+		default:
+			return false
+		}
+	})
+
+	// Most recent activity first.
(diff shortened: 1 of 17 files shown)


pkg/saga/nested.go
```diff
@@ -168,5 +168,5 @@ func UndoNested(ctx context.Context, executionID string) error {
 // createChildExecution's idempotency check to find the prior child execution.
 func deriveChildID(parentExecID, sagaName, actionName string) string {
 	h := sha256.Sum256([]byte(parentExecID + "\x00" + sagaName + "\x00" + actionName))
-	return "saga" + base58.Encode(h[:16])
+	return sagaIDKind + "/" + sagaIDName + "-" + base58.Encode(h[:16])
```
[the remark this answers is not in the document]
phinze: Good catch on the mechanism, and we're deliberately not fixing it.

You're right about how it breaks: `createChildExecution` uses the derived ID as its idempotency key, so a nested child created before this change won't be found after it, and a parent re-running that action during recovery would spawn a second child and re-execute its actions.

Worth being precise about the blast radius, though. Top-level sagas recover by reading their stored ID rather than deriving it, so only nested children created before the upgrade are exposed, and only if their parent is mid-flight across that exact restart.

We're prerelease, which is the reason the ID format is moving at all. A legacy fallback means keeping the old derivation alive in the saga hot path with no natural point where we'd feel safe deleting it, so we're taking the narrow window instead. It's noted in the PR description so it reads as a recorded decision rather than an oversight.

cli/commands/debug_saga.go
```diff
@@ -0,0 +1,590 @@
...
+	if opts.IsJSON() {
+		items := make([]sagaListJSON, 0, len(records))
+		for _, r := range records {
+			items = append(items, newSagaListJSON(r))
+		}
+		return PrintJSON(items)
+	}
+
+	if len(records) == 0 {
+		if opts.All || wantStatus != "" || opts.Definition != "" {
```
evanphx: Non-blocking, but this one will actually bite someone. `--definition X` on its own still filters to active statuses, so a definition whose sagas have all completed gets a flat "No sagas found" — with the `--all` hint suppressed exactly when the user most needs to know the default is still in play.

Dropping `|| opts.Definition != ""` gets it right: the hint should key off whether we're actually filtering to active, which is `!opts.All && wantStatus == ""`.

cli/commands/debug_saga.go
```diff
@@ -0,0 +1,590 @@
...
+
+		if result := r.exec.ExecutedActions[name]; result != nil {
+			action.ExecutedAt = sagaJSONTime(result.ExecutedAt)
+			if result.UndoneAt != nil {
+				action.UndoneAt = sagaJSONTime(*result.UndoneAt)
+			}
+			action.Error = result.Error
+			// Truncated output would not be valid JSON, so JSON callers get
+			// the whole thing or nothing. --full opts in.
+			if full && len(result.Output) > 0 {
```
evanphx: Non-blocking. The reasoning here is right — truncated output isn't valid JSON — but it doesn't carry to `initial_inputs` on line 551, which go out whole and unconditionally. Text mode truncates both; JSON mode truncates neither and silently drops one.

Net effect is that plain `--format json` hands a machine a partial record with nothing marking what's missing. Either always emit outputs in JSON, or hold inputs to the same rule. My weak preference is always-emit — a debug command's JSON is for piping into `jq`, and size feels like the human reader's problem, not the parser's.

cli/commands/debug_saga.go
```diff
@@ -0,0 +1,590 @@
...
+		return err
+	}
+
+	// Nested sagas record their parent, and that attribute is indexed, so the
+	// children of this execution are one query away. The ref value is hoisted
+	// into its own variable because the guardrail in guardrails/refenum_test.go
+	// flags entity.Id(…) sitting directly in an entity.Ref, which is how the
+	// short-enum-value bug in MIR-1288 looks. This is a real execution ID, not
+	// an enum value.
+	parentID := entity.Id(record.exec.ID)
```
evanphx, line 232: I checked this rather than take it on faith — inlined `entity.Id(record.exec.ID)` back into the `entity.Ref` and ran the guardrail, and it does fail with `entity.Id(ID)` pointing at this line. So the hoist is load-bearing.

No change wanted here. My worry is the next person who tidies away the "unnecessary" variable and gets a CI failure about MIR-1288 and short enum values, on a line containing neither. The guardrail's own doc comment already grants that `entity.Id(x.ID)` is legitimate — it just carves it out everywhere except inside a `Ref`. Feels like the matcher could learn the difference; happy to file that separately so it doesn't ride on this PR.

cli/commands/debug_saga_index_test.go
```diff
@@ -0,0 +1,118 @@
...
+		require.NoError(t, err)
+
+		records, err := listSagas(ctx, inmem.EAC, []entity.Attr{attr})
+		require.NoError(t, err)
+		require.NotEmpty(t, records)
+
+		// The saga schema has no timestamp fields, so the UPDATED column is
+		// only meaningful if the store stamps the entity itself.
+		//
+		// createdAt is not asserted, and the omission is deliberate: it passes
```
evanphx, line 66: 👏 This is the good stuff. The easy version of this test asserts `createdAt`, goes green, and quietly lies. Leaving it out and writing down that MockStore's replace carries `db/entity.created` where EtcdStore's doesn't is the harder call and the right one — and it means anyone reaching for `NewInMemEntityServer` to reason about persistence now has a heads-up.

cli/commands/debug_saga.go
```diff
@@ -0,0 +1,590 @@
...
+	return entityserver_v1alpha.NewEntityAccessClient(client), nil
+}
+
+// listSagas queries each index and returns the decoded executions, deduplicated
+// by ID. A saga can appear under more than one index at once (a stale pending
+// entry lingering after the transition to running), so the caller would
+// otherwise see the same execution twice.
+func listSagas(ctx *Context, eac *entityserver_v1alpha.EntityAccessClient, indexes []entity.Attr) ([]*sagaRecord, error) {
+	var records []*sagaRecord
+	seen := make(map[string]struct{})
```
evanphx, line 456: ✨ Nice that the read path picked up the same lesson `ListIncomplete` learned the hard way — a saga sitting under two status indexes mid-transition is the exact shape that caused double-recovery over in `storage.go`. Cheap here, and it means the listing can't ever show you the same wedged saga twice while you're trying to count them.

evanphx (approved):
Approved! Nice that a wedged saga is just readable now.

Three inline notes, all non-blocking — the `--all` hint one on `debug_saga.go:143` is the only one a user will actually hit. Traced the ID migration end to end and the blast radius is exactly what you described: top-level sagas recover from stored IDs, so only a pre-upgrade nested child resuming mid-flight is affected.

--e+🤖

cli/commands/debug_saga.go
phinze: Fixed, and you're right that this is the one that bites. The filter treats active-only as `!opts.All && wantStatus == ""`, and the message was restating that condition with `--definition` bolted on, so the two disagreed exactly when it mattered.

Rather than just dropping the clause I bound both to a single `filterActive` variable. The restatement is the actual bug and it could grow back the same way, so now the message can't disagree with the filter unless someone deliberately changes it.

cli/commands/debug_saga.go
phinze: Fixed, going with always-emit. My reasoning only covered half the payload: truncated output isn't valid JSON, sure, but `initial_inputs` went out whole regardless, so plain `--format json` handed a machine a partial record with nothing marking the gap.

Outputs are now unconditional in JSON and `--full` is purely a text-mode concern. Added a test pinning both halves present together so the asymmetry can't reappear, and said so in the flag help and the command docs.

cli/commands/debug_saga.go
phinze: Agreed, and thanks for re-running it rather than taking the comment on faith.

Your worry is the right one. The failure names MIR-1288 and short enum values on a line containing neither, so someone who tidies the variable away has nothing connecting the two. The comment already says the hoist is load-bearing and why, which helps whoever reads the code but not whoever only reads the CI output.

Grabbed the follow-up rather than making you file it: MIR-1539, with the multi-part-selector heuristic as the cheapest fix (a package-qualified enum is always exactly `pkg.NAME`, so `x.y.ID` could be read as a field access). Teaching the matcher the difference beats another allowlist entry keyed by line number.

PR #1028, opened 2026-08-05
Reference secrets from config and materialize them into sandboxes

evanphx:
Second of three stacked PRs implementing [RFD-90: Secret Backends and Materialization](https://github.com/mirendev/rfd/blob/main/rfds/0075/0090-secret-backends-and-materialization.md) ([MIR-1343](https://linear.app/miren/issue/MIR-1343)). Stacked on #1027, which built the storage half.

This PR makes a stored secret consumable by an app: a variable can point at a backend instead of holding a value, and the value arrives in the container's environment without ever being written down.

## Pinning

An authored reference usually floats — `payments/stripe-key`, meaning whatever is current. What a ConfigVersion records is what that resolved to at mint time: `payments/stripe-key@x1A`. This makes "which secret did this version actually ship with?" answerable for the most sensitive input a deploy has, and it makes a rollback come back on the value the old version ran with rather than today's.

Pinning is idempotent, because resolving an already-pinned reference returns the same reference. That single property is what lets a hand-set reference stay at the version it was set to while an `app.toml` reference re-pins on every deploy, with **no second piece of state to keep in sync**.

Per the RFD's open question, `miren env set` **pins at set-time**. A rotation therefore reaches a hand-set variable when you re-run `env set`, and reaches an `app.toml` reference on the next `miren deploy`.

## Not writing secrets down

A sandbox spec is a persisted entity. Resolving a secret while building it would put the plaintext in etcd — the exact thing referencing a secret rather than inlining it is meant to prevent. So the spec carries a reference, and the value comes into being in the sandbox controller at container creation, on its way into the process environment.

The reference doubles as the unit of change detection for pool reuse: it carries a concrete version, so it differs precisely when the bytes behind it would differ, without the pool comparison ever seeing a secret.

Two paths build sandbox specs — the deploy launcher and exec sandboxes — and initially I only fixed the first, which showed up as `miren app run -- printenv` printing a path where a credential belonged. Both now go through one function, so "does a secret ever get written into a spec?" has a single answer rather than one per call site.

## Failing closed

Materialization fails rather than degrading. An app handed a reference where its credential belongs does not fail at boot in a way anyone can read — it fails much later, inside whatever the credential was for. A missing resolver, an unknown backend, a revoked version, and an unreadable reference all stop the container, where the error can name the variable. Errors name the variable and the backend, never the value.

A config that references no backend never touches a resolver at all, so a cluster with no secrets in play pays nothing and cannot fail here.

## Authoring

```bash
miren env set -e STRIPE_API_KEY --backend cluster --ref payments/stripe-key
```

```toml
[[env]]
key = "STRIPE_API_KEY"
backend = "cluster"
ref = "payments/stripe-key"
```

**Deviation from the RFD, flagged deliberately**: RFD-90's examples use an `[env]` map form (`KEY = { backend = …, ref = … }`). The shipped format is `[[env]]` array-of-tables, so this follows the shipped format rather than introducing a second one.

A reference replaces the value rather than accompanying it, so supplying both is rejected: guessing which was meant is how a literal path ends up deployed where a credential belongs. `env list` shows the reference and its pin rather than a mask, since there is nothing local to hide.

## Note for reviewers

The frozen-`sandbox.go` guard asks for an audit of the saga path before rehashing. Both the direct and saga paths reach container creation through `SandboxController.BuildSpec`, and the materialization sits inside `buildSubContainerSpec` beneath it, so the saga controller picks up the same behavior with no parallel change. The last commit is just that rehash.

## Testing

Unit coverage for pinning (including idempotence and fail-closed), sentinel parsing, and env materialization. A launcher test asserts that a persisted sandbox pool contains only references and no resolved value, however the spec is serialized.

Verified end to end in the dev environment against a real app:

- `secret set` → `env set --backend --ref` → deploy → the app serves the real value
- rotating the secret leaves the running app untouched; re-running `env set` re-pins and rolls it out
- an `app.toml` reference floats and re-pins on a plain `miren deploy`
- `miren app run -- printenv` materializes (this is what caught the second spec path)
- disabling the pinned version makes the next deploy fail closed: `cannot resolve variable STRIPE_API_KEY (cluster:payments/stripe-key): secret version is not enabled` — while the already-running app keeps serving
- entity inspection shows each ConfigVersion permanently recording its own pin (`@QEa`, `@Bvm`, `@oSX`), which is what makes rollback restore the right secret

Full suite green (5667 tests).

## What's next

**PR 3** — runner-node resolution over RPC (required: a distributed runner holds no key material today), tmpfs file secrets, version reaping that respects pins, and docs.

### DIFF
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -27,9 +27,12 @@ type AppEnvVar struct {
 	// never appears in app.toml, only a pointer to it, so the file stays safe to
 	// commit.
 	//
-	// A reference authored here floats — it re-pins to whatever is current on
-	// every deploy, which is how a rotation reaches the app. Appending @version
-	// to Ref holds it at that exact version instead.
+	// A reference authored here floats: each new ConfigVersion resolves it to
+	// whatever is current, which is how a rotation reaches the app. Note that
+	// this is about minting a config, not running one — deploying or rolling
+	// back to an existing version keeps the reference that version recorded, so
+	// a rotation reaches an app only once a new config is created for it.
+	// Appending @version to Ref holds it at that exact version regardless.
 	Backend string `json:"backend,omitempty" toml:"backend,omitempty"`
 	Ref     string `json:"ref,omitempty" toml:"ref,omitempty"`
 }
--- a/appconfig/appconfig.go
+++ b/appconfig/appconfig.go
@@ -21,6 +21,17 @@ type AppEnvVar struct {
 	Required    bool   `json:"required,omitempty" toml:"required,omitempty"`
 	Sensitive   bool   `json:"sensitive,omitempty" toml:"sensitive,omitempty"`
 	Description string `json:"description,omitempty" toml:"description,omitempty"`
+
+	// Backend names the secret backend the value comes from, and Ref addresses
+	// the secret within it. Together they replace Value: the credential itself
+	// never appears in app.toml, only a pointer to it, so the file stays safe to
+	// commit.
+	//
+	// A reference authored here floats — it re-pins to whatever is current on
+	// every deploy, which is how a rotation reaches the app. Appending @version
+	// to Ref holds it at that exact version instead.
+	Backend string `json:"backend,omitempty" toml:"backend,omitempty"`
+	Ref     string `json:"ref,omitempty" toml:"ref,omitempty"`
 }
 
 type BuildConfig struct {
@@ -174,6 +185,9 @@ func (ac *AppConfig) Validate() error {
 		if ev.Key == "" {
 			return &ValidationError{KeyPath: "env", Message: fmt.Sprintf("env[%d]: key is required", i)}
 		}
+		if err := ev.validateReference(fmt.Sprintf("env[%d]", i)); err != nil {
+			return err
+		}
 	}
 
 	// Validate service configurations
@@ -269,6 +283,9 @@ func (ac *AppConfig) Validate() error {
 		// Validate service environment variables
 		// Note: empty values are allowed - secrets may be stored server-side
 		for i, ev := range svcConfig.EnvVars {
+			if err := ev.validateReference(fmt.Sprintf("%s.env[%d]", svcPrefix, i)); err != nil {
+				return err
+			}
 			if ev.Key == "" {
 				return &ValidationError{
 					KeyPath: svcPrefix + ".env",
@@ -551,4 +568,31 @@ func AliasLineNumbers(configPath string) map[string]int {
 	}
 
 	return result
+}
+
+// validateReference checks the secret-reference form of an env var. A reference
+// replaces the value rather than accompanying it, so accepting both would leave
+// it ambiguous which one is meant to reach the app — and the wrong guess ships a
+// literal path where a credential belongs.
+func (ev AppEnvVar) validateReference(keyPath string) error {
+	switch {
+	case ev.Ref == "" && ev.Backend == "":
+		return nil
+	case ev.Ref == "":
+		return &ValidationError{
+			KeyPath: keyPath + ".ref",
+			Message: fmt.Sprintf("%s: backend %q needs a ref naming the secret", keyPath, ev.Backend),
+		}
+	case ev.Backend == "":
+		return &ValidationError{
+			KeyPath: keyPath + ".backend",
+			Message: fmt.Sprintf("%s: ref %q needs a backend to resolve it against", keyPath, ev.Ref),
+		}
+	case ev.Value != "":
+		return &ValidationError{
+			KeyPath: keyPath + ".value",
+			Message: fmt.Sprintf("%s: set either value or ref, not both — a referenced secret gets its value from the backend", keyPath),
+		}
+	}
+	return nil
 }
--- a/cli/commands/env.go
+++ b/cli/commands/env.go
@@ -31,6 +31,19 @@ func maskURLUserinfo(value string) string {
 	return urlUserinfoRe.ReplaceAllString(value, "${1}"+envRedacted+"@")
 }
 
+// displayEnvValue renders what `env list`/`env get` shows for one variable.
+//
+// A backend-sourced variable has no local value to mask: the bytes never
+// travelled here, only the reference did. Showing the reference — backend,
+// path, and the version this config pinned — is both safe and the only useful
+// thing to show, so --unmask has nothing extra to reveal for one.
+func displayEnvValue(value, backend string, sensitive, unmask bool) string {
+	if backend != "" {
+		return "\u2192 " + backend + ":" + value
+	}
+	return maskEnvValue(value, sensitive, unmask)
+}
+
 // maskEnvValue is the single source of truth for how an env-var value is
 // displayed anywhere in the CLI (env list/get, deploy analysis, and any future
 // surface). Sensitivity is decided by EVIDENCE, in tiers, and this function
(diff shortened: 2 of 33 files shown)


servers/app/app.go
```diff
@@ -522,6 +529,7 @@ func (r *AppInfo) GetConfiguration(ctx context.Context, state *app_v1alpha.CrudG
 		env.SetSource(ev.Source)
 		env.SetRequired(ev.Required)
 		env.SetDescription(ev.Description)
+		env.SetBackend(ev.Backend)
```
phinze, line 541: This is the read side, and nothing reads it back. `SetConfiguration` (~353) does `spec.Variables = nil` and rebuilds from the client's `NamedValue`s copying Key/Value/Sensitive/Source/Required/Description, with no `Backend`. The per-service conversion near 389 has the same gap, and the two `EnvVarInput` adapters at 645/670 are the same shape.

Nothing reaches any of them right now, so there's nothing urgent here. But the failure shape is worth noting: a variable that loses its backend becomes an ordinary literal holding a reference path, and every guard downstream then reads it as fine. `pinSecrets` finds no refs to pin, `EnvValue` passes the value through, `MaterializeEnv` sees no sentinel prefix. Silent by construction, which argues for copying the field at all three sites while the context is loaded rather than leaving it for whoever wires up the next caller.

pkg/secret/sentinel.go
```diff
@@ -0,0 +1,72 @@
...
+	return SentinelScheme + backend + "/" + ref
+}
+
+// EnvValue renders what a config variable contributes to a sandbox spec: its
+// literal value, or a reference standing in for one when it is backend-sourced.
+//
+// Every path that builds a sandbox spec from a config goes through this, so
+// there is one answer to "does a secret ever get written into a spec?" rather
+// than one per call site.
+func EnvValue(backend, value string) string {
```
phinze, line 54: Non-blocking design note. Because backend identity is encoded in-band, the mapping breaks in both directions: lose the `Backend` field and a reference silently becomes a literal, and conversely a literal value that happens to start with `miren+secret://` gets materialized as if it were a reference. There's no guard at mint time for the second case.

No escalation in it, since anyone who can set env vars can set a ref anyway, and the sandbox spec's `[]string` env genuinely leaves nowhere structured to put the backend. Rejecting a literal carrying the scheme prefix at config-mint time would close the half that's cheap to close.

cli/commands/env.go
```diff
@@ -55,6 +56,19 @@ func maskURLUserinfo(value string) string {
 // literal bytes" — no masking, no escaping — for copying or scripting.
 // Without it, the value is rendered display-safe: control/ANSI sequences are
 // escaped so a malicious value can't corrupt terminal or CI output.
+// displayEnvValue renders what `env list`/`env get` shows for one variable.
```
phinze: Tiny one: this landed between `maskEnvValue` and its doc comment, so the whole MIR-1356 essay above ("Deliberately absent: key-name matching…") now reads as documentation for `displayEnvValue`, and `maskEnvValue` has none. Worth moving, since that comment is doing real work keeping the next person from growing it into a pile of format matchers.

phinze (approved):
LGTM. One thing worth closing while you're in the file: `Backend` gets emitted by `GetConfiguration` but dropped by all three write paths back into `ConfigSpec`. CodeRabbit caught two of them, the `EnvVarInput` adapters at 645 and 670. The third is `SetConfiguration` itself, which nils `spec.Variables` and rebuilds without it, same for the per-service twin near 389.

None of it is reachable today. `SetConfiguration`'s only non-test caller is `cli/commands/set.go`, which isn't in the dispatch table (MIR-796 has it as dead code), and nothing sets `Backend` on the app-crud path since `env set` goes through the deployment service. So this is latent rather than live, and a one-line copy at each site closes it.

Worth saying that `SetConfiguration` being a whole-blob upsert is exactly why a new field silently vanishes there, which is the thing MIR-380 was about. I'd rather see the method go than get patched, so I'm adding a note to MIR-796. Not this PR's problem either way.

The saga-path audit your reviewer note asked for does hold: both paths reach container creation through `buildSubContainerSpec`, so no parallel change needed. CodeRabbit's `env list --format json` note is legit too, the arrow ends up in machine output. Two minor notes inline.

`--p+🤖`

servers/app/app.go
evanphx: 🤖 Good catch. The nasty part is the one you named — it doesn't break, it quietly works as a literal. Covered by tests now.

pkg/secret/sentinel.go
evanphx: 🤖 Both directions closed. CodeRabbit found the mirror image: a backend name with a slash doesn't round-trip either. Agreed there's no escalation in it — just confusing, and cheap to drop.

cli/commands/env.go
evanphx: 🤖 Oops. That comment's doing real work, so thanks for spotting it before it sat there pointing at the wrong function for a year.

appconfig/appconfig.go
```diff
@@ -21,6 +21,17 @@ type AppEnvVar struct {
...
+	// Backend names the secret backend the value comes from, and Ref addresses
+	// the secret within it. Together they replace Value: the credential itself
+	// never appears in app.toml, only a pointer to it, so the file stays safe to
+	// commit.
+	//
+	// A reference authored here floats — it re-pins to whatever is current on
+	// every deploy, which is how a rotation reaches the app. Appending @version
+	// to Ref holds it at that exact version instead.
+	Backend string `json:"backend,omitempty" toml:"backend,omitempty"`
+	Ref     string `json:"ref,omitempty" toml:"ref,omitempty"`
```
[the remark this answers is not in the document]
evanphx: 🤖 Fair — reworded. A rotation only reaches an app once a new config is minted; deploying or rolling back an existing version keeps what it recorded.

pkg/secret/pin.go
```diff
@@ -0,0 +1,75 @@
...
+			return nil, fmt.Errorf("%s has no backend to resolve against", ref.describe())
+		}
+		if ref.Ref == "" {
+			return nil, fmt.Errorf("%s names backend %q but no secret", ref.describe(), ref.Backend)
+		}
+
+		val, err := resolver.ResolveRef(ctx, ref.Backend, ref.Ref)
+		if err != nil {
+			return nil, fmt.Errorf("cannot resolve %s: %w", ref.describe(), err)
+		}
```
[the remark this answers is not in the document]
evanphx: 🤖 Leaving this one. The `%w` is load-bearing for `errors.Is`, and the contract's documented on `ResolveRef` — backends mustn't put secret material in errors. The detail is also what makes a fail-closed deploy readable: "secret version is not enabled" beats a redacted shrug.

pkg/secret/sentinel.go
```diff
@@ -0,0 +1,72 @@
...
+// would sit in etcd in plaintext — exactly what referencing a secret is meant
+// to avoid. The spec therefore carries this placeholder, and the value is
+// substituted in memory at the moment the container is created.
+//
+// It doubles as the unit of change detection for pool reuse: the reference
+// carries a concrete version, so it differs precisely when the bytes behind it
+// would differ.
+func FormatSentinel(backend, ref string) string {
+	return SentinelScheme + backend + "/" + ref
+}
```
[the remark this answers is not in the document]
evanphx: 🤖 Both directions closed. This half — a backend name with a slash not round-tripping — plus the mirror image phinze spotted, a literal carrying the scheme getting materialized as a reference. Registration now refuses a name that can't survive the trip.

PR #1031, opened 2026-08-05
Secrets on distributed runners, version reaping, and docs

evanphx:
Third of three stacked PRs implementing [RFD-90: Secret Backends and Materialization](https://github.com/mirendev/rfd/blob/main/rfds/0075/0090-secret-backends-and-materialization.md) ([MIR-1343](https://linear.app/miren/issue/MIR-1343)). Stacked on #1028.

The pieces that make the feature correct everywhere rather than just on a single-node cluster, plus the docs.

## Distributed runners

A distributed runner can reach the entity store but not the cluster keyring, so it cannot decrypt a stored secret itself. It now resolves through the coordinator: the bytes are decrypted where the keyring lives and travel back over the authenticated connection the runner already holds, to sit in memory only long enough to be handed to a container.

This mirrors the remote workload identity issuer, with one deliberate difference: **a failure to wire it up is fatal rather than degrading.** A runner without token issuance still runs apps; a runner without secret resolution starts sandboxes whose referenced secrets can never materialize, which is worse than refusing to start. `distributedrunners` is on by default, so without this, secrets would break on any multi-node cluster.

`Resolve` is not audited at Info. A write or a revocation is what an operator goes looking for after an incident, but a resolve happens on every sandbox start — logging each one would drown the tier it sits in while adding nothing the deploy record does not already say.

## Version reaping

A secret version is reapable only when it is neither its secret's current version nor pinned by any live ConfigVersion. That second condition is what makes the cluster backend's retention promise real: a pinned reference stays resolvable, so a rollback comes up on the bytes that version originally shipped with rather than failing or, worse, silently resolving something else.

Every ConfigVersion counts toward that, not only those belonging to active app versions — a rollback target is by definition not active, and it is exactly the case this protects.

Unlike an app version, a secret version cannot be rebuilt, so the sweep is conservative in every direction:

- A failure to list the pins **abandons the sweep** rather than proceeding on a partial picture, which would look exactly like "nothing references this."
- A superseded version is kept for a window even with nothing referencing it, because a rotation is exactly when someone might need the old value back.
- Secret versions sweep after app versions on the same tick, so deleting an app version releases its pins in the sweep that can act on them rather than the next one.

## Docs

A new [Secrets](docs/docs/secrets.md) page, plus `backend`/`ref` in the `app.toml` reference. The page leads with the distinction that actually matters: `-s` masks display while the value still sits in the store, whereas a secret is encrypted and referenced. Someone reaching for `-s` on a real credential should find that out here rather than after an audit.

Two things are called out prominently because getting them wrong is expensive: **how a rotation propagates** depends on where the reference was authored (an `app.toml` reference floats and moves on the next deploy; one set with `env set` is pinned and moves when you re-run it), and **losing the keyring file makes every stored secret permanently unrecoverable**, so it needs to be in whatever backs up the data path.

## Deliberately not included: tmpfs file secrets

RFD-90 also describes materializing a secret as an in-memory *file* for cases like a certificate or a JSON key. **I have left that out**, and I would rather flag why than ship it half-safe.

Writing a decrypted secret to a host tmpfs and bind-mounting it needs a guaranteed unmount at teardown *and* recovery for mounts orphaned by a controller crash. The sandbox controller has no per-sandbox directory cleanup today — only the Tempdir path is removed in the destroy path — so there is nowhere to hang the unmount, and no reconciliation to catch what a crash leaves behind. A tmpfs holding a decrypted secret that survives a controller restart with nothing to clean it up is a security regression relative to not having the feature.

Doing it properly needs a `mount_path` on the variable schema, the mount and unmount, and startup reconciliation for orphans following the `tokenRefresher` re-registration pattern. That is its own piece of work rather than a tail-end addition here.

Environment-variable materialization — which RFD-90 calls the common case — ships in #1028 and is unaffected.

## Testing

Unit tests for the reaping rules: current version retained, pinned version retained (with an explicit assertion that it is still readable afterward), unreferenced version reaped, recently superseded version kept, disabled-but-unreferenced version reaped, and an empty cluster sweeping cleanly. Service tests for `Resolve`, including that a revoked version fails closed for a remote caller too — otherwise disabling a leaked credential would only take effect on the coordinator.

New blackbox coverage running the full loop against a real cluster: store, reference, read the materialized value out of the running container with `sandbox exec`, rotate without disturbing it, re-pin, and confirm a revoked version fails closed. It also asserts the value never appears in any listing.

Full unit suite green.

### DIFF
--- a/controllers/version/gc.go
+++ b/controllers/version/gc.go
@@ -149,6 +149,36 @@ func (c *GCController) runGCWithLogging(ctx context.Context) {
 			"retained", result.RetainedVersions,
 			"total", result.TotalScanned)
 	}
+
+	// Secret versions are reaped on the same tick, after app versions: deleting
+	// an app version drops the ConfigVersion pinning its secrets, so running in
+	// this order lets a secret version become reapable in the same sweep that
+	// released it rather than waiting for the next one.
+	c.runSecretGCWithLogging(ctx)
+}
+
+func (c *GCController) runSecretGCWithLogging(ctx context.Context) {
+	result, err := c.RunSecretGC(ctx)
+	if err != nil {
+		c.Log.Error("secret version GC failed", "error", err)
+		return
+	}
+
+	if result.DeletedVersions > 0 || result.FailedVersions > 0 {
+		c.Log.Info("secret version GC complete",
+			"deleted", result.DeletedVersions,
+			"failed", result.FailedVersions,
+			"retained_pinned", result.RetainedPinned,
+			"retained_current", result.RetainedCurrent,
+			"retained_recent", result.RetainedRecent,
+			"total", result.TotalScanned)
+	} else {
+		c.Log.Debug("secret version GC complete, nothing reaped",
+			"retained_pinned", result.RetainedPinned,
+			"retained_current", result.RetainedCurrent,
+			"retained_recent", result.RetainedRecent,
+			"total", result.TotalScanned)
+	}
 }
 
 type versionInfo struct {
--- a/controllers/version/gc.go
+++ b/controllers/version/gc.go
@@ -112,7 +112,7 @@ func (c *GCController) run(ctx context.Context) {
 	// Run an initial GC on startup after a short delay.
 	select {
 	case <-time.After(30 * time.Second):
-		c.runGCWithLogging(ctx)
+		c.sweep(ctx)
 	case <-ctx.Done():
 		c.Log.Info("version retention GC controller stopped")
 		return
@@ -121,12 +121,26 @@ func (c *GCController) run(ctx context.Context) {
 	for {
 		select {
 		case <-ticker.C:
-			c.runGCWithLogging(ctx)
+			c.sweep(ctx)
 		case <-ctx.Done():
 			c.Log.Info("version retention GC controller stopped")
 			return
 		}
 	}
+}
+
+// sweep runs both retention passes for one tick.
+//
+// App versions go first: deleting one drops the ConfigVersion pinning its
+// secrets, so a secret version can become reapable in the same tick that
+// released it rather than waiting for the next. The order is the only thing
+// they share — the secret sweep reads its own pin picture and never looks at
+// the app sweep's result, so a failing app sweep must not take it down too.
+// Coupling them would hide the secret sweep behind an unrelated failure, and
+// the symptom (versions quietly accumulating) has no signal of its own.
+func (c *GCController) sweep(ctx context.Context) {
+	c.runGCWithLogging(ctx)
+	c.runSecretGCWithLogging(ctx)
 }
 
 func (c *GCController) runGCWithLogging(ctx context.Context) {
@@ -150,11 +164,6 @@ func (c *GCController) runGCWithLogging(ctx context.Context) {
 			"total", result.TotalScanned)
 	}
 
-	// Secret versions are reaped on the same tick, after app versions: deleting
-	// an app version drops the ConfigVersion pinning its secrets, so running in
-	// this order lets a secret version become reapable in the same sweep that
-	// released it rather than waiting for the next one.
-	c.runSecretGCWithLogging(ctx)
 }
 
 func (c *GCController) runSecretGCWithLogging(ctx context.Context) {
--- a/controllers/version/secret_gc.go
+++ b/controllers/version/secret_gc.go
@@ -3,6 +3,7 @@ package version
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	coreutil "miren.dev/runtime/api/core"
@@ -75,6 +76,17 @@ func (c *GCController) RunSecretGC(ctx context.Context) (*SecretGCResult, error)
 
 	cutoff := time.Now().Add(-secretVersionRetention)
 
+	// Candidates cleared every retention rule on the snapshot. They are
+	// collected rather than deleted inline so the pre-delete re-check below can
+	// be one fresh read for the whole sweep instead of one per candidate.
+	type candidate struct {
+		id      entity.Id
+		path    string
+		version string
+		ref     string
+	}
+	var candidates []candidate
+
 	for _, e := range secretsResp.Values() {
 		var sec core_v1alpha.Secret
 		sec.Decode(e.Entity())
@@ -85,11 +97,21 @@ func (c *GCController) RunSecretGC(ctx context.Context) (*SecretGCResult, error)
 			return result, fmt.Errorf("failed to list versions of secret %s: %w", sec.ID, err)
 		}
 
+		versions := make([]*entity.Entity, 0, len(versionsResp.Values()))
 		for _, ve := range versionsResp.Values() {
-			result.TotalScanned++
+			versions = append(versions, ve.Entity())
+		}
+		result.TotalScanned += len(versions)
 
-			ent := ve.Entity()
-			versionID := ent.Id()
+		// Newest first, so each version's successor is the one before it. A
+		// version stopped being usable when its successor was created, which is
+		// what the retention window is actually measured from.
+		sort.Slice(versions, func(i, j int) bool {
+			return versions[i].GetCreatedAt().After(versions[j].GetCreatedAt())
+		})
+
+		for i, ve := range versions {
+			versionID := ve.Id()
 
 			// The version a floating reference resolves to is never reapable,
 			// whatever else is true of it.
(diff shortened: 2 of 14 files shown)


controllers/version/secret_gc.go
```diff
@@ -0,0 +1,150 @@
...
+				result.RetainedPinned++
+				continue
+			}
+
+			if ent.GetCreatedAt().After(cutoff) {
+				result.RetainedRecent++
+				continue
+			}
+
+			if _, err := c.EAC.Delete(gcCtx, versionID.String()); err != nil {
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch — same treatment `pinnedNow` gets next door. Worth it because the asymmetry stings: a wrongly reaped app version rebuilds, a wrongly reaped secret doesn't.

servers/secret/server.go
```diff
@@ -207,3 +207,35 @@ func encodeSummary(summary secret.Summary) *secret_v1alpha.SecretInfo {
...
+// A distributed runner holds no key material, so this is how a sandbox on one
+// materializes its secrets: the bytes are decrypted here, where the keyring
+// lives, and travel to the runner, which holds them only long enough to hand
+// them to a container.
+//
+// It is deliberately not logged at Info. Unlike a write or a revocation, a
+// resolve happens on every sandbox start, so recording each one would drown the
+// audit tier it sits in without telling an operator anything they did not
+// already know from the deploy.
+func (s *Server) Resolve(ctx context.Context, state *secret_v1alpha.SecretsResolve) error {
```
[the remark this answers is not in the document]
evanphx: 🤖 It is enforced — non-public methods don't dispatch without an authenticated identity. But you're right that it was invisible, so it says so now. Also noted what it *doesn't* cover: any authenticated caller can resolve any reference. Per-secret scoping is still an open RFD question.

docs/docs/secrets.md
```diff
@@ -0,0 +1,175 @@
...
+
+## How it is stored
+
+Each version is sealed with its own random data key, and only that key is encrypted by a cluster key. So:
+
+- Rotating the cluster key re-encrypts a few bytes per version, never the values themselves.
+- A leaked data key exposes exactly one version rather than every secret.
+- The cluster key only ever performs small fixed-size operations, so it can move behind a KMS later without your stored data changing shape.
+
+The plaintext exists only transiently, in memory, while a deploy resolves it and while the container that needs it is being created. It is never written to the cluster store, a sandbox spec, an image layer, or a log.
```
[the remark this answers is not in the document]
evanphx: 🤖 All three done. This one was a real catch rather than style — I'd said plaintext only exists during resolve and container creation, but an env var lives as long as the process does. Miren stops passing it around; it can't un-give it to the app.

docs/docs/secrets.md
```diff
@@ -0,0 +1,175 @@
...
+miren secret set payments/stripe-key
+```
+
+</CliCommand>
+
+You are prompted with masking; the value is never echoed, logged, or written to disk by the CLI. Each write mints an immutable version and prints its handle:
+
+```
+Stored payments/stripe-key@x1A  (cluster, enabled)
+```
```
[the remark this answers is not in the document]
evanphx: 🤖 Done — `text` on the output fences.

docs/docs/secrets.md
```diff
@@ -0,0 +1,175 @@
...
+
+```miren
+miren secret disable payments/stripe-key@x1A
+```
+
+</CliCommand>
+
+Disabling is reversible with `miren secret enable`; the value itself is untouched. Note that already-running sandboxes keep the value they started with — revoking prevents new resolutions, it does not reach into running processes.
+
+`miren secret destroy` goes further and deletes the value for good. Prefer `disable` when responding to a leak: it fails closed just as hard and leaves you able to recover if something still needed the value.
```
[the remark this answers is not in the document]
evanphx: 🤖 Done — a `warning` for running sandboxes keeping the old value, and a `danger` for destroy being one-way.

controllers/version/gc.go
```diff
@@ -149,6 +149,36 @@ func (c *GCController) runGCWithLogging(ctx context.Context) {
 			"retained", result.RetainedVersions,
 			"total", result.TotalScanned)
 	}
+
+	// Secret versions are reaped on the same tick, after app versions: deleting
+	// an app version drops the ConfigVersion pinning its secrets, so running in
+	// this order lets a secret version become reapable in the same sweep that
+	// released it rather than waiting for the next one.
+	c.runSecretGCWithLogging(ctx)
```
phinze: An app-sweep failure takes this down with it: the `return` up at line 136 sits above this call, and this call is the last statement in the function. The two sweeps are independent — the secret one reads its pin picture fresh and never touches `RunGC`'s result — so the coupling looks incidental rather than deliberate.

The failure mode is a quiet one. A persistently failing app GC means secret versions just accumulate, with no signal of their own to say so. Hoisting this above the early return, or moving both calls into `run()`, would decouple them.

controllers/version/secret_gc.go
```diff
@@ -0,0 +1,186 @@
...
+				continue
+			}
+
+			ref := secret.FormatRef(sec.Path, ent.ShortId())
+			if pinned[ref] {
+				result.RetainedPinned++
+				continue
+			}
+
+			if ent.GetCreatedAt().After(cutoff) {
```
phinze: This measures from creation, but the constant above describes the window as protecting a *superseded* version — "reaping the instant the last reference drops would make that window zero."

That's the case it misses. A credential that sat as current for a year and got rotated five minutes ago is already well past `cutoff` on the day it's superseded, so an unreferenced old version can go in the next hourly sweep. The long-lived secret is the shape the comment is written for, and it's the shape with no window at all.

The pin check is the load-bearing protection and it's solid, so this is the backstop rather than the guarantee. Which way to take it feels genuinely open: the current version's own `CreatedAt` is a cheap supersession proxy (retain everything while the secret was recently rotated), or compute it per-version from the next-newest sibling, or decide the behavior is what you want and it's the comment that's overclaiming. Any of those reads fine from here.

controllers/version/secret_gc.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	return pinned, nil
+}
+
+// secretVersionPinnedNow re-reads the live ConfigVersions and reports whether
+// any of them pins this reference.
+//
+// It runs only for versions that already cleared every other retention rule,
+// so the extra pass costs nothing on a healthy sweep and buys the guarantee
```
phinze: Worth tuning this claim. `pinnedNow` next door is explicitly point queries (a Get plus one indexed list), while this re-lists and decodes every ConfigVersion in the cluster, once per candidate.

At zero candidates it really is free, so the sentence holds in steady state. It's the first sweep after this ships on a cluster with history, or any bulk reap, where K candidates × N ConfigVersions has to fit inside the 5-minute budget — and blowing that aborts the sweep partway.

Not arguing for a redesign; the safety is worth more than the scan. Just that the `pinnedNow` analogy sets an expectation about query shape that this one doesn't match.

docs/docs/secrets.md
```diff
@@ -0,0 +1,183 @@
+---
+title: Secrets
+description: Store credentials encrypted in your cluster and reference them from app config, so the value never sits in plaintext and every deploy records exactly which version it used.
```
phinze: This still promises "the value never sits in plaintext," which line 166 now explicitly contradicts: "your application holds its own copy for as long as it runs." The body picked up that correction in the last round and the description didn't come along.

Worth fixing mostly because of where it renders — the description is the search result, the social card, and the docs index entry, so it's the sentence most readers see and often the only one they see.

Maybe: "...so the value is never stored in the clear, and every deploy records exactly which version it used."

docs/docs/secrets.md
```diff
@@ -0,0 +1,183 @@
...
+keywords: [secrets, credentials, encryption, rotation, env vars, secret backend]
+---
+
+import CliCommand from '@site/src/components/CliCommand';
+
+# Secrets
+
+A secret is a credential Miren holds **encrypted**, that your app config **points at** rather than contains. The value is decrypted only in memory, at the moment a container starts.
```
phinze: Two things about this opening paragraph.

The precision one: "decrypted only in memory, at the moment a container starts" is narrower than what actually happens, and line 164 has it right. `Pin` (`pkg/secret/pin.go:68`) calls `ResolveRef`, which decrypts, at ConfigVersion mint time — it just discards the bytes and keeps the resolved ref. So the opener contradicts the section that explains the model 150 lines further down.

The bigger one: the page defines before it hurts. The problem frame — someone reaching for `-s` on a real credential — doesn't land until line 22, underneath a table. You already wrote the sentence that fixes this, over in `app-configuration.md`: "For a real credential, `sensitive` is not enough: it masks display but the value still sits in your config." Opening with that makes the definition read as relief rather than vocabulary.

docs/docs/secrets.md
```diff
@@ -0,0 +1,183 @@
...
+Storing a value identical to the current version is reported as unchanged rather than minting a duplicate, so re-running the command is safe.
+
+## Using a secret in an app
+
+Point an environment variable at it. The value does not travel through your shell, your config file, or your terminal — only the reference does.
+
+<CliCommand context="client">
+
+```miren
+miren env set -e STRIPE_API_KEY --backend cluster --ref payments/stripe-key
```
phinze: `--backend` is optional here — `env.go:263` falls back to `cluster` when it's empty — so this example carries a flag it doesn't need, and nothing on the page mentions the default.

The asymmetry is worth a line too, since the page shows both forms side by side as though they behave the same: in `app.toml`, `backend` genuinely is required (`appconfig.go:589` rejects a `ref` without one). So `miren env set -e STRIPE_API_KEY --ref payments/stripe-key` works fine, but dropping `backend` from the TOML block below doesn't.

docs/docs/secrets.md
```diff
@@ -0,0 +1,183 @@
...
+
+## How it is stored
+
+Each version is sealed with its own random data key, and only that key is encrypted by a cluster key. So:
+
+- Rotating the cluster key re-encrypts a few bytes per version, never the values themselves.
+- A leaked data key exposes exactly one version rather than every secret.
+- The cluster key only ever performs small fixed-size operations, so it can move behind a KMS later without your stored data changing shape.
+
+Within Miren, the plaintext exists only transiently and only in memory: while a deploy resolves it, and while the container that needs it is being created. It is never written to the cluster store, a sandbox spec, an image layer, or a log.
```
phinze, line 166: This PR is what makes plaintext cross a network boundary, and the page doesn't mention it. "Within Miren, the plaintext exists only transiently and only in memory" stays true — a runner's memory is memory — but on a multi-node cluster the bytes now travel coordinator→runner, which is material to anyone reading this section to decide how far to trust the model. Before this change they couldn't; `MaterializeEnv` failed closed without a resolver.

The package doc in `pkg/secret/remote` already says it well. One sentence borrowed from it would close the gap — something like: "On a multi-node cluster the coordinator does the decrypting and returns the value to the runner over the cluster's authenticated connection, where it lives only long enough to hand to the container."

phinze (approved):
LGTM, nothing blocking here.

The one with actual teeth: the secret sweep never runs on any tick where the app sweep errors — `runGCWithLogging` returns early and the secret call is the last statement in the function. Everything else inline is either a comment promising a little more than the code does, or the plaintext claims on the docs page, where the body already corrects an overclaim the frontmatter description still carries.

The other thing worth naming: the distributed resolve path has no end-to-end coverage. `blackbox/secret_test.go` is single-node, `pkg/secret/remote` has no test, and the peers harness already has the pattern in `distributed_runner_test.go`. With `distributedrunners` on by default it seems worth having, but up to you whether that's a follow-up or something you'd rather fold in here.

`--p+🤖`

controllers/version/gc.go
evanphx: 🤖 Yep, that one's real — and the quiet failure mode is the bad part. Both sweeps now run from one place that can't swallow the second.

controllers/version/secret_gc.go
evanphx: 🤖 Sharp catch. Went with per-version supersession from the next-newest sibling, since that's what the comment was already promising. The long-lived-secret case gets its window now; the test fails against the old creation-based check.

controllers/version/secret_gc.go
evanphx: 🤖 Fair — the analogy set an expectation the query shape didn't meet. Now it collects candidates and re-reads once for the sweep instead of once per candidate, so two scans rather than K+1. Same protection, and the comment stops borrowing `pinnedNow`'s reputation.

docs/docs/secrets.md
evanphx: 🤖 Good catch on where it renders — took your wording. The body picked up that correction last round and the description didn't come along.

docs/docs/secrets.md
evanphx: 🤖 Both fair. Pinning does decrypt at mint, so the opener was contradicting the section that explains the model 150 lines down. And you're right it defined before it hurt — it now opens with the `-s`-isn't-enough framing from `app-configuration.md`.

docs/docs/secrets.md
evanphx: 🤖 Dropped the flag from the example and said what the default is. Called out the asymmetry too — optional on the CLI, required in `app.toml` — since showing both forms side by side implied they behave the same.

docs/docs/secrets.md
evanphx: 🤖 Right, and this is the PR that makes it true — before it, `MaterializeEnv` failed closed without a resolver. Borrowed the `pkg/secret/remote` framing as you suggested.

PR #1032, opened 2026-08-05
Name static reflection types directly

phinze:
#1020 deliberately left the broader `reflect.TypeFor` fixer output for its own review. This is that slice.

Several reflection call sites still construct zero values, empty containers, or nil pointers only to recover a static `reflect.Type`. The generic `reflect.TypeFor` helper names those types directly, so conversion checks, CBOR tag registration, and saga signature validation now state what they mean without the scaffolding.

All 18 replacements come directly from the `reflecttypefor` analyzer across 10 hand-written files. The resulting type identities are unchanged. Generated files, serialization policy changes, and the other standard-library fixers stay separate.

Stacked on #1025, which handles the loop-syntax cleanup.

### DIFF
--- a/cli/commands/app_test.go
+++ b/cli/commands/app_test.go
@@ -343,7 +343,7 @@ func TestCLIEnvTagsDoNotReadInjectedVars(t *testing.T) {
 		injected[n] = true
 	}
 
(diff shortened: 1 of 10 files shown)


evanphx (approved):
Image

PR #1036, opened 2026-08-06
Use direct bounds and string helpers

phinze:
The smaller Go fixer pass in #1034 left a medium-sized group of helpers that are still mechanical, but broad enough to benefit from their own review boundary.

This keeps three analyzer families in separate revisions. The first replaces follow-up clamp branches with the built-in `min` and `max` helpers across backoff, batching, quota, and sizing paths. The second moves repeated string construction onto `strings.Builder`. The third ranges over `strings.SplitSeq` and `bytes.SplitSeq` where callers consume a split result once, avoiding the intermediate slice.

The resulting values, output ordering, empty fields, and trailing-delimiter behavior are unchanged. The larger collection-membership sweep and the concurrency, test-lifecycle, and serialization fixers stay separate.

All packages compile, generation and module tidiness checks pass, and the affected unit tests pass. Service-backed tests that require bridge creation, containerd, etcd, or VictoriaLogs remain unavailable locally, so CI is authoritative for those paths.

Stacked on #1034, which handles the smaller direct standard-library helpers.

### DIFF
--- a/cli/commands/app.go
+++ b/cli/commands/app.go
@@ -413,7 +413,8 @@ func (m Model) View() string {
 			))
 	}
 
-	hdr := fmt.Sprintf("       name: %s\nlast update: %s %s\n",
+	var hdr strings.Builder
+	fmt.Fprintf(&hdr, "       name: %s\nlast update: %s %s\n",
 		bold.Render(m.status.Name()),
 		bold.Render(lastUpdate),
 		laExtra,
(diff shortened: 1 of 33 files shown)


phinze:
@evanphx looks like GitHub recorded your approvals for the four PRs above this one but skipped this bottom layer. Mind giving this one a quick check? Everything is green, and I’ll merge the remaining stack once it’s approved.

evanphx (approved):

PR #1042, opened 2026-08-06
Fix manual `cluster add` writing a config that can't connect, and tidy cluster selection around it

phinze:
We've been chipping away at making tailnet-only clusters work end to end, and there are a few more bugs in the way. This clears two of them, plus some cleanup found on the trip.

The first is the one that actually stops people. Adding a cluster by hand skips identity resolution on purpose, so `miren cluster add --cluster <name> --address <name>:8443` was writing an entry with no identity, no client certificate and no key, even for someone who'd already run `miren login`. That config falls through every auth branch to the deprecated client-cert path, which hands two empty strings to `tls.X509KeyPair`. Since `[]byte("")` is non-nil, it looks like a certificate that failed to parse rather than one that was never there, so you get `tls: failed to find any PEM data in certificate input`, blaming the certificate for a config that simply had none, four layers from the cause. The command that created it had already printed "using environment auth", naming a mode that exists nowhere in the tree.

The second is the one we thought was the first: a DNS name in `--address` never verifies. The API certificate carries `localhost`, `api.miren`, `additional_names`, and every discovered address as an IP SAN, so a MagicDNS name matches nothing. These two stack, which is worth knowing, because fixing only the hostname bug would have produced a byte-identical error message. The fix borrows `api.miren`, the same escape hatch sandboxes use to reach the API over a bridge address that didn't exist when the certificate was issued. Verification stays fully on, and it covers a NAT'd IP the server doesn't know it has for free.

The second rev is cleanup found on the way through. `cluster remove` refused to remove the active cluster in three separate places, which with a single cluster is a dead end, since there's nowhere to switch to first. It now drops the active pointer along with the cluster, takes the survivor when there's one left, and opens the switcher when there are several.

That made "no active cluster" an ordinary state rather than a rare one, and it landed badly: the dial fell through to the default address and failed as `no remote address specified`. It gets a real diagnostic now. Writing that message next to its three siblings made it obvious they didn't sound like the same product, so they got a pass against [VOICE.md](https://github.com/mirendev/brand/blob/main/VOICE.md): the transport diagnostics are active voice and say "we", while the config ones had drifted passive and used *active*, *selected*, and *switch* interchangeably for one concept. The golden-file catalog exists so wording changes read as a diff in review, but it only covered transport failures, so it now covers these too.

Also gone: `ConfigRemove`, an unregistered second implementation of cluster removal that nothing referenced and that contradicted the live behaviour twice over.


phinze:
@coderabbitai review

cli/commands/global.go
```diff
@@ -478,7 +487,21 @@ func (c *Context) RPCClient(name string) (*rpc.NetworkClient, error) {
...
+		// user never asked about.
+		//
+		// A config with no active cluster is the one case where continuing is
+		// right: there is no cluster to be wrong about, and the local-dev path
+		// below is what the user meant.
+		if active := c.ClientConfig.ActiveCluster(); active != "" {
+			return nil, c.unusableClusterError(active, err)
+		}
+
+		c.Log.Debug("Client config could not provide RPC state", "error", err)
```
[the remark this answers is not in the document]
phinze: Agreed this is a genuine weak spot, but declining it here as out of scope.

It predates this PR, and the change actually narrows it: we now refuse when the config names an active cluster, so the default-address fallback is reachable in strictly fewer cases than before. The default `--server-address` is also `127.0.0.1:8443`, so in practice it is already the loopback bootstrap you propose restricting it to.

That said, you are right that `--server-address` can point anywhere and credential-bearing commands share the path, so an explicit loopback restriction is real hardening rather than a no-op. That is a security change to shared connection code and deserves its own PR with its own testing, not a rider on a cluster-selection cleanup.

clientconfig/clientconfig.go
```diff
@@ -919,6 +930,7 @@ func (c *Config) RemoveCluster(name string) error {
 				}
 				c.unsavedLeafConfigs[leafName] = configData
 			}
+			c.clearActiveIfNamed(name)
```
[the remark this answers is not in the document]
phinze: Good catch, and it was worse than minor. Fixed in the latest push.

Reproducing it: removing an active cluster defined in `clientconfig.d/` wrote the leaf back as just `active_cluster: homelab` with the clusters section gone, and the next load failed `Validate` with `active cluster "homelab" not found in configured clusters`. The whole configuration became unloadable, not only the one entry.

`clearActiveIfNamed` now sweeps leaf actives as well as the main one, and runs before the `ConfigData` snapshot is built rather than after, since the snapshot copies `active` by value. Added `TestRemoveActiveClusterDefinedInLeaf`, which loads a leaf fixture that actually sets `active_cluster` and asserts the config still loads afterward. Our existing leaf test passed only because its fixture happened not to set it.

phinze:
Follow-up on the two findings that came through the review body rather than as threads:

**`rpc_errors.go` "nothing was contacted" claim** — fixed. You were right that this is false: `-C` accepts an ad-hoc address, and `LoadCluster` dials it via `setupClusterFromAddress` before failing, so we may well have contacted the thing the user named. Reworded to `We did't fall back to another cluster`, which is what the sentence was actually reaching for and is true in both the typo'd-name and typo'd-address cases. Test assertion and golden file updated to match.

**`local.go` CloudAuth via `dialOptions`** — declining, on the merits rather than scope. `CloudAuth` currently verifies against the system roots, and routing it through `dialOptions` would newly apply a stored `CACert` when one is present. A user on that deprecated path whose cluster has a publicly trusted certificate would start failing verification against a CA that never signed it. Consistency is not worth a live regression on a path we are trying to retire, so it stays as-is deliberately.

clientconfig/local.go
```diff
@@ -206,16 +207,35 @@ foundAddress:
...
+		opts = append(opts, rpc.WithCertPEMs(
 			[]byte(c.ClientCert),
 			[]byte(c.ClientKey),
-		),
-		rpc.WithCertificateVerification([]byte(c.CACert)),
-		rpc.WithEndpoint(hostname),
-		rpc.WithBindAddr("[::]:0"),
-	}, nil
+		))
+	case c.ClientCert != "":
```
evanphx, line 232: 👏 Rejecting the half-configured pair instead of quietly dropping it. Easy to skip past on the way to fixing the empty case, and skipping it would have traded a bad error message for a silent one.

clientconfig/clientconfig.go
```diff
@@ -432,6 +432,27 @@ func (c *Config) SetActiveCluster(name string) error {
...
+// the leaf gets written back naming a cluster that no longer exists anywhere,
+// and the next load refuses the whole configuration with "active cluster %q not
+// found in configured clusters". Callers must run this before snapshotting a
+// leaf for save, since the snapshot copies active by value.
+func (c *Config) clearActiveIfNamed(name string) {
+	if c.active == name {
+		c.active = ""
+	}
+
+	for _, leafConfig := range c.leafConfigs {
```
evanphx, line 476: Note for whoever's here next rather than something to change now: this clears every leaf's pointer in memory, but only the leaf that *owned* the removed cluster gets added to `unsavedLeafConfigs`, so that's the only one written back. If a different leaf names the removed cluster active, the clear never reaches disk.

I poked at it with a scratch test — main config owns `homelab`, a separate `clientconfig.d/pointer.yaml` carries just `active_cluster: homelab`:

```
leaf on disk after save: "active_cluster: homelab\n"
reload err = invalid configuration: active cluster "homelab" not found in configured clusters
```

Same unloadable-config symptom this function exists to prevent. Nothing in the tree writes `active_cluster` into a leaf file today, so getting there takes a hand-written config — which is why this is a note and not a request.

cli/commands/rpc_errors_test.go
```diff
@@ -225,10 +232,101 @@ func TestRPCClientRefusesBrokenActiveCluster(t *testing.T) {
...
+// Refusing here would break `miren server` on localhost.
+func TestRPCClientFallsBackWithNoClustersConfigured(t *testing.T) {
+	c := &Context{
+		Context:      context.Background(),
+		Stderr:       io.Discard,
+		Log:          slog.New(slog.NewTextHandler(io.Discard, nil)),
+		ClientConfig: clientconfig.NewConfig(),
+	}
+
+	// Reaching the default-address path at all is the assertion; whether a
```
evanphx: This comment claims more than the test checks. With a non-nil-but-empty config we don't reach the default-address path — `cs.Client(name)` fails first at `global.go:486`:

```
err = no remote address specified (*errors.errorString)
```

It passes because that isn't a `Diagnostic`. The real local-dev shape is `ClientConfig == nil`, which takes a different branch entirely.

The test isn't worthless — it does pin the `GetClusterCount() > 0` half of the new guard, and it'd fail if someone dropped that. Worth either rewording the comment to say that's what it's guarding, or pointing it at a nil `ClientConfig` so it covers the fall-through it names.

evanphx (approved):
Approved! Nice that it just works now.

Two non-blocking notes inline, neither with teeth.

--e+🤖

clientconfig/local.go
phinze: Thanks. That one was CodeRabbit-adjacent luck honestly, I only noticed the half-pair case while writing the test for the empty one and realised dropping it silently would just move the bad error somewhere less findable.

clientconfig/clientconfig.go
phinze: You're right, and I've fixed it rather than leaving it as a note. Reproduced your case exactly, same unloadable config.

The gap was that clearing and persisting were two different decisions and I'd only wired up one. `clearActiveIfNamed` now queues each leaf it clears via a new `markLeafDirty`, which is also what the owning-leaf branch of `RemoveCluster` uses, so there's one path for "this leaf changed, write it back" instead of an inline snapshot that only fired for the owner. `TestRemoveActiveClusterPointedAtByAnotherLeaf` pins your exact shape: main config owns the cluster, a separate leaf carries only `active_cluster`.

Agreed nothing in the tree writes `active_cluster` into a leaf today, but hand-written configs are exactly how someone pins a cluster across a `clientconfig.d` they don't otherwise own, so it felt worth closing while we were here.

cli/commands/rpc_errors_test.go
phinze: Good catch, the comment was describing a path the test never reached. Confirmed your read: with an empty-but-present config we stop at `cs.Client` for want of an endpoint, and it passed only because that isn't a `Diagnostic`.

Took both of your options rather than choosing. It's now `TestRPCClientDoesNotRefuseWithoutClusters` with two subtests: a nil `ClientConfig`, which actually exercises the fall-through the old name promised, and the empty config, which pins the `GetClusterCount() > 0` half of the guard as you noted it was already doing. The comment says which one does what, since the fact that they fail in different places is the whole trap.

PR #1043, opened 2026-08-07
run: the Run entity, its controller, and miren app run absorbed (RFD-97, 2/3)

evanphx:
Second of three stacked PRs implementing [RFD-97 "App Tasks"](https://github.com/mirendev/rfd/blob/main/rfds/0075/0097-app-tasks.md) (MIR-853). Stacked on #1019.

This is the core: the `Run` entity, its controller, and the user surface. The manual trigger works end to end on merge; `deploy` and `schedule` follow in PR 3.

## A Run

One execution of a task: one sandbox built from the app's deployed image and config, one command, one exit code, then teardown. One kind and one controller cover every trigger, because a deploy-gated migration, a scheduled cleanup, and someone typing `miren app run` differ only in what created the Run and what is done with the result.

The controller is the only writer of `Run.Status`. The exit code, a cancellation request, a deadline passing — all arrive as inputs it reads. Splitting that ownership would be a race with no lock available to fix it, since the store has no cross-entity transaction.

Two decisions worth reading the code for:

**Retries reuse the entity.** A scheduled run's id will be derived from its tick, and that derivation *is* the single-execution guarantee — so a second entity for attempt two would carry a name that isn't derived from the tick and would forfeit it. Each attempt's sandbox is kept in `attempt_record`, because a failed attempt's logs live only there and that failure is the reason anyone opens a retried run.

**Admission is exact at one and best-effort above it.** At `max_concurrent = 1` the property users depend on is "two deploys can't run my migration concurrently", and put-if-absent `Create` on a derived name gives it. The gate is a separate `run_slot` pointer, not the Run: gating the Run directly would mean reclaiming one entity per run and destroying the history runs exist to record. Above 1 it's a check-then-act count that can overshoot — an exact N would need a transaction this store can't express, and a counter would drift upward on any crash between teardown and decrement, failing closed permanently. That's stated in the code, not just the RFD.

## Absorbing `miren app run`

The command surface is nearly unchanged — no arguments opens a shell, arguments run that command, the exit code still propagates — but the invocation now leaves a durable record, which is what makes exit codes retrievable afterwards, `--detach` possible, and "who shelled into production" answerable.

New: `app runs`, `app runs cancel`, `app attach`, `logs run`, and `--task`/`--detach`. The task selector is a flag rather than positional because bare arguments already mean the command. `logs run` joins the logs family rather than hanging off `app runs`, since `app runs logs` is a double plural one character from `app run`.

`logs run` needed no server change: the run controller puts the run id in the sandbox spec's log attributes and the sandbox controller copies those onto every entry, so it's a client-side LogsQL filter.

The exec proxy goes from 383 lines to 124. `createEphemeralSandbox`, `waitForSandboxRunning`, `deleteSandbox`, and the drifted `buildSandboxSpec` copy are gone — along with the leak they carried, where teardown was a `defer` in the request handler so a proxy crash stranded a running sandbox with nothing that would ever collect it.

## Stdio fan-out

Attaching is now a subscription rather than a second process. An exec'd process belongs to the connection that started it — it dies on disconnect and takes its exit code with it — so "the run survives your client going away" isn't expressible that way.

Two rules fall out of containerd's `copyIO`, which closes the container's stdin FIFO as soon as the reader it was given returns EOF: detaching never closes the Hub's stdin (that would make disconnect equivalent to cancellation), and only teardown does. Output is dropped for a client that falls too far behind rather than blocking, since `Hub.Write` runs on containerd's stdout copy goroutine and blocking there would freeze the workload being watched.

`SandboxSpecContainer.Stdin` becomes real config — it was written by the exec proxy and read by nothing. containerd fixes the FIFO set at task creation, so attachability can't be granted retroactively; runs set it unconditionally, which is what lets `--detach` mean only "don't attach my terminal right now".

## Task-only apps

They have no pools, so they fell into the same `desired == 0` branch as an app that scaled away and reported `idle` forever — deploys completed by accident, on a code path meaning "deliberately asleep". New `apphealth.Ready`, updated in lockstep across both classifiers, the deploy poller, and the blackbox harness, since a disagreement shows up as a deploy that times out despite having succeeded.

## Deviations from the RFD

1. **`SandboxSpec.RestartPolicy`** is not in the RFD and should be. `SandboxController.Create` falls through to `createSandbox` for a `RUNNING` sandbox whose containers have vanished, so "one command, at most once" isn't true across a runner restart.
2. **Per-attempt records.** Overwriting `Run.Sandbox` on retry loses the failed attempt's logs.

`sandbox.go` is frozen pending the saga rewrite. The saga path was audited rather than the hash bumped blindly: it reaches `BootContainers`/`monitorTaskExit` through `sandbox_ops.go` and inherits those changes, but duplicates the `Create` dispatch and replaces `createSandbox` wholesale, so the restart-policy guards and the boot-failure Hub teardown are mirrored into `SagaSandboxController.Create` and `undoBootContainers`.

The exit code is reported to the `Run` rather than read back off the sandbox. The sandbox entity has several writers and `PatchEntity` is read-modify-write at revision 0, so a boot-path write computed from a pre-exit snapshot could land after the exit and take it; the run has a single writer plus that input, so nothing races it.

## Testing

`make lint` clean. `make test` clean apart from `pkg/addon/postgresql` and `pkg/addon/rabbitmq`, which fail identically on `main`.

Verified against a dev cluster: a task-only app deploys and reports `ready`; detached runs return an id and complete; cancellation ends a run and records no exit code; `miren logs run` reads output back by id; an attached run propagates its exit code to the caller's shell.

Running it for real is what found two bugs that unit tests missed, both now covered:
- The no-restart guard fired on a sandbox's first reconcile, where containers are absent because none exist *yet* — every task run was retired before executing anything. The unit test constructed a `RUNNING` sandbox directly and never saw it.
- The run controller read sandbox *status* to decide a run was over, but status has several writers. The exit component has one and is set once, so it's now authoritative.

New etcd-backed round-trip tests exist because the in-memory store the other tests use is more forgiving than etcd about exactly this, which is how both bugs got through.


evanphx:
Pushed two fixes for the CI failures.

**The exit-code race is fixed properly**, not narrowed. Reporting the exit to the `Run` entity instead of leaving the controller to read it off the sandbox — which is what RFD-97 specified all along (`Result` is an input with a `source`; the field existed, nothing filled it in).

The sandbox has several writers all patching with revision 0, and `PatchEntity` is read-modify-write, so a write computed from a pre-exit snapshot lands afterwards and takes the exit with it. Guarding the exit write only helped when it was the late writer. The run has one writer plus reported inputs, so there is no race to lose. Verified with 10 consecutive runs of both previously-flaky tests — `TestFailingRunRecordsItsExitCode` (detached) here and `TestAttachedRunPropagatesExitCode` (attached, which was failing on #1045).

One consequence needed deciding: cancelling now observes a real exit code, because the platform kills the process and the kernel reports something. That number is teardown noise rather than the task's outcome, so an exit code is surfaced only for runs that ended by their command exiting. The observation stays on the entity; the status carries the meaning for a run that was killed.

**Lint** was stale generated command docs — `main` added `miren app set-workload-role`, which re-sorts between the new `app runs` entries and `app status`. Regenerated.

The `test` job is an aggregate that was failing only because `test-blackbox-distributed` was; that was the same race.

api/app/rpc.yml
```diff
@@ -412,6 +412,47 @@ types:
...
+      - name: status
+        type: string
+        index: 3
+      - name: command
+        type: string
+        index: 4
+      - name: exit_code
+        type: int32
+        index: 5
+        doc: Set only when an exit was actually observed; a run that timed out or was canceled has none
```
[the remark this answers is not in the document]
evanphx: 🤖 Skipping — the generated code already does this. `exit_code` gets a `HasExitCode()` accessor alongside `ExitCode()`, and the server only calls `SetExitCode` when an exit was actually observed, so absence and a legitimate 0 are already distinguishable on the wire and at the client.

`cli/commands/app_runs.go` relies on exactly that: it emits `null` rather than `0` in the JSON output when `HasExitCode()` is false. There's a unit test for the three cases (observed non-zero, observed zero, not observed) in `servers/app/runs_test.go`.

evanphx:

**Three were real defects, not polish:**

- **Crash path.** `Hub.Close` and a subscriber's `unsubscribe` both closed the same channel under separate guards, and `servers/exec/attach.go` defers `unsubscribe` — so tearing down a sandbox with a client attached panicked the runner. Now a `sync.Once` owned by the subscriber; either order is safe.
- **Authorization.** None of the four `Runs` methods checked app access. `CreateRun` runs a command inside the target app's image, and `GetRun`/`CancelRun` take a bare run id. All four guard now; the id-based pair resolve the owning app from the run and refuse rather than defaulting open if that lookup fails.
- **Dead deadline.** `StartDeadline` measured from `StartedAt`, only written once a run is already running — so a run held by admission had no bound at all. Runs now carry `queued_at`, stamped on first sight. Relatedly, a sandbox-creation failure no longer spends the retry budget on a store blip.

Plus: manual runs get one attempt (matching the comment and the controller), `admitByCount` lists the running index instead of every run an app ever produced, `recordExit` backs off and honors shutdown, the CLI retries an attach only while the sandbox is genuinely absent and no longer exits 0 when the outcome is unknown, and the assertion-free concurrency test is folded into the guard test it was explaining.

Declined one with reasoning inline (`exit_code` optionality — the generated `HasExitCode()` already provides it, and there's a test for the three cases).

evanphx:
🤖 Picked up three findings that were in review bodies rather than inline threads, so they weren't in the earlier sweep:

- The `result` schema doc still claimed a canceled or timed-out run has no observed exit. That stopped being true once the runner started reporting exits to the run — cancellation kills the process and the kernel reports something. The distinction is that the number describes the *teardown* rather than the task, which is why callers suppress it; the doc says that now.
- The run list comparator subtracted two Unix-millisecond values and cast to `int`, which doesn't fit on a 32-bit build. Now `cmp.Compare`.
- The console migration test asserted the env var's key but not its value, so a migration that carried the name across and dropped the setting would have passed.

servers/app/runs.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+	}
+	return nil
+}
+
+// resolveCommand picks what a run executes: an explicit override, else the
+// task's declared command, else nothing -- which leaves the resolution to the
+// runner's console fallback chain, since what a bare `miren app run` should do
+// depends on what the image supplies and cannot be written down as a default.
+func resolveCommand(task *core_v1alpha.ConfigSpecTasks, override []string) (string, error) {
```
phinze: Bare `miren app run` doesn't reach a fallback chain from here. The console, entrypoint-wrapped-shell, `/bin/sh` chain lives in `servers/exec.Server.spec()` (exec.go:305), reachable only from `Exec`, which this PR removes for the `app` category. So an app with no `[tasks.console]` gets `command == ""` and `appspec.Build` leaves `appCont.Command` empty (appspec.go:292).

What happens next depends on the image, since `createSandbox` only adds `oci.WithProcessArgs` when `co.Command != ""` (sandbox.go:2844) and otherwise takes whatever `oci.WithImageConfig` supplies. On stack-built images that's nothing, and the container never starts:

```
runc create failed: OCI runtime create failed: args must not be empty
```

Reproduced on both a task-only app and an ordinary Go web app. The run goes `failed` with no exit code and the user sees nothing at all. On an image that does define a CMD you'd instead get the app's normal process attached to your terminal, which isn't a console either.

servers/app/runs.go
```diff
@@ -0,0 +1,400 @@
...
+	if len(override) > 0 {
+		return strings.Join(override, " "), nil
```
phinze: `strings.Join` loses argument boundaries. `miren app run -- echo "hello   world"` comes back as `hello world`, and `psql -c "SELECT 1"` won't survive. The Run entity stores the joined string faithfully, so the loss is downstream at command to argv.

RFD-97 models this as `Command []string`; `api/run/schema.yml:33` made it a string. Probably worth fixing at the schema while it's still unreleased.

servers/app/runs.go
```diff
@@ -0,0 +1,400 @@
...
+// both a long-running service the launcher keeps up and the command app run
+// executes -- and only the second was ever the point.
+const ConsoleTask = "console"
+
+// ConsoleMaxConcurrent is the ceiling on simultaneous console runs.
+//
+// It is a new limit on an existing command rather than a limit on new
+// functionality, so it is set well past what anyone reaches by hand. An app
+// that genuinely wants more can declare [tasks.console] and say so.
+const ConsoleMaxConcurrent = 10
```
phinze: This is never read. Concurrency is decided entirely by `Controller.maxConcurrent` (admission.go:218), which resolves from the task's config and returns 1 when the task isn't declared, which is the common case since `console` is undocumented. So `miren app run` is capped at one concurrent session where today it's unbounded, the opposite of the RFD's "a new limit on an existing command rather than a limit on new functionality".

Two of us on the same app: the second waits two minutes with no output, then gets an error that never mentions concurrency.

controllers/run/admission.go
```diff
@@ -0,0 +1,249 @@
...
+// check-then-act, so two invokes arriving at nine live runs can both pass and
+// produce eleven. Making it exact would need a transaction spanning the counter
+// and the run, and pkg/entity's store is single-entity with no cross-entity
+// compare-and-swap; the create guard is a key-absence predicate that cannot
+// express "count < N". A ceiling exists to stop runaway automation rather than
+// to enforce an invariant, so that trade is taken openly rather than papered
+// over with a counter entity that would drift upward on any crash between
+// teardown and decrement -- and a drifted counter denies service permanently,
+// which is the worst way for an admission control to fail.
+func (c *Controller) admit(ctx context.Context, r *run_v1alpha.Run) (bool, error) {
```
phinze, line 35: Worth a decision on queue versus reject. RFD "Concurrency and cost" is explicit that exceeding the limit should reject with a clear error "because a queue with no visible depth is worse than a refusal", and asks for wording like "9 of 10 concurrent runs of `console` are already active".

Queueing has a sharper consequence than the message, though. I watched a run whose CLI had already reported `did not start within 2m` go on to succeed once the slot freed. So a user who retries after that error gets two executions, which for a hand-invoked migration is the thing the rest of this design works hardest to prevent.

The error also carries a doubled prefix, `sandbox/sandbox/run-...`, from the proxy's bare-name retry re-prefixing an id that already had one.

controllers/run/controller.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	var ver core_v1alpha.AppVersion
+	ver.Decode(verResp.Entity().Entity())
+
+	cfgSpec, err := coreutil.ResolveConfig(ctx, c.EAC, &ver)
+	if err != nil {
+		return nil, "", fmt.Errorf("resolving config: %w", err)
+	}
+
+	spec, err := appspec.Build(c.Log, appspec.Options{
```
phinze, line 454: `[tasks.<name>.env]` never reaches the run. `buildTasksConfig` stores it in `ConfigSpecTasks.Env` (build.go:336) and nothing reads it back: `appspec.Build` merges `cfgSpec.Variables`, then a matching service's env, then `opts.ExtraEnv` (appspec.go:245-270). A run has no service and this call passes no `ExtraEnv`, so the declared env is silently absent.

Confirmed by declaring `TASK_SECRET` on a task and running `env`. The container gets `MIREN_APP` and friends, not `TASK_SECRET`.

Two things make it sting. It's the RFD's flagship example (`[[tasks.session.env]] ANTHROPIC_API_KEY required = true sensitive = true` on the agent sandbox), and `migrateConsoleService` (build.go:283) copies `EnvVars` from a legacy `[services.console]` into the task, so the migration promises to carry env that then gets dropped.

controllers/run/controller.go
```diff
@@ -0,0 +1,617 @@
...
+		// failing to bind one, and a miren disk is single-writer and would
+		// contend with the service already holding its lease.
+		SkipPorts: true,
+		SkipDisks: true,
+
+		// Attachable whether or not anyone attaches now. containerd fixes the
+		// stdin FIFO at task creation, so a run started detached could never be
+		// attached later otherwise -- and "start detached, attach later" is the
+		// point of separating the two.
+		Stdin: true,
```
phinze, line 477: `Stdin: true` without `Tty: true`, so the container never gets a pty. `miren app run -- tty` reports `not a tty` with the client on a real `/dev/pts/0`.

Meanwhile the CLI puts the local terminal in raw mode and streams window updates that `Hub.Resize` documents as a no-op without a TTY, and `setupExecIO` fills in a `ShellOptions` that `Attach` has no parameter for, so it's collected and dropped. For a console session that's no job control, no line editing, no prompt.

controllers/sandbox/sandbox.go
```diff
@@ -2167,7 +2277,24 @@ func (c *SandboxController) BootContainers(
...
 		}
-		cioOpts = append(cioOpts, cio.WithStreams(nil, sl, sl.Stderr()))
+
+		// An attachable container gets a Hub: its stdin FIFO exists from the
+		// start (containerd only wires one up when a reader is supplied here,
+		// and it cannot be added later), and its output is teed to whoever is
+		// attached as well as to the logs.
+		var hub *Hub
+		if container.Stdin {
+			hub = c.hubs.GetOrCreate(sb.ID, container.Name)
```
phinze, line 2309: The Hub is created before `cc.NewTask`, and `hubs.RemoveAll` only ever runs from `StopSandbox` (line 3152). `createSandbox`'s failure defer marks the sandbox DEAD without tearing it down, and its own comment says as much ("Unlike the graceful StopSandbox path, this defer doesn't tear the sandbox down"), so any run whose sandbox fails to boot leaves a Hub behind.

A client then finds that Hub and blocks in `io.Copy` against a container that never existed. That's the permanent hang on bare `miren app run`: the run had already failed, and the CLI was still stuck 3m21s later. The 2m deadline in `attachToRun` can't fire, because it's only evaluated after `Attach` returns.

Not specific to the empty-command bug either. Any boot failure leaves the same trap.

servers/exec/attach.go
```diff
@@ -0,0 +1,162 @@
...
+		}()
+	}
+
+	in := stream.ToReader(ctx, args.Input())
+	defer in.Close()
+
+	// Pump the client's input into the container until the client goes away.
+	// Reaching EOF here ends this attach and nothing else: the Hub's stdin
+	// stays open, so the container's shell doesn't see EOF and exit.
+	_, err := io.Copy(hubStdin{hub}, in)
```
phinze: This is the attach's termination condition, and it's the wrong one: the RPC ends when the client's stdin ends, not when the run does.

Non-interactively that's immediate, since stdin is `/dev/null` in CI, so `--task slow` returned in 3.2s with exit 0 while the run kept going. Interactively it never fires, so the call hangs after the command exits and the next keystroke lands on a closed pipe:

```
attach: relaying input: io: read/write on closed pipe
```

which the CLI surfaces as exit 1 for a run that recorded `succeeded` / `0`.

The Hub has no "container is gone" signal to select on. A `Done()` channel closed by `Hub.Close()` would give one. It'd also let the CLI stop guessing: `reportRunExit` reads the run once and treats a non-terminal status as "no code", which is where the 1-in-5 flake in `TestAttachedRunPropagatesExitCode` comes from.

cli/commands/app_run.go
```diff
@@ -1,22 +1,106 @@
...
+	}
+
+	return reportRunExit(ctx, runs, info.Id())
+}
+
+// attachToRun streams a run's terminal until the client disconnects or the run
+// ends.
+func attachToRun(ctx *Context, runs *app_v1alpha.RunsClient, runID, sandboxName string) error {
+	opt := new(exec_v1alpha.ShellOptions)
 	in, out, winUpdates, cleanup := setupExecIO(ctx, opt)
```
phinze, line 152: Flagging here because `exec_io.go` isn't in this diff, but this is the call that sets it up.

`setupExecIO` calls `stdinCon.SetRaw()` (exec_io.go:44), which clears `ISIG`, and nothing on this path handles SIGINT (the only `signal.Notify` there is SIGWINCH). So Ctrl-C reaches the container as a byte instead of interrupting the client, and there's no escape sequence either.

So you can't leave a run you've attached to, and killing the client from another terminal is the only way out. That sits badly against the help text ("Disconnecting is not cancelling... or leave it and let its timeout reap it") and is rough for the console task's 8h default: attach, and that terminal is gone for the day. Docker's Ctrl-P Ctrl-Q is the usual shape.

Same call site drops the `opt` it fills in, since `Attach` takes no options parameter.

phinze (changes requested):
Sorry in advance, this is a wall of text. REQUEST_CHANGES, and I feel bad about the volume, because the new machinery here is genuinely the good part. The Run entity, the controller, admission, the Hub, and the GC and scheduler over in #1045 all hold up. It's the absorbed `miren app run` that keeps falling over.

I tried not to waste your time with guesses. Everything below was reproduced on a dev cluster with the testdata apps, I spot-checked the findings by hand, and then ran the whole draft back through Codex to argue against it. That pass killed one of my claims (the `runs.go:305` note, where I had the mechanism wrong) and caught the task-env bug I'd missed, so I think what's left holds up.

The exit code, which RFD-97 files under "the behavior nobody should notice", is wrong in both directions:

| invocation | what actually happened | what the CLI said |
|---|---|---|
| `--task fail`, stdin at EOF | exit 3 | `0`, on 1 run in 5 |
| `--task slow`, stdin at EOF | still running | `0`, after 3.2s |
| `-- sleep 5`, real terminal | exit 0 | `1`, after hanging |
| bare `miren app run` | run failed | hangs forever, Ctrl-C dead |

Bare `miren app run` never gets you a console. The command resolves to `""`, and since the sandbox only sets `WithProcessArgs` for a non-empty command, the container falls through to the image's own configured process. On stack-built images that's nothing at all, so runc refuses with `args must not be empty`. On an image that defines a CMD you'd get the app's normal process instead of a shell. On a real terminal it then hangs permanently on a Hub that leaked before the container failed to boot, and Ctrl-C can't break out because the terminal is raw with no SIGINT handler. There's also no way to detach from a run you did attach to, which is the thing the help text promises most emphatically.

One more that isn't about the CLI at all: `[tasks.<name>.env]` never reaches the container. It's stored at build time and read back nowhere, so the RFD's own worked example (`ANTHROPIC_API_KEY` on the agent session task) comes up empty, and the `[services.console]` migration carries `EnvVars` into a task that drops them.

My read on how it got this far: every blackbox case drives `--task`, so nothing exercises the surface that already existed. Bare invocation, quoted args, a pty, and an exit code reaching a script have no coverage anywhere.

Hopefully you and your robo pal can chew through it together. The console fallback, `Tty`, argv as a list, task env, and wiring up `ConsoleMaxConcurrent` are all small, and I'm happy to push those if it saves you a pass. The attach lifecycle is the one I'd leave with you, since it's a question about who owns the Hub rather than a patch.

Housekeeping: the body's "Known issue" about losing one exit code in eight reads stale. `run: report the exit to the run, not just the sandbox` fixed it and says so in its own message, but it's still listed as future work here and in #1045.

`--p+🤖`

evanphx:
🤖 All nine reproduce, all nine are fixed, and the volume was the right call — thank you for running it rather than guessing. Your diagnosis of how it got here was also correct: every blackbox case drove `--task`, so the surface that already existed had no coverage. Bare invocation, quoted args, task env, a pty, and an exit code reaching a script all have tests now.

Fixing the attach lifecycle exposed a tenth, underneath the ones you listed: output written *before* a client attaches was dropped entirely, so `miren app run -- echo hi` printed nothing. The container starts on sandbox boot while the client is still connecting, and for a short command teardown can win that race too. The Hub keeps a bounded replay buffer now, and a torn-down Hub lingers a minute so a command that finished early still hands its output over. The old exec path could not lose this — it started the process with the client's streams already connected.

Also removed the stale "Known issue" from both PR bodies. You were right that it described something `run: report the exit to the run` had already fixed.

servers/app/runs.go
evanphx: 🤖 Confirmed, and the comment above it described a chain this PR had already removed. `resolveCommand` returns `/bin/sh` as the last resort; `appspec` prepends the entrypoint, so that is `<entrypoint> /bin/sh` where one exists and a bare shell where it does not — the two rungs the exec server applied. Covered by `TestBareRunOpensAConsole`.

servers/app/runs.go
evanphx: 🤖 Real, fixed — but I went a different way than the schema list, so flagging rather than burying it. The container's args are always built as `sh -c <string>` (`sandbox.go:2845`), so a list has to be shell-quoted into a string regardless. Quoting at the boundary keeps one field meaning one thing, and it is what `miren app runs` should display. It also converges the two paths correctly: a task's `command = "rails db:migrate"` stays shell-interpreted, while `-- echo "a  b"` becomes inert. Your call — say so and I will move it to the schema.

servers/app/runs.go
evanphx: 🤖 Right, never read. It is in `appconfig` now behind a shared `runapi.MaxConcurrent`, used by both the controller's admission and the server's refusal — so the number quoted to a user is the number enforced, which was the other half of it.

controllers/run/admission.go
evanphx: 🤖 Both parts fixed. Manual invokes refuse with your wording. But the refusal alone does not close the double-execution hole — a *pending* run is not counted — so the CLI also cancels a run it gave up waiting for. That is the part that actually prevents the retry-after-timeout case you hit. Doubled `sandbox/sandbox/` prefix fixed too: the proxy only prepends when the id is bare.

controllers/run/controller.go
evanphx: 🤖 Confirmed and fixed. `appspec` merges task env on the same footing as a service's — overriding globals, still losing to system-managed vars. Tested at both levels, and I checked the unit test fails without the fix rather than trusting it.

controllers/run/controller.go
evanphx: 🤖 Fixed: `tty` is a Run field the client sets when stdin is a terminal and it is going to attach, so deploy and schedule runs keep stdout and stderr separate and un-rewritten. containerd fixes this at task creation like stdin, so it cannot be settled on attach.

One limitation that falls out of that and is worth stating plainly: a run started with `--detach` has no pty, so attaching to it later gets you a working terminal but no job control. Fixing it properly would mean deciding the pty at create time for *any* run that might be attached, which costs every scheduled task a merged log stream. Happy to revisit if the detach-then-attach console turns out to be a real workflow.

controllers/sandbox/sandbox.go
evanphx: 🤖 This was the permanent hang, and it needed fixing twice: `saga_controller.go` replaces `createSandbox` wholesale, so it does not inherit the defer. The teardown went into `undoBootContainers` alongside the disk-lease and token-state releases duplicated there for the same reason. Recorded in the frozen-hash comment.

servers/exec/attach.go
evanphx: 🤖 Right on both the diagnosis and the shape of the fix. The Hub has a `Done()` closed at teardown, and the attach selects on that; the input pump ending now just stops the pump. That fixes your table in both directions and removes the guessing in `reportRunExit`, which waits for the run to settle instead of reading once — so the 1-in-5 flake in `TestAttachedRunPropagatesExitCode` is gone.

cli/commands/app_run.go
evanphx: 🤖 Fixed: Ctrl-P Ctrl-Q, docker's shape. A lone Ctrl-P still passes through, at the cost of one buffered byte until the next arrives — the same trade docker makes. Ctrl-C deliberately still reaches the command, and with the pty fix it is a real SIGINT there rather than a byte. Both are in the help text, which was promising detaching without saying how.

cli/commands/run_detach.go
```diff
@@ -0,0 +1,100 @@
...
+	d.buf = append(d.buf[:0], out[copied:]...)
+
+	if err != nil {
+		d.closed = true
+		if copied > 0 {
+			return copied, nil
+		}
+		return 0, err
+	}
+	return copied, nil
```
phinze: A held-back Ctrl-P returns `(0, nil)` here, and that's fatal one layer down. `stream.ServeReader` is in streaming mode on this path (no `WithBulkBatching`), so `serveReader.Recv` ships `buf[:0]`, and the server's `rscReader.Read` has `if len(data) == 0 { return 0, io.EOF }` (`pkg/rpc/stream/stream.go:84`). The input pump's `io.Copy` ends there and never restarts.

The `Done()` fix makes that quieter rather than better: output keeps flowing, but stdin is gone, and since the server has stopped calling `Recv` the detachReader is never read again — so Ctrl-P Ctrl-Q can't get you out either. Killing the client from another terminal is the only way left.

Ran the reader standalone to be sure of the zero-length read:

```
Read #1 -> n=0 err=<nil> bytes=""
Read #2 -> n=2 err=<nil> bytes="\x10l"
```

The byte itself isn't lost, so this just needs to never hand `(0, nil)` to the stream — loop back and read again when the filter consumed everything.

cli/commands/app_run.go
```diff
@@ -1,45 +1,316 @@
...
+		if runIsOver(ctx, runs, runID) {
+			return false, nil
+		}
+
+		if time.Now().After(deadline) {
+			// Cancel rather than abandon. The run is still pending, so leaving
+			// it would let it execute later, after this command has already
+			// told the caller it did not start -- and someone who retries on
+			// that message would get the command executed twice.
+			if _, cerr := runs.CancelRun(ctx, runID); cerr != nil {
```
phinze, line 215: Right for `AppRun`, but `attachToRun` is shared and `AppAttach` lands here too — so `miren app attach` on a run that takes more than two minutes to get a container (cold node, big image pull) cancels a run the caller only wanted to watch. `runIsOver` doesn't catch it, since a booting run isn't terminal, and "is not running yet" keeps `attachTargetMissing` true right up to the deadline.

`app-attach.md` promises the opposite three lines above where we'd be killing it: "Attaching and detaching are invisible to the run itself."

Probably just wants the cancel to be the caller's call — a bool on `attachToRun`, or hoist it into `AppRun`.

servers/exec/attach.go
```diff
@@ -0,0 +1,188 @@
...
+	copyErr := make(chan error, 1)
+	go func() {
+		_, err := io.Copy(hubStdin{hub}, in)
+		copyErr <- err
+	}()
+
+	select {
+	case <-hub.Done():
+		// The container is gone, which is the one thing that genuinely ends an
+		// attach. Drain whatever the client already sent so a command's last
```
phinze: Minor: the comment says drain, but this returns straight away and `in.Close()` fires on the way out, abandoning `io.Copy` wherever it is. I don't think it costs anything (the container's gone, so that input has nowhere to land), but it reads as a behavior that's already there.

servers/app/runs.go
```diff
@@ -0,0 +1,463 @@
...
+// Counts running rather than pending, matching what the controller's admission
+// counts, so the number quoted here is the number actually enforced. The check
+// is racy against a simultaneous invoke and deliberately so: it exists to give
+// a person a clear answer, while the guarantee that matters -- at most one run
+// of a task at max_concurrent = 1 -- is held by the controller's slot entity,
+// which is exact.
+func (r *AppInfo) refuseIfAtLimit(ctx context.Context, appID entity.Id, taskName string, task *core_v1alpha.ConfigSpecTasks) error {
+	maxConcurrent := runapi.MaxConcurrent(task, taskName)
+
+	results, err := r.EC.List(ctx, entity.Ref(run_v1alpha.RunAppId, appID))
```
phinze: Minor: this is the app index, which `admitByCount` in `controllers/run/admission.go` deliberately avoids — "Runs are durable history, so the app index grows without bound." Every manual invoke decodes the app's whole run history, deploy- and schedule-triggered ones from #1045 included, to count a handful of running ones.

Human-rate rather than a hot loop, so it's latency on `miren app run` rather than anything worse. But the running-status index is right there and the filter's already written.

phinze (approved):
Two stragglers inline, both small and both in the CLI, plus a couple of notes not worth holding anything up for. Start with the Ctrl-P one: a lone Ctrl-P is previous-line in readline, so bash and zsh out of the box, and pressing it leaves you attached with a dead keyboard and no way to detach — the trap the escape sequence was added to remove. It's a few lines in one file, so approving rather than bouncing it again.

The replay buffer wasn't on my list. Good catch, and the kind that only surfaces once the first fix is in.

`--p+🤖`

cli/commands/run_detach.go
evanphx: 🤖 Good catch, and worse than a nit — thank you for running the reader standalone. Confirmed against `pkg/rpc/stream/stream.go:84`: a zero-length payload is EOF to the transport, so one Ctrl-P ended stdin for the session, and `Done()` made that quieter rather than better exactly as you describe. A held byte now waits for the next one instead of reporting an empty read, so `Read` never returns `(0, nil)`. There is a test for the invariant, and I checked it fails against the old behaviour rather than trusting it.

cli/commands/app_run.go
evanphx: 🤖 Right, and the doc line three above is the tell. Cancelling is the caller's decision now: `app run` created the run, so a pending one is a command that never executed and would otherwise run after the caller was told it did not — while `app attach` only asked to watch, and killing a run over a slow image pull is not its business.

servers/exec/attach.go
evanphx: 🤖 The comment was wrong, and you are right that it costs nothing: the container is gone, so that input has nowhere to land. Reworded to say what actually happens rather than describing a drain that is not there.

servers/app/runs.go
evanphx: 🤖 Fixed — it lists the running index now, with the same reasoning admission gives. You are right that it was only latency, but the filter was already written two files over.

PR #1045, opened 2026-08-07
run: deploy and schedule triggers, with retention (RFD-97, 3/3)

evanphx:
Third and final PR implementing [RFD-97 "App Tasks"](https://github.com/mirendev/rfd/blob/main/rfds/0075/0097-app-tasks.md) (MIR-853). Stacked on #1043.

Everything that creates a `Run` without a person asking, plus the retention policy the schedule trigger's correctness depends on.

## `deploy` — gating the version flip

Creates a Run for every task with `trigger = "deploy"` and waits, between addons being ready and `ActiveVersion` flipping.

That placement is the design. Gating there means the runs complete or fail while the previous version is still the only thing serving: no partial rollout to unwind, no traffic to shift back, no compensation step. A failed deploy task is a failed deploy rather than a half-promoted one, and that holds for every rollout strategy because the gate sits ahead of all of them rather than inside any.

It also has a consequence the docs now state outright, because a user meets it on the first migration they write: the task runs on the **new** image while the **old** code is still serving every request. Schema changes have to be compatible with both at once. That was previously framed as a rollback concern, which was wrong — expand-contract is mandatory on the normal path.

The cost of gating whole-deploy is that a strategy can't gate per stage — "run the backfill after the canary validates" isn't expressible. When that's wanted, the right shape is a step in the strategy's sequence referencing a task by name, not another field on the task: the task says *what*, the strategy says *when*. Keeping that split means adding per-stage hooks later doesn't invalidate this.

The saga orders actions by data dependency rather than registration order, so the gate's position is asserted rather than assumed — a test reads `ExecutionOrder()` and pins that it lands after `provision-addons` and before `set-active-version`.

**Two paths, not one.** The saga builder is behind the `sagas` labs flag, so wiring only the saga meant the gate did nothing on the path that actually runs. The blackbox test caught exactly that — the deploy succeeded despite its task failing. Both paths gate now.

**Rollbacks deliberately don't run tasks.** That path activates a version that already exists, and re-running a migration against a database already migrated forward is more likely to be wrong than right, with no down-migrations to make it reversible. RFD-97 doesn't decide this either way, so the reasoning and the hook point are written down at the call site rather than left as a silent omission.

## `schedule` — one firing, one run

A tick is one scheduled firing, identified by app, task, and the instant it comes due. Because both scheduling forms are calendar expressions — `every` having been desugared at parse time — a tick is a pure function of the stored config, so every replica derives exactly the same ones without coordinating. The run's entity name is derived from the tick, which makes creating it a put-if-absent that etcd resolves to a single winner.

No leader election and no Valkey. RFD-79 left open whether to reuse cloud's `internal/bgtask/cron.go`; this says no for the runtime layer specifically — that code is good where Valkey is already a dependency, and the runtime's scheduler doesn't have one.

Deriving that name is load-bearing in a way worth stating, because the first version got it wrong. A run's identity is cluster-wide within its kind, and the readable prefix joins app and task with a hyphen that neither excludes — task names are TOML table keys, so `[tasks.db-migrate]` is ordinary. Hashing only the instant made app `api` + task `db-migrate` and app `api-db` + task `migrate` the same run whenever their schedules agreed, which for two daily jobs is daily. The failure is not a duplicate but a silent absence: the loser reads the conflict as "another replica claimed this tick", advances its frontier, and never retries. App and task are inside the digest now.

Three behaviors that needed deciding rather than defaulting:

- **Missed ticks are skipped, not backfilled.** The frontier starts at now, so an outage's worth of missed cleanups doesn't replay on recovery and turn an outage into an incident.
- **Overlapping ticks are skipped *and recorded*.** That falls out of the concurrency cap, but it earns its own status and a Warn: "my job stopped running" and "my job got slower than its interval" are indistinguishable from outside if the skip leaves a silent gap.
- **A replica only fires a tick its own clock has passed**, so skew makes a fast replica fire early rather than twice.

The frontier advances per fired tick rather than in one jump, so a tick whose fire fails is retried instead of being stepped over permanently. `Stop` waits for an in-flight sweep rather than only cancelling its context, since otherwise a sweep could still create runs after shutdown returned.

## Retention, which is a correctness constraint

The dedup guard for a scheduled tick *is* the run entity's existence. Deleting one while any replica could still evaluate that tick — partitioned, or restarted behind the others — lets create-if-absent succeed a second time and the job double-fires.

So schedule-triggered runs are **exempt from the count cap entirely** and retired only by age, well past any plausible partition. A count applied naively to a busy app would evict same-day ticks and silently reintroduce double-firing. That's why it's a constraint and not a tuning knob.

Everything else is count-and-age together, per **app** rather than per task (an app with 200 tasks would otherwise keep 200× the cap). Both conditions must hold: count alone evicts a burst of recent runs, age alone lets a busy app grow without bound. Console runs get a shorter window — their value is an audit record and a scrollback, not a permanent log.

**Sandbox reaping closes the leak for good.** Any sandbox tagged with a run id whose run no longer exists is stopped. That's what makes the old exec-sandbox leak un-leakable rather than merely less likely, since teardown used to be a deferred call in a request handler that a crash simply skipped.

This is another GC alongside the existing ones; @phinze filed MIR-1573 for a consolidation pass rather than blocking on it here.

## Docs

`docs/docs/tasks.md`: the three triggers, the scheduling behaviors above, disconnect-is-not-cancellation, task-only apps, and the `[services.console]` migration.

The limitations section describes behavior rather than absences, after review pointed out it was listing negative space. The two a user actually walks into:

- **A task cannot reach a Miren disk**, including one belonging to a service in the same app. Disks are single-writer and a run beside the lease-holder would block or race it, so runs skip them. It's silent — nothing warns, and the program reports its own "no such file or directory" — which makes a backfill over disk-backed data inexpressible today.
- **No ordering between deploy tasks.** They all start at once and the deploy waits for all of them; ordering means putting both steps in one task.

## Testing

`make lint` clean, docs lint clean. `make test` clean apart from `pkg/addon/postgresql` and `pkg/addon/rabbitmq`, which fail identically on `main`.

Thirteen scheduler tests cover dedup across two schedulers sharing a store, no-backfill, the skip being recorded and stamped, resumption once the predecessor finishes, the frontier not stepping over an unfired tick, `Stop` waiting for a sweep in flight, and the tick name distinguishing two jobs that share an instant while staying stable for one job across timezones. Seven GC tests cover the scheduled exemption, the age floor, console's shorter window, count-and-age together, and orphan reaping.

Eleven blackbox tests pass against a dev cluster, covering this PR and #1043 together: task-only deploy reporting `ready`, detached runs, exit-code recording, `logs run`, cancellation, attached exit-code propagation, an attached run waiting for its command rather than for its own stdin, bare `miren app run` opening a console, quoted arguments surviving as argv, declared task env reaching the container, and the deploy gate failing a deploy whose task fails.

### DIFF
--- a/controllers/run/scheduler.go
+++ b/controllers/run/scheduler.go
@@ -259,6 +259,16 @@ func (s *Scheduler) fireTick(
 		Tick:        tick.UTC().Format(time.RFC3339),
 	}
 
+	// A skipped tick is born terminal, so it never passes through the
+	// controller that would otherwise stamp its timestamps. Retention is
+	// measured from EndedAt, and a scheduled run with none is retained
+	// unconditionally -- so without this, skipped ticks accumulate forever on
+	// exactly the busy task that produces the most of them.
+	if status == run_v1alpha.SKIPPED {
+		run.StartedAt = tick
+		run.EndedAt = tick
+	}
+
 	// Create is put-if-absent on a name derived from the tick, which is the
 	// entire single-execution guarantee: every replica derives the same name,
 	// etcd admits one, and the rest learn the tick is taken.
--- a/controllers/run/scheduler.go
+++ b/controllers/run/scheduler.go
@@ -160,8 +160,10 @@ func (s *Scheduler) sweepTask(
 	expr, err := oncalendar.Parse(task.Schedule)
 	if err != nil {
 		// Config validation rejects unparseable schedules, so reaching here
-		// means a version predating that check. Skip rather than retrying every
-		// interval forever.
+		// means a version that predates that check. The error is returned and
+		// logged by the caller on every sweep, which is noisy but honest: a
+		// task the platform cannot schedule should keep saying so rather than
+		// going quiet and looking healthy.
 		return fmt.Errorf("parsing schedule %q: %w", task.Schedule, err)
 	}
 
@@ -200,16 +202,17 @@ func (s *Scheduler) sweepTask(
 		}
 	}
 
-	if len(fired) > 0 {
-		s.mu.Lock()
-		s.frontier[key] = fired[len(fired)-1]
-		s.mu.Unlock()
-	}
-
+	// Advance the frontier only past ticks that were actually handled. Moving
+	// it first would drop any tick whose fire failed -- permanently, since the
+	// frontier is what decides a tick is behind us.
 	for _, tick := range fired {
 		if err := s.fireTick(ctx, app, ver, appName, task, tick); err != nil {
 			return err
 		}
+
+		s.mu.Lock()
+		s.frontier[key] = tick
+		s.mu.Unlock()
 	}
 
 	return nil
--- a/controllers/run/scheduler.go
+++ b/controllers/run/scheduler.go
@@ -1,0 +1,312 @@
+package run
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/base32"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+	"sync"
+	"time"
+
+	coreutil "miren.dev/runtime/api/core"
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/api/entityserver"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	run_v1alpha "miren.dev/runtime/api/run/run_v1alpha"
+	"miren.dev/runtime/pkg/cond"
+	"miren.dev/runtime/pkg/entity"
+	"miren.dev/runtime/pkg/oncalendar"
+)
+
+// SchedulerInterval is how often the scheduler looks for ticks that have come
+// due. It bounds how late a run can start, not how precisely ticks are
+// computed: tick times come from the calendar expression, so a slow poll delays
+// a run without shifting the schedule.
+const SchedulerInterval = 15 * time.Second
+
+// Scheduler creates runs for tasks whose calendar expression has come due.
+//
+// It holds no leader election and no external coordination. A tick is a pure
+// function of the stored calendar expression, so every replica derives exactly
+// the same firing times; the run's entity id is derived from the tick, which
+// makes creating it a put-if-absent that etcd resolves to a single winner. The
+// losers see the run already exists and move on.
+//
+// The consequence worth stating: the dedup guard is the run entity's existence.
+// Deleting a tick's run resurrects that tick, which is why retention treats
+// scheduled runs as a correctness constraint rather than a preference.
+type Scheduler struct {
+	Log *slog.Logger
+	EC  *entityserver.Client
+	EAC *entityserver_v1alpha.EntityAccessClient
+
+	Interval time.Duration
+
+	cancel func()
+
+	// mu guards frontier, which records the newest tick already considered for
+	// each (app, task). It is in-memory on purpose: on startup the frontier
+	// begins at now, which is what makes missed ticks skipped rather than
+	// backfilled.
+	mu       sync.Mutex
+	frontier map[string]time.Time
+}
+
+func NewScheduler(log *slog.Logger, ec *entityserver.Client, eac *entityserver_v1alpha.EntityAccessClient) *Scheduler {
+	return &Scheduler{
+		Log:      log.With("module", "run-scheduler"),
+		EC:       ec,
+		EAC:      eac,
+		Interval: SchedulerInterval,
+		frontier: make(map[string]time.Time),
+	}
+}
+
+func (s *Scheduler) Start(ctx context.Context) {
+	s.Log.Info("starting run scheduler", "interval", s.Interval)
+
+	ctx, s.cancel = context.WithCancel(ctx)
+	go s.run(ctx)
+}
+
+func (s *Scheduler) Stop() {
+	if s.cancel != nil {
+		s.cancel()
+	}
+}
+
+func (s *Scheduler) run(ctx context.Context) {
+	ticker := time.NewTicker(s.Interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			s.Log.Info("run scheduler stopped")
+			return
+		case <-ticker.C:
+			if err := s.Sweep(ctx, time.Now()); err != nil {
+				s.Log.Warn("scheduler sweep failed", "error", err)
+			}
+		}
+	}
+}
+
+// Sweep fires every tick that has come due since the last sweep.
+//
+// now is a parameter so tests can drive it; in production it is time.Now.
+func (s *Scheduler) Sweep(ctx context.Context, now time.Time) error {
+	apps, err := s.EAC.List(ctx, entity.Ref(entity.EntityKind, core_v1alpha.KindApp))
+	if err != nil {
+		return fmt.Errorf("listing apps: %w", err)
+	}
+
+	for _, e := range apps.Values() {
+		var app core_v1alpha.App
+		app.Decode(e.Entity())
+		app.ID = entity.Id(e.Id())
+
+		if app.ActiveVersion == "" {
+			continue
+		}
+
+		if err := s.sweepApp(ctx, &app, now); err != nil {
+			s.Log.Warn("failed to sweep app for scheduled tasks", "app", app.ID, "error", err)
+		}
+	}
+
+	return nil
+}
+
+func (s *Scheduler) sweepApp(ctx context.Context, app *core_v1alpha.App, now time.Time) error {
+	var ver core_v1alpha.AppVersion
+	if err := s.EC.GetById(ctx, app.ActiveVersion, &ver); err != nil {
+		return fmt.Errorf("reading active version: %w", err)
+	}
+
+	cfgSpec, err := coreutil.ResolveConfig(ctx, s.EAC, &ver)
+	if err != nil {
+		return fmt.Errorf("resolving config: %w", err)
+	}
+
+	var md core_v1alpha.Metadata
+	if err := s.EC.GetById(ctx, app.ID, &md); err != nil {
+		return fmt.Errorf("reading app metadata: %w", err)
+	}
+
+	for _, task := range cfgSpec.Tasks {
+		if task.Schedule == "" {
+			continue
+		}
+		if err := s.sweepTask(ctx, app, &ver, md.Name, task, now); err != nil {
+			s.Log.Warn("failed to sweep task", "app", app.ID, "task", task.Name, "error", err)
+		}
+	}
+
+	return nil
+}
+
+func (s *Scheduler) sweepTask(
+	ctx context.Context,
+	app *core_v1alpha.App,
+	ver *core_v1alpha.AppVersion,
+	appName string,
+	task core_v1alpha.ConfigSpecTasks,
+	now time.Time,
+) error {
+	expr, err := oncalendar.Parse(task.Schedule)
+	if err != nil {
+		// Config validation rejects unparseable schedules, so reaching here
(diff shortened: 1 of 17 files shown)


evanphx:
Rebased on the updated #1043, which fixes the exit-code race that was failing `test-blackbox (shard 4)` here — `TestAttachedRunPropagatesExitCode` was the attached form of it. The stale-command-docs lint failure came from the same place and is fixed there too.

No changes to this PR's own content.

controllers/run/scheduler.go
```diff
@@ -0,0 +1,312 @@
...
+//
+// The timestamp is hashed rather than embedded: entity names are a single clean
+// segment, and an RFC 3339 string carries colons and a plus. The hash is
+// truncated because it only has to separate ticks of one task on one app, not
+// resist collision by an adversary.
+func tickRunName(appName, task string, tick time.Time) string {
+	sum := sha256.Sum256([]byte(tick.UTC().Format(time.RFC3339)))
+	digest := strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(sum[:6]))
+	return fmt.Sprintf("%s-%s-%s", appName, task, digest)
+}
```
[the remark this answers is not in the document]
evanphx: 🤖 Skipping this one — the collision isn't reachable. `tickRunName` returns `fmt.Sprintf("%s-%s-%s", appName, task, digest)`, so the app and task names are already the first two segments of the name; the digest only has to separate ticks *within* one app's task. Two apps with the same task name and the same tick get `app-a-cleanup-<d>` and `app-b-cleanup-<d>`, which are distinct entity ids.

Hashing the app and task into the digest as well would be harmless but wouldn't buy anything, and it would make the name less legible in `miren app runs` output.

evanphx:

- **The deploy gate accepted `SKIPPED` as success.** The gate exists to prove every task ran; a skipped run did not. It fails the deploy now, via a named predicate rather than an implicit branch.
- **The gate swallowed every run-read error**, turning a permission or transport failure into a two-hour wait ending in a misleading timeout. Only a not-yet-visible run is retried.
- **The scheduler advanced its frontier before firing.** Since the frontier is what decides a tick is behind us, a failure part-way through dropped the rest permanently. It advances one tick at a time, after each is handled.
- Run GC no longer re-reads every retained run's sandbox on every sweep; scheduler and GC start with the other background controllers after the reconcile manager; the parse-failure comment matches the code; a test asserts length before indexing; scheduling constraints are admonitions.

Declined one with reasoning inline (tick-name collisions — the app and task are already the first two segments of the name).

evanphx:
🤖 Two more from the review bodies, both staleness I introduced with my own earlier fixes:

- `Scheduler.Stop` cancelled the context and returned, so a sweep already running could still create runs after it returned — the opposite of what Stop appears to promise a caller shutting down in order. It waits for the loop now, with a test that pins it.
- The tick-cap comment still claimed the frontier advance moves past dropped ticks. It stopped doing that when the frontier began advancing per fired tick.

controllers/run/scheduler.go
```diff
@@ -0,0 +1,354 @@
...
+// creating it a deduplicating operation.
+//
+// The timestamp is hashed rather than embedded: entity names are a single clean
+// segment, and an RFC 3339 string carries colons and a plus. The hash is
+// truncated because it only has to separate ticks of one task on one app, not
+// resist collision by an adversary.
+func tickRunName(appName, task string, tick time.Time) string {
+	sum := sha256.Sum256([]byte(tick.UTC().Format(time.RFC3339)))
+	digest := strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(sum[:6]))
+	return fmt.Sprintf("%s-%s-%s", appName, task, digest)
```
phinze, line 375: You refuted 🐰 on this but both 🐰 and 🤖 came up with realistic-ish repro of a collision scenario, which makes it seems like might as well just make the change.

---

🤖 The name segments aren't unambiguous, and the blast radius is a silently missing job rather than a duplicated one.

Entity identity is `types.Keyword(sc.ShortKind() + "/" + name)` (`api/entityserver/client.go:202`) — scoped by kind, cluster-wide, not per app. `validateTasks` (`appconfig/appconfig.go:609`) puts no charset restriction on task names; they're TOML table keys, so `[tasks.db-migrate]` is ordinary rather than exotic. So `app "api" + task "db-migrate"` and `app "api-db" + task "migrate"` both render `api-db-migrate-<digest>`, and the digest is `sha256(tick)` alone, so it matches whenever the instants do. `every = "6h"` and midnight-daily schedules across two apps land on identical timestamps routinely.

The loser of that `Create` takes the `cond.ErrConflict{}` branch at `scheduler.go:303-306`: a Debug-level "tick already claimed by another replica", then `return nil`. The caller then advances the frontier past the tick (`scheduler.go:239-241`), so it never retries. One app's scheduled job silently stops running, and the only trace says something untrue about why.

Folding `appName` and `task` into the hashed input alongside the tick fixes it without changing the dedup property.

servers/build/build.go
```diff
@@ -1696,6 +1696,31 @@ func (b *Builder) buildFromDir(ctx context.Context, name string, path string,
...
+		// Gate the flip on deploy-triggered tasks. This mirrors the saga's
+		// runDeployTasks action and has to exist here too: the saga builder is
+		// behind a labs flag, so this is the path that actually runs today.
+		if len(deployTriggeredTasks(&configSpec)) > 0 {
+			appRec, appErr := b.appClient.GetByName(ctx, name)
+			if appErr != nil {
+				return nil, fmt.Errorf("reading app for deploy tasks: %w", appErr)
+			}
+
+			if taskErr := b.runDeployTasks(ctx, name, appRec.ID, id, &configSpec,
```
phinze, line 1708: I sicked 🤖 on this sort of odd looking signature asking if there was a cleaner expression, and lo and behold 🤖 found other patterns nearby we can borrow!

---

🤖 **Suggestion: have `runDeployTasks` take a `StatusSender` instead of the `report` callback.**

The adapter this closure hand-rolls already exists. `NewRPCStatusSender` (`servers/build/status_registry.go:74`) takes exactly this stream type, returns a no-op sender when it's nil, and its `SendMessage` (lines 90-94) is the same four lines as the closure below: build a `Status`, `SetMessage`, send, Warn on error.

That makes this call site a one-liner:

```go
b.runDeployTasks(ctx, name, appRec.ID, id, &configSpec, NewRPCStatusSender(status, b.Log))
```

And it simplifies the saga side too, which is the part that surprised me — `runDeployTasks` in `build_saga.go` already *holds* a real `StatusSender` from `deps.statuses.SenderFor(in.StreamID)`, and wraps it in a closure purely to fit the `report func(format string, args ...any)` shape. It would just pass it straight through:

```go
b.runDeployTasks(ctx, in.AppName, appRec.ID, entity.Id(in.AppVersionID), &spec, status)
```

Inside `deploy_tasks.go`, the two `report(...)` calls (lines 70 and 170) become `status.SendMessage(fmt.Sprintf(...))`.

**One caveat, so this isn't billed as a pure refactor:** `rpcStatusSender.send` uses `context.Background()` (`status_registry.go:85`) where this closure uses the request `ctx`. Status sends would survive a canceled request instead of aborting with it. Every send on the saga path already works that way, so it's a consistency change rather than a new behavior — but it is a change.

docs/docs/tasks.md
```diff
@@ -0,0 +1,151 @@
...
+command = "bin/rails console"
+```
+
+Until you do, Miren treats a `[services.console]` block as a task and warns at build time. That's the behavior you wanted anyway — the service was costing you a sandbox nobody used.
+
+## Current limitations
+
+- **No per-task image or disks.** A task runs in the app's image and touches no disks directly. It can still reach anything your addons expose, since those arrive as environment variables.
+- **No ordering between deploy tasks.** They run concurrently.
+- **No per-stage deploy hooks.** A deploy task gates the whole deploy, before any rollout begins; "run this after the canary validates" isn't expressible yet.
```
phinze: My initial take here was "hm this is being too hard on ourselves and calling out negative space, maybe we cut it?" but then digging in I think there are more concrete behavior-centric things we can highlight here. I think the below two things are what I'd highlight in this section instead. The first is a more precise version of "no ordering" and the second is "no disks" - I think "per stage deploy hooks" can just fall away as referencing future work.

---

🤖 **The previous version is live during a successful deploy task.** Line 49 says it keeps serving when the task *fails*; line 54 raises expand-contract only for rollbacks. But the gate runs before `SetActiveVersion`, the launcher only builds pools on that flip (`controllers/deployment/launcher.go:39`), and the run's sandbox is built from the new version (`createDeployRun` sets `Version: versionID`; `controllers/run/controller.go:442`). A migration runs on the new image against the same database while the old code is serving. RFD-65's own migration sequence puts `drain_old` after `shift_traffic`, so that stays true once strategies land — expand-contract is mandatory on the normal path, not a rollback concern.

**A task cannot reach a disk-backed service's data at all.** Declaring `disks` on a task is a clean build-time failure (`TaskConfig` has no such field and `appconfig.go:296` sets `DisallowUnknownFields`). But a task on an app whose *service* has a disk just runs without it, silently: `controllers/run/controller.go:470` sets `SkipDisks: true`, and `appspec.go:116` never resolves a service for a run anyway, so the disk loop at `:333` can't match. The `log.Warn` at `appspec.go:351` sits inside the `!SkipDisks` block, so nothing is logged. The user sees their own program's "no such file or directory" and nothing connects it to a platform decision. For an app whose data lives on a miren disk rather than behind an addon, a backfill task can't reach what it exists to process — which is a wall to know about up front, not a limitation to note in passing.

phinze (approved):
Looking good, Taskmaster!

Seeing _Yet Another GC_ slide in here, I filed MIR-1573 for an eventual consolidation pass.

A few inlines, nothing major.

controllers/run/scheduler.go
evanphx: 🤖 You are both right and I was wrong to refute it — my reasoning assumed run names were scoped per app, and they are cluster-wide within the kind. Hyphens in task names settle it: `[tasks.db-migrate]` is an ordinary table key, so the prefix genuinely is ambiguous.

What convinced me was the blast radius you and 🐰 landed on. A duplicate would be loud; this is a job that silently stops, with a log line claiming another replica took the tick. App and task are inside the digest now, separated by NUL so neither can forge the other, with a test that fails against the old derivation.

servers/build/build.go
evanphx: 🤖 Taken, and it simplified more than I expected — the saga side was wrapping a real `StatusSender` in a closure purely to fit the `report` shape, so it just passes it through now. Both call sites are one line.

Billing the caveat honestly, since you flagged it: status sends move from the request context to `context.Background()`, so a message survives a cancelled request instead of aborting with it. Every send on the saga path already worked that way, so this makes the two agree rather than inventing a third behaviour — but it is a change, and it is in the commit message rather than only here.

docs/docs/tasks.md
evanphx: 🤖 Agreed on all counts, and your instinct to dig rather than cut was the right one — the section was listing negative space. Both concrete behaviours are in now.

The deploy-task one turned out to need a fix above it too: the note framed expand-contract as a rollback concern, which is wrong. The gate runs before the flip but the task runs on the new image, so the previous version is serving every request while the migration executes. That is the normal path, so it is a warning next to the trigger rather than a line in a limitations list.

The disk one I have stated as a wall rather than a note, for the reason you give — nothing warns, and the user sees their own "no such file or directory". Per-stage hooks dropped from the docs; the design rationale for the split stays in the PR body where it belongs.

PR #1047, opened 2026-08-11
Report app health to cloud, and give the uplink its own package

phinze:
The runtime knows everything about a user's apps and cloud knows one integer. RFD-94 settles the contract that closes that; this is the runtime half of its first slice, reporting health over the cluster-channel uplink. Three revs: open the link up, use it, then put it in the right house.

The first opens the uplink. It was built for Miren Anywhere and closed around it, with the Connector keeping its router and client private so nothing outside `pkg/anywhere` could register a handler or send. The sharper edge was `OnConnect`, which stored a single callback that the Connector already claims for time sync and org info. A second registration would have replaced the first, silently disabling NTP sync with nothing failing and nothing logged. It fans out now, with a test, since that regression is invisible at runtime.

The second adds the reporter. It fires on connect rather than a timer because the uplink discards whatever was queued while it was down, so anything that happened during an outage never reached the wire at all and can only come back by re-deriving current state from the entity store. Health reads through a shared `ListApps` rather than a second implementation: it's a classification over sandbox pool state joined with active versions and config specs, not a stored field, so a copy would drift and let `miren app list` and the dashboard disagree about whether an app is fine.

The third is the one worth arguing about. `pkg/anywhere` held two different things: the control plane (one persistent payload-agnostic link per cluster, what RFD-94 calls the unified uplink) and the data plane (POPManager and `connection.request`, forwarding public traffic into a NAT'd cluster). They touch at one seam, where the data plane's setup handshake arrives over the control plane. Left alone, every future tenant would import a package named for NAT traversal to reach cloud, and deploy reporting was going to be the second one. A rename wouldn't fix it, since POPManager genuinely is Anywhere. So the control plane moves to `pkg/uplink`, Anywhere becomes a tenant holding a client it no longer owns, and `coordinate` owns the link because several features share it. Time sync and org info go along, being properties of the link rather than of Anywhere. The `app.snapshot` payload moves the opposite way, out of the pipe and in with the reporter, because the link should keep understanding nothing about apps.

Only connect triggers a snapshot in this slice, so a freshly deployed app doesn't reach cloud until the cluster next reconnects. Deltas are the next piece. Smoke tested end to end against a local cloud, before and after the move, including corrupting the cloud row by hand and watching a reconnect converge it back from runtime truth.

Pairs with mirendev/cloud#196, which should land first.

Part of MIR-1558, closes MIR-1568

### DIFF
--- a/components/coordinate/appreport.go
+++ b/components/coordinate/appreport.go
@@ -1,0 +1,63 @@
+package coordinate
+
+import (
+	"context"
+	"time"
+
+	"miren.dev/runtime/pkg/anywhere"
+)
+
+// startAppReporter arranges for the cluster's app state to be reported to
+// cloud on every (re)connection of the uplink.
+//
+// Reconnect is the trigger rather than a timer because the uplink discards
+// anything queued while it was down. Whatever cloud missed during an outage
+// never reached the wire at all, so it can only be recovered by re-deriving
+// the current state from the entity store once the link is back. That makes
+// the source, not the pipe, responsible for completeness — see RFD-94.
+func (c *Coordinator) startAppReporter(conn *anywhere.Connector) {
+	conn.OnConnect(func(ctx context.Context) {
+		// Run off the connection goroutine: this reads the entity store, and
+		// the callback returns before the read and write loops start.
+		go c.reportAppSnapshot(ctx)
+	})
+}
+
+// reportAppSnapshot sends the cluster's current app set to cloud.
+//
+// Failure here is deliberately non-fatal and unretried. Reporting is value
+// flowing up to a value-add control plane, never a dependency flowing down:
+// a cluster runs indefinitely with no cloud attached, and visibility simply
+// stops updating until the link returns. Nothing the cluster does for its own
+// users may degrade because this failed, so the loudest response available is
+// a warning.
+func (c *Coordinator) reportAppSnapshot(ctx context.Context) {
+	if c.appInfo == nil || c.anywhere == nil {
+		return
+	}
+
+	apps, err := c.appInfo.ListApps(ctx)
+	if err != nil {
+		c.Log.Warn("failed to list apps for cloud snapshot", "error", err)
+		return
+	}
+
+	snapshot := anywhere.AppSnapshot{
+		ObservedAt: time.Now().UTC(),
+		Apps:       make([]anywhere.AppState, 0, len(apps)),
+	}
+
+	for _, a := range apps {
+		snapshot.Apps = append(snapshot.Apps, anywhere.AppState{
+			Name:   a.Name(),
+			Health: a.Health(),
+		})
+	}
+
+	if err := c.anywhere.SendMessage(anywhere.TypeAppSnapshot, snapshot); err != nil {
+		c.Log.Warn("failed to send app snapshot to cloud", "error", err)
+		return
+	}
+
+	c.Log.Info("reported app snapshot to cloud", "apps", len(snapshot.Apps))
+}
--- a/components/coordinate/appreport.go
+++ b/components/coordinate/appreport.go
@@ -4,8 +4,59 @@ import (
 	"context"
 	"time"
 
-	"miren.dev/runtime/pkg/anywhere"
+	"miren.dev/runtime/pkg/uplink"
 )
+
+// TypeAppSnapshot carries the cluster's full current app set to cloud for
+// visibility reporting, one of the app.* family described in RFD-94.
+//
+// The type and its payload live with the reporter rather than in pkg/uplink
+// because the link is deliberately payload-agnostic: it moves opaque envelopes
+// and understands no apps. A tenant owns its own wire shape.
+const TypeAppSnapshot = "app.snapshot"
+
+// AppSnapshot is the cluster's view of every app it currently runs, sent on
+// each (re)connect. Cloud treats it as ground truth for that cluster.
+//
+// This is the ephemeral tier of RFD-94's volatility spectrum: the weakest
+// durability contract in the design. Delivery is best-effort because loss
+// self-heals — a dropped snapshot is repaired by the next one rather than
+// retried, so nothing here is worth an ack.
+//
+// Two things are deliberately absent and belong to the follow-up that adds
+// sync correctness:
+//
+//   - No epoch. Cloud upserts what it receives and never sweeps, so an app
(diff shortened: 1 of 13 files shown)


pkg/anywhere/anywhere.go
```diff
@@ -1,100 +1,62 @@
...
-
-	// On each new connection, request time sync and org info
-	client.OnConnect(func(ctx context.Context) {
-		c.log.Info("sending initial requests")
-		client.SendMessage(TypeTimeRequest, TimeRequest{
-			ClientTransmitTime: time.Now().UTC(),
-		})
-		client.SendMessage(TypeOrgInfoRequest, struct{}{})
-	})
+	cfg.Uplink.Handle(TypeConnectionRequest, c.handleConnectionRequest)
```
evanphx, line 52: This is the one I'd like fixed. Anywhere used to say "connector starting" and "connector stopped" on the way in and out; now it says nothing at all until a POP request arrives, which could be hours or never. An operator reading the log can't tell "Anywhere is wired and waiting" from "Anywhere never came up."

An Info line here — something like `log.Info("Miren Anywhere ready", "cluster", cfg.ClusterXID)` — plus one in `Close` gets the affirmative record back. Our own log guidance names this case: if a healthy service says nothing at Info, you can't tell working from never attempted.

components/coordinate/appreport.go
```diff
@@ -0,0 +1,114 @@
...
+			Health: a.Health(),
+		})
+	}
+
+	if err := c.uplink.SendMessage(TypeAppSnapshot, snapshot); err != nil {
+		c.Log.Warn("failed to send app snapshot to cloud", "error", err)
+		return
+	}
+
+	c.Log.Info("reported app snapshot to cloud", "apps", len(snapshot.Apps))
```
evanphx: `SendMessage` returns nil whether or not the envelope made it into the outbox — `Send` logs a warn and drops on overflow. So this line can claim we reported when we didn't. "queued app snapshot for cloud" would be honest about what we actually know. Non-blocking; at 64 deep it'd take a real stall to hit.

pkg/uplink/client.go
```diff
@@ -31,29 +33,139 @@ type Client struct {
...
+// Callbacks accumulate rather than replace: the connector registers its
+// own for time sync and org info, and feature reporters add theirs on top.
+// They run in registration order on the connection goroutine, so a
+// callback that needs to do real work should hand off to its own.
+//
+// The context is scoped to the connection, so work handed off that way is
+// cancelled when the connection drops. That is the right lifetime for it:
+// anything still queued at that point is discarded on reconnect anyway, so
+// finishing would only produce a message nobody sends.
 func (c *Client) OnConnect(fn func(ctx context.Context)) {
```
evanphx, line 147: 👏 A single-slot callback where the second registration silently wins is the kind of thing that never surfaces as a failure — clock sync just stops, and you're chasing skewed timestamps a month later with no reason to look here. Good that it came with a test.

pkg/uplink/client.go
```diff
@@ -31,29 +33,139 @@ type Client struct {
...
+	return c
+}
+
+// TimeOffset returns the estimated clock offset between this cluster and the
+// cloud, computed via simplified NTP.
+//
+// This is link-level state rather than a tenant's: cloud reconciles the
+// timestamps a cluster reports against its own clock using this offset, so any
+// tenant reporting timestamped state depends on it.
+func (c *Client) TimeOffset() time.Duration {
```
evanphx: `TimeOffset` and `OrganizationID` have no callers anywhere in the runtime — grepped, they're only defined here. The doc says any tenant reporting timestamped state depends on the offset, but the reporter this PR adds sends raw wall clock and lets cloud do the math, which is right. Worth either softening the doc or dropping the accessors until something wants them. Non-blocking.

pkg/uplink/client.go
```diff
@@ -1,12 +1,14 @@
-package anywhere
+package uplink
```
evanphx, line 1: Drive-by from the move: `blackbox/harness/cloud.go:500` still says the marker comes from `pkg/anywhere/client.go:163`. The string it greps for is unchanged so the test is fine, just the pointer sends you to a file that isn't there anymore.

evanphx (approved):
Approved! Nice that cloud is going to know more about a cluster than one integer.

Four inline notes, all non-blocking — the Anywhere logging one is the only one I'd like to see fixed.

--e+🤖

pkg/anywhere/anywhere.go
phinze: Fixed — Anywhere logs "Miren Anywhere ready" with the cluster on registration. You're right that this is a regression I introduced: it used to bracket its own Run loop, and moving the connection out took both lines with it. Registration is the moment worth recording now that there's no loop to bracket, and the log guidance in CLAUDE.md is explicit that a healthy service saying nothing at Info leaves an operator unable to tell working from never-attempted.

components/coordinate/appreport.go
phinze: Fixed, now "queued app snapshot for cloud", with a comment on why. Slightly embarrassing given the rest of this PR argues for being honest about delivery — `SendMessage` returning nil tells us the marshal worked and nothing more.

pkg/uplink/client.go
phinze: Dropped both, along with the fields behind them. You're right that the doc described a runtime that doesn't exist: the reporter sends raw wall clock and lets cloud reconcile, which is the correct split, so an accessor for a caller that never came is just a false claim about how this works. The exchanges stay — they surface skew and tenancy in the log, and cloud needs its half of the round trip — they just don't retain anything now.

If MIR-1559's freshness work turns out to want the offset runtime-side, it comes back then with a consumer attached.

pkg/uplink/client.go
phinze: Fixed, points at `pkg/uplink/client.go` now. My mess from the move.

pkg/uplink/client.go
phinze: Worth saying that this is the failure mode I'd have shipped without the review: not a wrong line, but a reassurance from one bot that the code already did the right thing. biscuit read the same callback and reported the context was connection-scoped when it wasn't — CodeRabbit caught it. Logged that on MIR-1467 as evidence for the biscuit eval, since "prove a finding before posting it" doesn't currently cover proving a clearance.

PR #1048, opened 2026-08-11
Recover addon associations stranded by a coordinator crash

phinze:
This has gone through a whole loop-de-loo. ➿ Analyzing the bug here caused me to realize that Addons diverged in how they were treating Sagas from other usages, which made me want to write up guidelines for us to use going forward, which became RFD-99.

RFD-99 evolved all the way to sketching a full FSM library for us, which I do think will help us make our entity -> controller -> saga relationships more stable in the long run, but is also a fairly large swing that touches our low level subsystems while the sagas we do have are still in a split flag situation.

So this PR went from being "the first demo of new-world FSM library" back to a ~minimal fix for the bug, which is still a nontrivial diff!

The rest of RFD-99 becomes guidelines for greenfield and someday backlog stuff.

---

🤖 Kill a coordinator halfway through provisioning an addon and the association never comes back. `provision()` writes `provisioning` before the work starts, so a crash leaves the association on that value, and reconcile's switch had no case for it. Every later pass fell through and returned in silence. The app doesn't recover either: the deployment launcher defers pool creation while an association is still provisioning, so it activates but never gets running workloads.

The missing case is one line, but on its own it would have traded a wedged association for a duplicated one. Starting a saga has always meant writing a fresh execution over anything sharing its name, so a second pass would have discarded the record of what the first attempt built and created a second set of backing resources beside it. So the first two revs are groundwork: `Execute` now treats an existing name as "continue this one" (resume an unfinished run, finish an interrupted rollback, return straight away if it already succeeded), sharing the code path `Recover` already used, and addon provisioning names its execution after the association so a later pass can find the first attempt. Sandbox creation picks this up for free, since it already named executions after the sandbox and had been racing recovery to clobber its log.

That resumability then forces the third rev. Every provisioning saga used to end with a `Build*Result` action that stashed the answer in a struct the provider read back out of memory, which stops working the moment a resumed execution completes without running anything. The provider now reads the recorded outputs back off the execution and assembles the result itself, deleting the action, its undo, and the capture struct from all seven provider paths.

Fixes MIR-1524

### DIFF
--- a/controllers/addon/controller.go
+++ b/controllers/addon/controller.go
@@ -50,7 +50,19 @@ func (c *Controller) Init(ctx context.Context) error {
 
 func (c *Controller) Reconcile(ctx context.Context, assoc *addon_v1alpha.AddonAssociation, meta *entity.Meta) error {
 	switch assoc.Status {
-	case "pending":
+	case "pending", "provisioning":
+		// "provisioning" belongs here because provision() writes it before the
+		// work starts, so a coordinator that dies mid-provision leaves the
+		// association on it. The switch used to have no case for that: every
+		// later pass fell through and returned in silence, nothing picked the
+		// work back up, and because the deployment launcher holds an app back
+		// while an association is still provisioning, the app stopped
+		// deploying too. That is MIR-1524.
+		//
+		// Re-entering is safe because of the revs below this one: the
+		// execution is named after the association, so a second pass continues
+		// the interrupted attempt rather than starting a fresh one beside
+		// whatever the first already built.
 		return c.provision(ctx, assoc, meta)
 	case "deprovisioning":
 		return c.deprovision(ctx, assoc, meta)
@@ -68,12 +80,15 @@ func (c *Controller) provision(ctx context.Context, assoc *addon_v1alpha.AddonAs
 	// resync, the reconcile event may carry an older revision than what's
 	// now in the store). If the user has already marked this association for
 	// deprovisioning, skip; the subsequent deprovisioning event will handle it.
+	//
+	// "provisioning" passes as well as "pending", because an attempt a crash
+	// interrupted is still an attempt that should finish.
 	var current addon_v1alpha.AddonAssociation
 	if err := c.ec.GetById(ctx, assoc.ID, &current); err != nil {
 		return fmt.Errorf("re-reading association: %w", err)
 	}
-	if current.Status != "pending" {
-		c.log.Info("association no longer pending, skipping provision",
+	if current.Status != "pending" && current.Status != "provisioning" {
+		c.log.Info("association no longer wants provisioning, skipping",
 			"association", assoc.ID, "status", current.Status)
 		return nil
 	}
--- a/controllers/addon/controller.go
+++ b/controllers/addon/controller.go
@@ -15,6 +15,7 @@ import (
 	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/entity"
 	"miren.dev/runtime/pkg/idgen"
+	"miren.dev/runtime/pkg/saga"
 )
 
 // Controller reconciles AddonAssociation entities, driving provisioning
@@ -26,6 +27,11 @@ type Controller struct {
 	ec       *entityserver.Client
 	eac      *entityserver_v1alpha.EntityAccessClient
 	registry *addon.Registry
+
+	// Providers build their own executors, so this is not for running sagas.
+	// It is here so the controller can retire a teardown record it owns; see
+	// deprovision.
+	sagaStorage saga.Storage
 }
 
 // NewController creates a new addon controller.
@@ -34,12 +40,14 @@ func NewController(
 	ec *entityserver.Client,
 	eac *entityserver_v1alpha.EntityAccessClient,
 	registry *addon.Registry,
+	sagaStorage saga.Storage,
 ) *Controller {
 	return &Controller{
-		log:      log.With("module", "addon"),
-		ec:       ec,
-		eac:      eac,
-		registry: registry,
+		log:         log.With("module", "addon"),
+		ec:          ec,
+		eac:         eac,
+		registry:    registry,
+		sagaStorage: sagaStorage,
 	}
 }
 
@@ -135,6 +143,14 @@ func (c *Controller) provision(ctx context.Context, assoc *addon_v1alpha.AddonAs
 	// just created. If compensation also fails, return the error without
 	// setting terminal "error" status so the controller retries.
 	if err := c.completeProvision(ctx, assoc, meta, provider, result); err != nil {
+		// Same reasoning as deprovision(): this compensation runs the teardown
+		// saga under the association's name, so a failed one would answer every
+		// later attempt from its own record and the retry promised above would
+		// never actually run.
+		if dropErr := saga.DropIfFailed(ctx, c.sagaStorage,
+			addon.DeprovisionExecutionID(assoc.ID)); dropErr != nil {
+			return fmt.Errorf("clearing failed teardown record: %w", dropErr)
+		}
 		depErr := provider.Deprovision(ctx, addon.AssociationFrom(assoc, meta.Entity))
 		if depErr != nil {
 			c.log.Error("compensation deprovision failed, will retry",
@@ -221,6 +237,19 @@ func (c *Controller) deprovision(ctx context.Context, assoc *addon_v1alpha.Addon
 	provider, _, ok := c.registry.Get(addonName)
 	if !ok {
 		return c.setError(meta, fmt.Errorf("unknown addon %q", addonName))
+	}
+
+	// Naming the execution after the association makes a teardown resumable,
+	// but it also makes a failed one permanent: Execute would hand back the
+	// recorded error on every later pass, so `addon destroy` would be one-shot
+	// until saga retention freed the name. Retrying a compensated saga is the
+	// owner's decision, and for teardown the controller is that owner. Every
+	// undo in these sagas is a no-op, so a failed run tore nothing down and put
+	// nothing back; a fresh attempt has strictly more to do, never something to
+	// redo.
+	if err := saga.DropIfFailed(ctx, c.sagaStorage,
+		addon.DeprovisionExecutionID(assoc.ID)); err != nil {
+		return fmt.Errorf("clearing failed teardown record: %w", err)
 	}
 
 	// Step 1: Call provider.Deprovision
--- a/controllers/addon/controller.go
+++ b/controllers/addon/controller.go
@@ -107,7 +107,7 @@ func (c *Controller) provision(ctx context.Context, assoc *addon_v1alpha.AddonAs
 		ID:   assoc.App,
 		Name: appName,
 	}
-	result, err := provider.Provision(ctx, app, addon.Variant{
+	result, err := provider.Provision(ctx, addon.AssociationFrom(assoc, meta.Entity), app, addon.Variant{
 		Name:   assoc.Variant,
 		Config: variantConfig,
 	})
@@ -120,13 +120,7 @@ func (c *Controller) provision(ctx context.Context, assoc *addon_v1alpha.AddonAs
 	// just created. If compensation also fails, return the error without
 	// setting terminal "error" status so the controller retries.
 	if err := c.completeProvision(ctx, assoc, meta, provider, result); err != nil {
-		depErr := provider.Deprovision(ctx, addon.AddonAssociation{
-			ID:      assoc.ID,
-			App:     assoc.App,
-			Addon:   assoc.Addon,
-			Variant: assoc.Variant,
-			Entity:  meta.Entity,
-		})
+		depErr := provider.Deprovision(ctx, addon.AssociationFrom(assoc, meta.Entity))
 		if depErr != nil {
 			c.log.Error("compensation deprovision failed, will retry",
 				"provision_error", err, "deprovision_error", depErr)
@@ -165,13 +159,7 @@ func (c *Controller) completeProvision(
 	envVars := result.EnvVars
 	collisions := findCollisions(existingVars, envVars)
 	if len(collisions) > 0 {
-		adjusted, err := provider.AdjustEnvVars(ctx, result, addon.AddonAssociation{
-			ID:      assoc.ID,
-			App:     assoc.App,
-			Addon:   assoc.Addon,
-			Variant: assoc.Variant,
-			Entity:  meta.Entity,
-		}, collisions)
+		adjusted, err := provider.AdjustEnvVars(ctx, result, addon.AssociationFrom(assoc, meta.Entity), collisions)
 		if err != nil {
 			return fmt.Errorf("adjusting env vars: %w", err)
 		}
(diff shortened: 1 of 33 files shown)


controllers/addon/controller.go
```diff
@@ -48,39 +48,57 @@ func (c *Controller) Init(ctx context.Context) error {
...
+	switch desired {
+	case addon.StateRemoved:
+		return c.deprovision(ctx, assoc, meta)
+	case addon.StateActive:
+		if current == addon.StateFailed {
+			// A saga tried, compensated, and gave up. Retrying is a transition
+			// this machine does not declare yet, so leave it for an operator.
+			return nil
+		}
+		return c.provision(ctx, assoc, meta)
```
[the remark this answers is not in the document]
phinze: Good catch, and biscuit landed on the same thing independently. Both paths now take `&live`, so `deprovision` gets the association's real `Variables` rather than whatever the event happened to be carrying.

We did not take the `meta` half, though. `meta` is not a second read of the association, it is the write handle the reconcile framework owns: `AdaptReconcileController` clones the entity before calling us and then patches `Diff(meta.Entity, orig)` after we return (`pkg/controller/controller.go:655-668`). A `Meta` built from a fresh `GetByIdWithEntity` would be a different allocation, so nothing would diff it and every `setLifecycle` and `setError` write would vanish with no error and no log line. Added a comment on the re-read saying so, since it is a trap worth naming.

controllers/addon/controller.go
```diff
@@ -253,11 +275,16 @@ func (c *Controller) deprovision(ctx context.Context, assoc *addon_v1alpha.Addon
...
 func (c *Controller) setError(meta *entity.Meta, err error) error {
 	c.log.Error("addon error", "error", err)
 
 	updateErr := meta.Update((&addon_v1alpha.AddonAssociation{
-		Status:       "error",
+		DesiredState: addon.StateActive,
+		CurrentState: addon.StateFailed,
+		Status:       addon.LegacyStatus(addon.StateActive, addon.StateFailed),
 		ErrorMessage: err.Error(),
 	}).Encode())
```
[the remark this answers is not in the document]
phinze: Real bug, thanks. Went with your second suggestion: `setError` carries the caller's desired state through untouched, and the `current == StateFailed` terminal check moved above the `switch desired` so failure stays terminal whichever direction was wanted. A removal that fails now records `removed`/`failed` instead of quietly becoming an active association nobody asked for.

`TestDeprovisionFailureLeavesTheRemovalWanted` pins both halves. Checked it against the old code and it fails there with `expected: removed, actual: active`.

pkg/saga/executor.go
```diff
@@ -61,6 +62,13 @@ type Executor struct {
 	storage  Storage
 	registry *Registry
 	log      *slog.Logger
+
+	// inFlight names the executions this process is currently driving. A
+	// caller that names its execution after the entity it belongs to will
+	// re-enter Execute on every reconcile pass, and without this the second
+	// caller would drive the same execution concurrently with the first.
+	inFlightMu sync.Mutex
+	inFlight   map[string]struct{}
```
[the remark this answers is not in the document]
phinze: You are right about the scope, and the comment was wrong. It now says what the map actually covers, which is this `Executor` value rather than the process.

Keeping the design, though. What serializes addon and sandbox work is a level up: the reconcile controller handles one event per entity at a time and queues the rest, so the per-instance map is a backstop rather than the mechanism. A durable lease or a compare-and-swap on the record is the right answer for anything driving an execution outside such a loop, and the only answer that holds across coordinator processes, but that is its own change rather than something to fold in here. The comment says that too now, so the next person knows what they are relying on.

pkg/saga/executor.go
```diff
@@ -141,6 +169,27 @@ func (e *Executor) execute(ctx context.Context, defName string, inputs map[strin
...
 	}
 
+	// A caller that names its execution after the entity it belongs to calls
+	// this on every reconcile pass. Only one of those may drive at a time.
+	if !e.claim(id) {
+		e.log.Debug("execution already being driven here, skipping",
+			"saga", defName, "execution", id)
+		return nil
+	}
+	defer e.release(id)
```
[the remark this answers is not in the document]
phinze: Agreed, `nil` there was a lie. Added `ErrExecutionInProgress`, and `execute` returns it when the claim fails.

Chasing the callers turned up a live one worth knowing about. `SagaSandboxController` keeps a single executor for its lifetime, so unlike the addon providers its claim genuinely can fire, and it marks the sandbox DEAD on any error out of `Execute`. It now checks for the sentinel and leaves the sandbox PENDING for the reconciler instead of killing it over work that is still going. The addon controller treats it the same way rather than landing the association on a terminal status.

On the related nitpick about supplied inputs differing from `InitialInputs`: took the documentation option rather than the log. `InitialInputs` round-trips through JSON, so comparing it against freshly supplied values reports mismatches that are only type coercion (an int coming back as a float64). `WithID` now spells out that a named re-execution resumes from what the first call recorded and ignores new inputs.

phinze:
Worked through the body nitpicks as well.

Taken: the `LegacyStatus` round-trip table (with a real name for the empty-status subtest instead of `t.Run("")`), saga tests for a failed execution reporting its recorded error and for a concurrent re-entry not running actions twice, the note marking `RotateCredential` as a reader of the legacy `status` field, and the comment explaining why only `desired_state` moves in the removal path. The five `dedicated_test.go` files now assert `def.ExecutionOrder()` rather than membership, which earned its keep immediately: the real order is not the narrative order those lists carried, since `create-dedicated-pool` and `create-dedicated-service` sit at the same dependency level as the server action and sort ahead of it alphabetically. A test named `TestDedicatedSagaActionOrder` now actually checks order.

Skipped two. The shared `RunProvision` helper is real duplication, but extracting it means touching seven provider files this PR does not otherwise go near, so it belongs in its own change. And threading a single association id through both `Provision` and `RotateCredential` in `rotate_integration_test.go` is churn in a build-tagged test for no behaviour difference.

pkg/saga/executor.go
```diff
@@ -141,6 +192,27 @@ func (e *Executor) execute(ctx context.Context, defName string, inputs map[strin
...
+	// built, leaving those resources with nothing that knows to undo them.
+	switch existing, err := e.storage.Get(ctx, id); {
+	case err == nil:
+		e.log.Info("continuing existing execution",
+			"saga", defName, "execution", id, "status", existing.Status)
+		return e.resume(ctx, def, existing)
+	case !errors.Is(err, ErrExecutionNotFound):
+		return fmt.Errorf("loading execution %q: %w", id, err)
+	}
+
```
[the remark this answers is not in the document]
phinze: Fixed, thanks. `claim` now says "this Executor" to match.

pkg/saga/executor.go
```diff
@@ -527,6 +581,42 @@ func (e *Executor) Recover(ctx context.Context) error {
...
+		}
+		return e.runExecution(ctx, def, exec)
+	case StatusUndoing:
+		return e.runUndo(ctx, def, exec)
+	case StatusCompleted:
+		return nil
+	case StatusFailed:
+		return fmt.Errorf("saga failed: %s", exec.Error)
+	}
+	return nil
```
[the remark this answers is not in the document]
phinze: Good one, and the sentinel we added this round made it sharper than it used to be. A stranded claim was previously invisible, since `execute` just returned `nil` and carried on; now it would report `ErrExecutionInProgress` under that name forever. `Recover` releases through `defer` now, same as `execute` already did.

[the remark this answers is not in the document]
phinze: You are right, and my last reply answered a different question than you asked. Sorry about that. The `defer` change was real but it was not this finding.

Fixed properly now: `resume` returns an error for a status it does not recognise instead of falling out of the switch into `return nil`. Hard to argue with, given this whole change exists because a value nobody listed fell through a switch in silence. `TestExecutor_ReExecuteRejectsAnUnrecognizedStatus` covers it.

pkg/addon/lifecycle.go
```diff
@@ -0,0 +1,82 @@
...
+	StateProvisioning   = "provisioning"
+	StateDeprovisioning = "deprovisioning"
+)
+
+// LifecycleOf reads an association's machine, filling in for records written
+// before it existed.
+//
+// Associations created before this change carry only the old status field, so
+// they are read through it once and written back in the new shape by the first
+// reconcile that touches them. The mapping is the obvious one: whatever the old
```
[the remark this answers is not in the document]
phinze: Good catch on the mismatch, but the fix goes the other way: the comment was wrong, not the code.

You are exactly right about the behaviour. `status == "active"` returns on the equality guard and `status == "error"` returns on the failed guard, so neither is ever rewritten. That is deliberate. Migration here is a side effect of doing work rather than a backfill, and since `LifecycleOf` is total over the old vocabulary an unmigrated record stays correct indefinitely. Forcing a write on every settled legacy association would be a patch per association for no behavioural gain.

What misled you is a stale claim in `LifecycleOf`'s own doc comment, which said records were "written back in the new shape by the first reconcile that touches them". That was never true for settled ones. We corrected the same sentence in the rev message and the PR description earlier and missed this third copy, which is the one that matters most since it sits next to the function. It now says what actually happens, and why, including the consequence you would want to know: retiring `status` later needs a deliberate backfill rather than a waiting period.

The tests you asked for are already there. `TestReconcileLeavesSettledLegacyAssociationsInTheOldShape` covers status-only `active` and `error` and asserts nothing is staged, which is the opposite of what you proposed but is the behaviour we want pinned.

controllers/addon/controller.go
```diff
@@ -221,13 +224,7 @@ func (c *Controller) deprovision(ctx context.Context, assoc *addon_v1alpha.Addon
...
 
 	// Step 1: Call provider.Deprovision
-	err := provider.Deprovision(ctx, addon.AddonAssociation{
-		ID:      assoc.ID,
-		App:     assoc.App,
-		Addon:   assoc.Addon,
-		Variant: assoc.Variant,
-		Entity:  meta.Entity,
-	})
+	err := provider.Deprovision(ctx, addon.AssociationFrom(assoc, meta.Entity))
```
evanphx, line 256: Non-blocking, but worth a ticket at least: destroy is one-shot now.

The part that isn't visible from the diff — every undo in the deprovision sagas is `return nil`, so a failed action always walks all the way to `StatusFailed` instead of parking in Undoing. Under the stable name that's permanent: `servers/app/addons.go:197` will happily set `"deprovisioning"` again, and each retry hands back the old error having touched nothing, until saga GC frees the name a week later.

`resume()`'s comment says retrying a compensated saga is the owner's call, which I agree with — there just isn't an owner yet. Dropping the failed execution here before re-driving is probably all it needs. Would also make the "so the controller retries" note at line 136 true again.

controllers/sandbox/saga_controller.go
```diff
@@ -171,6 +172,15 @@ func (s *SagaSandboxController) createSandboxViaSaga(ctx context.Context, co *co
 		WithID(fmt.Sprintf("create-sandbox-%s", co.ID)).
 		Execute(ctx)
 
+	if errors.Is(err, saga.ErrExecutionInProgress) {
```
evanphx, line 187: Not covered by this catch: a sandbox whose containers vanish under a live runner arrives here with a `Completed` execution, resumes to nil, and never gets recreated or marked DEAD. The old overwrite re-ran creation.

Mostly theoretical today — `labs.Sagas()` is off, and boot reconciliation marks these DEAD — so really a flag for whenever we flip it on.

pkg/saga/executor.go
```diff
@@ -527,6 +586,47 @@ func (e *Executor) Recover(ctx context.Context) error {
...
+// resume continues an execution from wherever it stopped. Both Recover and a
+// re-entered Execute go through here, so the two cannot drift apart on what
+// "continue this" means for a given status.
+//
+// A caller that reaches a Failed execution gets the recorded failure back
+// rather than a silent success: the saga tried, compensated, and gave up, and
+// retrying that is a decision for whoever owns the operation, not something to
+// do implicitly under the same name.
+func (e *Executor) resume(ctx context.Context, def *Definition, exec *Execution) error {
+	if def.Version != exec.DefinitionVersion {
```
evanphx, line 608: Non-blocking: worth erroring on a definition-name mismatch alongside the version warning. `execute()` resumes with the `def` from the requested name and never checks it against `exec.DefinitionName`.

Unreachable today — an association's variant is fixed, so nothing can swap definitions under one id — but stable naming makes id collisions the thing to guard, and it's two lines.

controllers/addon/controller_test.go
```diff
@@ -252,6 +252,46 @@ func TestProvisionCompensatesOnPostProvisionFailure(t *testing.T) {
 // association has since moved to "deprovisioning" (e.g. on startup resync
 // after the user ran `addon destroy`), provision() must be a no-op so the
 // subsequent deprovisioning event can run.
+// MIR-1524 itself. provision() writes "provisioning" before the work starts, so
```
evanphx, line 298: This landed between `TestProvisionSkipsWhenAssociationNoLongerPending` and its doc comment, so that paragraph now reads as the first half of this test's.

pkg/saga/executor.go
```diff
@@ -527,6 +586,47 @@ func (e *Executor) Recover(ctx context.Context) error {
...
+		return e.runExecution(ctx, def, exec)
+	case StatusUndoing:
+		return e.runUndo(ctx, def, exec)
+	case StatusCompleted:
+		return nil
+	case StatusFailed:
+		return fmt.Errorf("saga failed: %s", exec.Error)
+	}
+
+	// This whole change exists because a value nobody listed fell through a
```
evanphx, line 633: 👏 Not letting an unknown status pass as success. Easy to leave implicit and never find out.

pkg/saga/executor.go
```diff
@@ -490,34 +562,21 @@ func (e *Executor) Recover(ctx context.Context) error {
...
-			// Resume undo
-			recoverErrors = append(recoverErrors, e.runUndo(ctx, def, exec))
-		case StatusCompleted, StatusFailed:
-			// Terminal states; nothing to recover.
+		if !e.claim(exec.ID) {
+			e.log.Debug("execution already being driven here, leaving it alone",
+				"saga", exec.DefinitionName, "execution", exec.ID)
+			continue
+		}
+		// Released through defer so an action that panics cannot strand the
```
evanphx, line 570: ✨ Good catch on the panic path — a stranded claim would read as in-flight work forever.

evanphx (approved):
Approved. Nice that a half-finished addon just picks itself back up now.

Four non-blocking inline notes; the `deprovision()` one is the only one with teeth.

--e+🤖

controllers/addon/controller.go
phinze: 🤖 Fixed rather than ticketed, since sagas flip on by default soon.

`deprovision()` retires a failed teardown record before re-driving, via a new `saga.DropIfFailed`. The controller is the owner you said did not exist yet, and the reason a retry is safe here is at the call site rather than left to be re-derived: every undo in those sagas is a no-op, so a failed run tore nothing down and a second attempt has strictly more to do, never something to redo.

You were right about line 136 too, and it needed its own fix. That compensation calls `provider.Deprovision` directly rather than going through `deprovision()`, so it was replaying the same stale record and the retry it promises never actually ran. Both paths drop now.

`TestDeprovisionRetriesAfterAFailedTeardownSaga` pins it. Worth saying the load-bearing assertion is that the record is gone, not that the provider was called: the mock never runs a saga, so it would be called either way.

controllers/sandbox/saga_controller.go
phinze: 🤖 Taken now rather than ticketed, given sagas go default soon.

`createSandboxViaSaga` drops a completed record before re-driving. It is only reached once `CheckSandbox` has found the containers missing, so a record of a previous success describes resources that are no longer there, which is exactly the case you named.

pkg/saga/executor.go
phinze: 🤖 Done, and agreed on the reasoning: stable naming is what turns id collisions from theoretical into the thing worth guarding. It errors rather than warns, since resuming across a collision would run this definition's actions against another's recorded outputs, which is worse than a version skew.

controllers/addon/controller_test.go
phinze: 🤖 Good catch, entirely my doing. The edit that added the test matched on the `func` line, so it slid in underneath that paragraph. Doc comment is back on its own function.

PR #1049, opened 2026-08-11
Back up disks to miren.cloud from both universal and accelerator mode

evanphx:
The cloud side of this (mirendev/cloud#195) gave volumes a format-agnostic update stream, where each backup artifact carries a kind and an ordering key. This is the runtime half.

Two things were wrong here. Accelerator mode was uploading log segments without their TAI64N label, so the cloud had nothing to order them by and rejected them outright, which meant continuous backup didn't actually work. And universal mode had no cloud path at all: a loop device on a backing file and nothing else, so losing the host lost the data.

The label fix is small. It lives in the segment's filename, which the uploader already has in hand, and a segment without a usable one is now refused locally rather than discovered as a 400. Replay got the related fix: it used to list a volume's entire backup history and throw away everything at or below its horizon, and now passes that horizon to the server so the response scales with what's actually missing.

Everything then moved onto one client. Log segments used to speak a compatibility alias with its own request shapes; they now go through the same generic endpoints that image snapshots use, which is what makes a third kind cheap rather than another bespoke path. The migration is behaviour-preserving, so the replay tests carried over untouched, which was the point of doing it that way.

Universal mode gets snapshots via the existing `pkg/snapshot` format, uploaded as `loop_image` updates, plus a restore that rebuilds a missing backing file from the newest snapshot before attaching the loop device. An image that's already present is never touched, since the local copy is by definition newer than anything the cloud has.

The last commit walks something back. Snapshots started out on an hourly timer, and working through what that actually produces changed my mind: reading a mounted image while the loop device keeps writing to it gives you a smear across the read window, not a point-in-time copy. That's weaker than the power-loss state fsck and WAL recovery are built for, and taking it automatically while filing each one as a restore point is worse than not backing up, because it looks like protection. Doing it properly needs FIFREEZE with a copy-on-write clone, or dm-era change tracking to converge a live copy and freeze only for the final delta. Both are worth building and neither belongs in a guess, particularly since ext4 has no reflinks and that's what the host filesystem usually is. So snapshots are on demand only now, via `miren disk backup --cloud`, where a person decides it's safe. The warning on that path used to claim the backup would be crash-consistent, which was never true of anything this code does, and now says what actually happens.

The `hack/e2e_disk` driver runs against a real cloud rather than mocks: it authenticates as a service account, uploads both kinds, lists and downloads them, then rebuilds a deleted image and checks it byte for byte. Worth being precise about the scope, though. The driver creates its own cloud volume and calls the restore path directly, so it proves the wire format and the snapshot round trip, not the production path from "user creates a disk" to "a volume exists in the cloud". That path is covered by unit tests over the real reconcile loop.

Continuous backup remains accelerator mode's job, which is the shape RFD-0064 argued for. lbd captures writes in order inside the kernel, so its segments can't smear.

Depends on mirendev/cloud#195.

Closes MIR-1557


components/diskio/image_snapshot.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return "", fmt.Errorf("stat image: %w", err)
+	}
+
+	staged, checksum, err := s.stage(req, info)
+	if err != nil {
+		return "", err
+	}
+	defer os.Remove(staged.Name())
+	defer staged.Close()
```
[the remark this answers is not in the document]
evanphx: 🤖 I don't think the double-call happens. The two defers are registered *after* the error check on `stage()`:

```go
staged, checksum, err := s.stage(req, info)
if err != nil {
    return "", err        // no defers registered yet
}
defer os.Remove(staged.Name())
defer staged.Close()
```

So on a `stage()` failure the function returns before either defer exists, and `stage`'s own explicit Close+Remove is the only cleanup that runs. On success `stage` doesn't clean up at all, and the defers are the sole owners. The two paths are already disjoint.

Agreed the ownership split is easy to misread, which is presumably what flagged it. Leaving the code as is rather than restructuring something that works, but happy to move the cleanup entirely into `Snapshot` if you think the current shape is a trap for the next editor.

cli/commands/disk_backup.go
```diff
@@ -94,5 +109,74 @@ func DiskBackup(ctx *Context, opts struct {
...
+	}
+
+	ctx.Info("Uploading to %s...", cloudURL)
+
+	updates := diskio.NewCloudUpdatesClient(ctx.Log, cloudURL, authClient)
+	return updates.Upload(context.Background(), target.VolumeID, diskio.UploadRequest{
+		Kind:         diskio.KindLoopImage,
+		OrderingKey:  fmt.Sprintf("%016x", time.Now().UnixNano()),
+		SnapshotName: pin,
+		Metadata: map[string]any{
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch, aligned rather than documented. Both paths now write `image_size`, `image_sha256` and `compressed_size`, so a restore reading any of them gets a real value regardless of which produced the update. The command already had all three to hand, so there was no reason for the gap.

`was_attached` stays on this path only, since it is genuinely specific to it: the CLI is the one place a snapshot can be taken from a live disk, and it records that the image is a smear across the read rather than a point-in-time copy. Added a comment saying so.

components/diskio/segment_replay.go
```diff
@@ -17,33 +17,25 @@ import (
...
 		return fmt.Errorf("reading log horizon: %w", err)
 	}
 
-	// Filter to segments with labels newer than the horizon
-	var missing []LogSegmentInfo
-	for _, seg := range remoteSegments {
-		if seg.Label > horizon {
-			missing = append(missing, seg)
-		}
+	missing, err := c.cloudClient.ListLogSegments(ctx, volState.VolumeId, horizon)
```
phinze: Looks like a legit upload gap possibility to me?

---

🤖 The write side is `log_watcher.go:96-115` (not in this diff, so I can't anchor there). On upload failure it warns and `continue`s, which correctly leaves the segment on disk for a later retry. But it continues to the *next* segment, `os.ReadDir` returns entries sorted by filename, and TAI64N labels sort chronologically. So the newer segment uploads, and `updateLogHorizonFromPath` advances:

```go
if label > current {
    return writeLogHorizon(diskPath, label)
}
```

It's a high-water mark, not a cursor, so it steps over the gap. This call then passes that horizon as `after`, and the skipped segment is never requested. Restore applies later writes without earlier ones, with nothing reporting it.

The retry doesn't recover either, because of the cloud half. `BeginUpdateUpload` inserts the row *before* any bytes move (cloud `updates.go:227`), and `idx_volume_updates_volume_kind_key` is unique `WHERE state <> 'deleted'` (`db/schema.hcl:2175`), so a `pending` row holds the key. The ordering key is a pure function of the filename, so every retry sends the same one and gets a 409 from `conflictFromCreateError`. The presigned URL expires after 15 minutes and I couldn't find a reaper for stale pending rows, so it stays claimed.

Both halves are individually defensible and the combination loses data quietly. The horizon side seems like the cheaper fix: only advance it when the segment it names actually made it up, or track a low-water mark alongside it so a gap is detectable.

components/diskio/image_restore.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// This is the universal-mode counterpart to replayMissingSegments. It recovers a
+// lost disk; it never touches an image that is already here, because the local
+// copy is by definition newer than anything the cloud holds.
+func (c *DiskMountController) RestoreImageIfMissing(ctx context.Context, volState *VolumeState, imagePath string) error {
+	if c.updatesClient == nil {
+		return nil
+	}
+
+	if _, err := os.Stat(imagePath); err == nil {
```
phinze, line 37: Looks like 🤖 caught something real here?

---

🤖 Two things get in front of it.

The volume controller reconciles the `DiskVolume` entity. On a host that lost `/var/lib/miren`, the entity is still `DV_READY` while `VolumePathExists` is false, which routes to `createVolume` (`disk_volume_controller.go:153`), and that lays down a fresh sparse `disk.img` at lines 248-258. By the time the mount controller reaches this stat the image exists, so it returns early.

Separately, `alwaysMount` is exactly `VM_UNIVERSAL` (`disk_volume_controller.go:21`), so universal volumes are attached and formatted by `ensureVolumeMount`, which has no restore hook at all. The hook added here is on `DiskMountController.attachAndMount`, a different path.

I traced both mechanisms but not the reconcile ordering between the two controllers, so the first one is "very likely" rather than proven. The second stands on its own.

components/diskio/image_restore.go
```diff
@@ -0,0 +1,89 @@
...
+		// The image is present; leave it alone
+		return nil
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("stat image: %w", err)
+	}
+
+	updates, err := c.updatesClient.List(ctx, volState.VolumeId, ListOptions{Kind: KindLoopImage})
```
phinze: Looks like 🤖 caught a big scan here.

---

🤖 `List` with kind only, then `updates[len(updates)-1]` for the newest. Cloud side pages at 100 (`defaultUpdateListLimit`), so this is real round trips that grow with the volume's history in order to select one row. Commit 43bdd917 made exactly this argument for log segments.

Cloud `ListVolumeUpdates` is `ORDER BY kind, ordering_key`, so `until` won't help, but a `limit=1` with descending order would, if the cloud grows one.

phinze (approved):
I like the way this is coming together! But I don't think it's quite all-the-way-together yet. 

On code: I didn't go so far as to load the full disk stack back into my brain, got 🤖 to give me a brief tour and scanned through its findings.

On behavior: I went to take this for a spin locally and couldn't get to the point of exercising it. As far as I can tell there's no path from creating a disk to having a volume in the cloud, so --cloud has nothing to address. Is provisioning a follow-up I've missed? (detailed 🤖 version of the gap below)

Regardless, approving as a good next step!

---

🤖 Tracing the id plumbing end to end, I don't find a path from "user asks for a disk" to "a volume exists in miren.cloud":

- `controllers/disk/disk_controller.go:240` builds the `DiskVolume` entity with name, disk id, size, filesystem, mode, states and node id. No `MountId`.
- `disk_volume_controller.go:271` derives the volume's public id as `volume.MountId`, falling back to the entity id suffix (`disk-vol-*`, from `idgen.GenNS`).
- `rg 'MountId\s*[:=]'` across the repo, excluding generated schema code, returns exactly one hit: `ownerMountId`, a variable name in `disk_lease_controller_test.go:350`. So the fallback is the only branch, and `DiskVolume.VolumeId` is always the local id, which is also the on-disk directory name (`resolveImagePath` builds the path from it).
- The only code in the runtime that POSTs `/api/v1/disk/volumes` is `hack/e2e_disk/main.go:146`. Nothing in `pkg/registration` or `components/coordinate` references volumes.

So `AcquireLease`, `ListLogSegments`, `RestoreImageIfMissing` and `uploadSnapshotToCloud` all hand the cloud a `disk-vol-*` id, and `GetVolumeByXID` (cloud `updates.go:79`) turns that into `ErrVolumeNotFound` and a 404.

The e2e harness passes because it creates its volume through the cloud API and feeds the returned xid straight into the client, so it covers the wire protocol thoroughly and never exercises the runtime's id plumbing. That seems like a fair split for the first half of a two-repo change; it just isn't visible from the PR body, which reads as end-to-end verified.

`MountId` looks like the intended hook. Is provisioning a known follow-up?

components/diskio/segment_replay.go
evanphx: 🤖 Confirmed, and fixed in `2b4f3b8c`. The watcher now breaks out of a volume's scan on the first upload failure instead of moving to the next segment, so nothing later can advance the horizon past the gap. The horizon write gets the same treatment, for the same reason.

The test that existed used a single segment, which can only show the file survives. `1971ccec` adds a two-segment one, and I checked it fails on the old code: the second segment uploads, the horizon lands on it, and the first is stranded where replay will never ask for it again.

components/diskio/image_restore.go
evanphx: 🤖 It did. Fixed in `47a27e21`. `restoreImageIfMissing` is now a plain function called from `DiskVolumeController.ensureVolumeMount`, which is the path universal volumes actually take, since `alwaysMount` is exactly `VM_UNIVERSAL`. They never reached the mount controller's copy at all.

It runs before the format step, which is what made this quiet rather than loud. The old behaviour wasn't a failed restore, it was a freshly imaged empty disk.

`TestVolumeControllerRestoresImageBeforeMounting` drives `ensureVolumeMount` directly and checks the image comes back byte for byte.

components/diskio/image_restore.go
evanphx: 🤖 Agreed, and now fixed rather than deferred. It needed a change on both sides, so `mirendev/cloud#195` gains `order=desc` on the update listing (`ca41848`) and this branch uses it (`0911a7ff`). Restore asks for one row from the newest end instead of paging the whole history to keep the last one.

On the cloud side it's a second SQL statement rather than a conditional `ORDER BY`, so both directions still walk the `(volume_id, kind, ordering_key)` index instead of sorting in memory. On this side `ListOptions` grew `Descending` and `Limit`, and a bounded list stops instead of following the cursor.

Two things worth mentioning. Writing the shared mock turned up a real bug: equality is neither above nor below the cursor, and treating "not above" as "below" made a descending page hand back its own cursor row forever. And the restore test double had to start honouring the new options, because a double that ignores them keeps passing while the real path asks for the oldest image, which is exactly the failure worth catching.

There's also a contract test on the cloud side pinning the literal query this client sends. If `order` were dropped or renamed the request would still return 200 with the oldest image, and a restore would quietly rebuild a stale disk.

evanphx:
Provisioning wasn't a follow-up I'd left deliberately. It was a real hole and your trace of it was exactly right: nothing ever assigned `MountId`, so `VolumeId` was always the local `disk-vol-*` id and every cloud call addressed a volume that didn't exist. Only `hack/e2e_disk` worked, because it created the cloud volume itself. Fixed in `47a27e21`.

The two ids now sit side by side rather than one standing in for the other. `VolumeId` stays local, because it names the mount point and repointing it would relocate a live mount out from under whatever is using it. `CloudVolumeId` is new, persisted to the entity as `MountId`, and it's what every cloud call uses.

Registration is lazy: any volume that reconciles without a cloud id, on a cluster with a cloud configured, gets registered. That picks up disks predating enrolment and retries for free after an outage. The name is scoped to the cluster and the uuid is a hash of (cluster, volume), so two clusters can't collide on a shared disk name and a retry is idempotent instead of a constraint violation. Failures are logged rather than returned, since an unreachable cloud shouldn't stop a local disk working.

One consequence worth knowing: the log watcher now leaves segments alone for a volume with no cloud id, rather than uploading into the void or deleting something that was never backed up.

Your other three findings are fixed too, each in its own commit. You should be able to take it for a spin now, though the cloud half has to be deployed first for registration to have anywhere to go.

components/diskio/disk_volume_controller.go
```diff
@@ -152,21 +250,17 @@ func (c *DiskVolumeController) reconcileVolumePresent(ctx context.Context, volum
...
-			Name:       volume.Name,
-			DiskPath:   volumePath,
-			SizeBytes:  units.GigaBytes(volume.SizeGb).Bytes().Int64(),
-			Filesystem: volume.Filesystem,
-			Mode:       volume.VolumeMode,
+			EntityId: entityId,
+			VolumeId: localVolumeIDFromEntity(entityId),
+			// MountId carries this volume's id in miren.cloud, empty until it
+			// has been registered there.
+			CloudVolumeId: volume.MountId,
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch, fixed in `8fcc110b`. There's now a dedicated `cloud_volume_id` attribute on `disk_volume`, and `mount_id` goes back to meaning what the schema says.

You're right in both directions, and the second one is the part I'd missed. Not only was a cloud id sitting in a field documented as a mount-point override, but the override had stopped working at all: `localVolumeIDFromEntity` derived the local id from the entity suffix unconditionally, so a volume that set `mount_id` would have had its mount path move. `localVolumeID` now prefers `mount_id` when set, exactly as before, and `TestMountIdStillOverridesTheMountPoint` holds that line.

On the migration half specifically: no data needed moving. Nothing in the repo has ever assigned `mount_id` (phinze traced the same thing independently on this PR), so there are no existing entities carrying a custom value. The problem was the semantic collision rather than live data, and a dedicated attribute removes it either way.

components/runner/runner.go
```diff
@@ -886,10 +887,27 @@ func (r *Runner) SetupControllers(
...
+					r.dvc.SetUpdatesClient(updatesClient)
 
-					logUploader = diskio.NewCloudSegmentUploader(log, cloudURL, authClient, diskioState)
+					// Give local disks an identity in the cloud. Without this
+					// every backup call would carry an id the cloud has never
+					// heard of.
+					r.dvc.SetCloudVolumeRegistrar(
+						diskio.NewCloudVolumeRegistrar(log, cloudURL, authClient),
+						r.CloudAuth.ClusterID,
+					)
```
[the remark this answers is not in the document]
evanphx: 🤖 This one was a real bug and the most consequential thing found on this PR. Fixed in `8fcc110b`.

You're right that `ReconcileWithEntities` ran before any of the cloud wiring, and the consequence is worse than a delayed registration. Startup is precisely when a host that lost its disks has an image to restore. That first pass would find the image missing, see a nil updates client, skip the restore, and then continue into `ensureVolumeMount`, which formats an unformatted image. So the failure mode wasn't "recovery waits for the next reconcile", it was a fresh empty disk written over the gap, with the restore never attempted again because the image now exists.

The mount controller is now constructed and both controllers wired before either reconciles, so the first pass has everything it needs.

phinze (approved):
Changes look solid!

Just thinking through rollout here:

- When this lands, all accelerator managed disks begin continuous backups, but we believe there are ~0 of those in the wild.
- For most people it just makes `miren disk backup --cloud` work, though you can't see the results because the UI is still flagged off in Cloud... prolly wanna turn that on for everybody sometime soon.

evanphx:
> Changes look solid!
> 
> Just thinking through rollout here:
> 
> * When this lands, all accelerator managed disks begin continuous backups, but we believe there are ~0 of those in the wild.
> * For most people it just makes `miren disk backup --cloud` work, though you can't see the results because the UI is still flagged off in Cloud... prolly wanna turn that on for everybody sometime soon.

Agreed! There is still a bit more work on the universal front. I initially had some changes for doing periodic but couldn't yet come up with a way to get a consistent view of the file. More research is needed there.

PR #1050, opened 2026-08-12
Stop `miren logs` printing a stream-reset error after a successful run

phinze:
A bounded `miren logs` run would sometimes return every record, exit 0, and still print this to stderr:

```
[ERROR] rpc.callstream call: error decoding stream request │ error: "stream reset, but failed to convert stream error 386759528: error code outside of expected range"
```

That number is the whole story. It's webtransport-go's internal `sessionCloseErrorCode`, the code it uses to reset every open stream when a session closes. The client serves the log sink over an inline callback stream, and when the call finishes it closes its session out from under a decoder still parked on that stream. The reset arrives as an opaque transport error rather than a clean EOF, so the loop called it a failure. Whether the server's own clean close beat our teardown was a race, which is why it only showed up some of the time, and why it was never really about bounded queries.

`serveInlineCalls` now treats a read failure as expected when the stream's context is already canceled, and still reports at Error otherwise, so a callback stream that genuinely breaks mid-call isn't muted. In dev that took a looped query from 11 stderr hits in 40 runs to none in 60, with full records throughout.

The garbled wording is a webtransport-go bug, already fixed upstream in v0.12.0, where `Read` returns a typed `SessionError` instead. We're on v0.9.0, and upgrading is its own piece of work. It would improve the text but wouldn't remove this line, since the loop was treating expected teardown as a failure regardless of how it was worded.

Closes MIR-1570

### DIFF
--- a/pkg/rpc/inline_router.go
+++ b/pkg/rpc/inline_router.go
@@ -132,7 +132,14 @@ func (s *State) serveInlineCalls(ctx context.Context, dec *cbor.Decoder, enc *cb
 		var rs streamRequest
 
 		if err := dec.Decode(&rs); err != nil {
-			if !errors.Is(err, io.EOF) {
+			// A canceled context means the call these callbacks served has
+			// already finished and the caller is tearing down its session,
+			// which resets any stream still open here. That reset surfaces as
+			// a transport error rather than a clean EOF, and on WebTransport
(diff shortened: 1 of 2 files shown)


phinze:
@coderabbitai review

evanphx (approved):

PR #1054, opened 2026-08-12
Add a recipe for self-hosting headscale

evanphx:
Closes MIR-1575.

Nothing in the docs covered running [headscale](https://headscale.net), the open-source Tailscale control server. The closest page, `tailscale.md`, is the inverse topic — running a Miren *cluster* on a tailnet — and mentions headscale only in a callout, so a reader wanting to host it had nowhere to land. This adds a fourth recipe and cross-links the two.

## No shim required

The interesting part is that this needs no Dockerfile shim. headscale ships a distroless image with no shell, which used to be disqualifying — but since #951, a service that sets no `command` runs the image's `ENTRYPOINT` + `CMD` as argv in exec form. So the whole build is three lines on the upstream image:

```dockerfile
FROM docker.io/headscale/headscale:0.29.3
COPY config.yaml /etc/headscale/config.yaml
CMD ["serve"]
```

That makes the page a demonstration of bring-your-own-image (RFD-91) rather than a workaround for its absence. I originally wrote it the old way — a multi-stage lift of the binary onto `debian:bookworm-slim` — before checking whether the `/bin/sh -c` assumption still held. It doesn't, and the recipe is much better for it. The wrapper survives as an optional "Getting a shell" section, since that's the real tradeoff.

I verified all four combinations on a cluster rather than trusting the code read:

| | distroless, no `command` | with a `command` |
|---|---|---|
| Deploy and run | works — argv in exec form, PID 1, graceful SIGTERM | fails, no `/bin/sh` |
| `sandbox exec <id> -- headscale …` | works (argv, no shell) | works |
| `sandbox exec <id>` interactive | `stat /bin/sh: no such file or directory` | works |
| `miren app run` | `sandbox failed to start, status: status.dead` | reaches the socket error |

The third row is the non-obvious one, and `servers/exec/exec.go:296-311` explains it: with args present and no configured entrypoint, argv is used directly; with no args it falls back to `/bin/sh`. So named admin commands work fine on a shell-less image and only the interactive prompt doesn't.

## The ingress timeout

The one server-side change the page asks for. Tailscale clients hold a long-poll open on `/machine/map`; Miren's ingress applies an idle read deadline from `http_request_timeout` (default 60s), and headscale's keepalive fires every 50s plus up to 9s of jitter. The deadline resets on each read, so it survives — with as little as one second of margin, varying per session. That's the shape that produces intermittent, hard-to-attribute drops, so the recipe raises it to 120 before a client is ever connected.

## Verification

Deployed end to end on a dedicated `headscale` cluster: `--analyze`, build, boot, route, HTTPS health, `users create`, `preauthkeys create`, `nodes list`. State persistence was checked properly — the first redeploy reused the sandbox, so I forced a real rebuild, confirmed a new sandbox ID, and checked that `noise_private.key` kept its original timestamp. A regenerated noise key would break every enrolled client, so that's the part that matters. The user also survived a switch from the wrapper image to the distroless one.

Every error string quoted on the page was reproduced, not inferred.

The client join and the embedded DERP relay are drawn from headscale's own docs, and the page marks them as untested in a callout. Embedded DERP starts and advertises correctly (`stun server started at [::]:3478`, region node on `DERPPort: 443`), but I never joined a real node, so the page doesn't claim otherwise.

`bun run lint` and `bun run build` pass.

### DIFF
--- a/docs/docs/recipes/headscale.md
+++ b/docs/docs/recipes/headscale.md
@@ -352,29 +352,6 @@ relayed session that goes quiet long enough hits the timeout from
 rather than removing them.
 :::
 
-## Roadblock checklist
-
-1. Don't set a `command` — the image has no shell, and the sandbox dies at startup if you do.
-2. Give the image a `CMD ["serve"]`. Upstream sets an `ENTRYPOINT` but no `CMD`, so without
-   it headscale starts with no subcommand.
-3. Name the service **`web`**, or routing returns `error acquiring lease: app/headscale`.
-4. Set `listen_addr` to `0.0.0.0:8080`, never `127.0.0.1`.
-5. `HEADSCALE_SERVER_URL` must match the routed hostname exactly — clients are handed that
-   URL and keep using it.
-6. Include a `dns` block, or startup fails on
-   `dns.nameservers.global must be set when dns.override_local_dns is true`.
-7. Keep the `server_url` host out from under `dns.base_domain`, or you get
-   `server_url cannot be part of base_domain in a way that could make the DERP and headscale server unreachable`.
-   `headscale.example.com` with a base domain of `ts.example.com` is fine;
-   `example.com` as the base domain is not.
-8. Raise `http_request_timeout` before connecting a client — see
-   [Give clients a longer timeout](#longer-timeout).
-9. Run admin commands with `miren sandbox exec <id> -- <cmd>`, against the sandbox actually
-   serving headscale. `preauthkeys create` wants the user's numeric ID, not the name.
-10. `WRN listening without TLS but ServerURL does not start with http://` is expected.
-11. For the embedded DERP, open 3478/udp in the cloud firewall as well as declaring the
-    node port.
-
 ## Next steps
 
 - [App Configuration](/app-configuration) — the full `app.toml` reference in context
--- a/docs/docs/recipes/headscale.md
+++ b/docs/docs/recipes/headscale.md
@@ -1,6 +1,6 @@
 ---
 title: Run a headscale control server
-description: A worked example — self-host headscale, the open-source Tailscale control server, as a Miren app with its state on a persistent disk and its clients logging in over your own hostname.
+description: Self-host headscale, the open-source Tailscale control server, on Miren — reachable at your own hostname, with its database and keys on a persistent disk.
 keywords: [headscale, tailscale, tailnet, control server, vpn, wireguard, derp, self-hosted, example, deploy]
 ---
 
@@ -13,14 +13,8 @@ control server — the coordination plane your Tailscale clients log in to, exch
 through, and get their network map from. Self-hosting it means your tailnet's coordination
 belongs to you.
 
-The end state: headscale running as a single Miren `web` service on port 8080, reachable
-over HTTPS at a hostname you own, with its SQLite database and private keys on a persistent
-disk. You point `tailscale up --login-server` at it and register nodes with
-`miren sandbox exec`.
-
-Along the way it exercises a fair amount of Miren — running an upstream image on its own
-entrypoint, a local disk, environment-driven config, and (for the embedded relay) a UDP node
-port.
+By the end you'll have headscale answering over HTTPS at a hostname you own, its database
+and keys on a disk that survives redeploys, and a first node joined to your tailnet.
 
 :::info[This is an application recipe, not a language guide]
 For getting your own source code onto Miren, start with [Deployment](/deployment) and the
@@ -29,20 +23,6 @@ For the opposite topic — running a Miren **cluster** on a tailnet — see
 [Running Miren on a Tailnet](/tailscale).
 :::
 
-## How it works
-
-1. A three-line Dockerfile adds a `config.yaml` and a `CMD` to the official headscale image.
-   The service sets no `command`, so the image's own `ENTRYPOINT` + `CMD` run as the process
-   argv — which is what lets a shell-less image work at all.
-2. The `web` service listens on `0.0.0.0:8080`. Miren's ingress terminates TLS at your
-   hostname and proxies to it.
-3. A `config.yaml` baked into the image holds everything static; the values that change per
-   deployment come from `HEADSCALE_*` environment variables in `app.toml`.
-4. The database and the noise private key live on a local disk at `/data`, so they survive
-   redeploys.
-5. Admin commands (`headscale users create` and friends) run *inside the live sandbox*,
-   because the headscale CLI talks to the running server over a unix socket.
-
 ## Prerequisites
 
 - `miren` CLI installed and authenticated (`miren whoami`).
@@ -50,6 +30,17 @@ For the opposite topic — running a Miren **cluster** on a tailnet — see
 - A hostname you control, pointed at the cluster — see
   [Custom Domains](/traffic-routing#custom-domains) or claim one through
   [Miren Cloud subdomains](/miren-cloud/subdomains).
+- **The ability to edit the cluster's server config.** Tailscale clients hold a connection
+  to the control server open far longer than the ingress allows by default, so this recipe
+  needs one cluster-wide setting changed before a client will stay connected. Details in
+  [Give clients a longer timeout](#longer-timeout).
+
+:::warning[That timeout is cluster-wide]
+`http_request_timeout` applies to every app on the cluster, not just headscale. If you can't
+change it — someone else runs the cluster, or other apps depend on the current value — nodes
+will keep dropping their connection to headscale, and there's no way to work around it from
+headscale's side.
+:::
 
 ## Select the target cluster
 
@@ -66,28 +57,25 @@ miren whoami
 
 ## The Dockerfile
 
-The official headscale image is **distroless** — it contains no shell, no package manager,
-nothing but the binary at `/ko-app/headscale` and a CA bundle. That's fine, as long as you
-don't ask Miren to run a `command`:
+You only need to add your config file and tell the image what to run:
 
 ```dockerfile
 FROM docker.io/headscale/headscale:0.29.3
 
 COPY config.yaml /etc/headscale/config.yaml
 
-# The image sets ENTRYPOINT ["/ko-app/headscale"] but no CMD, so give it one.
+# The image sets an ENTRYPOINT but no CMD, so give it one.
 CMD ["serve"]
 ```
 
-That's the whole thing. The upstream image already ships the CA bundle headscale needs to
+That's the whole build. The upstream image already carries the CA bundle headscale needs to
 fetch the DERP map, so there's nothing to install.
 
-:::info[Why there's no `command` in the app.toml]
-When a service sets no `command`, Miren runs the image's `ENTRYPOINT` + `CMD` directly as
-the process argv, the way `docker run IMAGE` does — no shell involved, the entrypoint gets
-PID 1, and signals reach it. Set a `command` and Miren runs it through `/bin/sh -c` instead,
-which a distroless image has no shell for. That's the whole reason this recipe leaves
-`command` out.
+:::warning[Don't set a `command` for this app]
+The headscale image ships without a shell, and a service that sets `command` needs one.
+Leave it out — as the `app.toml` below does — and the image runs what its own `CMD` says.
+This also means you can't open an interactive shell in the container; see
+[Getting a shell](#getting-a-shell) if you want that.
 :::
 
 Add a `.dockerignore` so local files stay out of the build context:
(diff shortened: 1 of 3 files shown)


docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+control server — the coordination plane your Tailscale clients log in to, exchange keys
+through, and get their network map from. Self-hosting it means your tailnet's coordination
+belongs to you.
+
+The end state: headscale running as a single Miren `web` service on port 8080, reachable
+over HTTPS at a hostname you own, with its SQLite database and private keys on a persistent
+disk. You point `tailscale up --login-server` at it and register nodes with
+`miren sandbox exec`.
+
+Along the way it exercises a fair amount of Miren — running an upstream image on its own
```
phinze: I think this line is the seed of how things get impl-centric. A small sweep on the siblings is prolly in order if we agree on the pivot.

---

🤖 The same sentence appears at `hermes-agent.md:20` and `amp-runner.md:21`, so removing it here leaves this page out of step with two of its three neighbors until the sweep lands. `openhands-agent-server.md` already omits it, so the sweep is two one-line deletions.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+COPY config.yaml /etc/headscale/config.yaml
+
+# The image sets ENTRYPOINT ["/ko-app/headscale"] but no CMD, so give it one.
+CMD ["serve"]
+```
+
+That's the whole thing. The upstream image already ships the CA bundle headscale needs to
+fetch the DERP map, so there's nothing to install.
+
+:::info[Why there's no `command` in the app.toml]
```
phinze: Example of iml-centric framing. It's maybe worth the user knowing the rule about no-command-on-distroless images? idk I'd consider dropping it or chopping it down to just the user-facing bits.

---

🤖 The rule itself checks out. `controllers/sandbox/sandbox.go:2608` is `oci.WithProcessArgs("/bin/sh", "-c", co.Command)`, so a set `command` really does go through a shell, and `servers/build/build.go:588-592` documents the other half — an image/Dockerfile build leaves the entrypoint empty "so the image's own ENTRYPOINT+CMD run in exec form at launch (MIR-1444)". No objection to the content, just its weight and its location.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+and will use it for every subsequent request.
+
+<CliCommand context="client">
+```bash
+miren route set headscale.example.com headscale
+miren route list
+```
+</CliCommand>
+
+## Raise the ingress timeout
```
phinze: Since this is server config it should get raised in prominance and highlighted. Seems like it would belong prerequisites up near top. Also per impl de-centering can drop the keepalive arithmatic and simplify the language to like "headscale needs reqs longer than default timeout" 

---

🤖 Two things worth knowing here.

**It's cluster-global.** `http_request_timeout` is read once at `cli/commands/server.go:805` and handed to the single ingress (`components/coordinate/coordinate.go:1498`), so raising it to 120 changes the idle deadline for every app on the cluster, not just headscale. Prerequisites currently asks only for "access to the target cluster and its org", which doesn't imply rights to edit `/etc/miren/server.toml`.

**There's no headscale-side alternative.** `keepAliveInterval` is a package const in headscale v0.29.3 (`hscontrol/poll.go:24`), and the jitter is drawn once per session at line 56 — `rand.IntN(9000)` ms, so 50.000–58.999s. Your numbers are exact, including the per-session part. But nothing in `hscontrol/types/config.go` or `config-example.yaml` exposes it, so there's no `HEADSCALE_*` override to reach for either. A reader who can't change server config can't run this recipe reliably at all, which seems worth saying out loud.

Minor: the arithmetic is stated twice, here and again in roadblock item 8 at line 417.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+```bash
+SANDBOX=$(miren sandbox list --format json | jq -r '.[0].id')
+
+miren sandbox exec "$SANDBOX" -- headscale users create alice
+miren sandbox exec "$SANDBOX" -- headscale preauthkeys create --user 1 --expiration 24h
+miren sandbox exec "$SANDBOX" -- headscale nodes list
+```
+</CliCommand>
+
+Passing a command after `--` runs it as argv, with no shell in the way, which is why this
```
phinze: The 🤖 detail distinction looks correct but I wonder if pulling up the level of detail on this and the two callouts will help make it all simpler.

---

🤖 The commands work, but not quite for the reason given. `servers/exec/exec.go:309-311`: with args present **and** a configured entrypoint, exec builds `[]string{"/bin/sh", "-c", "exec " + ep + " " + strings.Join(args, " ")}` — shell very much in the way. It only runs argv directly when `ep` is empty.

`ep` is empty here because `ConfigSpec.Entrypoint` is populated solely on the stackbuild path (`servers/build/buildkit.go:363`, `res.Entrypoint = stack.Entrypoint()`), and a Dockerfile build never reaches `DetectStack`. The comment at `servers/build/build.go:588-592` says so directly.

So the sentence is true of this app but not of `miren sandbox exec` in general — a source-deployed Go or Node app would take the shell path. Deleting it costs the reader nothing, since the commands above already demonstrate that it works.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+:::
+
+Then, on the machine joining the tailnet, point Tailscale at your server and use the
+preauth key:
+
+```bash
+tailscale up --login-server=https://headscale.example.com --authkey=<key>
+```
+
+:::info[What wasn't exercised]
```
phinze: Process notes, cut! ✂️ 

---

🤖 None of the three sibling recipes carry an authoring-process note like this one. The untested-ness is worth keeping where it's load-bearing (the client join and the DERP section could each say they follow upstream docs), but as a standalone callout about the page's own provenance it's addressed to reviewers rather than readers.

phinze (changes requested):
As per live convo - I think we should scrub this to de-center Miren and impl details and focus on what the user is trying to get done.

---

🤖 Structural context for the sweep. The "doubles as a tour of Miren" line is inherited template, present in two of the three existing recipes:

- `docs/docs/recipes/hermes-agent.md:20` — "it exercises a lot of real Miren behavior … so it doubles as a tour of what matters when you bring your own image"
- `docs/docs/recipes/amp-runner.md:21` — "it exercises real Miren behavior … so it doubles as a tour of running an outbound-only worker"

`openhands-agent-server.md` is the counterexample, and it landed in the same batch (all four recipes were added within days of each other in mid-July). It carries no tour framing and is the shortest of the set at 243 lines, against 255 and 310 for the two that do — and 431 here.

Otherwise this page is a faithful match to its siblings: same section order, 8 admonitions against their 6–7. The delta is almost entirely the three added sections and the mechanism explanations inside them.

docs/docs/recipes/headscale.md
evanphx: 🤖 Dropped. Agreed it was the seed — the whole intro now says what you'll have at the end rather than which Miren features get exercised on the way. Leaving the sibling sweep for a follow-up per your call.

docs/docs/recipes/headscale.md
evanphx: 🤖 Chopped to the user-facing rule: this image has no shell, so don't set a command, and here's the consequence you'll actually feel (no interactive shell — pointer to the variant that gives you one). The PID-1 and `/bin/sh -c` mechanics are gone.

docs/docs/recipes/headscale.md
evanphx: 🤖 Moved to Prerequisites and cut to "clients hold connections longer than the default allows." Your 🤖 note convinced me to go further: since it's cluster-global and there's no headscale-side override, the prerequisite now states plainly that you need rights to edit the server config, with a callout that nodes will keep dropping if you can't. Also deduped — the arithmetic in roadblock item 8 is now a pointer.

docs/docs/recipes/headscale.md
evanphx: 🤖 Deleted, and thanks — the note caught a real error, not just excess detail. I'd written it as a property of `sandbox exec`, but exec only skips the shell when the config spec has no entrypoint, which holds for a Dockerfile build and not a stack build. It was wrong in general, and the commands above already show it works.

docs/docs/recipes/headscale.md
evanphx: 🤖 Cut.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+
+The headscale CLI talks to the running server over the unix socket in
+`/var/run/headscale`, so these have to run **in the live sandbox**:
+
+<CliCommand context="client">
+```bash
+SANDBOX=$(miren sandbox list --format json | jq -r '.[0].id')
+
+miren sandbox exec "$SANDBOX" -- headscale users create alice
+miren sandbox exec "$SANDBOX" -- headscale preauthkeys create --user 1 --expiration 24h
```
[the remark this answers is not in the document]
evanphx: 🤖 Right about the problem, but that fix doesn't work — in 0.29.3 the flag is uint-only, so `--user alice` fails with `invalid argument "alice" for "-u, --user" flag: strconv.ParseUint`. Fixed by adding a `headscale users list` step and passing the ID it reports.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+# Overridden by HEADSCALE_SERVER_URL from app.toml.
+server_url: http://127.0.0.1:8080
+
+# Bind on all interfaces: Miren health-checks and routes from outside the container,
+# so a 127.0.0.1 listener never comes up healthy.
+listen_addr: 0.0.0.0:8080
+metrics_listen_addr: 0.0.0.0:9090
+
+# The CLI reaches the running server over this socket.
+unix_socket: /var/run/headscale/headscale.sock
```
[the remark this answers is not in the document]
evanphx: 🤖 Checked this against 0.29.3 — headscale creates the directory and binds the socket itself, so no mount or `mkdir` is needed. Verified on the bare distroless image with nothing but `/data` mounted: the server comes up and `headscale users create` succeeds over that socket. It's also already exercised by the live deploy this recipe was written against, where the admin commands work.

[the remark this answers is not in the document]
evanphx: 🤖 Not needed — headscale creates `/var/run/headscale` itself at startup. Verified on the plain 0.29.3 image with only `/data` mounted: it binds the socket and `headscale users create` works. The analysis above checked whether the directory pre-exists in the `-debug` image, which it doesn't, but that's a different question from whether the server creates it.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+:::
+
+## Create a user and register a node
+
+The headscale CLI talks to the running server over the unix socket in
+`/var/run/headscale`, so these have to run **in the live sandbox**:
+
+<CliCommand context="client">
+```bash
+SANDBOX=$(miren sandbox list --format json | jq -r '.[0].id')
```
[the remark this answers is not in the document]
evanphx: 🤖 Real bug, thanks — on our test cluster `.[0]` was a different app's sandbox. The suggested filter wouldn't have caught it though: `app` serializes as `app/headscale`, not `headscale`, so the select matches nothing and the `jq -e` errors out. Went with reading the ID off `miren sandbox list` instead, which fixes it without a jq incantation in a doc.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,431 @@
...
+string — see [Server Configuration](/server-config#server).
+
+## Verify
+
+<CliCommand context="client">
+```bash
+miren app status -a headscale   # Current Version + active
+miren sandbox list              # one running sandbox, service "web"
+
+curl -s https://headscale.example.com/health   # {"status":"pass"}
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, now `curl -fsS`.

docs/docs/recipes/headscale.md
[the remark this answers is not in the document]
evanphx: 🤖 Addressed — the recipe now runs `headscale users list` and passes the ID it reports, rather than assuming `1`.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,384 @@
...
+
+## The app.toml
+
+This file lives at `.miren/app.toml`.
+
+```toml
+name = "headscale"
+
+[build]
+dockerfile = "Dockerfile"
```
phinze, line 152: 💭 I wonder if we should default to this eventually... lots of implications and footguns though.

docs/docs/recipes/headscale.md
```diff
@@ -0,0 +1,384 @@
...
+you, and [Firewall](/firewall) covers the inbound rules.
+
+:::warning[Keep the public relays as a fallback]
+Relayed traffic reaches your hostname over the same HTTPS path as everything else, so a
+relayed session that goes quiet long enough hits the timeout from
+[earlier](#longer-timeout). Leave Tailscale's servers in `derp.urls` alongside your own
+rather than removing them.
+:::
+
+## Roadblock checklist
```
phinze: Ah missed this in my first read-through, IMHO this can be cut too, sort of a weird summary section.

phinze (approved):
Much better! One more recommended cut inline but not blocking.

docs/docs/recipes/headscale.md
evanphx: Agreed, worth discussing!

docs/docs/recipes/headscale.md
evanphx: 🤖 Cut. Two things went with it that only lived there — the verbatim `base_domain` overlap error and `error acquiring lease: app/headscale`. Both rules are still stated as comments on the config they concern, so what's lost is the exact text someone might paste into a search. Easy to put back inline if that turns out to matter.

Worth noting the other three recipes all still carry a checklist (7–8 items), so this page is now the odd one out until a sweep follows — same shape as the tour-line question.

PR #1056, opened 2026-08-13
Reach a cluster's RPC through the link it already holds up to cloud

evanphx:
Today `miren` reaches a cluster by dialing it, which quietly assumes you have a route to it. A cluster behind NAT can accept nothing, so an operator without a path to it cannot run a single command, even though the cluster is sitting there holding an authenticated WebSocket up to cloud that it dialed outbound. This turns that link into a way in.

The other half of the setup already landed: `pkg/rpc` grew a message transport that runs the whole RPC protocol over any pipe of discrete byte messages, explicitly including "a socket carrying somebody else's envelope protocol". The uplink is exactly that, so `pkg/cloudrpc` becomes a tenant of it in the same shape as Miren Anywhere, and each relayed session turns into an `rpc.MessageConn` carrying ordinary RPC frames.

The part worth being precise about is authentication, because the detour changes nothing. The caller's own bearer rides inside every operation frame and is judged by the cluster's existing authenticator and RBAC, exactly as on a direct connection. Cloud never sees that token, it is inside a CBOR frame cloud does not parse, so cloud is a byte relay and vouches for nothing. The copy on the WebSocket handshake exists only so cloud can decide whether to carry the traffic at all.

Three supporting changes fell out. The uplink capped reads at coder/websocket's 32 KiB default and dropped on a full outbox, both of which are right for telemetry that self-heals and fatal for RPC, which cannot resynchronise after a hole. `ws://` meant TLS, the same as `wss://`, and now means plaintext, which is what it says and what a local development cloud needs. And a `wss://` remote naming a path keeps it, so a remote can point at somebody else's endpoint instead of requiring that endpoint to mimic our route.

Clusters opt in with `miren cluster add --via-cloud`. A fallback from a failed direct dial would make every command wait out a timeout first and leave you unsure which path answered.

Two later commits are hardening from review: bounding what a session can hold (by bytes, since the peer picks the frame size), capping concurrent sessions, refusing to put credentials on an unencrypted socket outside loopback, and stopping a session refusal from stalling the link's shared read loop.

Needs mirendev/cloud#201 to answer. Without it these config entries have nothing to connect to.

Part of MIR-1580

### DIFF
--- a/cli/commands/config_bind_helpers.go
+++ b/cli/commands/config_bind_helpers.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -15,6 +16,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"miren.dev/runtime/clientconfig"
+	"miren.dev/runtime/pkg/rpc"
 	"miren.dev/runtime/pkg/theme"
 	"miren.dev/runtime/pkg/ui"
 )
@@ -322,6 +324,78 @@ func clusterOnlineInCloud(
 
 	return response.Online, nil
 }
+
+const (
+	// cloudProbeTimeout bounds the relayed probe. The failure it exists to
+	// catch is silence, so without a deadline the probe inherits the very hang
+	// it is there to prevent. Longer than the presence check, because this one
+	// opens a session and makes a call rather than asking a question.
+	cloudProbeTimeout = 20 * time.Second
+
+	// cloudProbeObject is what the probe resolves. Any exposed object would do;
+	// this one is the object nearly every command reaches for, so a cluster
+	// that cannot produce it is not usable through cloud in any case.
+	cloudProbeObject = "dev.miren.runtime/app"
+)
+
+// cloudRouteWorks reports whether a cloud-routed entry for this cluster would
+// actually work, by building the entry that would be written and using it.
+//
+// Cloud reporting a cluster online is a weaker claim than it appears. It says a
+// link exists; it does not say what is on the far end of that link will answer
+// a relayed call. A cluster running a runtime from before the relay shipped is
+// online and silent: it has no handler for the session, so nothing refuses the
+// call and nothing answers it either. An entry written on the strength of
+// presence alone would hang rather than fail.
+//
+// So the cloud route is probed the same way a direct address is — by using it,
+// and believing the result rather than a claim about it. This costs a round
+// trip at add time and settles the question for every command afterwards.
+func cloudRouteWorks(
+	ctx *Context,
+	config *clientconfig.Config,
+	identityName string,
+	identity *clientconfig.IdentityConfig,
+	cluster *ClusterResponse,
+) bool {
+	// The entry that would be written, not an approximation of it. Probing
+	// anything else would leave the thing actually written untested.
+	probe := &clientconfig.ClusterConfig{
+		ViaCloud: true,
+		XID:      cluster.XID,
+		Identity: identityName,
+	}
+	if identity != nil && strings.HasPrefix(identity.Issuer, "http://") {
+		probe.Insecure = true
+	}
+
+	probeCtx, cancel := context.WithTimeout(ctx, cloudProbeTimeout)
+	defer cancel()
+
+	state, err := probe.State(probeCtx, config, rpc.WithLogger(ctx.Log))
+	if err != nil {
+		ctx.Warn("Could not reach %s through cloud: %v", cluster.Name, err)
+		return false
+	}
+	defer state.Close() //nolint:errcheck // a probe's teardown tells us nothing
+
+	// Resolving the object is a round trip the cluster itself has to answer.
+	// Cloud relays it and cannot satisfy it, which is what makes this evidence
+	// about the cluster rather than about the link.
+	if _, err := state.Client(cloudProbeObject); err != nil {
+		ctx.Warn("Could not reach %s through cloud: %v", cluster.Name, err)
+		return false
+	}
+
+	return true
+}
+
+// cloudRouteProbe is the check the fallback decision runs, as a variable so a
+// test can substitute one. The real probe needs a live relayed session, which
+// is the blackbox suite's job; what is worth testing here is the decision made
+// from its answer. Tests in this package run sequentially, so swapping it is
+// safe as long as none of them opts into t.Parallel.
+var cloudRouteProbe = cloudRouteWorks
 
 // cloudRoutableClusters reports which of the given clusters cannot be dialed
 // directly but can be reached through cloud, keyed by XID.
--- a/cli/commands/config_bind_helpers.go
+++ b/cli/commands/config_bind_helpers.go
@@ -51,6 +51,12 @@ const (
 	// port is what's necessary and sufficient, not TCP) and points at the
 	// on-host diagnostic that reproduces the exact decision.
 	unreachableAddressHelp = "open UDP 8443 (QUIC) on the host or set additional_ips, then restart miren; run 'miren debug advertise' on the host to see why"
+
+	// viaCloudNote replaces that note for a cluster nothing can dial but cloud
+	// can still reach. Not a degraded state to remediate: it is a working
+	// cluster on a network that does not accept inbound connections, which is a
+	// reasonable way to run one.
+	viaCloudNote = "via Miren Cloud"
 )
 
 // printUnreachableClustersHelp prints a warning header, a bulleted list of the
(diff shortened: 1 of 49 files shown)


pkg/rpc/message_ws.go
```diff
@@ -40,20 +40,56 @@ func (w *wsConn) Close() error {
...
+		scheme, remote = "ws", rest
+	} else {
+		remote = strings.TrimPrefix(remote, "wss://")
+	}
+
+	host, path, _ := strings.Cut(remote, "/")
+	if path == "" {
+		return scheme + "://" + host + wsMessagePath
+	}
+	return scheme + "://" + host + "/" + path
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, though not quite as suggested. Rejecting the remote outright would make `pkg/rpc` enforce policy, and it already has a legitimate plaintext case in tests. Instead `dialWSMessageConn` now declines to *send the bearer* over an unencrypted hop to anything but loopback, which removes the exposure without the transport deciding who you are allowed to talk to. A caller that meant it gets an auth failure from the far end rather than a leaked token. `clientconfig` still refuses the config outright, so the common path gets the clear error. Covered by `TestBearerTravelsSafely`.

pkg/cloudrpc/cloudrpc.go
```diff
@@ -0,0 +1,289 @@
...
+		// Ordinary at the tail of a teardown: the far end had frames in flight
+		// when we closed. Nothing to do about them, and nothing alarming.
+		s.log.Debug("frame for unknown relayed RPC session", "session", msg.SessionID)
+		return nil
+	}
+
+	if !sess.reserve(len(msg.Payload)) {
+		s.log.Warn("relayed RPC session is holding too much, closing it",
+			"session", msg.SessionID, "limit", maxPendingBytes)
+		s.forget(sess.id)
```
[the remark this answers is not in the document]
evanphx: 🤖 Agreed, and done: there is now a single `teardown(sess)` that shuts the session down first and then forgets it, used by every path. It also only removes the session that is actually registered, so a late second teardown cannot evict a successor holding the same id — which was a latent version of the same "the ordering happens to work" problem.

pkg/cloudrpc/cloudrpc.go
```diff
@@ -0,0 +1,289 @@
...
+		sess.shutdown()
+	}()
+
+	err := s.state.ServeMessageConn(ctx, sess, rpc.WithMaxFrameSize(maxFrameData))
+
+	// Tell the far end, best effort: if the uplink is what failed, this goes
+	// nowhere, and the caller learns from its own broken connection instead.
+	//nolint:errcheck // best-effort teardown notice
+	s.send(ctx, TypeClose, Close{SessionID: sess.id, Reason: "session ended"})
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch, that was a real inconsistency rather than just a confusing one: `sendBestEffort` was added for exactly this and only `handleOpen` was switched over. `serve` uses it now, so the code and the comment agree.

clientconfig/local.go
```diff
@@ -238,6 +246,142 @@ foundAddress:
...
+func cloudRPCEndpoint(cloudURL, clusterXID string, allowPlaintext bool) (string, error) {
+	scheme, base := "wss://", cloudURL
+	switch {
+	case strings.HasPrefix(base, "http://"):
+		scheme, base = "ws://", strings.TrimPrefix(base, "http://")
+	case strings.HasPrefix(base, "https://"):
+		base = strings.TrimPrefix(base, "https://")
+	}
+
+	base = strings.TrimRight(base, "/")
```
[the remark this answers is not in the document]
evanphx: 🤖 Intended, and now written down. A cloud mounted under a path is a real shape (an install behind a shared domain), so the relay route is appended to where that cloud actually lives rather than to its host. Added cases to `TestCloudRPCEndpoint` covering it with and without the trailing slash.

pkg/cloudrpc/cloudrpc.go
```diff
@@ -0,0 +1,304 @@
...
+	// An ordinary ending arrives as EOF or a cancelled context: the caller hung
+	// up, or this side tore the session down. Attaching an error field to those
+	// makes every healthy session read as a fault to an operator scanning for
+	// one, so the field is kept for endings that are actually unexpected.
+	switch {
+	case err == nil, errors.Is(err, io.EOF), errors.Is(err, context.Canceled):
+		s.log.Info("relayed RPC session closed", "session", sess.id)
+	default:
+		s.log.Info("relayed RPC session closed", "session", sess.id, "error", err)
+	}
```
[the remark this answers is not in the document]
evanphx: 🤖 Went with Warn rather than Error here. CLAUDE.md draws the line at whether the system failed at something it is responsible for: a relayed session dying on a transport error is usually the far end's network, which is degraded-and-handled rather than this cluster failing. Ordinary endings (EOF, cancelled context) stay at Info with no error field, so an operator scanning for faults doesn't see every clean hangup as one.

pkg/rpc/audit_test.go
```diff
@@ -495,3 +495,44 @@ func TestLogCertAuthDeduplicates(t *testing.T) {
...
+	}
+}
+
+type plainMessageConn struct{}
+
+func (plainMessageConn) Send([]byte) error     { return nil }
+func (plainMessageConn) Recv() ([]byte, error) { return nil, nil }
+func (plainMessageConn) Close() error          { return nil }
+
+type namedMessageConn struct{ plainMessageConn }
```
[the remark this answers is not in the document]
evanphx: 🤖 We think this one's a false positive. The two helpers have different names and live in different packages: `namedMessageConn` is in `pkg/rpc/audit_test.go` (package `rpc`), `namedConn` is in `pkg/rpc/message_adopt_test.go` (package `rpc_test`). `go vet ./pkg/rpc/` compiles clean, so there's no redeclaration to fix.

pkg/cloudrpc/cloudrpc.go
```diff
@@ -0,0 +1,309 @@
...
+
+	go s.serve(ctx, sess)
+
+	return nil
+}
+
+func (s *Server) serve(ctx context.Context, sess *session) {
+	defer s.teardown(sess)
+
+	err := s.state.ServeMessageConn(ctx, sess, rpc.WithMaxFrameSize(maxFrameData))
```
phinze: 🤖 This path currently loses method-parameter introspection. `Server.msgMethods` builds `methodsResponse{Methods: methods}` instead of using `newMethodsResponse`, so a cloud-routed client's `HasMethodParam` always returns false even against a current server. In practice, `miren logs --until` rejects a compatible cluster as too old, and deploy falls back to the deprecated client-owned deployment path. Could we return the shared `newMethodsResponse(hc.methods)` from the message transport and cover it with a cloud-routed introspection test?

pkg/cloudrpc/session.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Recv returns the next frame, or io.EOF once the far end is gone and nothing
+// buffered remains.
+func (s *session) Recv() ([]byte, error) {
+	// Buffered frames outrank a close that has already been signalled: the far
+	// end may have sent its last reply and then hung up, and returning EOF while
+	// that reply sits in the channel would lose it.
+	select {
+	case b := <-s.inbound:
+		s.release(len(b))
```
phinze, line 96: 🤖 This releases the 8 MiB accounting as soon as `msgmux` takes a frame, but `msgmux` then appends the payload to `msgStream.buf` without any byte ceiling. That means `maxPendingBytes` does not bound the relay session's actual memory: a caller can keep one stream open and send DATA faster than its handler consumes it, or continue after the handler rejects the operation, while `readPump` keeps `session.inbound` drained and moves the bytes into unaccounted buffers. Could we enforce a session-wide buffered-byte limit inside `msgmux` and tear down the session when it is exceeded, with a test that holds one stream open past the limit?

pkg/cloudrpc/cloudrpc.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	if req.SessionID == "" {
+		return fmt.Errorf("rpc.open with no session id")
+	}
+
+	sess := &session{
+		id:      req.SessionID,
+		srv:     s,
+		ctx:     ctx,
+		inbound: make(chan []byte, inboundDepth),
```
phinze: 🤖 This context is scoped to the entire uplink, so ending one relay session does not cancel RPC handlers already dispatched from it. `teardown` removes the session from the map and closes its message pipe, but `sessionRouter`'s handler goroutines retain this context until the cluster-cloud link itself reconnects. A long-running build or log stream can therefore survive its caller, and because its session no longer counts toward `maxSessions`, repeated open-and-disconnect cycles bypass that resource ceiling. Could each relay session own a child context that `session.shutdown()` cancels, and pass that context to `ServeMessageConn`? A handler that blocks on `ctx.Done()` would make a focused teardown test.

pkg/uplink/client.go
```diff
@@ -190,6 +205,26 @@ func (c *Client) Send(env *Envelope) {
...
+// broke" has to notice the break — which, for anything session-shaped, means
+// tearing the session down and letting the caller retry.
+func (c *Client) SendContext(ctx context.Context, env *Envelope) error {
+	select {
+	case c.outbox <- env:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
phinze, line 237: 🤖 Could this use the same upfront `ctx.Err()` guard as `SendBlocking` below? With outbox room and an already-canceled connection context, both select cases are ready and Go may enqueue the envelope instead of returning the cancellation. A stale relay sender that races after the next connection's `drainOutbox` can then put `rpc.data` from an ended session onto the new uplink while Cloud may still route that session ID, violating the no-resume-on-link-break contract. Mirroring `TestSendBlockingRefusesADeadContextEvenWithRoom` for `SendContext` should pin the fix.

cli/commands/config_bind_helpers.go
```diff
@@ -252,6 +258,128 @@ func fetchAvailableClusters(ctx *Context, config *clientconfig.Config, identityN
...
+	}
+
+	var response struct {
+		Online bool `json:"online"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return false, fmt.Errorf("failed to parse response: %w", err)
+	}
+
+	return response.Online, nil
```
phinze, line 326: 🤖 `online` only proves that the cluster has an uplink; it does not prove that its runtime has registered the new `rpc.open` handler. During rollout, a new CLI can therefore auto-write `via_cloud: true` for an older online cluster. Cloud accepts the caller and forwards `rpc.open`, but the old runtime only logs “no handler registered”; with no open acknowledgement the caller can wait indefinitely. Since the upcoming uplink handshake gives us a natural place to advertise an `rpc_relay` capability, could we keep automatic selection disabled and require explicit `--via-cloud` until that lands? An open acknowledgement or timeout would still help stale and hand-edited configs fail fast.

phinze (changes requested):
Spent a good long while with 🤖 on this pair of PRs and the notes from our convo. Very excited about the feature! Inlines below are 🤖-surfaced issues that rose to the level of fix-before-merge.

---

🤖 Directionally, `MessageConn` feels like the right seam to preserve. It lets us ship this WebSocket relay without binding RPC semantics to it, and leaves room for a more QUIC-native transport later if bulk traffic makes the shared TCP path’s head-of-line behavior material.

The cloud-routed happy path is well covered. I found five correctness, resource-lifetime, and rollout gaps that need closing before landing: parameter introspection is incomplete over `MessageConn`, the advertised per-session byte limit ends before `msgmux`’s unbounded buffers, session teardown does not cancel dispatched handlers, `SendContext` can enqueue onto a later connection after cancellation, and automatic selection treats any online runtime as relay-capable. The resource-lifetime cases let work or frames escape the session and connection lifetimes meant to bound them. The rollout gap can auto-write a broken config for an older cluster, so this first release should stay explicitly opt-in until the upcoming uplink handshake can advertise the relay capability.

pkg/cloudrpc/cloudrpc.go
evanphx: 🤖 Fixed. `msgMethods` now returns `newMethodsResponse(hc.methods)`, the same builder the HTTP endpoint uses, so the two can't report different things. Added `TestRelayedIntrospectionReportsMethodParameters` covering it over the relay; confirmed it fails against the old code with the message "the relayed client sees the method but not its parameters."

pkg/cloudrpc/session.go
evanphx: 🤖 Fixed, in msgmux, since that's where the buffers are. Each session now tracks delivered-but-unread bytes across all its streams and tears down when a frame would exceed the limit — tearing down being the only lever, since there's no way to apply backpressure retroactively and dropping the frame leaves a hole RPC can't resync after. Accounting is per stream so releasing is idempotent by construction, and streams release what they hold when removed, or a long connection running short ops would ratchet itself shut. The relay passes its own `maxPendingBytes` down, so the advertised number is the one both halves enforce. Test holds a stream open past the limit; confirmed it fails without the ceiling.

Also corrected the msgmux header, which claimed no flow control was needed — true of throughput, not of buffering, which was the distinction it missed.

pkg/cloudrpc/cloudrpc.go
evanphx: 🤖 Fixed. Each session owns a child context that `shutdown()` cancels, and that's what's passed to `ServeMessageConn`. A constructor builds the session and its cancel together, since a session missing its cancel is exactly this bug. Test parks a handler on `ctx.Done()` and tears the session down under it; confirmed it fails without the change.

pkg/uplink/client.go
evanphx: 🤖 Fixed, mirroring `SendBlocking`'s guard, and added `TestSendContextRefusesADeadContextEvenWithRoom` alongside the existing one. Confirmed it fails without the guard (returns nil about half the time, so the test repeats 200×).

cli/commands/config_bind_helpers.go
evanphx: 🤖 You're right that `online` is the wrong evidence. We went at it from the other side rather than gating the feature off: the direct path doesn't ask whether an address would work, it connects to it, so the cloud route is now probed the same way.

`cluster add` builds the entry it would write, opens a session through cloud, and resolves an object the cluster itself has to answer — cloud relays that round trip and can't satisfy it, which is what makes it evidence about the cluster rather than about the link. Bounded by a timeout, since silence is the failure being caught. Both automatic paths go through it; the no-address case used to trust the picker hint alone and now confirms too, with an error naming the upgrade if the cluster turns out to be too old. The presence check stays as the picker's hint, where being wrong is cheap because the probe is what commits.

This seemed better than waiting for the handshake since it tests the actual route rather than a capability bit, but if you'd still rather have it explicit-only until `rpc_relay` lands, say so and I'll gate it.

evanphx:
Went back through the nitpick sections from the earlier reviews.

**Presence checks run concurrently** — taken. They ran one at a time with a 10s timeout each while somebody waits at a prompt, so several undialable clusters meant tens of seconds of blank screen. Now parallel, matching how `tryConnectToCluster` already probes addresses. `-race` promptly caught that the test server recorded requests without a lock, which it now takes.

**Audit records at the Info floor** — skipping. Loopback successes drop to Debug deliberately, and the comment right there says why: same-host component chatter, where the audit signal we care about is a network peer. Raising them to Info would flood the stream with lines identical on every firing, which is what CLAUDE.md's log-level guidance puts at Debug. The `certAuthDeduper` idea is reasonable but is separate work, and this predates the branch.

**15m timeout for `TestDeployViaCloud`** — skipping; it runs ~95s locally and the whole pop job finishes in about 5m in CI, so there is ample headroom.

**Docs admonitions** — skipping, style preference.

**`TestRoutedEntryKeepsNoDialableFields` doesn't call `addCluster`** — fair, and still true. An end-to-end version needs to drive the interactive picker, which is the same gap that means there is no non-interactive way to add a cloud-routed cluster at all. Left as-is by decision rather than oversight.

cli/commands/config_bind_helpers.go
phinze: Ah nice fix! I didn't see a way to make the experience better without the handshake, but with the way you tweaked it now old clusters are no worse than unreachable clusters before. Strictly better so we don't need to fall back to opt-in-only.

phinze (approved):
All fixes looking good! Let's get it out there and take it for a spin! 🚢

PR #1057, opened 2026-08-13
Let a route carry its own request timeout

evanphx:
Ingress gives up on a proxied request once it has gone quiet for `server.http_request_timeout`, 60 seconds by default. That covers most apps, but it cuts off the ones that are silent on purpose: long polling, an endpoint that waits on a slow job before writing its first byte, a report that takes minutes to assemble. Those got killed at 60s and reconnected in a loop. The only lever was raising the setting, which is cluster-wide, so one long-poll app meant weakening the timeout for every other app on the cluster.

A route can now carry a `request_timeout` of its own, set with `miren route timeout longpoll.example.com 10m`. `--clear` puts it back on the cluster default.

The interesting part was where the timeout actually lives. It's two knobs on `http.Transport`: `ResponseHeaderTimeout` bounds the wait for the app's response headers, and the read deadline `idleTimeoutConn` sets bounds the gap between reads once bytes are flowing. Neither is settable per request, and Go's connection pool keys on (scheme, host) alone, so a connection dialed under a 10 minute deadline would happily get handed to the next request that wanted 60 seconds. Threading the value down through the request context runs into the same wall. So rather than mutate one shared transport, ingress keeps one transport per distinct timeout value, built lazily and memoized. Each gets its own connection pool, which is the entire point. The count is bounded by how many distinct overrides an operator has configured, so in practice one or two.

A stored value that will not parse, or that parses non-positive, falls back to the server default and is otherwise ignored. That matches `resolvePortWaitTimeout` and `getShutdownTimeout`, which handle the same shape of user-supplied duration string the same way, and `SetRouteRequestTimeout` validates at the write boundary so a typo can only arrive by editing the entity directly.

Verified end to end with the server default dropped to 5 seconds against an app whose endpoint sleeps for 15. The plain route 503s at 5s, the route carrying a 60s override returns 200 at 15s, and clearing the override puts it back to 503. The transport gets built once no matter how much traffic runs through it.

Two things I left alone on purpose. The JSON output reports `request_timeout` as a Go duration string like `10m`, which a non-Go consumer can't compare or sort without reimplementing `ParseDuration`, and the transport map never evicts, so churning a route's override repeatedly leaves small orphans behind. Both are noted in my local task list rather than pretending they're urgent.

Closes MIR-1576

### DIFF
--- a/blackbox/route_timeout_test.go
+++ b/blackbox/route_timeout_test.go
@@ -1,0 +1,95 @@
+//go:build blackbox
+
+package blackbox
+
+import (
+	"encoding/json"
+	"testing"
+
+	"miren.dev/runtime/blackbox/harness"
+)
+
+func TestRouteTimeoutSetShowClear(t *testing.T) {
+	c := harness.NewCluster(t)
+	m := harness.NewMiren(t, c)
+
+	name := harness.DeployApp(t, m, harness.AppOptions{
+		Testdata: "go-server",
+	})
+
+	host := name + ".timeout.test.local"
+
+	m.MustRun("route", "set", host, name).RequireSuccess(t)
+
+	// No override to begin with
+	r := m.MustRun("route", "timeout", host)
+	r.RequireSuccess(t)
+	r.RequireContains(t, "server default")
+
+	// Set an override
+	r = m.MustRun("route", "timeout", host, "10m")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "10m")
+
+	// It shows up on the route
+	r = m.MustRun("route", "show", host)
+	r.RequireSuccess(t)
+	r.RequireContains(t, "10m")
+
+	r = m.MustRun("route", "show", host, "--format", "json")
+	r.RequireSuccess(t)
+
+	var shown struct {
+		RequestTimeout string `json:"request_timeout"`
+	}
+	if err := json.Unmarshal([]byte(r.Stdout), &shown); err != nil {
+		t.Fatalf("failed to parse route show JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if shown.RequestTimeout != "10m" {
+		t.Errorf("expected request_timeout 10m, got %q", shown.RequestTimeout)
+	}
+
+	// And in the list
+	r = m.MustRun("route", "list")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "TIMEOUT")
+
+	// Invalid values are rejected without disturbing the stored one
+	r = m.Run("route", "timeout", host, "10")
+	if r.Success() {
+		t.Errorf("expected a bare number to be rejected\nstdout: %s\nstderr: %s", r.Stdout, r.Stderr)
+	}
+
+	r = m.MustRun("route", "timeout", host, "--format", "json")
+	r.RequireSuccess(t)
+	if err := json.Unmarshal([]byte(r.Stdout), &shown); err != nil {
+		t.Fatalf("failed to parse route timeout JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if shown.RequestTimeout != "10m" {
+		t.Errorf("expected request_timeout to stay 10m, got %q", shown.RequestTimeout)
+	}
+
+	// Clearing puts the route back on the server default
+	r = m.MustRun("route", "timeout", host, "--clear")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "server default")
+
+	r = m.MustRun("route", "timeout", host, "--format", "json")
+	r.RequireSuccess(t)
+	// Unmarshal into a fresh value: request_timeout is omitted once cleared, so
+	// reusing `shown` would leave the previous value in place.
+	var cleared struct {
+		RequestTimeout string `json:"request_timeout"`
+	}
+	if err := json.Unmarshal([]byte(r.Stdout), &cleared); err != nil {
+		t.Fatalf("failed to parse route timeout JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if cleared.RequestTimeout != "" {
+		t.Errorf("expected request_timeout to be cleared, got %q", cleared.RequestTimeout)
+	}
+
+	// Clearing again is a no-op
+	r = m.MustRun("route", "timeout", host, "--clear")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "No request timeout override")
+}
--- a/blackbox/route_timeout_test.go
+++ b/blackbox/route_timeout_test.go
@@ -93,3 +93,63 @@ func TestRouteTimeoutSetShowClear(t *testing.T) {
 	r.RequireSuccess(t)
 	r.RequireContains(t, "No request timeout override")
 }
+
+// TestRouteTimeoutDefaultRoute covers the --default form, where there is no
+// hostname to give and the sole positional is the duration. That shift is easy
+// to get wrong in argument parsing and cannot be exercised by the named-host
+// path above.
+func TestRouteTimeoutDefaultRoute(t *testing.T) {
+	c := harness.NewCluster(t)
+	m := harness.NewMiren(t, c)
+
+	name := harness.DeployApp(t, m, harness.AppOptions{
+		Testdata: "go-server",
+	})
+
+	m.MustRun("route", "set-default", name).RequireSuccess(t)
+
+	r := m.MustRun("route", "timeout", "--default")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "server default")
+
+	// The duration lands in the right slot despite being the only positional.
+	r = m.MustRun("route", "timeout", "--default", "5m")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "5m")
+
+	r = m.MustRun("route", "show", "--default", "--format", "json")
+	r.RequireSuccess(t)
+
+	var shown struct {
+		RequestTimeout string `json:"request_timeout"`
+	}
+	if err := json.Unmarshal([]byte(r.Stdout), &shown); err != nil {
+		t.Fatalf("failed to parse route show JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if shown.RequestTimeout != "5m" {
+		t.Errorf("expected request_timeout 5m on the default route, got %q", shown.RequestTimeout)
+	}
+
+	// A hostname alongside --default is a contradiction, not a second reading.
+	r = m.Run("route", "timeout", "--default", "example.com", "5m")
+	if r.Success() {
+		t.Errorf("expected --default with a hostname to be rejected\nstdout: %s\nstderr: %s", r.Stdout, r.Stderr)
+	}
+
+	r = m.MustRun("route", "timeout", "--default", "--clear")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "server default")
+
+	r = m.MustRun("route", "timeout", "--default", "--format", "json")
+	r.RequireSuccess(t)
+
+	var cleared struct {
+		RequestTimeout string `json:"request_timeout"`
+	}
+	if err := json.Unmarshal([]byte(r.Stdout), &cleared); err != nil {
+		t.Fatalf("failed to parse route timeout JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if cleared.RequestTimeout != "" {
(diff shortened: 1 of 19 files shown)


blackbox/route_timeout_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if cleared.RequestTimeout != "" {
+		t.Errorf("expected request_timeout to be cleared, got %q", cleared.RequestTimeout)
+	}
+
+	// Clearing again is a no-op
+	r = m.MustRun("route", "timeout", host, "--clear")
+	r.RequireSuccess(t)
+	r.RequireContains(t, "No request timeout override")
+}
```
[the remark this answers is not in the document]
evanphx: Good call, took this. The `--default` form is the one place where there is no hostname to give, so the sole positional is the duration and `resolveRouteTimeoutArgs` shifts it into place. That shift was in fact wrong on the first pass and I only caught it driving the CLI by hand, so it deserves coverage that goes through the real argument parser rather than just the unit case.

Added `TestRouteTimeoutDefaultRoute` in 8ffe433e covering show, set, the JSON shape, and clear on the default route, plus the contradiction case of passing a hostname alongside `--default`. Passes locally against a dev cluster alongside the existing named-host test.

docs/docs/command/route-timeout.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Global Options
+
+- `--options` — Path to file containing options
+- `--server-address` — Server address to connect to (default: `127.0.0.1:8443`)
+- `--verbose, -v` — Enable verbose output
+
+## Examples
+
+**Give a long-poll route a 10 minute timeout:**
```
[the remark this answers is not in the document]
evanphx: Fixed in 8ffe433e. Worth noting for future passes that this file is generated by `hack/gen-command-docs`, so the edit went into the example name in `cli/commands/commands.go` and the doc was regenerated from there.

servers/httpingress/httpingress.go
```diff
@@ -189,6 +196,114 @@ func NewServer(
...
+
+// routeRequestTimeout resolves a route's request timeout override. It returns 0
+// when the route carries no override or carries one we cannot honor, which
+// tells transportFor to use the server-wide default.
+func (h *Server) routeRequestTimeout(route *ingress_v1alpha.HttpRoute) time.Duration {
+	if route == nil || route.RequestTimeout == "" {
+		return 0
+	}
+
+	d, err := time.ParseDuration(route.RequestTimeout)
```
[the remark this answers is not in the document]
evanphx: Going to pass on this one. The parse is real work, but I think the cache would cost more than it saves.

A mutex-guarded map lookup means acquiring a lock and hashing the key. `time.ParseDuration` on a three-character string like `10m` is a short scan of a tiny string with no allocation. Those are close enough that the cache is plausibly the slower of the two, and it adds shared mutable state to the request path to get there.

The bigger issue is what it sits next to. Every request through this path already does `LookupWithWildcard` against the entity store to resolve the route, which is a network round trip. The parse is orders of magnitude below that, so removing it does not move the number an operator would notice.

I did add per-route state right next to this, but for a different reason: the warning on an unusable value needed collapsing, since `routeRequestTimeout` runs per request and a bad value never self-heals, so warning unconditionally would flood the Warn tier for as long as the misconfiguration lasted. That one is about log volume rather than CPU, which is why it earns its lock.

servers/httpingress/httpingress.go
```diff
@@ -189,6 +196,114 @@ func NewServer(
...
+
+// warnBadRouteTimeout warns about an unusable request_timeout once per route,
+// and again only if that route later carries a different unusable value.
+//
+// The warning is worth keeping at Warn — a route that silently ignores its
+// configured timeout is exactly the kind of thing an operator needs to see. But
+// routeRequestTimeout runs on every request through the route, and a bad value
+// never self-heals, so warning unconditionally would emit the same line for as
+// long as the misconfiguration lasts and drown the tier it sits in.
+func (h *Server) warnBadRouteTimeout(route *ingress_v1alpha.HttpRoute, msg string, extra ...any) {
```
phinze: ✂️ I'd cut this and match what we do for `resolvePortWaitTimeout`

---

🤖 Agreed, and there's a house pattern to match. `resolvePortWaitTimeout` (`controllers/sandbox/sandbox.go:2158`) is the entire read-side story for the same shape of attribute:

```go
func resolvePortWaitTimeout(spec string) time.Duration {
	if spec == "" {
		return defaultPortWaitTimeout
	}
	d, err := time.ParseDuration(spec)
	if err != nil || d <= 0 {
		return defaultPortWaitTimeout
	}
	return d
}
```

`getShutdownTimeout` (`sandbox.go:2935`) does the same. The other duration-string attributes (`api/run/schema.yml:85`, `api/compute/schema.yml:26`, `api/core/schema.yml:273`) all lean on write-boundary validation the way `SetRouteRequestTimeout` does here — the only writers that skip it are the `debug entity` commands. So six lines of silent fallback is right-sized insurance; 45 lines plus a mutex on a per-request path isn't.

One more reason to cut rather than shrink it: `badTimeouts` is process-local state describing a cluster-global condition. Under MIR-904, N ingress instances means N warnings for one misconfigured route, with nothing in the logs to say it's one problem — the dedup gets weaker exactly as we scale out. If we do want visibility here, a metric aggregates correctly by construction.

(If it stays: entries are never removed, so the 1024 slots fill with historical bad routes and `httpingress.go:289` then goes quiet on genuinely new ones.)

servers/httpingress/httpingress.go
```diff
@@ -69,6 +69,11 @@ const (
 	// apps from having their leases evicted on every 30s tick, which would
 	// force every request through the full entity store + activator pipeline.
 	minLeaseTTL = 5 * time.Minute
+	// maxBadTimeoutsTracked caps the per-route memory of which unusable
+	// request_timeout we have already warned about. Past the cap we stop
+	// warning rather than start flooding — a config error that repeats forever
+	// is worth one line, never a stream of them.
+	maxBadTimeoutsTracked = 1024
```
phinze: Micro-nit here, stale comment just above the diff line.

---

🤖 `leaseAcquisitionTimeout` at line 66, just above this hunk:

```go
// leaseAcquisitionTimeout is the maximum time to wait for sandbox boot
// This is longer than request timeout to prevent dangling resources
leaseAcquisitionTimeout = 2 * time.Minute
```

True when the only request timeout was the 60s default, false once a route can carry 10m. Behavior is unchanged — lease acquisition always ran on its own `context.Background()` budget and never consulted the request timeout — so it's only the comment that needs the edit.

servers/httpingress/httpingress.go
```diff
@@ -189,6 +196,114 @@ func NewServer(
...
+	}
+
+	h.transportMu.Lock()
+	defer h.transportMu.Unlock()
+
+	if t, ok := h.transports[timeout]; ok {
+		return t
+	}
+
+	// Tests build a Server directly and skip this map; writing to a nil one
```
phinze: Dead code maybe?

---

🤖 Checked every caller. In production `NewServer` always initializes the map (`httpingress.go:173`). The three existing tests that reach `transportFor` via `proxyToLease` all pass `0` for the timeout (`httpingress_test.go:432`, `:462`, `:489`), which returns at the `timeout <= 0` early return on line 224 before the map is touched. And `newTimeoutTestServer` initializes `transports` itself (`route_timeout_test.go:25`).

So the only thing that reaches this branch is `TestTransportForToleratesNilMap`, which exists to test it. Its comment says "Several existing test helpers build a Server without the transports map" — true that they build one, but none of them get here.

Note the sibling guard at line 281 is *not* the same case: `newTimeoutTestServer` leaves `badTimeouts` nil, and `TestRouteRequestTimeout` drives bad values through it, so that one is genuinely load-bearing today.

phinze (approved):
Looking good! As much as it unnerves me to see more _behavior_ in our already bloated `httpingress` I think the multi-transport stuff holds up as the only way we'll be able to reliably support multiple timeout routes to the same backend. 

On the other hand, IMHO the `badTimeouts` stuff doesn't earn its keep. It's a lot of machinery to respond to a malformed timeout we already guard at write. I think "silent fallback to default" is good enough and way less code.

servers/httpingress/httpingress.go
evanphx: 🤖 Agreed on all counts, cut in 55b22b0b. `routeRequestTimeout` is now the same six-line shape as `resolvePortWaitTimeout`: parse, fall back on error or non-positive, no state.

The parenthetical is the part that settled it. You're right that entries are never removed, so once the cap filled with historical bad routes the warning would go quiet on genuinely new ones. That inverts the signal the thing existed to protect, and I built the cap as "safer failure" without noticing. The multi-ingress point compounds it, and agreed a metric is the right tool if we want visibility later.

Net 158 lines removed across the three threads.

servers/httpingress/httpingress.go
evanphx: 🤖 Fixed in 55b22b0b. Confirmed the behavior claim before touching it: lease acquisition builds its own `context.WithTimeout(context.Background(), ...)` and never reads the request timeout, so this was comment-only. Reworded to say what the constant is actually for and to note it's unrelated to the request timeout.

servers/httpingress/httpingress.go
evanphx: 🤖 It was, and the caller analysis checks out. Gone in 55b22b0b along with `TestTransportForToleratesNilMap`.

I added it against my own review note about a future test tripping over the nil map, which is exactly the speculative guard worth cutting: the only thing reaching the branch was the test written to reach it. The sibling guard you flagged as genuinely load-bearing went away too, since it belonged to the `badTimeouts` map that came out in the other thread.

blackbox/route_timeout_test.go
```diff
@@ -0,0 +1,155 @@
...
+	r.RequireSuccess(t)
+
+	var cleared struct {
+		RequestTimeout string `json:"request_timeout"`
+	}
+	if err := json.Unmarshal([]byte(r.Stdout), &cleared); err != nil {
+		t.Fatalf("failed to parse route timeout JSON: %v (stdout: %s)", err, r.Stdout)
+	}
+	if cleared.RequestTimeout != "" {
+		t.Errorf("expected request_timeout to be cleared, got %q", cleared.RequestTimeout)
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch, fixed in 55b22b0b. Both clear assertions now decode into a `map[string]json.RawMessage` and require the key to be absent, since the field carries `omitempty` and an omitted field, `""` and `null` all decode to the same empty string. Verified against a dev cluster, so the CLI is genuinely omitting it rather than emitting an empty value.

servers/httpingress/httpingress.go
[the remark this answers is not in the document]
evanphx: 🤖 Following up to close this out. We went the other direction on this path in 55b22b0b: @phinze asked us to cut the per-route warning state entirely and match `resolvePortWaitTimeout`, which is the house pattern for the same shape of attribute. `routeRequestTimeout` is now a bare parse with a silent fallback and no state at all, so a parse cache would be the only shared mutable state left on the request path. Sticking with the parse.

evanphx:
Picking up the outside-diff-range finding from the latest CodeRabbit pass, on HTTP/2 stream multiplexing vs `idleTimeoutConn` (`httpingress.go:194-205`).

The mechanism described is real in general: `idleTimeoutConn` sets a deadline on the underlying connection, so if many HTTP/2 streams shared one TCP connection, an active stream would keep resetting the deadline while another stalled body went unnoticed.

It does not apply here, for two reasons.

HTTP/2 never engages on this transport. Sandbox lease URLs are built by `netutil.BuildHTTPURL`, which returns `fmt.Sprintf("http://%s", hostPort)` (`pkg/netutil/netutil.go:36`), so every proxied request is cleartext. Go negotiates HTTP/2 through TLS ALPN; `ForceAttemptHTTP2` governs whether that negotiation is allowed once a custom `DialContext` is set, and it has no effect on `http://` requests. Cleartext h2c requires an explicit opt-in this code does not make. One connection therefore carries one request at a time and the deadline is unambiguous.

It is also not something this PR changes. `idleTimeoutConn` and the transport construction predate it; `newProxyTransport` is the same code lifted out of `NewServer` so the timeout could be parameterized per route. Whatever is true of h2 here was equally true before.

Skipping it on that basis. Worth revisiting if we ever terminate TLS to sandboxes or turn on h2c, at which point the per-response idle reader would be the right shape.

PR #1058, opened 2026-08-13
Let `miren sandbox exec` pick an app's sandbox for you

evanphx:
Getting a shell in a running instance had an annoying preamble. You'd run `miren sandbox list`, scan the table for a row belonging to your app that was actually RUNNING, copy the ID, and only then get to the thing you wanted to do. That's a lookup the CLI has everything it needs to do itself, so now it does: `miren sandbox exec --app myapp -- ls -la` picks one for you.

The selection happens entirely in the CLI. It picks an ID and hands it to the exec RPC through the path that already existed, so there's no server or protocol change in here at all. The proxy still resolves the target's app from the sandbox itself and re-checks access, which is the guard that was actually load-bearing.

The interesting decision was how a sandbox gets connected back to its app. The obvious route is `spec.version` to the app version to the app, which is what `sandbox list` uses for its APP column. I went through the pool instead, since the pool carries the indexed app ref and, more importantly, ephemeral sandboxes don't have one. `miren app run` creates those for console sessions, and picking one would drop you into somebody else's shell.

Among what's left, service is a requirement and version is a preference, in that order, and the asymmetry is the interesting part. Which service you get is the difference between a web process and a background worker, so substituting one for the other is never a kindness: without `--service` you get web, and if web has nothing running you get told that rather than quietly handed a worker. Which version you get is the difference between newer and older code for the same job, where insisting is what hurts, since a failed deploy leaves the current version pointing at something that never came up while the previous instances keep serving traffic. Refusing those would lock you out of a shell exactly when a deploy has just broken.

That ordering matters more than it sounds. Web has to bind its declared port to go ready and workers don't, so a rolling deploy routinely has the worker on the new version while web is still on the old one. Taking version first would hand you a worker there, and silently, since only one candidate survives and the "which one did I pick" notice stays quiet at one. Taking service first gives you old web instead. An app with no web service at all is exempt, since there's nothing to substitute. Whatever survives, we pick from at random, and name the choice on stderr when there was more than one to choose from (on a terminal only, since stdout there belongs to the sandbox).

`sandbox list` picked up matching `--app` and `--service` flags along the way. That started as a nicety and turned out to be load-bearing for the error messages: when exec can't find anything running, "go look at `miren sandbox list --app foo`" is a useful thing to say, and "go look at every sandbox on the cluster" isn't. Those filters match the values shown in the APP and SERVICE columns, so what you filter on is what you see.

Closes MIR-1577

### DIFF
--- a/cli/commands/sandbox_exec.go
+++ b/cli/commands/sandbox_exec.go
@@ -6,19 +6,84 @@ import (
 
 	"miren.dev/runtime/api/exec/exec_v1alpha"
 	"miren.dev/runtime/pkg/rpc/stream"
+	"miren.dev/runtime/pkg/ui"
 )
 
-func SandboxExec(ctx *Context, opts struct {
+// sandboxExecOptions is a named type rather than an inline struct so the flag
+// parsing, which now has rules of its own, can be tested directly.
+//
+// App deliberately has no env:"MIREN_APP" tag. Inside a sandbox that variable
+// is injected as the running app's own name, and picking it up here would
+// silently switch the command into --app mode and reinterpret the first
+// positional as part of the command instead of a sandbox id.
+type sandboxExecOptions struct {
 	ConfigCentric
-	Id string `short:"i" long:"id" description:"Sandbox ID"`
+
+	Id      string `short:"i" long:"id" description:"Sandbox ID"`
+	App     string `short:"a" long:"app" description:"Pick a running sandbox belonging to this app"`
+	Service string `long:"service" description:"With --app, restrict the choice to one service (default: web)"`
 
 	Args []string `rest:"true"`
-}) error {
+}
+
+// validate covers the flag combinations mflags can't express. It is called from
+// the command body rather than through OptsValidate, which RunCommand skips.
+func (o sandboxExecOptions) validate() error {
+	if o.App != "" && o.Id != "" {
+		return fmt.Errorf("--app and --id both name a sandbox; pass one or the other")
+	}
+
+	if o.Service != "" && o.App == "" {
+		return fmt.Errorf("--service selects among an app's sandboxes; it needs --app")
+	}
+
+	return nil
+}
+
+// execChoiceNotice describes where --app landed the caller, or returns "" when
+// there was nothing to report. One qualifying sandbox means we didn't choose
+// anything — naming it would be noise in front of every single-instance app.
+func execChoiceNotice(chosen sandboxCandidate, considered int) string {
+	if considered <= 1 {
+		return ""
+	}
+
+	return fmt.Sprintf("%s (%s), picked from %d running", chosen.Brief, chosen.Service, considered)
+}
+
+func SandboxExec(ctx *Context, opts sandboxExecOptions) error {
+	if err := opts.validate(); err != nil {
+		return err
+	}
+
 	id := opts.Id
 	args := opts.Args
-	if id == "" {
+
+	// Non-nil only when we picked the sandbox ourselves, which changes how a
+	// failure downstream has to be explained.
+	var chosen *sandboxCandidate
+
+	switch {
+	case opts.App != "":
+		// Every positional is part of the command here: there is no sandbox id
+		// to consume, because that's the whole point of --app.
+		pick, considered, err := selectAppSandbox(ctx, opts.App, opts.Service)
+		if err != nil {
+			return err
+		}
+		chosen = &pick
+
+		// Only to a terminal: stdout belongs to the sandbox, and plenty of
+		// callers merge stderr into it. This has to happen before setupExecIO
+		// puts the terminal into raw mode.
+		if msg := execChoiceNotice(pick, considered); msg != "" && isInteractiveOutput(ctx.Stderr) {
+			ctx.Begin("%s", msg)
+		}
+
+		id = pick.ID.String()
+	case id == "":
 		if len(args) == 0 {
-			return fmt.Errorf("sandbox ID is required (pass as first positional arg or via --id)")
+			return fmt.Errorf("sandbox ID is required (pass as first positional arg, via --id, or use --app)")
 		}
 		id, args = args[0], args[1:]
 	}
@@ -46,9 +111,34 @@ func SandboxExec(ctx *Context, opts struct {
 		stream.ChanReader(winUpdates),
 	)
 	if err != nil {
+		if chosen != nil {
+			return chosenSandboxExecError(opts.App, *chosen, err)
+		}
 		return err
 	}
 
 	ctx.SetExitCode(int(res.Code()))
 	return nil
+}
+
+// chosenSandboxExecError explains a failed exec against a sandbox the caller
+// never named. Selecting and connecting are separate round trips, so the
+// sandbox we picked can be gone by the time exec looks for it — scaled down,
+// crashed, or reaped along with its pool by a deploy. Left unwrapped, the
+// caller gets an entity id they have no way to connect to the -a they typed.
+//
+// This deliberately doesn't try to tell "it vanished" from any other failure:
+// the cause is shown either way, and retrying is the right first move for both.
+func chosenSandboxExecError(appName string, chosen sandboxCandidate, cause error) error {
+	return &ui.Diagnostic{
+		Summary: fmt.Sprintf("couldn't exec into the sandbox picked for %q", appName),
+		Detail: fmt.Sprintf("We picked %s (%s) from the sandboxes running at the time. "+
+			"If it has gone away since, running the same command again picks from "+
+			"whatever is running now.", chosen.Brief, chosen.Service),
+		Actions: []ui.Action{
+			{Command: "miren sandbox list --app " + appName, Note: "see what's running for it"},
+		},
+		Cause:     cause,
+		ShowCause: true,
+	}
 }
--- a/cli/commands/sandbox_list.go
+++ b/cli/commands/sandbox_list.go
@@ -12,12 +12,50 @@ import (
 	"miren.dev/runtime/pkg/ui"
 )
 
+// sandboxFilter narrows a sandbox listing to one app and/or service. Both
+// match against the values shown in the APP and SERVICE columns, so what you
+// filter on is what you see — including the "-" cases, which simply match
+// nothing rather than matching everything.
+type sandboxFilter struct {
+	app     string
+	service string
+}
+
+func (f sandboxFilter) keep(app, service string) bool {
+	if f.app != "" && app != f.app {
+		return false
+	}
+
+	if f.service != "" && service != f.service {
+		return false
+	}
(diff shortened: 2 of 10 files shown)


cli/commands/sandbox_select.go
```diff
@@ -0,0 +1,242 @@
...
+//
+// The result is sorted so that the only nondeterminism left is the caller's
+// random pick.
+func narrowSandboxCandidates(candidates []sandboxCandidate, activeVersion entity.Id, service string) []sandboxCandidate {
+	candidates = slices.Clone(candidates)
+
+	// A deploy leaves the previous version's pool running until the new one is
+	// healthy, so without this a pick during a rollout can drop you into the
+	// code you just replaced.
+	if !entity.Empty(activeVersion) {
```
phinze: Here's the bit I think we should flip. Allowing for a silent web->worker switcheroo seems like the wrong tradeoff to make. IMHO better to throw up our hands and ask user to try again.

---

🤖 For an app with `web` and `worker`:

```
# normal times — correct, unchanged
$ miren sandbox exec -a myapp -- hostname
myapp-web-8c1d

# while a deploy is rolling
$ miren sandbox exec -a myapp -- hostname
myapp-worker-4f2a          ← a worker, and nothing on stderr said so
```

The two preferences run version-first, so when web is still on the old version and worker has already rolled, the version rule keeps only the worker and the web rule has no web left to prefer. The "which one did I pick" notice stays quiet too, since it only speaks when more than one candidate survived and by then exactly one has.

Not a freak race, either: web has to bind its declared port to go ready and workers don't, so worker-first is ordinary.

**The behavior I'd want:** when there's no web instance to give you, say that, rather than substituting a different service.

```
$ miren sandbox exec -a myapp -- hostname
error: myapp has no web instance on its current version

  usually a deploy in progress — retrying may find one
  or pick a service: miren sandbox exec -a myapp --service worker -- ...
```

One trap worth avoiding: this can't become "refuse anything on an old version." A *failed* deploy leaves `ActiveVersion` pointing at a version that never came up while the old instances keep serving traffic. Those are production, and refusing them would lock you out of a shell exactly when a deploy just broke.

`sandbox_select_test.go:95` pins the current precedence as `"version wins over service"`, so that case flips rather than gets added.

cli/commands/sandbox_exec.go
```diff
@@ -46,9 +111,34 @@ func SandboxExec(ctx *Context, opts struct {
...
+// This deliberately doesn't try to tell "it vanished" from any other failure:
+// the cause is shown either way, and retrying is the right first move for both.
+func chosenSandboxExecError(appName string, chosen sandboxCandidate, cause error) error {
+	return &ui.Diagnostic{
+		Summary: fmt.Sprintf("couldn't exec into the sandbox picked for %q", appName),
+		Detail: fmt.Sprintf("We picked %s (%s) from the sandboxes running at the time. "+
+			"If it has gone away since, running the same command again picks from "+
+			"whatever is running now.", chosen.Brief, chosen.Service),
+		Actions: []ui.Action{
+			{Command: "miren sandbox list", Note: "see what's running"},
```
phinze: Nit. Might as well suggest the app scoped one since we have the app name right here.

---

🤖 `noRunningSandboxError` gets this right at line 238 with `miren sandbox list --all --app <app>`. `appName` is already a parameter here, so it's just the interpolation.

cli/commands/sandbox_list.go
```diff
@@ -122,6 +160,18 @@ func SandboxList(ctx *Context, opts struct {
...
 
+			// Extract pool label from metadata
+			var md core_v1alpha.Metadata
+			md.Decode(e.Entity())
+			poolLabel, _ := md.Labels.Get("pool")
+
+			// Get service from pool
+			service := poolServiceMap[poolLabel]
+
+			if !filter.keep(ui.CleanEntityID(versionAppMap[sandbox.Spec.Version.String()]), service) {
```
phinze, line 171: Nit. Seems good to make the exec behavior and the list output make sense together. Agree it's maybe simpler to just clarify the error message.

---

🤖 Confirmed the divergence is real: `controllers/run/controller.go:412-423` creates ephemeral run sandboxes with `app`/`run`/`task` labels and **no `pool` label**, while `pkg/appspec/appspec.go:124` sets `Spec.Version` on them anyway. So `exec --app` (pool-keyed) skips them and `list --app` (version-keyed) includes them.

Where it bites is the handoff between them: `noRunningSandboxError` says "nothing is running" and sends you to `sandbox list --all --app foo`, which will show a RUNNING sandbox if you have an `app run` console open on a scaled-to-zero app.

CodeRabbit's proposed fix (blank the APP column for unpooled sandboxes) trades away the only thing telling you which app a console sandbox belongs to, in exchange for filter tidiness. If this is worth fixing, the error text seems like the cheaper side to fix.

phinze (approved):

phinze:
Fighting GitHub API/UI inconsistencies w/ top level comments in reviews... I wrote this comment on the main PR thread and the review submission clobbered it?! Anyways, belongs at top of review:

---

Great behavior improvement here. It does open a small can-o-worms on mid-deploy behavior races (which was already there for regular `sandbox exec` just didn't require us to make any calls about it). I think the selection logic could be made a little safer in this version, which is the first comment.

Noodling even-better UX, all follow-up shaped:

 * Run `sandbox exec --app` mid deploy? CLI say "oh sorry, hold plz, new one coming up" then hands it to you
 * Run `sandbox exec --app` and catch a sandbox that gets marked for doom soon after? We notify you on stderr, gracefully shutdown the shell.

cli/commands/sandbox_list.go
```diff
@@ -122,6 +160,18 @@ func SandboxList(ctx *Context, opts struct {
...
+				continue
+			}
```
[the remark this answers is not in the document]
evanphx: You're right about the behavior: `sandbox list --app` derives the app from `Spec.Version`, so it does include ephemeral `app run` sandboxes, and `sandbox exec --app` doesn't. Leaving that as-is deliberately, though.

The pool route matters in `exec` for a specific reason: picking an ephemeral sandbox would drop you into somebody else's live console session. That's an argument about *selection*, not about *display*. Listing an ephemeral sandbox harms nobody, and knowing which app it belongs to is exactly what an operator wants when they spot an unfamiliar sandbox.

Moving the APP column and the JSON `app` field onto the pool map would mean ephemeral sandboxes render as `-` in the table and carry an empty `app` in JSON. That's a loss of information the listing currently provides, and a behavior change for anyone reading that field.

The filter is deliberately defined as "match what the APP and SERVICE columns show," so that filtering never hides a row whose displayed app is the one you asked for. Changing the filter without changing the column would break that; changing both costs the information above.

Worth noting the two views already reconcile visually: an ephemeral sandbox shows `-` for both POOL and SERVICE, since it has no pool. So a row that `exec` would never pick is distinguishable in the listing, and `--service web` excludes them outright.

cli/commands/sandbox_select.go
evanphx: Good catch, and you're right that this is the wrong tradeoff. Flipped it.

Service is now a requirement and version is a preference, in that order. Which service you get is the difference between a web process and a background worker, so substituting is never a kindness. Which version you get is the difference between newer and older code for the same job, where insisting is what hurts, for exactly the failed-deploy reason you flagged.

That ordering also handles your rolling-deploy example better than erroring would: service filters first, so you get old web rather than new worker, and the version rule then finds nothing on the current version and falls back rather than refusing. You only get the error when the web service has genuinely nothing up.

An app with no web service at all stays exempt, since there's no substitution to worry about there.

Staged it in dev with a web+worker app, scaled web to zero:

```
$ miren sandbox exec -a multisvc -- hostname
ERROR: no running web instance for app "multisvc"

  Without --service, exec goes to the web service, and none of its sandboxes are
  up. This is usually a deploy in progress, in which case trying again shortly
  will find one. Still running: worker.

  Try
    miren sandbox exec -a multisvc --service worker    exec into that service instead
    miren sandbox list --all --app multisvc            see this app's sandboxes and their state
```

Exits 1. Scaling web back up returns the default to web.

`sandbox_select_test.go` flips as you predicted: the old `version wins over service` case is now `a rolling deploy gives you old web, never a new worker`, with your readiness reasoning in the comment. Added two more for the substitution refusal and the failed-deploy case, so the asymmetry can't get "tidied" back into symmetry later.

cli/commands/sandbox_exec.go
evanphx: Done, it now interpolates the app name the same way `noRunningSandboxError` does.

cli/commands/sandbox_list.go
evanphx: Went with clarifying the error text, agreeing with you over CodeRabbit for the reason you both landed on: blanking the APP column trades away the only thing telling you which app a console sandbox belongs to.

Thanks for digging up `controllers/run/controller.go` too. I only knew about the `app run` path, so I'd have written a caveat that quietly missed task runs. `noRunningSandboxError` now names both:

> Nothing is running and placed on a node right now. A sandbox that is still starting up doesn't count, since exec needs one that's already up. The listing below can still show a sandbox here: consoles from `miren app run` and task runs aren't part of a service, so exec won't pick one.

cli/commands/sandbox_select_test.go
```diff
@@ -0,0 +1,301 @@
...
+			// version first would hand over a worker, silently, since only one
+			// candidate survives and the notice stays quiet at one.
+			name: "a rolling deploy gives you old web, never a new worker",
+			candidates: []sandboxCandidate{
+				{ID: "sandbox/a", Service: "web", Version: old},
+				{ID: "sandbox/b", Service: "worker", Version: active},
+			},
+			activeVersion: active,
+			appHasWeb:     true,
+			want:          []string{"sandbox/a"},
```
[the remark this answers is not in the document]
evanphx: Keeping the test as it is, and I've fixed what led you here: the PR description was stale. It still described the original version-first ordering, so you read the code as contradicting the contract when it was the contract that was out of date. Updated now.

The precedence was deliberately flipped during review, because version-first has a bug. Web has to bind its declared port to go ready and workers don't, so a rolling deploy routinely has worker on the new version while web is still on the old one. Taking version first keeps only the worker, the web preference then has no web left to prefer, and you get a shell in a background worker while asking for your app. Silently, too, since exactly one candidate survives and the notice only speaks when more than one does.

The asymmetry between the two rules is the point rather than an oversight. Service is a requirement because web-versus-worker is a difference in kind. Version is only a preference because a failed deploy leaves ActiveVersion pointing at a version that never came up while the old instances keep serving traffic, and refusing those would lock you out of a shell exactly when a deploy has just broken.

Applying the web preference "only among remaining active-version candidates" as suggested is the version-first ordering under another name, and reintroduces the same substitution.

PR #1059, opened 2026-08-14
Report app state as snapshots that can sweep

phinze:
Another step in getting a real sync protocol running over our uplink! We spent a lot of time looking at the scaling properties of the protocol here: one cluster many apps, many clusters checking into cloud, that kind of thing. That's what landed the snapshot batching and startup jitter into the mix, as well as the follow up issues to think through other aspects of this uplink protocol scaling up.

If "epoch" and "mark-and-sweep" read as jargon: an epoch is which snapshot a message belongs to, and mark-and-sweep is the GC trick, mark what the snapshot mentions and delete what it doesn't.

There's deliberately no delete message anywhere, because that's the one message whose loss you can never recover from. (Another way to design around this would be to change `app delete` to record tombstones but that would be its own can-o-worms.)

---

🤖 The reporter sent one snapshot when the uplink connected and then went quiet, so cloud's picture aged until the next reconnect and it never learned that an app had been deleted. Deletes are the hard half. A delete message can be dropped, and unlike a stale health value its loss is permanent, so nothing later repairs it. An app's absence from a completed snapshot proves the deletion instead.

Here's the conversation that produces, with the two pre-existing exchanges at the top for context:

```mermaid
sequenceDiagram
    autonumber
    participant R as runtime
    participant C as cloud

    R->>C: WebSocket upgrade (bearer token)
    C-->>R: 101 Switching Protocols
    Note over C: token resolves to clusterXID,<br/>the only identity that counts

    R->>C: time.request (t1)
    C-->>R: time.response (t1, t2, t3)
    R->>C: org.info.request
    C-->>R: org.info.response (organization_id)

    Note over R: wait a random slice of 60s,<br/>so the fleet does not report in unison

    R->>C: app.snapshot (epoch, observed_at, up to 100 apps)
    R->>C: app.snapshot (epoch, observed_at, remaining apps)
    R->>C: app.snapshot.complete (epoch, app_count)
    Note over C: marks == app_count ? sweep absent apps<br/>: create only, sweep nothing

    loop every 15s, only when something moved
        R->>C: app.status (observed_at, changed apps)
    end

    Note over R: an app disappeared,<br/>or the 15 minute floor expired
    R->>C: app.snapshot ... complete (fresh epoch)
```

The epoch is what lets cloud recognize a snapshot as a whole. Batches carry it, the closing message carries the count of what was sent, and cloud sweeps only when the two agree. A snapshot interrupted by a disconnect never sends its closing message and so sweeps nothing, which is what lets the design skip timeouts and partial-apply logic entirely: the missing message is already the safe state. The count covers the case the epoch can't, where a batch dropped on outbox overflow looks exactly like the apps in it having been deleted.

Between snapshots the reporter samples every fifteen seconds and sends only what moved, so an idle cluster stays quiet. A removal can't ride a delta, so it promotes that tick to a full snapshot rather than waiting out the floor. That's what lets the floor sit at fifteen minutes: it's a backstop for a lost sample store rather than the main repair path. Instance counts come along because they were already on `AppInfo` and cost nothing.

The first rev is the uplink rather than the reporter, because two other things had to be true first. `Send` drops on outbox overflow, which is right for a lone message whose loss self-heals and wrong for a sequence, so streaming tenants get a blocking send. And reconnect backoff had no jitter, which matters much more now that every cluster opens a snapshot the moment it connects: clusters don't disconnect independently, so the fleet was marching back in lockstep. MIR-1560 wants both, which is why they're separated out.

One thing worth flagging for review: the sixty-second connect spread is chosen generously rather than measured. Widening it later means shipping a release to every cluster in the fleet, so the cost of overshooting is a slightly later first snapshot and the cost of undershooting is a stampede we can't easily fix. MIR-1587 covers handing that number to cloud instead.

Part of MIR-1559.

### DIFF
--- a/components/coordinate/appreport.go
+++ b/components/coordinate/appreport.go
@@ -2,37 +2,103 @@ package coordinate
 
 import (
 	"context"
+	"crypto/rand"
 	"time"
 
+	"github.com/oklog/ulid/v2"
+
+	app_v1alpha "miren.dev/runtime/api/app/app_v1alpha"
 	"miren.dev/runtime/pkg/uplink"
 )
 
-// TypeAppSnapshot carries the cluster's full current app set to cloud for
-// visibility reporting, one of the app.* family described in RFD-94.
+// The app.* message family carries the cluster's current app state to cloud
+// for visibility reporting, as described in RFD-94.
 //
-// The type and its payload live with the reporter rather than in pkg/uplink
-// because the link is deliberately payload-agnostic: it moves opaque envelopes
-// and understands no apps. A tenant owns its own wire shape.
-const TypeAppSnapshot = "app.snapshot"
+// These types and their payloads live with the reporter rather than in
+// pkg/uplink because the link is deliberately payload-agnostic: it moves
+// opaque envelopes and understands no apps. A tenant owns its own wire shape.
+const (
+	// TypeAppSnapshot carries one batch of the cluster's full app set.
+	TypeAppSnapshot = "app.snapshot"
 
-// AppSnapshot is the cluster's view of every app it currently runs, sent on
-// each (re)connect. Cloud treats it as ground truth for that cluster.
+	// TypeAppSnapshotComplete closes an epoch and is what authorizes cloud to
+	// sweep. It is a separate message rather than a flag on the final batch so
+	// that a cluster running no apps can still close an epoch, which is the
+	// case where sweeping matters most: every app was deleted.
+	TypeAppSnapshotComplete = "app.snapshot.complete"
+
+	// TypeAppStatus carries one app whose sample changed between snapshots.
+	TypeAppStatus = "app.status"
+)
+
+const (
+	// snapshotBatchSize bounds one message so a large cluster's snapshot
+	// arrives as a slide rather than a jump (RFD-94). Reporting must never take
+	// meaningful horsepower from the cluster's real work.
+	snapshotBatchSize = 100
+
+	// batchPause spaces batches out for the same reason, and leaves the link
+	// available to other tenants during the burst that follows a reconnect.
+	batchPause = 100 * time.Millisecond
+
+	// sampleInterval is how often the reporter re-derives app state and emits
+	// deltas for whatever moved. It sets typical freshness.
+	sampleInterval = 15 * time.Second
+
+	// snapshotFloor is the safety net, not the main repair path, which is why
+	// it is this loose.
+	//
+	// It once had three jobs and is down to one. Deltas are sent with the
+	// blocking send, so they are not silently dropped; if a connection dies
+	// mid-send, the reconnect opens a fresh snapshot anyway. Deletes no longer
+	// wait for it either — a removal triggers a snapshot on the next sample.
+	// What remains is a store that lost its samples without the connection
+	// dropping, which on cloud's disk-backed Valkey survives an ordinary
+	// restart and so is genuinely exceptional.
+	//
+	// The cost of a longer floor is how long that exceptional case shows apps
+	// as awaiting a report. The cost of a shorter one is re-sending unchanged
+	// state across the whole fleet forever. For a rare event, the trade favors
+	// the fleet.
+	snapshotFloor = 15 * time.Minute
+
+	// connectSpread is how far the opening snapshot is scattered across the
+	// fleet.
+	//
+	// Clusters reconnect for reasons that are the same for all of them at once
+	// — a cloud deploy, a load balancer rotation — so without this every
+	// cluster's opening snapshot lands on cloud in the same instant. Jittering
+	// the reconnect alone would only move that spike, since the snapshot
+	// follows the connect immediately.
+	//
+	// Chosen generously rather than tightly, because the cost is asymmetric.
+	// This constant ships inside runtimes customers deploy, so widening it later
+	// means getting a release onto every cluster in the fleet, and the fleets
+	// that would most need it are the slowest to upgrade. Overshooting costs a
+	// slightly later first full picture and nothing else, since this is the
+	// ephemeral tier and deltas keep flowing during the wait.
+	connectSpread = 60 * time.Second
+)
+
+// AppSnapshot is one batch of the cluster's view of every app it currently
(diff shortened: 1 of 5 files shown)


components/coordinate/appreport.go
```diff
@@ -56,63 +161,264 @@ type AppState struct {
...
+}
+
+// sendAppSnapshot streams a full epoch: every app in bounded batches, then the
+// complete message that authorizes cloud to sweep.
+//
+// Batches use the blocking send. A snapshot is a sequence, and a silently
+// dropped batch in a sequence does not look like a lost sample that the next
+// report repairs — it looks exactly like those apps having been deleted.
+func (c *Coordinator) sendAppSnapshot(ctx context.Context, link reportLink, state map[string]AppState) error {
+	epoch := ulid.MustNew(ulid.Now(), rand.Reader).String()
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, fixed. Switched to `ulid.New` with the error logged and returned.

Your point about the goroutine is what made this worth doing rather than shrugging at: reporting runs on its own goroutine, and the file states that nothing the cluster does for its own users may degrade because reporting failed. A panic there takes the whole process down, which is about the loudest possible violation of that. An entropy failure is close to impossible on a real host, which is exactly what would make it an absurd way to lose a runtime.

components/coordinate/appreport.go
```diff
@@ -56,63 +161,264 @@ type AppState struct {
...
+	observedAt := time.Now().UTC()
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed, along with both nitpicks from the review body.

`ulid.New` now returns the error rather than panicking. The delta log dropped to Debug, since an operator cannot act on the fact that three apps changed and the snapshot log already marks the transitions worth reading.

On the `jittered` doc comment: you were right that it described the wrong symbol, and rather than delete the paragraph I moved it onto `SpreadOnConnect`, which is what it was actually explaining. The rationale is worth keeping, it was just attached to its neighbor after a refactor.

components/coordinate/appreport.go
```diff
@@ -56,63 +161,278 @@ type AppState struct {
...
+			return
+		case <-ticker.C:
+		}
+
+		state, err := c.deriveAppState(ctx)
+		if err != nil {
+			continue
+		}
+
+		if lastSent == nil || removedAny(lastSent, state) || time.Since(lastSnapshot) >= snapshotFloor {
```
evanphx: Non-blocking, and maybe a follow-up rather than this PR: this three-way decision is the one piece no test drives directly — `sampleInterval` and `snapshotFloor` being consts means a test would have to wait real minutes. Everything it calls is covered well, so the specific gap is "a failed snapshot leaves `lastSent` alone and retries" and "a removal promotes the tick end to end." Happy to file a ticket if you'd rather not touch it here.

components/coordinate/appreport.go
```diff
@@ -56,63 +161,278 @@ type AppState struct {
...
+	}
+	return false
+}
+
+// deriveAppState lists the cluster's apps and reduces them to the sample
+// fields cloud stores, keyed by name.
+func (c *Coordinator) deriveAppState(ctx context.Context) (map[string]AppState, error) {
+	apps, err := c.appInfo.ListApps(ctx)
+	if err != nil {
+		c.Log.Warn("failed to list apps for cloud report", "error", err)
```
evanphx: Non-blocking: if the store stays unhappy this warns every 15s indefinitely, where the old code warned once per connect. CLAUDE.md's logging section calls out the retry-loop-that-can't-converge case specifically. Dropping repeats to Debug after the first, or collapsing with a count the way `certAuthDeduper` does, would keep the signal without the volume.

components/coordinate/appreport.go
```diff
@@ -56,63 +161,278 @@ type AppState struct {
...
+	})
+	if err != nil {
+		// Without the complete message cloud sweeps nothing, which is the safe
+		// outcome: the apps it received still upsert, and the next epoch closes
+		// properly. Worth a warning because deletes stop propagating until then.
+		c.Log.Warn("failed to close app snapshot epoch", "epoch", epoch, "error", err)
+		return err
+	}
+
+	c.Log.Info("sent app snapshot to cloud", "epoch", epoch, "apps", len(state))
```
evanphx: Non-blocking: the code this replaced said "queued app snapshot" with a comment about why the word mattered, and the warn paths just above still say "failed to queue". The blocking send tightened the claim but didn't close it — `drainOutbox` can still throw it away. Worth keeping the old word so someone reading "sent app snapshot" against an empty console isn't misled.

pkg/uplink/client.go
```diff
@@ -186,6 +200,42 @@ func (c *Client) SendMessage(msgType string, data any) error {
...
+// snapshot that is indistinguishable from apps having been deleted.
+//
+// Blocking makes the write loop's drain rate the natural throttle, which is
+// also what lets two streaming tenants share the link without coordinating:
+// neither can starve the other by filling the outbox, because filling it just
+// slows the filler down. Callers must pass the connection-scoped context so a
+// sender parked here is released when the connection drops rather than waking
+// up to write into a socket that is gone.
+func (c *Client) SendBlocking(ctx context.Context, env *Envelope) error {
+	select {
```
evanphx, line 235: Both cases can be ready at once and Go picks at random, so a sender whose connection just died can still land an envelope in the outbox.

I traced how far that gets. Usually the next connect's `drainOutbox` clears it. The case that escapes is `ListApps` straddling the reconnect: the reporter comes back from a slow read with a dead context, and for a cluster under one batch there's no `batchPause` between the final flush and the complete, so the whole abandoned epoch can land on the *new* connection and authorize a sweep against a few-seconds-stale picture. Next epoch repairs it, so genuinely minor.

A `ctx.Err()` check ahead of the select would make the contract deterministic rather than probabilistic — feels worth it for something other tenants will build on.

evanphx (approved):
Approved! Really nicely done.

The 60s connect spread you flagged reads fine to me — right hedge against a stampede.

Four notes inline, all non-blocking. The `SendBlocking` one is the only one with any teeth and even that's narrow.

--e+🤖

PR #1060, opened 2026-08-14
Keep dispatched release rebuilds out of the default branch's cache scope

phinze:
🤖 Code scanning has been sitting on seven high-severity cache-poisoning alerts in `test.yml`, and they all trace back to one line in `release.yml`: the `workflow_dispatch` `ref` input took any git ref and handed it straight to the test workflow, which checked it out and ran `make` and `bun run` against it.

What makes it more than theoretical is that a run's Actions cache scope follows the ref it was dispatched on, not the code it checks out. Dispatching from main with a ref pointing somewhere else runs that code while holding a write token for the default branch's cache scope, which every later run restores from. The entry I'd least like poisoned is `~/go/bin/iso`, a binary the test jobs go on to execute.

That input is worth keeping, since it's the escape hatch we added after a release mispackaged and there was no way to rebuild it, so this narrows it rather than deleting it. Only version tags are accepted now, tests skip on that path, and every job checks out a sha resolved once up front instead of a mutable name. CodeQL only flagged `test.yml`, but `package` and `build-binaries` run the tag's Makefile and the Docker job builds its Dockerfile, so the Go module cache and the buildx layer cache both go read-only on the same path.

Closes MIR-1588

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        description: 'Git ref to build (tag, branch, or SHA). Leave empty to use current ref.'
+        description: 'Version tag to rebuild, e.g. v1.2.3. Leave empty to build the ref this run was dispatched on.'
         required: false
         default: ''
       update_latest:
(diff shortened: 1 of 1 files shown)


phinze:
Good catch, and both of you landed on the same thing independently. Fixed in 642535d.

`cache-to` is now `${{ needs.init.outputs.cache == 'true' && 'type=gha,mode=max' || '' }}`, so a dispatched rebuild builds the tag's Dockerfile without a write token for the layer cache. I left `cache-from` reading: a rebuild consuming a stale layer is the ordinary cache-trust question every push already shares, and gating it only on this path wouldn't change that, so it costs rebuild time for no gain.

I also swept the rest of the release path rather than fixing just the reported line. `setup-go` in `package` and `build-binaries`, plus this `cache-to`, are now the complete set of cache writes in the workflow.

One correction on the `update-homebrew-tap` note, biscuit: it *can* run on a dispatch-with-ref rebuild, since its condition is `github.event_name == 'push' || inputs.update_latest == true` and the second clause is reachable from a dispatch. Right conclusion for a different reason though, because that job never checks out the repo and has no cache of its own, so there's nothing to gate.

phinze:
Thanks both. biscuit's ✅ and the Docker `cache-to` gate are confirmed on 642535d.

**CodeRabbit's `ACTIONS_CACHE_MODE: read` nitpick: declining, though it does check out.** I verified it rather than assuming: the variable is real, it landed in `@actions/cache` 6.2.0, and the string is present in the pinned setup-go v7.0.0's bundled `dist/cache-save/index.js`. So it would work today.

The reason not to take it is where the security property ends up living. `cache: false` is a documented input on the action itself, and it fails closed. `ACTIONS_CACHE_MODE: read` pushes the boundary down into an env var read by a bundled transitive dependency, brand new as of 6.2.0, and its failure mode is silent: if a future setup-go bump vendors an older or renamed toolkit, the var stops being honored, `cache: true` starts saving again, and nothing tells us the protection is gone. What we'd buy for that is a warm module cache on a path that runs a couple of times a year and nobody is waiting on. Bad trade for a boundary this PR exists to establish.

On biscuit's forward-looking note about `test.yml`'s six hardcoded `cache: true`: agreed that they're only safe because the `test` job is skipped wholesale, and that the coupling is implicit. The comment on the skip in `release.yml` names that dependency, so a reader changing the condition should trip over it. Not worth widening the diff further here.

.github/workflows/release.yml
```diff
@@ -66,13 +122,16 @@ jobs:
...
 
     - name: Set up Go
       uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
         go-version-file: 'go.mod'
-        cache: true
+        # `make release-data` below runs the checked-out tag's Makefile, so this
+        # job needs the read-only cache on a dispatched rebuild. CodeQL only
+        # flagged this shape in test.yml, but it is the same shape here.
+        cache: ${{ needs.init.outputs.cache }}
```
evanphx, line 148: setup-go's `cache: false` skips the restore too — it has no read-only mode — so a dispatched rebuild pulls every module fresh. Only the Docker job is genuinely read-only. Doesn't change what happens, but these comments are what the next person reasons from, so worth a word tweak. Same wording at 40-42 and 190.

(CodeRabbit suggested `ACTIONS_CACHE_MODE=read` as an alternative here — I don't think that env var exists.)

.github/workflows/release.yml
```diff
@@ -31,23 +31,79 @@ jobs:
...
+        id: resolve
+        env:
+          DISPATCH_REF: ${{ inputs.ref }}
+          PUSH_REF: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+
+          if [ -z "$DISPATCH_REF" ]; then
+            echo "ref=$PUSH_REF" >> "$GITHUB_OUTPUT"
+            echo "checkout=$PUSH_REF" >> "$GITHUB_OUTPUT"
```
evanphx: The dispatch path below qualifies `refs/tags/` so a branch can't shadow the tag; this one still hands checkout a bare name, so a tag push carries the same hazard. `${{ github.ref }}` is already fully qualified if we want them to match. Non-blocking — takes someone making a branch named `v1.2.3`.

.github/workflows/release.yml
```diff
@@ -292,6 +352,14 @@ jobs:
 
   build-and-push-docker:
     needs: [init, test, package]
+    # `test` is skipped on a dispatched rebuild, and a skipped dependency would
+    # otherwise skip this job too, leaving the rebuild with nothing to publish.
+    # Naming the results keeps a genuine test failure blocking while letting the
+    # deliberate skip through.
+    if: >-
+      !cancelled() &&
```
evanphx, line 375: 👏 easy one to get wrong.

evanphx (approved):
Approved! Really nicely done.

Two non-blocking notes inline, both about comments rather than behavior. One thing I couldn't check from here: whether the seven alerts actually close, since they're anchored in `test.yml` and the PR's CodeQL check only says "no new alerts." The watch job will tell us either way within a day.

--e+🤖

.github/workflows/release.yml
phinze: 🤖 Right on both counts, and the wording is fixed in all three places. The `cache` output, `package`, and `build-binaries` now say the rebuild is kept off the shared cache entirely and resolves modules cold, and the output comment notes the Docker job is the only genuinely read-only one since its `cache-from` survives.

On the parenthetical: `ACTIONS_CACHE_MODE` does exist, which surprised me too. It landed in `@actions/cache` 6.2.0 and the string is present in the pinned setup-go v7.0.0's bundled `dist/cache-save/index.js`, so it would be honored today. I still didn't take it, for the reason in the earlier thread: it moves the boundary into an env var read by a bundled transitive dep, where a future setup-go bump could stop honoring it silently. Worth revisiting if the cold restore ever actually costs us something.

.github/workflows/release.yml
phinze: 🤖 Taken, thanks. The push path now passes `${{ github.ref }}` through a `PUSH_CHECKOUT` env, so both branches of that `if` emit a fully qualified ref and neither can be shadowed by a same-named ref of the other kind.

.github/workflows/release.yml
```diff
@@ -31,23 +31,86 @@ jobs:
...
+          # newline cannot satisfy this and go on to inject extra step outputs.
+          if [[ ! "$DISPATCH_REF" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.]+)?$ ]]; then
+            echo "::error::Dispatch ref must be a version tag such as v1.2.3 or v1.2.3-rc1, got '$DISPATCH_REF'. To build a branch, dispatch the workflow on that branch and leave this input empty."
+            exit 1
+          fi
+
+          echo "ref=$DISPATCH_REF" >> "$GITHUB_OUTPUT"
+          # Fully qualified so a branch sharing the tag's name cannot shadow it.
+          echo "checkout=refs/tags/$DISPATCH_REF" >> "$GITHUB_OUTPUT"
+          echo "cache=false" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, taken. The regex now mirrors `pkg/release/semver.go`:

```
^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$
```

That fixes it in both directions. It now accepts hyphens inside a prerelease identifier (`v1.2.3-rc-1` would previously have been refused a rebuild despite being a perfectly valid tag here), and it now rejects the empty-identifier forms `pkg/release/semver_test.go` already treats as invalid: `v1.0.0-test..1`, `v1.0.0-.test`, `v1.0.0-test.`.

Worth noting the second half is a small security win as well as a correctness one. Requiring every identifier to be non-empty means the value can't contain `..` at all, so `refs/tags/$DISPATCH_REF` has no traversal shape available to it even before git rejects the ref.

Checked against 17 cases: the repo's real tags (`v0.0.0-test.1`, `v0.13.0`), the three fixtures `semver_test.go` rejects, and the adversarial set (`main`, `release/1.2.3`, embedded `;`, path traversal, trailing space, embedded newline). All behave as intended.

PR #1061, opened 2026-08-14
Report the deploy log to cloud with a resumable watermark

phinze:
🤖 Deploys are the durable half of RFD-94's contract, and they behave nothing like app state. A dropped health sample self-heals on the next report; a dropped deploy is gone, because the runtime prunes its own records and cloud becomes the book of record once it does. So completeness here has to come from reconciliation rather than from delivery.

The shape that falls out: on each connection the runtime asks cloud where its copy left off, then decides for itself whether that answer is usable. Three of the four reasons to distrust a cursor (compacted away, above the store's head, incoherent) are only visible from the cluster side, and the fourth is just "cloud has none", which arrives as a zero. Putting the decision here keeps all four in one function instead of splitting them across the wire. A usable cursor means the watch from that revision is simultaneously the catch-up and the live tail, so there is no second stream to run. An unusable one costs an oldest-first re-list and nothing else.

What the runtime never gets to do is tell cloud where the frontier is. The link drops on outbox overflow, so a "watermark is now W" message surviving while the batch before it died would move cloud past deploys it never received, and nothing ever looks below the frontier again. Instead a batch carries the range it covers, which is the claim that nothing in that range was omitted. Cloud can't verify that claim, since revisions are store-global and a legitimate gap between two deploys looks exactly like a missing one, but it can refuse to act on half a message. A batch that never arrives leaves the frontier where it was and gets re-sent next connect.

That is also why this is `deploy.batch` rather than RFD-94's sketched `deploy.upsert`. A message named for the write invites a second sender that upserts without a range, and that sender advancing the frontier is precisely the permanent gap the design exists to prevent.

One find worth flagging, since it's load-bearing: `deployed_at` is cloud's partition key and part of the upsert's conflict target, so it has to be identical in every report of a deploy or the same deploy writes two rows. Records written before the server owned the deploy lifecycle can carry an empty timestamp, and the backfill reaches back far enough to meet them. Entity ids turn out to be base58 UUIDv7, so creation time is recoverable from the identity itself, which is the one other immutable thing on the record. That decoding went into `idgen` rather than the reporter, since that package owns the format.

Two things landed after review that are worth pointing at, because they are the sharpest part of the diff and neither is in the story above.

Evan caught that a full batch does not fit a websocket frame. coder/websocket defaults to a 32 KiB read limit and a hundred deploys carrying git provenance measure 105,924 bytes worst case, so a cluster past roughly eighty deploys would not have dropped a batch, it would have wedged: the read fails, the connection drops, the runtime reconnects and re-sends the same oversized batch forever, ingesting nothing. Cloud now sets the limit explicitly, and `TestFullBatchFitsInOneFrame` builds a worst-case batch and asserts it fits, failing at the old default with "105924 is not less than 32768".

Chasing that turned up a worse one. `error_reason` was completely unbounded, while the comment directly above it and RFD-94 both say short reason and never raw build output. A deploy failing with a stack trace shipped all of it, which blows the frame and quietly walks build logs past the custody boundary the rest of the file is careful about. Now capped, deterministically, so the same record always produces the same payload.

The cloud half is mirendev/cloud#203, merged. It uses `SendMessageBlocking` and `SpreadOnConnect` from #1059, which is also merged, so this now sits on main rather than on that branch.

Part of MIR-1560.

### DIFF
--- a/components/coordinate/deployreport.go
+++ b/components/coordinate/deployreport.go
@@ -1,0 +1,777 @@
+package coordinate
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"log/slog"
+	"sort"
+	"time"
+
+	"miren.dev/runtime/api/core/core_v1alpha"
+	esv1 "miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/deploylifecycle"
+	"miren.dev/runtime/pkg/entity"
+	"miren.dev/runtime/pkg/idgen"
+	"miren.dev/runtime/pkg/rpc/stream"
+	"miren.dev/runtime/pkg/uplink"
+)
+
+// The deploy.* message family carries the cluster's deploy log to cloud, the
+// durable half of the visibility contract in RFD-94.
+//
+// These types are one half of a wire contract. The other half is
+// DeployBatch/DeployState in mirendev/cloud, services/cluster_channel/
+// deploy_handlers.go, and nothing enforces that the two agree: cloud does not
+// import this module, so renaming a json tag here leaves both repos building
+// and every test on both sides passing while the field silently stops
+// arriving. Change a tag, a type or a message name here and you must change it
+// there in the same breath. MIR-1601 is the fix that makes this checkable.
+//
+// It differs from app.* in what a lost message costs. App state is an ephemeral
+// sample and a dropped one self-heals from the next report. A deploy is an event
+// that happened once, and the runtime prunes its own records over time, so once
+// cloud has a deploy it becomes the book of record for it. Nothing here can be
+// left to best effort.
+const (
+	// TypeDeploySyncHello opens deploy sync on a connection, asking cloud where
+	// its copy left off.
+	//
+	// The runtime asks rather than cloud volunteering because three of the four
+	// reasons to distrust a cursor are only visible from this side: whether the
+	// revision has been compacted away, whether it sits above the store's
+	// current head, and whether it is coherent at all. Only "cloud has none"
+	// is visible from over there, and that arrives as a zero. Putting the
+	// question here keeps one decision in one function.
+	TypeDeploySyncHello = "deploy.sync.hello"
+
+	// TypeDeploySyncCursor is cloud's answer: the revision below which it
+	// believes its copy is complete.
+	TypeDeploySyncCursor = "deploy.sync.cursor"
+
+	// TypeDeployBatch carries a contiguous run of deploys and the range it
+	// covers.
+	//
+	// Named for the contiguity claim rather than for the write it performs —
+	// RFD-94 sketches this as deploy.upsert — because the claim is the part
+	// that matters. A message called "upsert" invites a second sender that
+	// upserts without a range, and cloud advancing its watermark on that would
+	// open exactly the permanent gap this design exists to prevent.
+	TypeDeployBatch = "deploy.batch"
+)
+
+const (
+	// deployBatchSize bounds one message so a backfill arrives as a slide
+	// rather than a jump (RFD-94). Reporting must never take meaningful
+	// horsepower from the cluster's real work.
+	deployBatchSize = 100
+
+	// deployBatchPause spaces batches out for the same reason, and leaves the
+	// link available to its other tenants during a long backfill.
+	deployBatchPause = 250 * time.Millisecond
+
+	// deployConnectSpread is how far the opening handshake is scattered across
+	// the fleet, and it deliberately starts after the app reporter's own
+	// spread rather than overlapping it.
+	//
+	// A cluster that just came online should finish painting its app list
+	// before a backfill's long tail starts competing for the same link. That
+	// ordering is a preference rather than a requirement: both tenants send
+	// with backpressure, so an overlap would resolve into them sharing the
+	// pipe rather than one starving the other. Starting later simply makes the
+	// common case the intended one.
+	deployConnectSpread = 90 * time.Second
+
+	// maxErrorReasonBytes bounds the failure reason put on the wire.
+	//
+	// RFD-94 asks for a short reason and never raw build output, and until now
+	// that was a sentence in a comment with nothing enforcing it: the entity's
+	// error message went out whole. A deploy that failed with a stack trace or
+	// a wall of compiler output would ship all of it, which smuggles build logs
+	// past the custody boundary the rest of this file is careful about, and
+	// blows the frame budget on top.
+	//
+	// Truncation is deterministic, so the same record always produces the same
+	// payload. That matters more than it looks: a value that varied per report
+	// would defeat the upsert the way a drifting deployed_at would.
+	maxErrorReasonBytes = 512
+
+	// deployCursorTimeout bounds the wait for cloud's answer.
+	//
+	// A cloud that never answers is treated as a cloud with no cursor, which
+	// costs a re-list and no correctness. Waiting forever instead would mean a
+	// cluster silently stops reporting deploys against a cloud that is up
+	// enough to hold a socket but not to answer.
+	deployCursorTimeout = 30 * time.Second
+)
+
+// DeploySyncHello opens the exchange. It carries no fields today; the runtime
+// is asking a question, not asserting anything.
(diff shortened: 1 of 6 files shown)


phinze:
🤖 Both of you landed on the same thing in `run()`, so that is fixed: the fallback to a re-list is now gated on `errors.Is(err, errCompacted)`, and any other stream error warns and returns to wait for the next connection. Re-listing a cluster's whole history on a transport blip was expensive for nothing, and the comment claiming compaction was the only case no longer contradicts the guard. `TestNonCompactionStreamErrorDoesNotRelist` pins it.

`idgen.TimeOf` now requires exactly sixteen decoded bytes rather than at least six. base58 drops leading zero bytes, so a truncated id could decode short and have its leading bytes read as a timestamp, which is a confident wrong answer rather than a refusal. Regression case added for a truncated id.

Added the progress-notification test: an op with no entity still has to move the frontier, which is what keeps a quiet cluster from freezing its watermark at its last deploy.

On batching the delta stream: leaving that for now. The premise is a little off, since `WatchIndex` is scoped to the deployment kind index rather than the store, so it fires on deploy changes and periodic progress notifications, not once per store revision. Catch-up from a very old watermark on a busy cluster is still the case where accumulating would help, and `SendMessageBlocking` already supplies the backpressure that keeps it safe rather than merely chatty. Worth doing if it shows up in practice.

/biscuit review

phinze:
🤖 All three taken, they all made the tests prove more than they were proving.

The custody test now uses a distinct sentinel per excluded field, including `deployed_by.user_name`. The old version shared one substring across the deployer fields, so a leak of the name could hide behind the email being correctly excluded. Added user id and the git commit message to the set while I was in there.

The resume test now has the watch replay the deploys the interrupted backfill never sent, not just one live op, and asserts the batches tile without a gap from the frontier the first connection left through to the newest. As written before it only proved the stream opened in the right place, which is the weaker half of what the test is named for.

`deployedAtFor` is now pinned against `idgen.TimeOf` directly rather than only checked for stability, so the fallback is asserted to be the id's own creation time rather than some other clock reading that happens to be stable.

/biscuit review

components/coordinate/deployreport.go
```diff
@@ -0,0 +1,757 @@
...
+// message can carry a range and advance the watermark. The two-stream split
+// RFD-94 describes is only needed when there is no usable cursor, which is what
+// relist handles.
+//
+// Returns nil when the connection ends, and an error when the stream must fall
+// back to a re-list.
+func (r *deployReporter) stream(ctx context.Context, link deployLink, from, head int64) error {
+	sent := from
+
+	return r.source.Watch(ctx, from, func(op *esv1.EntityOp) error {
```
evanphx, line 479: This block has a second empty-batch case the comment doesn't name: `HasEntity()` is true but `stateFrom` returns `ok=false` — a deploy with no stable `deployed_at`. It ships an empty batch too, so the frontier advances past a deploy it drops, and the range claims "no deploys here" when there was one.

Advancing is right, not a bug — holding `sent` back would wedge the watermark forever, since the next connect re-streams the same revision and skips it again, and the dropped deploy has no partition key for cloud to hold anyway. It's also near-unreachable: a real id decodes via `idgen.TimeOf`. Just worth a word in the comment so the empty batch reads as "a deploy we dropped," not only "nothing happened here." One asymmetry while we're here — `relist` logs a `skipped` count, but this path drops silently; a `Debug` here would match them.

components/coordinate/deployreport.go
```diff
@@ -0,0 +1,757 @@
...
+func shortReason(msg string) string {
+	if len(msg) <= maxErrorReasonBytes {
+		return msg
+	}
+	return msg[:maxErrorReasonBytes] + "… (truncated)"
+}
+
+// deployedAtFor returns the moment a deploy was initiated, which must be stable
+// across every report of it.
+//
```
evanphx, line 710: `HeadRevision` and `Watch` are the seam where the tested sync logic meets real etcd, and they lean on two store behaviors that are true today but would fail *quietly* if they ever drifted: `List(...).Revision()` being the global head regardless of index (if it went per-index, `decideSyncMode` would relist on every connect, forever), and compaction surfacing as an `IsCompacted` op rather than an error (if it errored, `stream` would take the non-compaction branch and never re-list). The fakes encode both as their own behavior, so the unit suite can't catch a drift in the thing it's standing in for. A thin integration test over `entityDeploySource` — head ≥ a deploy's revision, a compacted watch yields an `IsCompacted` op — would pin them. Non-blocking; good follow-up ticket.

components/coordinate/deployreport.go
```diff
@@ -0,0 +1,757 @@
...
+		}
+	}
+
+	return state, true
+}
+
+// shortReason trims a failure message to something that fits the contract.
+//
+// The marker matters: a reader who sees a reason end mid-sentence should be
+// able to tell that cloud has a summary rather than that the deploy failed
```
evanphx, line 698: `log` is set in `startDeployReporter` but nothing here reads it. Drop it, or wire it into the `List`/`Watch` error paths if that was the intent.

evanphx (approved):
Approved — the watermark holds end to end. I chased the reconnect and cursor edges hard, and they all bottom out the same way: a wrong or crossed cursor costs a re-list, never a gap.

Inline notes are all non-blocking. The `entityDeploySource` seam is the only one with any teeth — happy to file it as a follow-up rather than hold the PR. The other two are a silent skip and a dead field.

--e+🤖

phinze:
🤖 All three addressed, two in code and one filed.

**The second empty-batch case** is now named in the comment, and you talked me into it being interesting rather than just undocumented. The empty batch on that path means "a deploy was here and is not coming," which reads differently from "nothing happened," and the comment now says so along with why advancing anyway is still right: holding the range back wedges the watermark permanently, since the next connection re-streams the same revision and drops it again.

You were also right about the asymmetry, and it was worse than cosmetic. `relist` counted skips while the tail said nothing, so the same event was visible on one path and invisible on the other, which is exactly the kind of gap that makes someone trust the wrong log. There is a `Debug` there now with the entity id and revision, and `TestUnkeyableDeployInTheTailStillAdvancesTheFrontier` pins the behaviour rather than leaving it to the prose.

**The dead `log`** is gone. The comment on the struct now says it carries no logger on purpose, since every method hands its error back to the reporter and that is the layer that decides how loud a failure to report deploys should be.

**The `entityDeploySource` seam** is MIR-1609, and digging into it made your point sharper. My first instinct was to reach for `testutils.NewInMemEntityServer`, which would have been worse than nothing: it is backed by `MockStore`, whose `ListIndexRevision` returns the maximum revision across all entities rather than etcd's `Header.Revision`. That lands on the same answer by a different mechanism, so the test would have passed while pinning something production does not do, and it has no notion of compaction at all. Pinning these needs real etcd, which `pkg/etcdtest.TestEtcdClient` gives us. The issue spells out both assertions and, more usefully, what each drift would look like in production: a permanent re-list on every connect for the first, and deploy sync silently stopping with only a warning for the second.

Full package is 98 tests green in iso, lint clean.

components/coordinate/deployreport.go
phinze: 🤖 Named it, and added the `Debug` so the tail matches `relist`. The asymmetry was worse than cosmetic: the same event was countable on one path and invisible on the other. `TestUnkeyableDeployInTheTailStillAdvancesTheFrontier` pins that the frontier still passes it.

components/coordinate/deployreport.go
phinze: 🤖 Dropped. The struct comment now says it carries no logger deliberately, since every method hands its error back to the reporter and that is the layer that decides how loud a reporting failure should be.

components/coordinate/deployreport.go
phinze: 🤖 Filed as MIR-1609. Worth knowing: `testutils.NewInMemEntityServer` would have been worse than nothing here. It is backed by `MockStore`, whose `ListIndexRevision` returns max-revision-across-entities rather than etcd's `Header.Revision` — same answer, different mechanism — and it has no notion of compaction. Real etcd via `pkg/etcdtest` is the only thing that pins either.

PR #1063, opened 2026-08-17
Resolve addon env vars from the association instead of copying them into versions

phinze:
This is an alternative to #1055. Same bug, same Linear issue, different layer, and Evan's PR is a correct fix for the model we have today. This one argues the model is the thing to change.

## The diagnosis

An addon's variables are recorded in two places. The `AddonAssociation` holds them, which is where RFD-59 says they live. Provisioning then also copies them into the app's config, by minting a new version.

That copy is the bug. It gets made once and never maintained, so a version built before the addon was attached simply doesn't have it. Activating such a version — which is exactly what `deploy -V` and `app rollback` do — hands the app a config with no `DATABASE_URL`, and nothing puts it back. Rebuilds copy config forward from whatever is active, so they inherit the gap. The addon controller only injects at provision time, so it never re-injects. The only recovery anyone found was destroying the addon, which drops the database.

Every other symptom is the same copy showing up somewhere else: deprovision has to hunt down copies by exact string match to delete them, rotation has to mint a version to update them, and two addons updating copies at once needed a CAS retry loop to keep from clobbering each other.

So instead of teaching one more write path to maintain the copy, this deletes the copy. Bindings are read from the association at resolve time. They follow the app rather than the version, and no version has to have been built with them.

## What that buys

`deploy -V g8W` activates `g8W` — not a version derived from it, `g8W` itself. Rollback stops minting anything. `app restart` on a version without the binding works, which a fix in `DeployVersion` can't reach.

And about 250 lines go away. With nothing maintaining copies, `updateAppActiveVersion`, `createVersionWithVars`, `removeEnvVars` and `mergeAddonVars` have no callers. The MIR-1458 race goes with them: two associations used to concurrently swing `active_version` and clobber each other, and now each writes only its own entity, so there's nothing to race over. That retry loop is gone rather than merely unused.

## The commits

**Sync addon association variables at boot.** Rotation used to write the new credential into a version and leave the association holding the old one. Harmless while nothing read the values, but the next commit makes them load-bearing, so the record has to be made true first. One pass over active associations, bounded by how many addons exist, idempotent. No version history is walked and no ConfigVersion is rewritten. Worth knowing: replacing the variables needs a full-entity `Replace`, because `Patch` and `Update` both *merge* multi-valued attributes and would leave the superseded credential beside the new one. That's most likely why the field was left frozen in the first place.

**Resolve addon bindings from the association at runtime.** The resolver, plus switching the runtime consumers to it. Write paths deliberately stay on `ResolveConfig`, since resolving through the overlay and storing the result would bake the copy straight back in.

**Stop the addon controller from writing app versions.** The deletions. Provisioning, rotation and deprovision each write only the association now. They still propagate: the launcher already watches `AddonAssociation`, and pool identity is keyed on the resolved sandbox spec rather than the version, so a changed environment produces a new pool. Minting a version was only ever the trigger; the config diff always did the work.

**Cover the MIR-1579 recovery path end to end.** Blackbox, run against a real cluster in both directions: it passes here and fails deterministically on trunk.

## Sequencing

The boot sync ships in the same release as the resolver that depends on it, and runs before any controller starts. That ordering is the whole safety argument, so it's worth a look in review.

Rotation has only been out since v0.12.1 with near-zero usage, which is what makes the sync's blast radius small: for any app that never rotated, it's a no-op.

## Left alone

Addon credentials still travel as plaintext values, so they land in the sandbox pool spec and therefore in etcd. RFD-90 already flagged this: it asks whether addons should `Put` into the cluster backend and contribute a reference, so their secrets get the same at-rest protection and versioning as any other secret, and calls the convergence attractive but contract-changing. This PR changes the addon contract anyway, which makes that a shorter step from here than from trunk, but it doesn't attempt it.

Nothing here touches `source`. RFD-90 settled that it is the ownership axis (`config | manual | addon`) rather than RFD-55's storage-backend axis, and this PR uses it exactly that way.

Closes MIR-1579 if we take this route instead of #1055.

### DIFF
--- a/api/core/addonvars.go
+++ b/api/core/addonvars.go
@@ -1,0 +1,178 @@
+package compute
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"miren.dev/runtime/api/addon/addon_v1alpha"
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/entity"
+)
+
+// ResolveRuntimeConfig returns the config an AppVersion runs with: the version's
+// stored config, plus the addon bindings the app's associations currently supply.
+//
+// This reads the app's addon associations, so it costs one indexed List per call
+// and fails if that read fails. Callers that write a ConfigVersion must use
+// ResolveConfig instead. Storing the result of this function would put addon
+// bindings back into a version, which is what this removes.
+//
+// Addon variables stored on the version are dropped, and the live bindings are
+// used instead. This keeps the invariant in one place: an AppVersion records
+// user config, and addon state resolves at runtime. A version that stores a copy
+// never serves it, whether the copy predates this design or a later change
+// writes one.
+//
+// Precedence:
+//
+//   - A binding replaces an app.toml declaration of the same key. Declaring
+//     DATABASE_URL in app.toml and then attaching a database addon means the
+//     addon supplies the value. Required and Description carry from the
+//     declaration onto the binding, because `miren env list` shows them and the
+//     next build validates against them.
+//   - A binding never replaces a variable an operator set. See RFD-59.
+//
+// Before this, an addon contributed its variables by minting a successor
+// version. A version built before the addon existed therefore never contained
+// DATABASE_URL, and activating one left the app without its credentials
+// permanently (MIR-1579).
+func ResolveRuntimeConfig(
+	ctx context.Context,
+	eac *entityserver_v1alpha.EntityAccessClient,
+	ver *core_v1alpha.AppVersion,
+) (*core_v1alpha.ConfigSpec, error) {
+	spec, err := ResolveConfig(ctx, eac, ver)
+	if err != nil {
+		return nil, err
+	}
+
+	if ver.App == "" {
+		return spec, nil
+	}
+
+	bindings, err := addonBindings(ctx, eac, ver.App)
+	if err != nil {
+		return nil, err
+	}
+
+	// The common case: no addons, no stored copies. Return the spec untouched
+	// rather than rebuilding an identical slice.
+	if len(bindings) == 0 && !hasAddonVars(spec.Variables) {
+		return spec, nil
+	}
+
+	// Only operator-set keys block a binding. An app.toml declaration does not,
+	// because blocking it would hand the app a placeholder instead of a credential.
+	kept := make([]core_v1alpha.ConfigSpecVariables, 0, len(spec.Variables)+len(bindings))
+	blocked := make(map[string]struct{}, len(spec.Variables))
+	for _, v := range spec.Variables {
+		if v.Source == SourceAddon {
+			continue
+		}
+		kept = append(kept, v)
+		if operatorOwned(v.Source) {
+			blocked[v.Key] = struct{}{}
+		}
+	}
+
+	for _, b := range bindings {
+		if _, ok := blocked[b.Key]; ok {
+			continue
+		}
+		blocked[b.Key] = struct{}{}
+
+		bound := core_v1alpha.ConfigSpecVariables{
+			Key:       b.Key,
+			Value:     b.Value,
+			Sensitive: b.Sensitive,
+			Source:    SourceAddon,
+		}
+
+		// Replace a shadowed declaration in place. This keeps one entry per key and
+		// preserves its position.
+		if i := indexOfKey(kept, b.Key); i >= 0 {
+			bound.Required = kept[i].Required
+			bound.Description = kept[i].Description
+			kept[i] = bound
+			continue
+		}
+		kept = append(kept, bound)
+	}
+
+	spec.Variables = kept
+	return spec, nil
+}
+
+// operatorOwned reports whether a person set this value, in which case an addon
+// must not replace it. An empty source predates the field and counts as
+// operator-set, matching mergeVariablesFromAppConfig.
+func operatorOwned(source string) bool {
+	return source == SourceManual || source == ""
+}
+
+func indexOfKey(vars []core_v1alpha.ConfigSpecVariables, key string) int {
+	for i := range vars {
+		if vars[i].Key == key {
+			return i
+		}
+	}
+	return -1
+}
+
+func hasAddonVars(vars []core_v1alpha.ConfigSpecVariables) bool {
+	for _, v := range vars {
+		if v.Source == SourceAddon {
+			return true
+		}
+	}
+	return false
+}
+
+// addonBindings returns the variables every active addon association contributes
+// to an app.
+//
+// Only "active" associations count. A pending or provisioning one has no final
+// values yet, and the launcher defers pool creation until they settle (see
+// addonsReady). A deprovisioning one is being torn down, so serving its
+// credentials would point the app at a database that is about to disappear. An
+// "error" one never finished provisioning, so its values were never known good.
+func addonBindings(
+	ctx context.Context,
+	eac *entityserver_v1alpha.EntityAccessClient,
+	appID entity.Id,
+) ([]addon_v1alpha.Variables, error) {
+	results, err := eac.List(ctx, entity.Ref(addon_v1alpha.AddonAssociationAppId, appID))
+	if err != nil {
+		return nil, fmt.Errorf("listing addon associations for %s: %w", appID, err)
+	}
+
+	type contribution struct {
+		assoc entity.Id
+		vars  []addon_v1alpha.Variables
+	}
+
+	var found []contribution
+	for _, ent := range results.Values() {
+		var assoc addon_v1alpha.AddonAssociation
+		assoc.Decode(ent.Entity())
+
+		if assoc.Status != "active" {
+			continue
+		}
+		found = append(found, contribution{assoc: assoc.ID, vars: assoc.Variables})
+	}
+
+	// The first contributor to supply a key wins, so sort by association id to make
+	// that a function of stored state rather than of List order. Two addons
+	// supplying the same key is rare, because AdjustEnvVars asks the second to
+	// rename at provision time, but it is reachable.
+	sort.Slice(found, func(i, j int) bool { return found[i].assoc < found[j].assoc })
+
+	var out []addon_v1alpha.Variables
+	for _, f := range found {
+		out = append(out, f.vars...)
+	}
+	return out, nil
+}
--- a/components/activator/activator.go
+++ b/components/activator/activator.go
@@ -606,7 +606,7 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
 	// register a freshly-discovered pool on the cache-miss path. This is the
 	// scale-up path (only reached when no sandbox has capacity), so the extra
 	// entity store reads are acceptable.
-	spec, err := coreutil.ResolveConfig(ctx, a.eac, ver)
+	spec, err := coreutil.ResolveRuntimeConfig(ctx, a.eac, ver)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve config: %w", err)
 	}
@@ -1416,7 +1416,7 @@ func (a *localActivator) watchSandboxes(ctx context.Context) {
 			appVer.Decode(verResp.Entity().Entity())
 
 			// Resolve config from ConfigVersion if available
-			spec, err := coreutil.ResolveConfig(ctx, a.eac, &appVer)
+			spec, err := coreutil.ResolveRuntimeConfig(ctx, a.eac, &appVer)
 			if err != nil {
 				a.log.Error("failed to resolve config for sandbox", "error", err, "sandbox", sb.ID)
 				return nil
@@ -1610,7 +1610,7 @@ func (a *localActivator) recoverSandboxes(ctx context.Context) error {
 		appVer.Decode(verResp.Entity().Entity())
 
 		// Resolve config from ConfigVersion if available
-		spec, err := coreutil.ResolveConfig(ctx, a.eac, &appVer)
+		spec, err := coreutil.ResolveRuntimeConfig(ctx, a.eac, &appVer)
 		if err != nil {
 			a.log.Error("failed to resolve config for sandbox", "error", err, "sandbox", sb.ID)
 			continue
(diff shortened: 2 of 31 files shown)


phinze:
Whoops this got pushed out of draft too early. Pulling back to keep scrubbing it a little more. Will re-request review when I'm 🤝 with the 🤖

components/activator/activator.go
```diff
@@ -606,7 +606,7 @@ func (a *localActivator) requestPoolCapacity(ctx context.Context, ver *core_v1al
 	// register a freshly-discovered pool on the cache-miss path. This is the
 	// scale-up path (only reached when no sandbox has capacity), so the extra
 	// entity store reads are acceptable.
-	spec, err := coreutil.ResolveConfig(ctx, a.eac, ver)
+	spec, err := coreutil.ResolveRuntimeConfig(ctx, a.eac, ver)
```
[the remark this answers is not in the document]
phinze: Not taking this one, though the reasoning is sound as far as it goes.

You are right that several of these sites read only concurrency, services or tasks, and that routing them through `ResolveRuntimeConfig` makes them newly dependent on an association read that could fail. But the split is per-call-site rather than per-type, and it breaks down exactly where the two sit next to each other: `servers/exec` and the run controller genuinely need variables, since they launch processes with the app environment, while the activator sites beside them do not. Encoding that as "call A here and B there" leaves a rule with no local justification, and the next person adding a resolve call has no way to tell which one they want. Getting it wrong silently means an app running without its database credentials, which is the bug this PR exists to remove.

One resolver that always returns the configuration a version actually runs with is the property worth keeping. If association reads turn out to be a real availability risk, the fix belongs inside `ResolveRuntimeConfig` — caching against the `AddonAssociation` watch the launcher already runs, or degrading to the stored config on a list failure rather than propagating the error — and it would then apply everywhere rather than at whichever sites someone remembered to annotate.

Worth saying plainly that the read is one indexed lookup by app and is unmeasured. If it shows up in practice that is a good argument for the cache, and a better one than fragmenting the API pre-emptively.

components/coordinate/coordinate.go
```diff
@@ -1099,6 +1099,18 @@ func (c *Coordinator) Start(ctx context.Context) error {
...
+	// Report associations left stale by a rotation that predates bindings being
+	// resolved from them. This only reports; see ReportStaleAssociationVariables
+	// for why fixing it here would be worse than naming it.
+	assocStale, assocChecked, assocErr := addonctrl.ReportStaleAssociationVariables(ctx, c.Log, ec, eac)
+	if assocErr != nil {
+		c.Log.Warn("addon association variable check did not complete",
+			"stale", assocStale, "checked", assocChecked, "error", assocErr)
+	} else if assocStale > 0 {
+		c.Log.Warn("addon associations disagree with their apps' stored config",
+			"stale", assocStale, "checked", assocChecked)
```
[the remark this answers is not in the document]
phinze: Taken — folded into the commit that added the call rather than stacked on top, so it reads as one change.

`ReportStaleAssociationVariables` now runs under `maintCtx`, the same two-minute startup-maintenance deadline the entity-store and short-id migrations already share, rather than the unbounded request context. That keeps the whole startup-maintenance phase under one budget instead of introducing a second timeout with its own tuning story.

A sweep cut short by the deadline behaves the same way an unreadable app already did: it warns and boot continues. Worth stating why that degradation is safe here specifically — the scan only reports, so an association it never reached means a missing warning, never a missing binding. Nothing about what an app runs depends on this completing.

api/core/addonvars.go
```diff
@@ -0,0 +1,163 @@
...
+		if operatorOwned(v.Source) {
+			blocked[v.Key] = struct{}{}
+		}
+	}
+
+	for _, b := range bindings {
+		if _, ok := blocked[b.Key]; ok {
+			continue
+		}
+		blocked[b.Key] = struct{}{}
```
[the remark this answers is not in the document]
phinze: Fixed, and folded into the commit that introduced `addonBindings` so it reads as one change rather than a correction on top.

Associations are now sorted by id before their variables are concatenated, so the "first contributor wins" rule resolves to the same answer regardless of what order the entity store returned. biscuit raised the same point independently, which was a good signal it was worth doing rather than waving off as unreachable.

It is still an unusual case — `AdjustEnvVars` asks the second addon to rename at provision time when keys collide — but "the resolved config is a function of stored state" is a property worth being able to state without qualification.

controllers/addon/migrate.go
```diff
@@ -0,0 +1,137 @@
...
+			skipped++
+			continue
+		}
+
+		if err := setAssociationVars(ctx, eac, assoc.ID, desired); err != nil {
+			log.Warn("failed to sync association variables",
+				"association", assoc.ID, "error", err)
+			skipped++
+			continue
+		}
```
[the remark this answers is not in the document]
phinze: This was the best catch of the review, and the fix was to delete the thing rather than gate it.

You were right that it was not an edge case. After this branch, rotation writes only the association, while a version predating the upgrade still stores the old copy — so the next coordinator restart would read that copy and write it back over the rotated credential. Rotate then restart is completely routine, and the app would come back holding a password that no longer works.

The trap underneath it is that there is no reliable way to tell which of the two records is fresher. Before this branch the version was, after it the association is, and nothing durable distinguishes them. I looked at precedence by entity timestamp and rejected it: it is not a migration, it is a permanent tiebreaker with no removal date.

So there is no automatic repair now. `ReportStaleAssociationVariables` scans, names the associations that disagree and the keys involved, never the values, and writes nothing. `miren addon rotate` reconciles both records, which makes the repair one documented command against a population we expect to be empty — rotation shipped in v0.12.1 and is barely used. Because it writes nothing it can be deleted outright once v0.12.x and v0.13.x are out of upgrade range, which was the test that settled the design.

controllers/addon/assocvars.go
```diff
@@ -0,0 +1,93 @@
...
+	byKey := make(map[string]addon_v1alpha.Variables, len(a))
+	for _, v := range a {
+		byKey[v.Key] = v
+	}
+	for _, v := range b {
+		prev, ok := byKey[v.Key]
+		if !ok || prev != v {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
phinze: Taken. Folded into the commit that added `setAssociationVars`.

`setAssociationVars` now collapses repeated keys before comparing or writing, last one wins. A container environment is a map, so a repeated key has no meaning to store, and collapsing is more honest than faithfully preserving a multiset of something ill-defined.

That also makes the comparison sound rather than needing to change. Once the incoming set has one entry per key, a stored set holding a repeat has fewer distinct keys at the same length, so it can never match, and it gets rewritten and cleaned up instead of being mistaken for a no-op.

Worth recording how the failure actually presents, because it is not the obvious one. The map is built from the *stored* set and looked up with the incoming one, so the duplicate has to arrive on the incoming side: stored `{A=1, B=2}` against incoming `{A=1, A=1}` matches on length and on every lookup, the write is skipped, and `B` survives after the addon stopped supplying it. My first two tests missed that and passed against the unfixed code. The one that ships, `TestSetAssociationVarsDoesNotSkipWhenIncomingHasDuplicates`, was checked against the unfixed code and fails there.

controllers/addon/rotation_controller.go
```diff
@@ -128,19 +128,21 @@ func (c *RotationController) rotate(ctx context.Context, req *addon_v1alpha.Rota
...
+	// AddonAssociation watch turns the change into a reconcile that relaunches
+	// the consumer with the new secret. No version is minted.
+	//
+	// This runs after the credential already changed on the engine, so a failure
+	// must stay retryable rather than terminal. Returning an error re-reconciles,
+	// and re-running the idempotent rotation above converges.
 	if len(result.EnvVars) > 0 {
-		if err := createVersionWithVars(ctx, c.log, c.ec, c.eac, assoc.App, result.EnvVars); err != nil {
-			return fmt.Errorf("redeploying consumer (will retry): %w", err)
+		if err := setAssociationVars(ctx, c.eac, req.Association, result.EnvVars); err != nil {
```
evanphx, line 140: Non-blocking, but worth a look before merge, and it lands in files this PR doesn't touch.

`activeConfigVar` in `pkg/addon/postgresql/rotate.go` and its mysql twin read the app's active *stored* ConfigVersion. Now that this PR strips `source=addon` from every path that writes a version, that read comes back empty for anything built after the upgrade. Three things follow.

The one with teeth: `CaptureDedicatedConnInfo` falls back to `activeConfigVar("MYSQL_USER")` / `("PGUSER")` for dedicated associations predating the stored `username`/`database_name` attrs — a fallback that exists precisely to keep those rotatable. After the first redeploy post-upgrade the active version no longer carries the key, so rotation hard-fails with "could not determine dedicated mysql user". Loud rather than silent, at least. Worth checking whether any pre-attr dedicated associations are actually out there.

Second: the old-password capture for saga compensation degrades to a no-op. I don't think it can fire today — nothing after `AlterSharedUserPassword` returns an error — so it's a trap for the next fallible step someone adds rather than a live bug.

Third: the doc comment on `activeConfigVar` now says the reverse of the truth — "unlike association.variables, which is deliberately not rewritten on rotation and can hold a stale password."

All three want the same change: read from `assoc.Variables`, which is the fresh record now and already carries these keys.

api/core/addonvars.go
```diff
@@ -0,0 +1,178 @@
...
+	// Only operator-set keys block a binding. An app.toml declaration does not,
+	// because blocking it would hand the app a placeholder instead of a credential.
+	kept := make([]core_v1alpha.ConfigSpecVariables, 0, len(spec.Variables)+len(bindings))
+	blocked := make(map[string]struct{}, len(spec.Variables))
+	for _, v := range spec.Variables {
+		if v.Source == SourceAddon {
+			continue
+		}
+		kept = append(kept, v)
+		if operatorOwned(v.Source) {
```
evanphx, line 75: Non-blocking, and not a request to change the precedence: a `manual` var blocking the binding is inherited behavior and I think it's the right call, especially now that it's reversible — unsetting brings the binding straight back, which it didn't before this PR.

The gap is that nothing tells you it happened. `m env set DATABASE_URL=...` on an app with a live database addon silently covers the real credential; rotation keeps writing the association and none of it reaches the app. `mergeIntoSpec` stamps `manual` unconditionally, so it also relabels a legacy `source=addon` entry on the way through. And `env list` gives no hint there's a suppressed value underneath.

Much more fixable now than it was before, since the shadowed value is sitting right there in the association at resolve time. Happy to file it as a follow-up rather than grow this PR.

controllers/addon/migrate.go
```diff
@@ -0,0 +1,119 @@
...
+
+// ReportStaleAssociationVariables warns about associations whose recorded
+// variables disagree with the copy in the app's active version.
+//
+// Rotation used to write the new credential into a new app version and leave the
+// association holding the provision-time value. The runtime now resolves through
+// the association, so one left stale by a rotation on v0.12.x names a password
+// that no longer works.
+//
+// This only reports; it never writes. There is no reliable way to tell which
```
evanphx, line 24: 👏 Report-only, and the reasoning for it. Repairing this would have looked obviously right, and would have undone a rotation on every restart that followed one.

controllers/addon/assocvars.go
```diff
@@ -0,0 +1,110 @@
...
+	"miren.dev/runtime/api/entityserver/entityserver_v1alpha"
+	"miren.dev/runtime/pkg/addon"
+	"miren.dev/runtime/pkg/cond"
+	"miren.dev/runtime/pkg/entity"
+)
+
+// setAssociationVars replaces the association's record of what the addon
+// supplies. Every other attribute on the entity is left as it was.
+//
+// This rewrites the whole entity instead of patching it. Patch and Update both
```
evanphx, line 18: 👏 The `Replace`-not-`Patch` catch. A merge here would have quietly left the superseded credential sitting next to the new one, and everything downstream would still have looked fine.

evanphx (approved):
Good change.

Nothing blocking. Start with the note on `rotation_controller.go` — the legacy-association fallback in the providers is the one with actual teeth. Rest is non-blocking.

--e+🤖

controllers/addon/rotation_controller.go
phinze: Taken, all three, in a new commit (`25ed2611`). You were right that it has teeth, and right that the fix lands in files this PR did not otherwise touch.

The dedicated fallback would have broken exactly as you describe. `CaptureDedicatedConnInfo` reads PGUSER and PGDATABASE from the active config for associations predating the `username`/`database_name` attrs, and after the first post-upgrade redeploy the version no longer carries them, so it fails outright.

The reassuring part is that the legacy population is covered by construction rather than by luck. `association.variables` landed 2026-02-13 and provisioning has written it on every association ever since; the `username` and `database_name` attrs only arrived 2026-07-22. So an association old enough to be missing the attrs necessarily still carries the variables, and reading them covers precisely the window the fallback exists for. I checked both dates with `jj file annotate` rather than assuming.

Old-password capture moves to the same source, and it is arguably better there than it was: the association is the record the rotation controller rewrites, and this reads it before that write lands, so it holds what the app is actually running with instead of whatever a concurrent deploy last stored. Agreed it cannot fire today, so this is about not leaving the trap armed.

The doc comment is gone with the function it was describing.

One thing worth flagging, because it went the other way from what I expected. The legacy-fallback integration tests failed after the change, and my first instinct was that I had broken something. They were seeding the app config and leaving `association.variables` empty, which is a shape provisioning never produced. Fixture updated to the real legacy shape; both still rotate against a live engine and assert the new password authenticates and the old one is rejected.

Leaving the `env set` shadowing note for a follow-up as you suggested, and I will file it rather than let it evaporate.

api/core/addonvars.go
phinze: Filed as MIR-1600 so it does not evaporate: https://linear.app/miren/issue/MIR-1600/env-set-on-an-addon-supplied-key-silently-shadows-the-live-binding

Kept your framing that this is not a precedence change. Manual winning is inherited and right; the gap is that nothing surfaces the shadow. Recorded all three symptoms, including `mergeIntoSpec` stamping `manual` unconditionally and so relabelling a legacy `source=addon` entry on the way through, which is the part I would most expect someone to hit and misread.

Also noted why it is worth doing now rather than before: it became reversible (unsetting brings the binding straight back, which it could not do when the value was baked into a version) and knowable (the suppressed value is on the association at resolve time, so `env list` can say so and `env set` can warn at the moment it happens).

PR #1067, opened 2026-08-19
Verify networking commands before install; document host command needs

evanphx:
## Summary

Closes MIR-1612.

Miren's server and runner shell out to external system commands that aren't in the release bundle and weren't checked or documented. If a host is missing them, networking silently breaks at runtime — hard to diagnose. This adds an install-time preflight check for the always-required networking tools and documents the full host command inventory.

## Changes

**Preflight check** — `miren server install` and `miren runner install` now verify `iptables` and `nft` (nftables) are on `PATH` before installing, and fail early with per-distro (apt/dnf) install guidance if either is missing.

- `cli/commands/install_requirements.go` — pure, testable `requiredNetCommands()` and `findMissingCommands(names, lookup)` helpers.
- `cli/commands/server_install.go` — `missingNetCommands` on `installPrerequisites`, the gate, and a shared `printMissingNetCommandGuidance`.
- `cli/commands/runner_install.go` — same gate + guidance for the runner path.
- `cli/commands/install_requirements_test.go` — `TestFindMissingCommands`.

**Docs**

- `docs/docs/system-requirements.md` — new "Required host commands" section (iptables + nftables), plus a `:::note` on feature-conditional tooling (disk tools for block-device volumes; SELinux tools at install time — both fail-soft).
- `docs/docs/firewall.md` — `:::info` clarifying both iptables and nftables are required.

## Scope

Only `iptables` and `nft` are always exercised and fail hard, so those are what the check gates on. Conditional commands (disk tooling, SELinux tools) and bundled ones (`containerd`, `ctr`) are documented rather than gated. The `shasum`/perl item from the issue lives in the bash installer in the separate `mirendev/cloud` repo and is out of scope here.

## Testing

- `go vet ./cli/commands/` — clean
- `go test ./cli/commands/ -run TestFindMissingCommands` — PASS
- `make lint` — 0 issues (Go) and docs lint passed

### DIFF
--- a/cli/commands/install_requirements.go
+++ b/cli/commands/install_requirements.go
@@ -25,3 +25,27 @@ const systemResourceTolerance = 0.90 // accept down to 90% of the threshold
 func meetsThreshold(observed, threshold int64) bool {
 	return float64(observed) >= float64(threshold)*systemResourceTolerance
 }
+
+// requiredNetCommands lists the external commands the server and runner always
+// shell out to for container networking: iptables for the bridge and sandbox
+// NAT rules, and nft (nftables) for services and NodePorts. Neither ships in the
+// release bundle, so a host missing them fails only later, when networking
+// silently breaks — the install-time check catches it up front instead. Disk and
+// SELinux tooling is deliberately excluded here: it's needed only for optional
+// features and those paths already degrade gracefully.
+func requiredNetCommands() []string {
+	return []string{"iptables", "nft"}
+}
+
+// findMissingCommands returns the subset of names that lookup can't resolve on
+// PATH, preserving input order. lookup is normally exec.LookPath; taking it as a
+// parameter keeps this pure and testable.
+func findMissingCommands(names []string, lookup func(string) (string, error)) []string {
+	var missing []string
+	for _, name := range names {
+		if _, err := lookup(name); err != nil {
+			missing = append(missing, name)
+		}
+	}
+	return missing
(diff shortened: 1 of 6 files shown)


phinze (approved):
🤖 This looks good overall. The shared prerequisite check keeps the server and runner install paths aligned, and the docs give people a useful failure path instead of leaving networking to break later.

Two non-blocking follow-ups may be worth tightening: bridge setup also resolves `ip6tables`, though mainstream distributions install it with the `iptables` package, and the `nft` preflight could eventually validate the same version and capabilities that the runtime requires instead of checking only `PATH`. Neither needs to hold this change.

evanphx:
🤖 Thanks @phinze — follow-ups from your review:

- **ip6tables**: good catch. Bridge setup drives it for the IPv6 rules, so I added it to the required host commands table in `system-requirements.md` and noted it ships with the `iptables` package. I deliberately left it out of the blocking preflight check: `firewall.md` notes IPv6 bridge errors are usually harmless when IPv4 works, so gating on `ip6tables` would over-block IPv4-only hosts that have IPv6 disabled.
- **nft version/capability check**: agreed this is worth doing eventually. Leaving it as future work — the PATH check covers the common failure (the tool missing entirely), and validating version/capabilities is a larger piece.

Also folded in biscuit's wording nit (the stale "systemd" in the root-privilege guidance) in the same push.

PR #1069, opened 2026-08-20
Carry deploy references as id and short-id pairs

phinze:
The console's app detail view renders versions as `app_version/discoveryagent-vCdjv…` where RFD-96 asks for `8kd0`. Cloud renders what it's sent, and what it's sent is the deployment entity's `app_version` attribute: the full entity id. The short id lives on the `app_version` entity rather than on the deploy, so cloud was never going to recover it. The runtime already does this join for the CLI in the deployment server; the reporter just never did.

Rather than bolting a sibling field onto each reference, `app_version` and `source_deploy_id` become `EntityRef` values carrying the entity id and the short id together. Cloud needs both and neither substitutes for the other: short ids are allocated per cluster against the entities alive at the time, so they collide across clusters and are freed for reuse once their entity is pruned, and cloud becomes the book of record for a deploy after the runtime prunes it. A history keyed on short ids is one whose identifiers can later name something else.

That breaks the wire, which is why `source_deploy_id` came along now rather than waiting for someone to want its short id. Deploy visibility is unreleased, so a reshape costs nothing today and costs compatibility handling forever once clusters we don't control are running the released reporter.

A first backfill walks every deploy a cluster has ever run, so resolution takes the cheapest source per kind: versions cost one bulk read, deploys come free from the listing that already carries them, and the live tail resolves one at a time against a connection-scoped cache that remembers misses as well as hits.

Cloud's half is mirendev/cloud#211. Neither is much use without the other.

MIR-1615

### DIFF
--- a/components/coordinate/deployreport.go
+++ b/components/coordinate/deployreport.go
@@ -182,10 +182,16 @@ type DeployState struct {
 	// consulting any clock.
 	Revision int64 `json:"revision"`
 
-	AppName    string `json:"app_name"`
-	AppVersion string `json:"app_version"`
-	Status     string `json:"status"`
-	Phase      string `json:"phase"`
+	AppName string `json:"app_name"`
+
+	// Version is the app version this deploy shipped, nil when it never got
+	// one — a deploy that failed before a build produced a version has no
+	// version to name, and saying so with an absent object is honest where an
+	// empty string is a value that has to be recognized as meaning nothing.
+	Version *EntityRef `json:"version"`
+
+	Status string `json:"status"`
+	Phase  string `json:"phase"`
 
 	// DeployedAt is when the deploy was initiated, and must be identical in
 	// every report of this deploy: cloud partitions on it, so a value that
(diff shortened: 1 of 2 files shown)


components/coordinate/deployreport.go
```diff
@@ -768,6 +878,64 @@ func (s *entityDeploySource) List(ctx context.Context) ([]*deploylifecycle.Recor
...
+	if ent == nil {
+		return "", ""
+	}
+	return ent.Id(), shortIDOf(ent)
+}
+
+// shortIDOf reads the db/short-id attribute off an RPC entity, or "" when it
+// carries none. That attribute is string-kinded, so String is the right reader
+// for it.
+func shortIDOf(ent *esv1.Entity) string {
```
evanphx, line 947: Non-blocking: this is byte-for-byte `shortIDFromRPCEntity` over in `servers/deployment/server.go:1328`. Two readers of the same `db/short-id` attribute in different packages will drift eventually — might be worth a shared home next to `entity.Entity.ShortId()` someday.

components/coordinate/deployreport.go
```diff
@@ -539,17 +589,29 @@ func (r *deployReporter) stream(ctx context.Context, link deployLink, from, head
...
+	//
+	// A re-list walks every deploy a cluster has ever run, and each one names a
+	// version and possibly a source deploy. Resolved one at a time that is two
+	// reads per deploy against a store that can answer for a whole kind at
+	// once, which is the difference between a backfill that is a slide and one
+	// that is a load. Anything created after this read still falls through to a
+	// single lookup, so the bulk read is an optimization and never the source
+	// of truth.
+	ids.preload(ctx, core_v1alpha.KindAppVersion)
+	ids.preload(ctx, core_v1alpha.KindDeployment)
```
evanphx: Non-blocking: `List` above already handed back every deployment with its `.Entity` attached, so every deployment's short id is already in memory — this preload re-lists the whole kind again just to resolve `source_deploy`. The `KindAppVersion` preload earns its read (those aren't in the list); this one could come straight from `records`. Doesn't matter much on the cold backfill path, just noticed it cuts against the "slide, not a load" goal.

evanphx (approved):
Approved! Really nicely done.

Two non-blocking inline notes below, both minor — neither needs to hold the merge.

--e+🤖

PR #1070, opened 2026-08-20
MIR-1047: Publish app.toml JSON Schema

maryelizbeth:
## Summary

- publish a Draft 2020-12 JSON Schema for `.miren/app.toml`
- link the machine-readable schema from the app.toml reference
- add a reflection-based drift test covering supported TOML fields

## Validation

- `go test ./appconfig`
- `bun run lint`
- `bun run build`

Closes MIR-1047

### DIFF
--- a/appconfig/schema_test.go
+++ b/appconfig/schema_test.go
@@ -35,6 +35,8 @@ func TestPublishedSchemaFieldsMatchConfig(t *testing.T) {
 		got    map[string]any
 		skip   map[string]bool
 	}{
+		// post_import is intentionally absent: it is legacy dead config that is
+		// parsed but never used. Keep the public schema from advertising a no-op.
 		{name: "root", typeOf: reflect.TypeFor[AppConfig](), got: schema.Properties, skip: map[string]bool{"post_import": true}},
 		{name: "env", typeOf: reflect.TypeFor[AppEnvVar](), got: schema.Defs["env"].Properties},
 		{name: "build", typeOf: reflect.TypeFor[BuildConfig](), got: schema.Defs["build"].Properties},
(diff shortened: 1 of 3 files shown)


appconfig/schema_test.go
```diff
@@ -0,0 +1,74 @@
...
+		t.Fatalf("invalid app.toml JSON Schema: %v", err)
+	}
+
+	tests := []struct {
+		name   string
+		typeOf reflect.Type
+		got    map[string]any
+		skip   map[string]bool
+	}{
+		{name: "root", typeOf: reflect.TypeFor[AppConfig](), got: schema.Properties, skip: map[string]bool{"post_import": true}},
```
evanphx, line 40: Skipping `post_import` is the right call — it's dead config: parsed and validated but never read, and #884 pulled it from the docs for exactly that reason. Letting the editor flag it steers people off a field that silently does nothing.

Raising it only because CodeRabbit read the exclusion as a bug and wants it added back. A one-line comment on this skip would keep the next reader (human or bot) from re-adding it. And if we want to close the loop for real, deleting `PostImport` from the struct makes the skip moot — glad to file that as a follow-up.

evanphx (approved):
Approved! Clean and tidy.

One inline note below on `post_import` — it's confirming the exclusion is right, not asking to change anything (a bot suggested adding it back). Happy to file a follow-up to delete the dead field for real if you want it gone.

--e+🤖

appconfig/schema_test.go
maryelizbeth: i'll add the comment and linear the follow-up!

maryelizbeth: Added the explanatory comment in 47b7f4fa and filed MIR-1624 to remove the dead field for real: https://linear.app/miren/issue/MIR-1624/remove-dead-post-import-apptoml-field

PR #1074, opened 2026-08-21
Use the cloud volume id for accelerator lease calls

evanphx:
## Summary

Accelerator disk backups stopped as soon as the runtime mounted a volume and acquired its cloud lease. Every upload after that failed with `invalid update request: volume has an active lease: nonce required` and retried every five seconds indefinitely. Seen in production on `disktest2` / volume `vol-ac3q8p1f72o1` (MIR-1634).

The runtime keeps two identifiers for a volume and confused them:

- `MountState.VolumeId` — the local `disk_volume/…` entity id.
- `VolumeState.CloudVolumeId` — the miren.cloud `vol-…` id.

The uploader looked up the lease nonce by comparing the cloud id against the local id, so they never matched and the nonce was omitted — the cloud then rejected the write. The three lease-**release** paths (shutdown, unmount, orphan cleanup) had the mirror of the same bug: they passed the local id to the cloud release call, so they could fail to release indefinite leases.

## What changed

- **`leaseNonceFor`** now resolves the nonce through a new `State.LeaseNonceForCloudVolume` helper that bridges cloud id → volume → mount.
- The **three release paths** resolve the cloud id via a `cloudVolumeIdFor` helper and skip when the volume was never cloud-registered.
- **`MountState.CloudVolumeId`** is now persisted, captured at mount-creation time. This makes orphan cleanup robust: the lease is released with the right id even after the backing `VolumeState` is gone. Mounts persisted before this field fall back to a volume-state lookup, so upgrades don't regress.

## Testing

- The regression tests previously masked the bug by using one string for both ids; they now use deliberately different local and cloud ids.
- The mock cloud server now enforces the nonce on leased volumes, and a new integration test uploads a segment through the real `LogWatcher` → uploader path against an active lease.
- Sanity-checked: reverting only the production `.go` files makes the integration test fail with the exact production error.
- `hack/it components/diskio` — 169 tests pass. `make lint` — 0 issues.

https://claude.ai/code/session_01YNNxdoTF7YUsFaxiJnhsp9

### DIFF
--- a/components/diskio/state.go
+++ b/components/diskio/state.go
@@ -74,6 +74,13 @@ type MountState struct {
 	// VolumeId is the ID of the disk_volume entity
 	VolumeId string `json:"volume_id"`
 
+	// CloudVolumeId is the backing volume's miren.cloud identifier, captured
+	// when the mount is created. Every cloud lease call keys off it, and
+	// recording it here means lease release still works after the volume's
+	// VolumeState is gone (e.g. orphan cleanup). Empty for mounts persisted
+	// before this field existed, or volumes never registered with the cloud.
+	CloudVolumeId string `json:"cloud_volume_id,omitempty"`
+
 	// DevicePath is the path to the loop device node
 	DevicePath string `json:"device_path"`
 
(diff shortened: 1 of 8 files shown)


components/diskio/state.go
```diff
@@ -129,9 +136,41 @@ func LoadState(dataPath string) (*State, error) {
...
 	state.path = path
+
+	// Backfill the cloud id onto mounts that predate MountState.CloudVolumeId.
+	// Without this, a legacy mount that stays mounted across the upgrade never
+	// records its cloud id, and once its VolumeState is gone (orphan cleanup)
+	// the lease can no longer be released. Persist so the migration survives a
+	// restart even if no mount mutation happens in between.
+	if state.backfillMountCloudVolumeIds() {
+		if err := state.Save(); err != nil {
+			return nil, fmt.Errorf("failed to persist mount cloud id backfill: %w", err)
```
phinze: 🤖 Small resilience thought: could we keep the successfully loaded state if only this migration save fails? `LoadState` returns `nil` here, so the runner replaces otherwise valid state with `NewState()`, including dropping active lease nonces from memory. This only affects a degraded startup where the existing file is readable but its atomic rewrite fails, so best-effort persistence seems sufficient: keep the backfill in memory and let a later `Save` retry.

phinze (approved):
🤖 The cloud/local ID split is handled consistently across upload nonce lookup and lease release, and the migration preserves the cloud ID for orphan cleanup. The end-to-end regression coverage exercises the real watcher path with distinct IDs and an enforced active lease. I left one small, nonblocking resilience comment; otherwise this looks good.

components/diskio/state.go
evanphx: 🤖 Good call — done in the latest push. Dropped the `return nil, err` so a failed migration rewrite no longer discards the valid state we just loaded. The backfill stays in memory, and since `Save` serializes volumes and mounts together, a later `Save` can't persist a volume deletion without also persisting the backfilled mounts — so we stay consistent without failing startup.

PR #1076, opened 2026-08-21
Fix rollback lookup for canonical app version IDs

phinze:
Rollback selects an app version from deployment history, where normal deploys record the canonical `app_version/...` entity ID. The deployment service treated that value as a bare name and prefixed the kind again, so valid rollback targets came back as missing.

App-version lookup now accepts bare, short, and canonical references, then records canonical IDs going forward. History also resolves short IDs for older bare-name records, which makes mixed deploy and env-mutation rows render consistently. The same normalization keeps source-app ownership checks effective for canonical references.

Closes MIR-1635

### DIFF
--- a/api/app/envvar.go
+++ b/api/app/envvar.go
@@ -454,6 +454,7 @@ func createNewVersion(ctx context.Context, ec *entityserver.Client, resolver sec
 	if err != nil {
 		return nil, err
 	}
+	appVer.ID = avid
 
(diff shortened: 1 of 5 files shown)


evanphx (approved):
Approved! Nice to have rollback finding its own versions again. Traced it end to end — the canonical/bare handling holds on both the create and deploy paths, and the cross-app tests go red without the fix, which is the good kind.

--e+🤖

PR #1077, opened 2026-08-21
Wait for disk sandbox cleanup before starting replacements

phinze:
Disk-backed deployments currently treat STOPPED as if teardown were finished. But STOPPED only requests teardown; the sandbox controller may still be shutting down the process and releasing resources. This lets a replacement start against the same local disk while the retiring process is still using it. We hit this when a retiring process deleted a WAL its replacement had just created.

Keep the old pool draining until all its sandboxes reach DEAD, then create the replacement. The tests cover STOPPED blocking replacement, DEAD unblocking it, and the complete rapid-redeploy lifecycle.

Closes MIR-1641

### DIFF
--- a/controllers/deployment/launcher.go
+++ b/controllers/deployment/launcher.go
@@ -1627,8 +1627,9 @@ func (l *Launcher) findStalePoolsForService(
 	return stale, nil
 }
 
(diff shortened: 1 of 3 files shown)


evanphx (approved):
Approved! Nice and tight — the way it should've worked all along.

--e+🤖

PR #1079, opened 2026-08-24
Inherit base image environment in stack builds

evanphx:
When Miren auto-builds an app that has no Dockerfile, `pkg/stackbuild` assembles the final image config from scratch. It seeds the environment with just a default `PATH`, stacks layer on a few of their own vars, and that config gets exported verbatim. The catch: buildkit doesn't merge the base image's own config into a config you hand it explicitly, so every `ENV` the upstream base image's Dockerfile set was silently dropped from the finished image.

This bit us with Bun. Recent Bun images relocate `node` into a non-standard bin dir and prepend that dir to `PATH`. Because we overwrote `PATH` with our default, `node` vanished at runtime and anything that shelled out to it broke. And it wasn't just `PATH`: `LANG`, `GEM_HOME`, the language version vars, everything the official images declare was getting thrown away too.

The fix routes every stack's final base image through a new `baseImage` helper. It resolves the base image config and folds that environment into the final config, so the built image behaves like `docker run <base>` would. `PATH` is always taken from the base, since its directories are a superset of our default. Every other var fills in only where the stack hasn't deliberately set one, so Miren's own choices (Ruby's `BUNDLE_PATH`, and so on) still win regardless of ordering. Putting it in one helper is the point: inheritance is now automatic and uniform, and a stack added later gets it for free rather than having to remember to opt in.

One deliberate exception: the Go builder image doesn't inherit, since we don't want the toolchain's env leaking into the tiny distroless runtime image. Only the runtime base's env comes through there.

Closes MIR-1631

https://claude.ai/code/session_0197pxQ5MX8GgzCQdsgXfLiv

### DIFF
--- a/pkg/stackbuild/stackbuild.go
+++ b/pkg/stackbuild/stackbuild.go
@@ -1,6 +1,8 @@
 package stackbuild
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -10,6 +12,8 @@ import (
 
 	"github.com/containerd/platforms"
 	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/client/llb/imagemetaresolver"
+	"github.com/moby/buildkit/client/llb/sourceresolver"
 	"github.com/moby/buildkit/util/system"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"miren.dev/runtime/pkg/imagerefs"
@@ -17,7 +21,10 @@ import (
 
 // BuildOptions contains configuration for stack builds
 type BuildOptions struct {
-	Log interface{ Info(string, ...any) }
+	Log interface {
+		Info(string, ...any)
+		Warn(string, ...any)
+	}
 
 	// Name is the name of the application being built
 	Name string
@@ -55,7 +62,7 @@ type Stack interface {
 	// Init is called after detection to perform common initialization
 	Init(opts BuildOptions)
 	// GenerateLLB creates the BuildKit LLB for building this stack
-	GenerateLLB(dir string, opts BuildOptions) (*llb.State, error)
+	GenerateLLB(ctx context.Context, dir string, opts BuildOptions) (*llb.State, error)
 
 	Image() ocispecs.Image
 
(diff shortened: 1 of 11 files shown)


pkg/stackbuild/stackbuild.go
```diff
@@ -179,15 +183,108 @@ func (s *MetaStack) setupResult() {
...
+			opts.Log.Info("could not resolve base image config for env inheritance; keeping defaults", "ref", ref, "error", err.Error())
+		}
+		return
+	}
+
+	var img ocispecs.Image
+	if err := json.Unmarshal(cfg, &img); err != nil {
+		if opts.Log != nil {
+			opts.Log.Info("could not parse base image config for env inheritance; keeping defaults", "ref", ref, "error", err.Error())
+		}
```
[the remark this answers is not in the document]
evanphx: Good catch, and it was worse than misclassified: neither caller was wiring `BuildOptions.Log`, so these fallback lines never surfaced at all. Extended `Log` with `Warn`, switched both fallbacks (resolve + parse failure) to `Warn`, and passed `b.Log` from both callers so the degraded path is now actually visible at the right level. Folded into the existing commit.

pkg/stackbuild/stackbuild.go
```diff
@@ -179,15 +186,108 @@ func (s *MetaStack) setupResult() {
...
+	}
+	return ""
+}
+
+// baseImage builds the LLB state for a base image ref and inherits the image's
+// environment into the final image config. Every stack should build the base(s)
+// that become its final image through this helper so env inheritance is
+// automatic and uniform — a new stack gets it for free.
+func (s *MetaStack) baseImage(ctx context.Context, ref string, opts BuildOptions) llb.State {
+	mr := imagemetaresolver.Default()
```
phinze: 🤖 `imagemetaresolver.Default()` caches resolved image metadata for the lifetime of the process, keyed by reference and platform. Since these refs are mutable tags, a later build can get newly resolved layers from BuildKit while inheriting `Env` cached from an older image revision. That can recreate the kind of PATH/config mismatch this helper is meant to prevent.

Could we use `imagemetaresolver.New()` here? The resolver would still be shared between `inheritBaseEnv` and this build’s LLB marshal, but each build would resolve a coherent current config.

phinze (approved):
Nice fix overall! 🤖 caught a good caching thing we should fix before merge

pkg/stackbuild/stackbuild.go
evanphx: 🤖 Great catch, this is a real trap with the long-lived build server. Switched `baseImage` to `imagemetaresolver.New()` so each build gets a fresh cache and resolves a coherent current config, while still sharing that one resolver between `inheritBaseEnv` and the LLB marshal within the build. Also swapped the Go builder's resolver to `New()` for consistency (it doesn't feed the final `Env`, but same staleness reasoning). Folded into the existing commit.

PR #1080, opened 2026-08-24
Keep a compatibility window for `app run` across the durable-runs switch

evanphx:
When we absorbed `miren app run` into durable runs, we swapped its wire protocol in both directions at the same time. That left no working upgrade order across the boundary. A current CLI can't resolve the `app-runs` capability against a released v0.13 server, and a v0.13 CLI's old exec-by-app request gets rejected outright by a current server. Production runs v0.13, so this actually blocked `app run` against prod, and the only workaround was to keep an old CLI lying around.

This restores a compatibility window in both directions.

On the client side, when `app-runs` resolution comes back saying the capability is specifically absent (the server answered, it just doesn't have it), the CLI falls back to the legacy exec-by-app path. That's the exact request a released v0.13 server already knows how to serve, so an older cluster handles it with its own code. The distinction matters: a transport failure or an auth rejection is a real error and surfaces unchanged, rather than being misread as an old server and quietly retried against a healthy one.

On the server side, the current server accepts the old exec-by-app request again. Rather than restoring the leak-prone ephemeral sandbox that the durable-runs work deliberately removed, it translates the request onto a durable run, created with the caller's own identity so the app-scoping guard still means something. The run controller owns the sandbox's lifecycle, so a request handler that dies mid-stream leaks nothing. The features the old protocol can't express, `--task` and `--detach`, are refused with a clear message instead of silently dropped.

Both shims are temporary and commented as such, to be removed once v0.13 is out of the support window.

Closes MIR-1633

https://claude.ai/code/session_01SnGq1uY7vDZSkxhnw2r4rc

### DIFF
--- a/servers/exec_proxy/exec_proxy.go
+++ b/servers/exec_proxy/exec_proxy.go
@@ -20,26 +20,31 @@ import (
 	"miren.dev/runtime/pkg/rpc/stream"
 )
 
-// RunCreator records a durable run in-process and returns its id and the name
-// its first attempt's sandbox will have. The exec proxy uses it on the legacy
-// exec-by-app compatibility path so the run is created with the caller's own
-// identity rather than the coordinator's. *app.AppInfo satisfies it.
-type RunCreator interface {
+// RunManager creates and cancels durable runs in-process. The exec proxy uses
+// it on the legacy exec-by-app compatibility path so a run is created (and, when
+// abandoned, canceled) with the caller's own identity rather than the
+// coordinator's. *app.AppInfo satisfies it.
+type RunManager interface {
+	// CreateRunEntity records a run and returns its id and the name its first
+	// attempt's sandbox will have.
 	CreateRunEntity(ctx context.Context, app, task string, command []string, tty bool) (entity.Id, string, error)
+	// CancelRunEntity requests cancellation of a run, reporting whether the
+	// request was recorded (false if it was already terminal).
+	CancelRunEntity(ctx context.Context, runID string) (bool, error)
 }
 
 type Server struct {
 	Log  *slog.Logger
 	EAC  *entityserver_v1alpha.EntityAccessClient
 	rs   *rpc.State
-	runs RunCreator
+	runs RunManager
 }
 
 func NewServer(
 	log *slog.Logger,
 	eac *entityserver_v1alpha.EntityAccessClient,
 	rs *rpc.State,
-	runs RunCreator,
+	runs RunManager,
 ) *Server {
 	return &Server{
 		Log:  log,
@@ -182,13 +187,12 @@ func (s *Server) execByApp(ctx context.Context, req *exec_v1alpha.SandboxExecExe
 	args := req.Args()
 
 	var command []string
-	tty := false
-	if opts := args.Options(); opts != nil {
+	opts := args.Options()
+	if opts != nil {
 		command = opts.Command()
-		tty = opts.Terminal()
 	}
 
-	runID, sandboxName, err := s.runs.CreateRunEntity(ctx, args.Value(), "", command, tty)
+	runID, sandboxName, err := s.runs.CreateRunEntity(ctx, args.Value(), "", command, execTTY(opts))
 	if err != nil {
 		return err
 	}
@@ -201,12 +205,20 @@ func (s *Server) execByApp(ctx context.Context, req *exec_v1alpha.SandboxExecExe
 	outW := stream.ToWriter(ctx, args.Output())
 	defer outW.Close()
 
-	// winCh is fed once and read by a fresh ChanReader per attach attempt. While
-	// an attempt is returning early (not ready), nothing drains it, so a resize
-	// that arrives in that gap can be dropped. That is harmless here: the next
-	// successful attach reads the current terminal size anyway, and this is a
-	// compatibility shim rather than a long-lived interactive session.
+	// winCh is read by a fresh ChanReader per attach attempt. ChanReader is lazy,
+	// so a not-ready attempt does not drain it and a queued resize survives to the
+	// next attempt.
+	//
+	// A v0.13 client reports its initial terminal size through the shell options,
+	// not the update stream (its stream only carries later SIGWINCH events), so
+	// seed that size as the first resize -- the old exec server read the same
+	// field to size the pty. Seed before wiring the update stream so it is first
+	// in line; the buffer of one holds it until the first successful attach reads
+	// it.
 	winCh := make(chan *exec_v1alpha.WindowSize, 1)
+	if opts != nil && opts.HasWinSize() {
+		winCh <- opts.WinSize()
+	}
 	if args.HasWindowUpdates() {
 		stream.ChanWriter(ctx, args.WindowUpdates(), winCh)
 	}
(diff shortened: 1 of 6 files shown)


servers/exec_proxy/exec_proxy.go
```diff
@@ -139,6 +158,250 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+// it would report a failing command as success to whatever invoked the old CLI.
+// A run that ended without a task exit code (canceled, timed out) reports 0
+// here; its status, not this number, carries that meaning, and the old protocol
+// has nowhere to put it.
+func (s *Server) runExitCode(ctx context.Context, runID entity.Id) (int32, error) {
+	deadline := time.Now().Add(runSettleTimeout)
+	for {
+		run, err := s.getRun(ctx, runID)
+		if err == nil {
+			if reportsExitCode(run.Status) && !run.Result.At.IsZero() {
```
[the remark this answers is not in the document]
evanphx: Good call. Added a Warn log of the last read error on deadline expiry (with the run id), so a corrupt or permanently-missing run entity leaves a trace instead of being an indistinguishable zero-exit. Return value unchanged.

servers/exec_proxy/exec_proxy.go
```diff
@@ -139,6 +158,250 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+				return fmt.Errorf("failed to connect to node %s: %w", node.ApiAddress, err)
+			}
+
+			sec := &exec_v1alpha.SandboxExecClient{Client: rcl}
+			_, aerr := sec.Attach(
+				ctx,
+				sandboxName, "",
+				stream.ServeReader(ctx, inR),
+				stream.ServeWriter(ctx, outW),
+				stream.ChanReader(winCh),
```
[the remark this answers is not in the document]
evanphx: Added a comment explaining it: winCh is fed once and read by a fresh ChanReader per attempt, so a resize arriving while an attempt returns early can be dropped, which is harmless because the next successful attach reads the current terminal size anyway.

servers/exec_proxy/exec_proxy.go
```diff
@@ -139,6 +158,250 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+	deadline := time.Now().Add(runSettleTimeout)
+	for {
+		run, err := s.getRun(ctx, runID)
+		if err == nil {
+			if reportsExitCode(run.Status) && !run.Result.At.IsZero() {
+				return clampInt32(run.Result.Code), nil
+			}
+			if isTerminalRunStatus(run.Status) {
+				return 0, nil
+			}
```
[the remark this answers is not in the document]
evanphx: I looked into this and I don't think the intermediate state is reachable, and the proposed fix would introduce a hang.

The run controller writes status and result in a single atomic patch (controllers/run/controller.go:300-312): when a code exists, Status and Result{Code, At} land at the same entity revision. So a terminal SUCCEEDED/FAILED run that carries a code always has Result.At set at the revision a reader sees it, there is no torn read where the status is terminal but the code is still missing. The settle-wait the function needs is already covered because during the container-exit-to-finish gap the status is still RUNNING (not terminal), so the loop keeps polling and only returns once the atomic finish lands.

The proposed change also regresses a legitimate case: a run that ends terminal with no command exit code (finish(FAILED, nil) when the sandbox disappears, controller.go:183) has reportsExitCode(FAILED) == true but a Result.At that will never be set. Under the diff it would poll the full runSettleTimeout (15s) before returning 0, instead of returning immediately.

So I'm keeping the current shape. I did add a Warn log on deadline expiry (from a related review note) so a persistent read failure is at least traceable.

servers/exec_proxy/exec_proxy.go
```diff
@@ -139,6 +158,266 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+// ephemeral sandbox this replaces, an abandoned run stays owned by the run
+// controller, so a handler that dies mid-request leaks nothing.
+func (s *Server) execByApp(ctx context.Context, req *exec_v1alpha.SandboxExecExec) error {
+	args := req.Args()
+
+	var command []string
+	tty := false
+	if opts := args.Options(); opts != nil {
+		command = opts.Command()
+		tty = opts.Terminal()
```
phinze: The v0.13 client never sets `ShellOptions.Terminal`; its `setupExecIO` signals that the remote process needs a PTY by setting `WinSize`, which the old exec server uses to set `proc.Terminal` and the initial console dimensions. This path reads only `Terminal`, so it creates a non-TTY run and discards those dimensions for every interactive v0.13 request. I smoke-tested the exact v0.13.0 binary against this head: remote `test -t 0` reported false and `stty size` failed with “Not a tty.” Could we derive the compatibility TTY signal from `WinSize` and pass its initial dimensions into the attach resize stream?

--p+🤖

servers/exec_proxy/exec_proxy.go
```diff
@@ -139,6 +158,266 @@ func (s *Server) Exec(ctx context.Context, req *exec_v1alpha.SandboxExecExec) er
...
+
+		// A short command can finish before the terminal ever attaches, taking
+		// its container with it. That is not a failure -- the run did what it was
+		// asked and recorded an outcome -- so stop waiting and report it.
+		if s.runIsTerminal(ctx, runID) {
+			break
+		}
+
+		if time.Now().After(deadline) {
+			return fmt.Errorf("run %s for app %s did not start within %s", runID, args.Value(), execByAppTimeout)
```
phinze: This returns an error to the legacy client but leaves the durable run alive, so the command can execute after the caller was told it failed. I smoke-tested this by cordoning the only runner: v0.13.0 received this exact two-minute timeout while the run remained live; after I uncordoned the runner, that same run succeeded and its logs recorded `DELAYED_EXECUTED`. The legacy client has no run ID to inspect or cancel, and a natural retry can therefore execute the command twice. Could we cancel the run before returning this timeout error?

--p+🤖

phinze (approved):
Two things 🤖 found that we should fix before merge. Otherwise looking good!

servers/exec_proxy/exec_proxy.go
evanphx: 🤖 Good catch, and thanks for smoke-testing the actual v0.13.0 binary. Fixed: `execTTY` now derives the pty signal from `WinSize` presence (what the old exec server keyed on) rather than `Terminal`, and we seed that initial `WinSize` into the attach resize stream so the terminal comes up sized. Added a unit test covering the WinSize-only signal. In `697b8cd9`.

servers/exec_proxy/exec_proxy.go
evanphx: 🤖 Agreed, that's the double-execution hazard the durable design exists to prevent. Fixed: we now cancel the run before returning the timeout error, mirroring what the CLI's `attachToRun` does for a run that never started (`cancelIfNeverStarts`). Needed a small `CancelRunEntity` core extracted from `AppInfo.CancelRun` so the proxy can cancel with the caller's identity. In `697b8cd9`.

PR #1081, opened 2026-08-24
Mount accelerator disks without udev; shore up disk restore testing

evanphx:
## What

Three related changes that let accelerator-mode disks mount and restore reliably, and make the disk restore path actually exercisable in tests.

- **`diskio`: create the lbd device node when udev is absent.** `LbdAttach` relied on udev to materialize `/dev/lbdN` after `lbdctl add`. In a container whose `/dev` is a private tmpfs there is no udev, so the node never appeared and the follow-up format/mount failed with `ENOENT`. We now read the `major:minor` the kernel reports in `/sys/block/<name>/dev` and `mknod` the node ourselves when it's missing.
- **`e2e_disk`: set `CloudVolumeId` so the restore step runs.** `restoreImageIfMissing` early-returns when `CloudVolumeId` is empty, which it always was in the driver, so the restore step silently did nothing (and the later `ReadFile` failed). Passing the volume id through makes the round trip actually exercise restore.
- **`testdata/miren-disk-app`: a managed-disk fixture.** A minimal app that declares a `provider = "miren"` disk and stores a payload at `/data`, exposing its sha256 on `GET` so a restore can be verified.

## Why

Without the `mknod`, accelerator disks can't mount in the iso dev container (no udev), so the whole managed-disk path was untestable there.

## Validation

Exercised end to end against a real local cloud, driven entirely through the normal deploy path (`provider = "miren"` disk, no debug commands):

1. **Create** — deploy auto-creates the disk in accelerator mode, provisions it, registers the volume with the cloud.
2. **Mount** — auto-leased and mounted via `/dev/lbd1` (node created by this change), formatted, mounted at `/data`.
3. **Backup** — app writes 80 MiB over HTTP → real lbd log segments → LogWatcher uploads them to the cloud.
4. **Restore** — simulate node loss (zero the local image + reset the horizon) → remount replays the segments from the cloud → `GET /data` returns the original sha256, byte-identical.

## Companion fix

The real lbd segment path also needed a kernel fix in the `lbd` repo (`vfs_rename` faulted on NULL parent dentries during segment rotation on newer kernels); that's merged separately. Both are required for accelerator rotation + restore to work on kernels that carry `old_parent`/`new_parent` in `struct renamedata`.

## Not included

The dev container needs `lbdctl` on its `PATH` for accelerator mode; I left that out of this PR (it was a local `.iso/config.yml` bind) since it wants a proper answer — baking `lbdctl` into the `.iso` image — rather than a bind-mount.

### DIFF
--- a/components/diskio/disk_ops_linux.go
+++ b/components/diskio/disk_ops_linux.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
@@ -258,7 +259,11 @@ func (r *realDiskMountOps) LbdAttach(ctx context.Context, imagePath, logDir stri
 		// lbdctl add already attached the device, but the caller receives
 		// only an error (no path), so it cannot detach. Clean up here, or a
 		// repeated failure leaks a kernel device with no state to reclaim it.
-		if derr := r.LbdDetach(ctx, result.Device); derr != nil {
+		// Give the detach an independent context so it can still complete even
+		// if the caller's ctx has already been canceled.
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		if derr := r.LbdDetach(cleanupCtx, result.Device); derr != nil {
 			return "", fmt.Errorf("lbdctl add: %w (detach after failure also failed: %v)", err, derr)
 		}
 		return "", fmt.Errorf("lbdctl add: %w", err)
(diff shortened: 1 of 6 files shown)


components/diskio/disk_ops_linux.go
```diff
@@ -249,9 +250,60 @@ func (r *realDiskMountOps) LbdAttach(ctx context.Context, imagePath, logDir stri
 		return "", fmt.Errorf("lbdctl add returned empty device path")
 	}
 
+	// The kernel registers the block device in sysfs, but the /dev node is
+	// normally created by udev. Containers whose /dev is a private tmpfs have
+	// no udev, so the node never appears and the mount that follows fails with
+	// ENOENT. Create it ourselves from the major:minor sysfs reports.
+	if err := ensureDeviceNode(result.Device); err != nil {
```
phinze, line 258: 🤖 This failure path needs to remove the device that `lbdctl add` just created before returning, otherwise each repeat of this failure leaves another kernel device attached with no state that can clean it up. There is a wrinkle in using `LbdDetach` for that cleanup: it currently passes the full `/dev/lbdN` path to `lbdctl remove`, but `lbdctl` expects a numeric index and parses its argument with `atoi`. For example, `/dev/lbd1` becomes index 0, so cleanup can remove the wrong volume. Please fix the detach invocation to pass the actual index, then use it here while preserving the node-setup error.

phinze (approved):
Good change overall! My 🤖 found an issue that seems legit. I'm approving since I trust you'll address it but recommend treating that like a blocker.

components/diskio/disk_ops_linux.go
evanphx: 🤖 Fixed in bae1889d. `LbdDetach` now extracts the numeric index and passes that to `lbdctl remove` — it was handing over the `/dev/lbdN` path, which atoi turned into 0, detaching the wrong volume. And `LbdAttach` now calls `LbdDetach` when `ensureDeviceNode` fails (preserving the original error), so a repeated failure no longer leaks the attached device.

PR #1083, opened 2026-08-25
Stop hardcoding DNS for the builder

evanphx:
The embedded BuildKit daemon pinned every build's DNS to `1.1.1.1` and `8.8.8.8`. It did this through a `[dns]` block in the generated `buildkitd.toml`, which BuildKit writes into the resolv.conf of each build sandbox it spawns for `RUN` steps. That's fine on a plain internet connection, but on a network that needs a specific internal resolver (split-horizon DNS, a private package mirror, a corporate network) it meant builds couldn't resolve internal names and quietly sent every lookup off to public resolvers.

Reading the vendored BuildKit source showed that when no nameservers are configured, buildkitd already does the right thing: it derives each build sandbox's resolv.conf from the host, systemd-resolved aware, strips loopback entries that wouldn't be reachable from the sandbox's network namespace, and falls back to public DNS only if nothing usable remains. So the hardcoded block wasn't adding anything, it was the one thing overriding that behavior. The fix is to delete it and let BuildKit pick up the host's DNS.

There was no blackbox test exercising the builder's DNS, so this adds one. The `build-dns` fixture's Dockerfile runs `apk add`, which only succeeds if the builder handed the sandbox a working, reachable resolver. Deploy it and the build is the assertion. There's also a small unit test guarding that the generated config no longer carries hardcoded nameservers, so nobody reintroduces the block.

One behavior note: with the block gone, the empty-list fallback becomes libnetwork's default (`8.8.8.8`/`8.8.4.4`) instead of the old `1.1.1.1`/`8.8.8.8`, which only matters on a host whose only resolver is loopback.

Closes MIR-1643

### DIFF
--- a/components/buildkit/buildkit.go
+++ b/components/buildkit/buildkit.go
@@ -321,10 +321,6 @@ insecure-entitlements = [ "network.host", "security.insecure" ]
 [log]
   format = "text"
 
-[dns]
-  nameservers=["1.1.1.1","8.8.8.8"]
-  options=["edns0"]
-
 [grpc]
   address = [ "unix:///run/buildkit/buildkitd.sock" ]
   uid = 0
--- a/components/buildkit/config_test.go
+++ b/components/buildkit/config_test.go
@@ -86,4 +86,14 @@ func TestGenerateConfig(t *testing.T) {
 		defaultConfig := c.generateConfig(10*1024*1024*1024, 86400, "")
 		r.Contains(defaultConfig, `[registry."cluster.local:5000"]`)
 	})
+
+	t.Run("does not hardcode DNS nameservers", func(t *testing.T) {
+		r := require.New(t)
+		// MIR-1643: DNS is delegated to buildkitd, which derives each build's
+		// resolv.conf from the host. A hardcoded nameserver directive would
+		// override that and break builds on hosts with an internal resolver.
+		r.NotContains(config, "nameservers=")
+		r.NotContains(config, "1.1.1.1")
+		r.NotContains(config, "8.8.8.8")
+	})
 }


blackbox/build_dns_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+package blackbox
+
+import (
+	"testing"
+
+	"miren.dev/runtime/blackbox/harness"
+)
+
+// TestBuildResolvesDNS exercises the builder's build-time DNS end to end. The
```
phinze: This doesn’t exercise the regression: with the old `[dns]` block, CI can still reach Alpine through the pinned public resolvers, so this deploy passes before and after the change. `TestDeployImageEntrypointNoCommand` already builds `alpine:3.20` and runs `apk add --no-cache python3`; it passed on the preceding PR with the old config. Could we drop this test and its fixture? It adds another blackbox deploy while only proving that some DNS works, not that BuildKit inherited the host resolver.

--p+🤖

components/buildkit/buildkit.go
```diff
@@ -321,9 +321,11 @@ insecure-entitlements = [ "network.host", "security.insecure" ]
...
   format = "text"
 
-[dns]
-  nameservers=["1.1.1.1","8.8.8.8"]
-  options=["edns0"]
+# DNS is intentionally left unconfigured here: buildkitd derives each build's
+# resolver list from the host (systemd-resolved-aware), drops loopback entries
+# for the sandbox netns, and falls back to public DNS if none remain. Pinning
+# resolvers here would override that and break builds on hosts with an internal
+# resolver (MIR-1643).
```
phinze: Since host-local resolver reachability depends on the worker using host networking, could we make the current behavior explicit with `networkMode = "host"` under `[worker.oci]`? The shipped image already auto-selects host mode; pinning it prevents a future CNI config from silently changing resolver reachability.

Also this is a nit but, could we drop this comment? The rationale belongs in the commit message, and the loopback/sandbox explanation describes BuildKit’s non-host path rather than Miren’s actual host-network worker.

--p+🤖

phinze (approved):
Looks like the right fix! A few tweaks recommended inline.

components/buildkit/buildkit.go
evanphx: I'd want to do more research about the impact of networkMode = "host", but we can consider that in the future.

phinze: Fair to pause on that! I had misunderstood the bot finding as "encoding the default into the config, to record our intentional choice / dependency on host for DNS" but peeking at the code it's a slight change from `auto` to `host`.

Here's the buildkit code for it, where auto does selection logic and host just sticks at host:

https://github.com/mirendev/buildkit/blob/fbd602b3d268ea5be08f05be64f8f9adc8c48ebc/util/network/netproviders/network.go#L30-L61

For us `auto` ends up resolving to `host` because we don’t enable bridge auto-selection and the shipped image has no CNI config. Pinning `host` would preserve today’s behavior, and prevent a future CNI config from taking effect automatically, which on further thought is probably the opposite of our likely future direction, which will be to isolate the build network further prolly via CNI.

Good to drop it!

blackbox/build_dns_test.go
evanphx: 🤖 Agreed, dropped the test and its fixture. You're right it only proved that some DNS works, not that BuildKit inherited the host resolver, since the old pinned resolvers reach Alpine on CI just the same. `TestDeployImageEntrypointNoCommand` already covers a build doing DNS, and the unit test asserting no hardcoded nameservers is the real guard against reintroduction.

components/buildkit/buildkit.go
evanphx: 🤖 Good catch, dropped the comment. It described BuildKit's sandbox/CNI path (loopback stripped) rather than our host-network worker, where the host's resolv.conf is used as-is, so it was misleading. Rationale lives in the commit message now. Leaving `networkMode = "host"` for follow-up research per evanphx's note above.

PR #1085, opened 2026-08-26
Negotiate capabilities before starting uplink tenants

phinze:
The runtime currently starts uplink tenants as soon as the WebSocket opens, before cloud has supplied identity, time, or a reliable statement of what the connection supports.

This adds the runtime half of the RFD-112 hello/welcome handshake. Capability offers stay owned by their tenants: Miren Anywhere offers POP connect and the cloud RPC package offers the relay, while future entity sync can carry its schema offer and config without widening the bootstrap.

The negotiated path stays behind the disabled `appvisibility` labs flag, and the default path preserves the existing time and organization requests. That lets cloud support deploy first and gives us an internal rollout before any runtime requires a welcome.

Part of MIR-1587

### DIFF
--- a/pkg/uplink/client.go
+++ b/pkg/uplink/client.go
@@ -23,11 +23,12 @@ const (
 	// backpressure does not immediately push the best-effort tenants over their
 	// drop threshold. The two coexist in one queue: a blocked sender waits for
 	// room, while Send discards as soon as there is none.
-	outboxSize     = 256
-	initialBackoff = 1 * time.Second
-	maxBackoff     = 60 * time.Second
-	wsEndpoint     = "/api/v1/cluster-channel/ws"
-	writeTimeout   = 10 * time.Second
+	outboxSize       = 256
+	initialBackoff   = 1 * time.Second
+	maxBackoff       = 60 * time.Second
+	wsEndpoint       = "/api/v1/cluster-channel/ws"
+	writeTimeout     = 10 * time.Second
+	handshakeTimeout = 10 * time.Second
 
 	// backoffJitter is the fraction of a delay that is randomized away.
 	//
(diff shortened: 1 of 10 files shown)


pkg/uplink/client.go
```diff
@@ -391,6 +471,53 @@ func (c *Client) runOnce(ctx context.Context) error {
...
+	}
+	receivedAt := time.Now().UTC()
+
+	switch env.Type {
+	case TypeSessionReject:
+		var reject SessionReject
+		if err := json.Unmarshal(env.Data, &reject); err != nil {
+			return Session{}, nil, fmt.Errorf("decode session rejection: %w", err)
+		}
+		return Session{}, nil, fmt.Errorf("cloud rejected session: %s", reject.Reason)
```
evanphx, line 510: Non-blocking: `SupportedHandshakeVersions` rides along in the reject but nothing here reads it, so a version mismatch just becomes an endless reconnect loop. That's the right call for v1 with a single handshake version — flagging it only to confirm it's deliberate and not a loose end for when a version 2 shows up.

pkg/uplink/session.go
```diff
@@ -0,0 +1,141 @@
...
+		return Session{}, fmt.Errorf("session welcome has incomplete clock timestamps")
+	}
+
+	offered := make(map[string]CapabilityOffer, len(hello.Capabilities))
+	for _, capability := range hello.Capabilities {
+		if capability.Name == "" {
+			return Session{}, fmt.Errorf("session hello contains unnamed capability")
+		}
+		if _, exists := offered[capability.Name]; exists {
+			return Session{}, fmt.Errorf("session hello contains duplicate capability %q", capability.Name)
```
evanphx, line 112: Non-blocking: a duplicate offer — two tenants naming the same capability, or one offering twice — makes `validateWelcome` reject every welcome, so the uplink never connects in negotiated mode. Today's two offers are distinct so it can't fire, but a self-inflicted config error surfacing as a silent reconnect loop is cheaper to catch at `OfferCapability` time than on the wire.

pkg/uplink/client.go
```diff
@@ -187,6 +236,19 @@ func (c *Client) connectCallbacks() []func(ctx context.Context) {
...
 }
 
+func (c *Client) sessionSnapshot() ([]CapabilityOffer, []func(context.Context, Session)) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	offers := make([]CapabilityOffer, len(c.capabilities))
+	for i, offer := range c.capabilities {
+		offers[i] = offer
+		offers[i].Versions = slices.Clone(offer.Versions)
```
evanphx, line 251: 👏 the per-offer deep copy is subtle and easy to get wrong.

evanphx (approved):
Approved! Nice that the tenants get to say who they are before they start talking. This reads clean and everything I traced holds up.

Two non-blocking notes inline (reject-version and duplicate-offer) plus one bit of applause — none of them block the merge.

--e+🤖

pkg/uplink/client.go
phinze: 🤖 Deliberate. A no-common-version rejection means this binary has already offered everything it can speak, so the returned list is diagnostic rather than an alternate retry. We keep reconnecting because cloud is independently deployed and may become compatible again. A future v2 should revisit logging and backoff here, but there is no different hello this process can send today.

pkg/uplink/session.go
phinze: 🤖 Good catch. I made duplicate capability names panic at registration time, where they are a startup programming error, instead of allowing a hello that can only reconnect-loop. The wire-side validation remains as defense against malformed state.

pkg/uplink/client.go
phinze: 🤖 Thanks! I wanted the session snapshot to stay genuinely connection-scoped even if a tenant retained and mutated its offer slices.

pkg/uplink/client_test.go
```diff
@@ -346,6 +400,134 @@ func TestOnConnectContextCancelledOnDisconnect(t *testing.T) {
...
+	WithSession("v1.2.3")(client)
+	client.OfferCapability(CapabilityOffer{Name: CapabilityPopConnect, Versions: []uint{1}})
+	client.OnSession(func(_ context.Context, session Session) { sessionStarted <- session })
+	client.OnConnect(func(context.Context) { connectStarted <- struct{}{} })
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	go client.Run(ctx) //nolint:errcheck
+
+	hello := <-helloRead
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed. The test now selects between `helloRead` and `ctx.Done()`, so a missing hello fails at the local two-second deadline instead of the package timeout.

PR #1087, opened 2026-08-27
Forward managed application metrics from the runtime

phinze:
Applications can now opt a service into managed metrics without exposing its endpoint publicly. The coordinator discovers each running sandbox, scrapes it privately with resource limits, and sends samples to the configured Prometheus Remote Write destination using a rotating workload identity.

Miren owns the canonical app, version, service, sandbox, runner, and cluster labels. Registered clusters use their cloud ID, while standalone clusters use `server.config_cluster_name`, giving shared destinations a stable identity. Scrape and delivery failures appear in `miren logs system vmagent`.

A reusable two-replica smoke app and inspection receiver make the happy path and principal failure modes repeatable. Focused Go and race tests, lint, generation, docs build, and the dev smoke pass. The vmagent proxy alias is already live and recorded in mirendev/infra#105.

Closes MIR-1654


servers/httpingress/httpingress.go
```diff
@@ -630,14 +639,138 @@ func (h *Server) serveHTTPWithMetrics(w http.ResponseWriter, req *http.Request,
 	// A wildcard route also serves as a multi-tenant subdomain, so a failed
 	// ephemeral lookup should fall back to the active version rather than 404.
 	wildcardRoute := route != nil && ingress.IsWildcardHost(route.Host)
+	target, ok := h.resolveIngressTarget(w, req, targetAppId, ephemeralLabel, wildcardRoute)
```
evanphx: Hoisting `resolveIngressTarget` above `buildRouteHandler` moves the app/version/config resolution ahead of the maintenance middleware, which short-circuits on `route.Maintenance` alone.

So a route in maintenance whose active version can't resolve right now — cleared, or a transient `eac.Get` on the ConfigVersion — hits the `ActiveVersion == ""` 404 or the config-resolve 500 here and returns before maintenance ever runs. The visitor gets a raw error instead of the 503 holding page, in the window maintenance is there to cover. The isolated maintenance tests drive the middleware directly, so they don't catch it.

The private-metrics 404 doesn't need to sit ahead of maintenance — a holding page for `/metrics` leaks nothing — so the resolution and check could move below the maintenance middleware. Non-blocking, but wanted it on your radar.

evanphx (approved):
Approved! Really nicely done.

One inline note on `httpingress.go`, non-blocking — hoisting the version resolution above the middleware chain opens a gap for routes in maintenance. Your call on fix-now vs. track it.

--e+🤖

servers/httpingress/httpingress.go
phinze: 🤖 Good catch, confirmed. We moved target resolution behind WAF and maintenance while keeping the private-metrics check ahead of auth, and added a route-level regression test with an unresolvable app.

appconfig/appconfig.go
```diff
@@ -899,6 +960,56 @@ func (ac *AppConfig) ResolveDefaults(services []string) {
...
+	}
+
+	if svc.Port > 0 && (svc.PortType == "" || svc.PortType == "http") {
+		return svc.Port
+	}
+	if serviceName == "web" {
+		return 3000
+	}
+	return 0
+}
```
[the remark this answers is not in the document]
phinze: Fixed in abd579f4. Untyped entries in `ports` now participate in metrics port defaulting as HTTP, with a regression test covering the omitted-type form. `go test`, the race detector, and lint all pass for `./appconfig`.

phinze:
@coderabbitai review

PR #1088, opened 2026-08-27
Shard the distributed blackbox suite in CI

evanphx:
## Why

`test-blackbox-distributed` runs the whole blackbox suite serially against a coordinator+runner topology in a single job — ~20 minutes, right at the edge of its timeout, and it gates every release.

Measuring a green run (CI run 32993856200) showed where the time actually goes:

| Phase | Time | Share |
|-------|------|-------|
| Infra setup (checkout, Go, iso, cache restore, docker load) | ~61s | 5% |
| `make dev-distributed` (peers up + build + join) | ~53s | 4% |
| **Serial test execution — 67 tests, `-p 1`** | **~1028s** | **~87%** |

So the wall clock is almost entirely serial test execution. The fix is to shard the suite across parallel jobs, the same way the standalone blackbox suite already does — the sharding tooling was written with a `peers` environment dimension for exactly this.

## What changed

- **`hack/measure-blackbox-times.py`** — peers-mode timing (`BLACKBOX_MODE=peers` / `--env`), rows tagged by environment and **merged** rather than overwritten so one file holds both the standalone and peers pools. Also parses `gh run --log` dumps, so timings can be seeded straight from a CI run.
- **`hack/blackbox-test-times.json`** — seeded with 67 peers timings from a green distributed run (33 standalone rows preserved).
- **`hack/calc-blackbox-groups.py`** — exclude the cloud-backed tests (`TestRPCViaCloud`, `TestDeployViaCloud`) in every environment, matching the `-skip` list; `-list` discovery would otherwise fold them into a shard.
- **`hack/blackbox-groups-distributed.json`** (new) — four balanced shards, ~257s each. No duplicate assignments; zero ungrouped tests.
- **`Makefile`** — `test-blackbox-distributed` takes a `BLACKBOX_RUN` subset (whole suite locally, one shard's subset in CI); adds `blackbox-groups-distributed` and `measure-blackbox-times-distributed`.
- **`.github/workflows/test.yml`** — a `blackbox-groups-distributed` builder job owns the run/skip gate (preserving the `!cancelled()` release-coverage guarantee and the `ci:distributed` opt-in), `test-blackbox-distributed` fans out over its shard matrix, and the all-green gate is updated so neither a builder failure nor a shard failure slips through.

**Projected ~20 min → ~6 min per shard (N=4), full suite coverage preserved.**

## Testing this PR

The `changes` filter already flags this PR as distributed-relevant (it touches `test.yml`), and the `ci:distributed` label is applied — so the sharded distributed job runs here. Watching for: the matrix fans out to four shards, each green, each well under the old 20-minute wall clock.

## Follow-ups (not in this PR)

- Build the miren binary once per `dev-distributed` bring-up instead of once per peer. Marginal per the measurement above (setup is a small slice of the job) and needs a live peers env to verify `/src` sharing and the `rebuild` path, so it's left as a separate change.
- Cache the compiled binary across shards; auto-refresh peers timings from CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -466,33 +466,81 @@ jobs:
       if: always()
       run: make dev-stop || true
 
-  test-blackbox-distributed:
-    # Exercises the multi-node distributed runner topology (a coordinator plus a
-    # separate runner peer) with distributedrunners enabled, running the whole
-    # blackbox suite against it. Setup is heavy (iso peers, a binary build on
-    # each peer, the join dance), so it runs on main/release builds rather than
-    # every PR push: release.yml (push to main and tags) calls this workflow via
-    # workflow_call, where github.event_name is the caller's event ('push'), so
-    # excluding 'pull_request' runs it there while skipping PRs. It would also
-    # run in a merge queue if one were enabled ('merge_group').
+  blackbox-groups-distributed:
+    # Builds the shard matrix for the distributed suite and owns the decision of
+    # *whether* that suite runs at all — the shard jobs below gate on this job's
+    # result, so the condition lives here in one place.
     #
-    # On a PR it runs when the diff touches code that can break the multi-node
-    # topology, or when someone labels the PR `ci:distributed` to ask for it
-    # explicitly. This job is the only place a second node exists, so a bug
-    # there is invisible until it is already on main blocking a release, which
-    # is exactly how MIR-1469 got in.
+    # The distributed suite exercises the multi-node topology (a coordinator plus
+    # a separate runner peer) with distributedrunners enabled. Setup is heavy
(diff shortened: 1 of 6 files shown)


phinze (approved):
We spent a brief moment evaluating the utility of running the whole dang blackbox suite twice. The history makes a surprisingly strong case for it: since the distributed job landed, five PR runs have had green standalone shards but failed ordinary blackbox tests under peers, and MIR-1469 was likewise exposed by ordinary `TestDiskUndelete`, not a `TestDistributed*` case. Preserving that breadth seems worthwhile, and this four-way split brings its wall-clock cost back into a reasonable range. The matrix construction and skip/failure propagation look sound, and the latest shard run is green.

nice speedup!

--p+🤖

PR #1089, opened 2026-08-27
Scope sandbox saga recovery to its owning runner

phinze:
Every sandbox executor starts against the same durable saga store. That meant a runner restart could claim another runner’s incomplete `create-sandbox` execution and continue it against the wrong local networking and container runtime.

This persists a recovery scope on each execution and scopes sandbox executors to their node ID. Recovery and named re-entry reject mismatches before claiming work. Normal routed reconciliation adopts legacy unscoped records, so upgrades preserve standalone and single-runner behavior without making an empty scope a recovery wildcard. Scope also appears in the saga debug output.

In the distributed rig, restarting the non-owner left an interrupted saga untouched. Restarting the owner then resumed it from the next action and brought the sandbox up.

Closes MIR-1657


evanphx (approved):
Approved! Clean, focused one — and the recovery test nails exactly the behavior that matters. Saw the CodeRabbit def-name catch; already handled in the head commit, so nothing outstanding from me.

--e+🤖

PR #1090, opened 2026-08-27
Deploy configured images without BuildKit

phinze:
An app can already configure `services.web.image`, but deploying one still tried to detect source and launch BuildKit unless a Dockerfile was present. That made a complete runnable image insufficient on its own.

This treats an explicit web image as build-source intent and runs it directly. Dockerfiles still win, while a lone non-web service image remains a fallback after source detection so a database sidecar cannot hijack a normal source build. Direct-image versions reference the normalized upstream image and have no synthetic Artifact.

Analyze and deploy now report `image` and `Use image` rather than pretending a zero-step build was cached. The deployment, service, and `app.toml` docs spell out the same precedence. Both saga and legacy paths are covered, with a config-only nginx blackbox deploy exercising the full path.

Closes MIR-1452
Closes MIR-1684

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -35,6 +35,7 @@ import (
 	"miren.dev/runtime/components/netresolve"
 	"miren.dev/runtime/observability"
 	"miren.dev/runtime/pkg/cond"
+	"miren.dev/runtime/pkg/containerdx"
 	"miren.dev/runtime/pkg/deploylifecycle"
 	"miren.dev/runtime/pkg/entity"
 	ephemeralx "miren.dev/runtime/pkg/ephemeral"
@@ -1337,7 +1338,7 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
 	}
 
 	if status != nil {
-		so.Update().SetMessage("Launching builder")
+		so.Update().SetMessage("Preparing deployment")
 		_, _ = status.Send(ctx, so)
 	}
 
@@ -1489,7 +1490,7 @@ func (b *Builder) BuildFromPrepared(ctx context.Context, state *build_v1alpha.Bu
 
 	if status != nil {
 		so := new(build_v1alpha.Status)
-		so.Update().SetMessage("Launching builder")
+		so.Update().SetMessage("Preparing deployment")
 		_, _ = status.Send(ctx, so)
 	}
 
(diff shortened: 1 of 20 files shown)


servers/build/build_saga_buildkit.go
```diff
@@ -65,6 +66,12 @@ func buildImage(ctx context.Context, in buildImageIn) (buildImageOut, error) {
 	b := deps.builder
 	status := deps.statuses.SenderFor(in.StreamID)
 
+	if in.BuildStack.Stack == "image" {
+		image := containerdx.NormalizeImageReference(in.BuildStack.Input)
+		status.SendImage(image)
+		return buildImageOut{FinalImageURL: image}, nil
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 Verified and covered in the latest push. Deployment phases are progress labels, not required state transitions: SetPhase only accepts labels while the record is in_progress, and FailIfUnsettled keys off status rather than phase. We intentionally keep direct-image deploys from reporting building or pushing because neither happens. TestBuildSaga_TrackedImage_ActivatesWithoutBuildPhases now proves the record advances from preparing to activating, records the version, reaches active, and releases its lock.

servers/build/build.go
```diff
@@ -2163,27 +2182,49 @@ func (b *Builder) AnalyzeApp(ctx context.Context, state *build_v1alpha.BuilderAn
 		stackName = "dockerfile"
 	}
 
-	// Always try to detect stack for analysis (env vars, events, etc.)
-	// even when dockerfile is present
+	// Match deploy's source precedence. An explicit Dockerfile wins; an explicit
```
evanphx: Non-blocking, and filed as MIR-1684 so it's tracked rather than lost. This precedence ladder is walked independently from `detectBuildStack` on the deploy path. The two share `webImageSource`/`fallbackImageSource`, but the ordering itself lives in both places with no test asserting they agree — and they do agree today, right down to checking both `Dockerfile.miren` and `[build].dockerfile`. Worth collapsing to one resolver, or a table test run through both, so `--analyze` can't quietly drift from what deploy actually picks.

evanphx (approved):
Approved! Nice that a named image just runs now, the way pointing at one always implied it would.

One non-blocking note inline, filed as MIR-1684 so it stays tracked rather than lost. The build-phase-skip question the bots raised is settled — the new tracked-image test pins it down.

--e+🤖

servers/build/build.go
phinze: 🤖 Good catch. This was small enough to fold in. Analyze and deploy now both call `resolveBuildSource`; deploy turns an unresolved source into an error, while analyze reports it as `unknown`. The refactor also uncovered a real mismatch: when both `[build].dockerfile` and `Dockerfile.miren` were present, deploy chose the configured file but analyze reported `Dockerfile.miren`. They now agree, and a regression test pins that ordering. This covers MIR-1684.

PR #1091, opened 2026-08-27
Remove dead post_import app.toml field

evanphx:
`AppConfig.PostImport` was still parsed and validated even though nothing in the runtime ever read it. #884 took it out of the docs so we stopped advertising a step that silently did nothing; this takes it out of the config.

- drop the field from `AppConfig`
- drop `post_import` from the root `validFields` list that drives "did you mean?" suggestions
- drop the explicit schema-test exclusion added in MIR-1047, since there is no longer a field to hide from the published JSON Schema

One thing worth a second pair of eyes: unknown keys are a hard parse error, not a warning. So an `app.toml` that still carries a `post_import` line goes from "quietly ignored" to "fails validation". That seems right — the value never did anything, and failing loudly is better than the current silence — but it is a visible change, so there is a changelog entry saying to delete the line and use a deploy task instead.

`go build ./...` and `go test ./appconfig/...` pass.

Closes MIR-1624

### DIFF
--- a/docs/docs/changelog.md
+++ b/docs/docs/changelog.md
@@ -11,6 +11,9 @@ All notable changes to Miren Runtime will be documented in this file.
 ## Unreleased
 *main*
 
+**Improvements**
+- **`post_import` is gone from app.toml** - The key was parsed and validated but never read by anything, so a deploy that set it reported success while the command silently never ran. It came out of the docs in [#884](https://github.com/mirendev/runtime/pull/884); now it's out of the config itself. If your `app.toml` still has a `post_import` line, delete it — unknown keys are a parse error, so it will now fail validation instead of being quietly ignored. Move the work to a [deploy task](/tasks).
+
 ---
 
 ## v0.14.0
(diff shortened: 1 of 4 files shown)


docs/docs/changelog.md
```diff
@@ -11,6 +11,9 @@ All notable changes to Miren Runtime will be documented in this file.
 ## Unreleased
 *main*
 
+**Improvements**
+- **`post_import` is gone from app.toml** - The key was parsed and validated but never read by anything, so a deploy that set it reported success while the command silently never ran. It came out of the docs in [#884](https://github.com/mirendev/runtime/pull/884); now it's out of the config itself. If your `app.toml` still has a `post_import` line, delete it — unknown keys are a parse error, so it will now fail validation instead of being quietly ignored. Move the work to a [deploy task](/tasks).
+
```
phinze: Nit: I don't think we need to populate changelog as we go - gets conflict-y once things are cooking, and we have good release-time catchup skills.

phinze (approved):
I WELCOME THE CLEANSING FIRE OF DELETION

PR #1092, opened 2026-08-27
Make `miren app runs` easier to scan

evanphx:
Two small changes to the run history table.

**Short ids in the table.** `RunInfo` now carries the run entity's `short_id`, and the CLI renders `ui.DisplayShortID(short, full)` the way `app history` and `app status` already do. `--format json` is untouched and still emits full ids, since that is what scripts pass back to `app runs cancel`.

The list path reads the short id straight off the entity it already has (`results.Entity().ShortId()`), so this costs no extra queries per row. `getRun` has no entity in hand, so it does one lookup — worth it to keep the two paths from disagreeing about what a run's id looks like.

**One consistent order.** Runs are now sorted by `started_at` descending, full stop. The old sort pulled failures to the front, which meant the position of the run you just triggered depended on how older, unrelated runs ended — you could trigger something and find it halfway down the table. The status column already says which runs failed.

A run recorded before short-ids existed has none; those fall back to the full id rather than rendering blank.

`go build ./...` and `go test ./servers/app/... ./cli/commands/...` pass. The two `...Etcd` failures in `servers/app` are the pre-existing ones that need `make dev` for etcd, unrelated to this change.

Closes MIR-1656

### DIFF
--- a/servers/app/runs.go
+++ b/servers/app/runs.go
@@ -212,21 +212,18 @@ func (r *AppInfo) ListRuns(ctx context.Context, state *app_v1alpha.RunsListRuns)
 		if args.Task() != "" && run.Task != args.Task() {
 			continue
 		}
-		runs = append(runs, runInfo(&run))
+		runs = append(runs, runInfo(&run, results.Entity().ShortId()))
 	}
 
-	// Failures first, then newest. The reason someone opens this list is almost
-	// always a failure, and a list where they are buried under successful
-	// console sessions is a list nobody reads.
+	// Newest first, with no special place for failures. Grouping failures ahead
+	// of everything else made the most recent run's position depend on the
+	// outcome of older ones, so the row someone had just triggered could sit
+	// halfway down the table. One consistent order is easier to scan, and the
+	// status column already says which runs failed.
+	//
+	// cmp.Compare rather than subtraction: these are Unix milliseconds, and
+	// their difference does not fit an int on a 32-bit build.
 	slices.SortStableFunc(runs, func(a, b *app_v1alpha.RunInfo) int {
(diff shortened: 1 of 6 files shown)


servers/app/runs.go
```diff
@@ -247,7 +244,7 @@ func (r *AppInfo) GetRun(ctx context.Context, state *app_v1alpha.RunsGetRun) err
 		return err
 	}
 
-	state.Results().SetRun(runInfo(run))
+	state.Results().SetRun(runInfo(run, r.runShortId(ctx, run.ID)))
```
phinze: Could the short ID stay attached to the run lookup result instead of being threaded into `runInfo` separately and then recovered with another fetch here? Both `ListResults.Entity()` and `GetByIdWithEntity(...).Entity()` expose the same `*entity.Entity`, so the list and get paths can carry a small `{run, entity}` or `{run, shortID}` result into rendering. Right now `lookupRun` discards the entity metadata, then `runShortId` rereads the same run to recover it. Because `GetRun` is polled every 250 ms while attach waits and exit status settles, that discard-and-refetch doubles the run reads throughout those loops.

--p+🤖

phinze (approved):
🤖 Looks good overall. One non-blocking note inline about carrying the run’s short ID through the existing lookup.

servers/app/runs.go
evanphx: Good catch, and the 250ms polling is the part I had not weighed — thank you. Fixed in ae0fd48.

`lookupRun` now uses `GetByIdWithEntity` and returns the short id alongside the run, so the get path reads once instead of twice. `runShortId` is gone.

Bonus: the old second return value was `run.ID`, which both callers discarded and which the caller already had on the run anyway, so the short id replaces it rather than widening the signature.

PR #1093, opened 2026-08-27
Remove legacy top-level Configuration.Concurrency and dead `set` command

evanphx:
Vestigial code from before concurrency became per-service, noticed while reviewing #667.

- **`cli/commands/set.go` is gone.** It defined `m set -c <n>`, writing the top-level `Configuration.Concurrency`. It was never registered in the dispatch table, so there was no way to invoke it — confirmed by grepping for any reference to `commands.Set`.
- **The `HasConcurrency()` checks in `app_status.go` are gone,** in both the human and JSON paths. `GetConfiguration` never populated that field, so both checks were always false: the "Concurrency:" line could not print, and the `concurrency` key could not appear in `--format json`. The JSON field was `omitempty`, so nothing consuming that output sees a change.
- **`concurrency` and `auto_concurrency` are out of `Configuration` in `api/app/rpc.yml`,** along with the now-unreferenced `AutoConcurrency` type. Indices 1 and 2 are left unused with a comment rather than reassigned, so an old client cannot read a new field through an old index.

Per-service concurrency (`services.*.concurrency`, `SetConcurrencyMode` / `SetNumInstances`) is untouched — that is the live path and this removes none of it.

`go build ./...`, `go test ./cli/commands/... ./appconfig/...` and `go generate ./api/app/...` are clean.

Closes MIR-796


phinze (approved):
🤖 I chased the three claims and they all hold. `Set` never made it into `RegisterAll`, `GetConfiguration` never populated the field, and regenerating the schema gives back exactly what's committed.

The one thing I'd note for later is that the app.toml side of this outlives the wire side. `appconfig.go:285` still parses a top-level `concurrency` that nothing reads, and it's fairly stuck there, because `DisallowUnknownFields()` means dropping the field would turn any app.toml still setting the key into a parse error.

PR #1095, opened 2026-08-28
Route HTTP hosts to named app services

evanphx:
## Why this change

HTTP host routes currently always reach an app’s `web` process. That prevents an app with a separately named HTTP API or dashboard process from exposing it through a hostname, even when that process is already part of the app’s active configuration. This change lets one hostname select one named HTTP-capable app service.

## Approach and choices

A route now stores an optional service name. An absent value still means `web`, which keeps existing stored routes and existing `miren route set <host> <app>` behavior compatible. The new `--service` flag defaults to `web`; route creation validates the selection against the active configuration, including both current `ports[]` declarations and legacy scalar ports.

The scope stays intentionally narrow. This does not add path-based routing, arbitrary port selection, automatic service selection, TCP/UDP hostname routing, or multi-service policies. Public ingress, internal requests, and tunnels use the selected service. Their lease-cache keys also include it, since reusing a lease across two services could proxy a request to the wrong process.

## Review guide

- Check the empty-service fallback for legacy routes and the route-set validation boundary.
- Check that cache isolation holds for active and ephemeral versions across service selection.
- Check whether applying the selected service to tunnels is the right consistent behavior.

## Validation

- Passed `go test ./api/ingress ./cli/commands ./servers/httpingress`.
- Passed `go vet ./api/ingress ./cli/commands ./servers/httpingress`.
- Ran `go generate ./api/ingress` and `git diff --check`.
- Added `TestRouteSetService` black-box coverage. It could not run in this sandbox because `hack/dev-exec` exits `127` before deployment; the CI black-box job provides the required development container.


evanphx:
Addressed the actionable review feedback in e4e74f0:

- Added the service cell to plain-text `route list` rows, so it now matches the `SERVICE` header and remaining columns stay aligned.
- Extended the service-route black-box test to check text output as well as JSON.
- Removed the redundant route-update comment.
- Put the route-set service constraint in a Docusaurus `info` admonition.

Validated with `go test ./api/ingress ./cli/commands ./servers/httpingress`, `go vet ./api/ingress ./cli/commands ./servers/httpingress`, and `git diff --check`. The black-box test still requires the project development container, which this sandbox does not have (`hack/dev-exec` exits 127 before deployment).

blackbox/route_service_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	shown := m.MustRun("route", "show", host, "--format", "json")
+	shown.RequireContains(t, `"service":"echo"`)
+
+	listed := m.MustRun("route", "list", "--format", "json")
+	listed.RequireContains(t, `"service":"echo"`)
+
+	textList := m.MustRun("route", "list")
+	textList.RequireContains(t, "SERVICE")
+	textList.RequireContains(t, "echo")
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed in `3b80f17`: the black-box test now finds the row for its hostname and asserts that the third rendered field, SERVICE, is `echo`.

evanphx:
🤖 Addressed the field-aware text table assertion in `3b80f17`: the black-box test now finds the route’s own row and verifies its `SERVICE` field equals `echo`.

The earlier table alignment, documentation, and redundant-comment feedback remains addressed in `e4e74f0`. I made no change for the two non-blocking observations about storing `web` explicitly and zero-port validation because both are intentional behavior.

Validated with `go test ./api/ingress ./cli/commands ./servers/httpingress`, `go vet ./api/ingress ./cli/commands ./servers/httpingress`, and `git diff --check`. GitHub CI has started for `3b80f17` and is still pending. The previous run had multiple failed black-box and lint jobs, but GitHub did not make logs available while it remained in progress. - e+🤖

evanphx:
🤖 Addressed the CI regression in `8d4aea0`. Route validation now honors the deployment launcher’s existing implicit HTTP default for a `web` app service with no declared port, so current `miren route set <host> <app>` behavior remains compatible. A non-`web` service still needs an explicitly declared HTTP port.

I also removed the ineffective tunnel-service initialization that `golangci-lint` reported. The black-box and POP failures all shared the same rejected implicit-`web` route setup; the aggregate test check failed only because those child jobs failed.

Validated with `go test ./api/ingress ./cli/commands ./servers/httpingress`, `go vet ./api/ingress ./cli/commands ./servers/httpingress`, and `git diff --check`. GitHub CI has started for `8d4aea0` and is pending. - e+🤖

evanphx:
🤖 Follow-up CI investigation found that `generate-check` also requires generated command reference pages to match CLI metadata. Commit `f11cd51` regenerates the three route command pages, removing manual prose that the generator does not own. The detailed named-service guidance remains in `traffic-routing.md`, which is authored documentation.

The local full `make generate-check` still rewrites unrelated encoded schemas because this sandbox’s generation toolchain produces different bytes. I restored those unrelated files; the route command docs and ingress schema output are generated and committed. GitHub CI has started for `f11cd51` and is pending. - e+🤖

evanphx:
🤖 Fixed `TestRouteSetService` in `84614e8`. I reproduced it locally with Docker and the ISO development environment. The route behavior was correct; the test failed because it searched pretty-printed JSON for whitespace-free substrings and expected `ok` without the fixture’s trailing newline. The test now decodes JSON and validates the route’s own host/service fields, while trimming the fixture response before comparison.

Validated with the targeted black-box command in the ISO/Docker environment. GitHub CI has started for `84614e8` and is pending. - e+🤖

api/ingress/client.go
```diff
@@ -250,11 +251,46 @@ func (c *Client) List(ctx context.Context) ([]*RouteWithMeta, error) {
...
+			// The deployment launcher gives a web service with no port declaration
+			// its historical HTTP default of port 3000.
+			if service == "web" && svc.Port == 0 {
+				return nil
+			}
+			if svc.Port > 0 && (svc.PortType == "" || svc.PortType == "http") {
+				return nil
+			}
+		} else {
+			for _, port := range svc.Ports {
```
phinze, line 265: 🤖 This check is stricter than the code that actually starts the service. A port with no `type` is valid and defaults to HTTP, but we reject it here. `web` also still gets its usual HTTP port 3000 when its configured ports are all TCP or UDP, and we reject that too. That means a plain `route set` can fail for configurations that already work. Could we reuse the launcher’s rules here, or cover the same cases and add tests?

cli/commands/route_set.go
```diff
@@ -46,13 +50,31 @@ func RouteSet(ctx *Context, opts struct {
 		return fmt.Errorf("failed to find app %q: %w", appName, err)
 	}
 
-	// Create/update the route
+	if appEntity.ActiveVersion == "" {
+		return fmt.Errorf("app %q has no active configuration", appName)
```
phinze: 🤖 Could this active-version lookup and service validation move behind the ingress client, perhaps as a service-aware `SetRoute` method? `RouteSet` now knows how app versions are stored, decodes core entities, and resolves runtime config only to enforce an ingress invariant. Keeping that orchestration at the write boundary would leave the command responsible for CLI input/output and make the same validation available to non-CLI route writers.

phinze (approved):
Hello you watched me look at this review and here it is! Two inline 🤖 findings for you to peek at, otherwise good to go.

api/ingress/client.go
evanphx: 🤖 Fixed in `e341d3d`: validation now follows `appspec.Build`, so an omitted port type defaults to HTTP and `web` remains HTTP-capable through its implicit port 3000 even when configured ports are non-HTTP.

cli/commands/route_set.go
evanphx: 🤖 Fixed in `e341d3d`: active-version lookup, runtime-config resolution, and HTTP-service validation now live in `ingress.Client.SetRoute`, keeping the CLI focused on input and output while making the invariant available to any route writer.

PR #1099, opened 2026-08-28
Default web ports from image EXPOSE

phinze:
Images already advertise their listening port through `EXPOSE`, but Miren ignored it and always resolved an unspecified web port to 3000. Common upstream images therefore needed duplicate port config or waited for the alternative-port fallback.

Builds now carry OCI exposed-port metadata into the resolved service config. A single TCP declaration fills an otherwise unspecified web port, while explicit port config still wins and ambiguous declarations stay conservative. Persisting the result as the ordinary scalar port keeps routing, health checks, and `$PORT` on the same value.

A Docker-backed smoke deploy with `EXPOSE 4321` came up directly on 4321 and served through the default HTTPS route.

Closes MIR-1445

### DIFF
--- a/servers/build/build.go
+++ b/servers/build/build.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Oh, very nice and friendly!

PR #1100, opened 2026-08-28
Model server startup as a typed boot dataflow graph

phinze:
Server startup was one long ordered function, so dependencies lived in line placement, sleeps, and background work that could fail after startup had already moved on. That made the sequence hard to reason about and unsafe to parallelize.

This moves lifecycle work out of `commands` into focused server boot components and a small generic `pkg/boot` package. Components publish typed outputs; passing an output to a constructor declares both the value it needs and the graph edge that orders it. The graph validates one fixed component set, runs independent branches in parallel, and uses the same edges for reverse shutdown. It stays a one-way boot sequencer, not a liveness or restart system.

Build and deployment handlers are assembled with the coordinator but exposed only by a final work-admission component. Build saga recovery likewise waits for BuildKit, the OCI registry, and registry host mapping. These hard boot boundaries replace the earlier bounded, fail-open readiness conditions.

The same pass tightens nearby lifecycle boundaries: workload identity fails closed when its issuer cannot initialize, server and local-client addresses share normalization, sandbox cleanup reports expired deadlines, and OCI requests retain a graceful drain window during shutdown.

Supersedes #1096, which Linear closed when MIR-1290 became a project and its placeholder issue was canceled. Verified with targeted package tests and vet, the coordinator integration test under `dev-exec` with `-race`, and a live dev-server restart.

Part of MIR-1688


pkg/boot/provide.go
```diff
@@ -0,0 +1,49 @@
+package boot
+
+import "context"
+
+func Provide0[O any](name string, start func(context.Context) (O, error), options ...Option) (*Component, Output[O]) {
```
evanphx, line 5: oooooooohhh yyeeeeeahhh

components/ocireg/registry.go
```diff
@@ -57,14 +59,32 @@ func (r *Registry) Start(ctx context.Context, addr string) error {
...
+		}
+	}()
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := r.server.Shutdown(shutdownCtx); err != nil {
+			r.Log.Warn("OCI registry shutdown", "error", err)
+		}
+	}()
```
evanphx: 🤖 the drain has two overlapping shutdown paths. The registry is only ever started by the boot component. On teardown, graph.Stop calls cancel() (which fires the internal
  ctx.Done() goroutine's Shutdown with a 5s budget) and runs the component's stop, which calls Shutdown again with the 30s componentStopTimeout — concurrently, on the same server. A healthy in-flight layer
  push/pull that takes 5–30s makes the internal goroutine log Warn "OCI registry shutdown" error=context deadline exceeded even though the 30s component budget would drain it cleanly. Since nothing outside the
  boot graph uses this registry anymore, the internal goroutine looks redundant with the component stop — we could drop it and let the component lifecycle own the drain, or at least align the two budgets. Minor
  log-noise-on-healthy-path thing, not a correctness bug.

evanphx (approved):
Love how it come together!

🤖 had one thing to mention, non-blocking.

components/ocireg/registry.go
phinze: 🤖 Fixed. The boot component now exclusively owns registry shutdown and its 30-second drain budget. `Registry.Start` no longer starts a competing five-second shutdown goroutine, and the standalone registry test now shuts it down explicitly.

PR #1102, opened 2026-08-28
Support unattended cluster enrollment with an enroll token

evanphx:
Miren Cloud's enrollment service is done: it mints single-use enroll tokens and generates a cloud-init payload whose last step runs `miren server install --enroll-token <token>`. The runtime was the missing half. Registration could only go the interactive way, where the CLI prints an auth URL and polls until a human approves in a browser. A machine booting itself from cloud-init has no human at a browser, so that path was a dead end for it.

This builds the runtime side. The registration client now sends the token on the initiate request and understands the synchronous "registered" response cloud returns for a valid token, which carries the cluster identity directly and has nothing to poll for. `server install` and `server register` take an `--enroll-token` flag and route to that path, saving an approved registration in one shot. A token cloud rejects is terminal on purpose: falling back to the interactive flow would leave an unattended machine waiting for an approval nobody knows to give. There's no new secret at rest, since auth after enrollment is a challenge-response against the public key the node supplies, and on an interrupted retry the saved keypair is reused so cloud can replay the original registration instead of refusing a spent token.

The second commit adds an end-to-end test. None of the existing cloud-backed blackbox tests actually drive the runtime's registration command (they hand-write registration.json), so this is the first to exercise the real handshake. It mints a token against a live cloud and runs `miren server register --enroll-token`, then checks the approved registration and that a spent token is refused. It only needs the cloud control plane, so it adds a lighter `NewCloudControlPlane` harness helper alongside the existing full-environment one. Wiring it into CI took a beat: cloud-backed tests run in their own job with the cloud repo checked out and are excluded from the sharded suite that runs without cloud, so the test is added to that job and to the exclusion lists rather than being silently skipped forever.

One thing this deliberately leaves for cloud: a token-enrolled cluster currently anchors its workload identity at the cluster rather than at cloud, because cloud's registered response omits the `identity_issuer_url` that the interactive responses carry. That's tracked in MIR-1691.

Closes MIR-1687

https://claude.ai/code/session_0185j32Ei3k78Cfh8QMGAPbX


blackbox/enroll_test.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// This is the first blackbox test that exercises the runtime's registration
+// CLI end to end. The others hand-write registration.json; here the runtime
+// produces it, so the whole initiate handshake is under test.
+func TestServerEnrollWithToken(t *testing.T) {
+	c := harness.NewCluster(t)
+	m := harness.NewMiren(t, c)
+	cloud := harness.NewCloudControlPlane(t, m)
+
+	token, _ := cloud.MintEnrollToken(t, map[string]string{"env": "bb"})
```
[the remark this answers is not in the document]
evanphx: Good catches, both addressed in the latest push. The test now keeps `orgID` from the mint and asserts `reg.OrganizationID == orgID`, so a token registering into the wrong org would fail. The spent-token replay now also asserts the stderr contains "already been used", so it has to fail for the right reason (cloud reporting the token spent) rather than on any nonzero exit.

docs/docs/command/server-install.md
```diff
@@ -46,6 +47,12 @@ miren server install
 miren server install --without-cloud
 ```
 
+**Install with an unattended enroll token:**
+
+```bash
+miren server install --enroll-token "$(cat /etc/miren/enroll-token)"
```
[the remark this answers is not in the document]
evanphx: Fair point about argv exposure, but not something to fix here. This file is generated (`hack/gen-command-docs` from the CLI help output), so a hand-added admonition gets overwritten on the next `go generate`. The example also mirrors the exact command Miren Cloud's cloud-init payload runs, which deliberately stores the token in a root-only file and `shred`s it right after registration, so the window is small and the token is single-use. The real fix is a non-argv input path (a `--enroll-token-file` or env var, so the cloud command drops the `$(cat ...)`), which also touches the cloud side that generates the command. Leaving that as a follow-up rather than a doc-only warning.

cli/commands/server_install.go
```diff
@@ -397,6 +398,7 @@ func ServerInstall(ctx *Context, opts struct {
 				CloudURL:    opts.CloudURL,
 				Tags:        opts.Tags,
 				OutputDir:   "/var/lib/miren/server",
+				EnrollToken: opts.EnrollToken,
```
phinze, line 401: 🤖 `ServerInstall` still swallows the `Register` error below, so an invalid or spent enroll token returns success and leaves an unregistered server installed. This is the path Cloud invokes, which means callers cannot detect failed unattended enrollment. Could we return the registration error when `opts.EnrollToken` is set, while keeping the existing best-effort behavior for interactive registration?

Makefile
```diff
@@ -204,6 +204,7 @@ test-blackbox-pop: ## Run cloud-backed blackbox tests (requires `make dev` and c
 	go test -tags blackbox -timeout 15m -v -count=1 -p 1 -run '^TestPOP$$' ./blackbox/...
 	go test -tags blackbox -timeout 15m -v -count=1 -p 1 -run '^TestRPCViaCloud$$' ./blackbox/...
 	go test -tags blackbox -timeout 15m -v -count=1 -p 1 -run '^TestDeployViaCloud$$' ./blackbox/...
+	go test -tags blackbox -timeout 15m -v -count=1 -p 1 -run '^TestServerEnrollWithToken$$' ./blackbox/...
```
phinze, line 207: 🤖 `test-blackbox-distributed` below says it skips the same cloud-backed tests as `test-blackbox`, but its `-skip` expression does not include `TestServerEnrollWithToken`. As a result, the local distributed target now runs this test without the Cloud setup it requires. Could we add it to that exclusion too?

phinze (approved):
🤖 The unattended registration path is straightforward and well covered, including a live Cloud handshake. I left two notes on preserving failure signaling through `server install` and keeping the local distributed suite from running the Cloud-backed test. CodeRabbit’s current-head private-key logging comment also looks worth fixing.

cli/commands/server_install.go
evanphx: 🤖 Good catch, fixed in 3cbcfdc7. When `opts.EnrollToken` is set, `ServerInstall` now returns the registration error (wrapped as `unattended enrollment failed`) instead of warning and continuing, so cloud-init gets a nonzero exit and the failure surfaces in the bootstrap log. Interactive registration keeps its best-effort behavior.

Makefile
evanphx: 🤖 Fixed in 3cbcfdc7. Added `TestServerEnrollWithToken` to the `test-blackbox-distributed` `-skip` list so it matches the comment and the other cloud-backed tests. The CI workflow's distributed shard already excluded it; this was the local Make target that had drifted.

PR #1103, opened 2026-08-30
Point builds at miren's bridge DNS instead of public DNS

evanphx:
## Problem

Container image builds silently resolve DNS against public servers (`8.8.8.8` / `8.8.4.4`) instead of the host's configured resolver. On a host that uses a systemd-resolved stub — `nameserver 127.0.0.53` in `/etc/resolv.conf` — with an internal-only or egress-filtered upstream, builds can't resolve names at all. Reported by a user whose build environment couldn't do DNS.

## Root cause

Build `RUN` steps run in the host network namespace (with no CNI config, buildkit's "auto" mode falls back to host networking), but buildkit generates each build's resolv.conf as if it were sandboxed. The step's `NetMode` is `UNSET`, not `HOST`, so `GetResolvConf` strips every loopback nameserver (the whole `127.0.0.0/8` range, `127.0.0.53` included) and, finding none left, substitutes `8.8.8.8`/`8.8.4.4`. It can't recover the real upstream either, since `/run/systemd/resolve/resolv.conf` isn't mounted into buildkitd — only the host's `/etc/resolv.conf` is, via `oci.WithHostResolvconf`.

So even though the build sits in the very netns where `127.0.0.53` would resolve, buildkit discards it and hardcodes public DNS — the "breaks builds on hosts with an internal resolver" case the earlier MIR-1643 note meant to avoid.

## Fix

Add a `[dns]` section to the generated `buildkitd.toml` pointing at miren's own bridge DNS server, wired in from the network boot node's router address. That resolver already serves runtime sandboxes: it forwards to the host's real upstream and answers `*.app.miren`. The bridge gateway is a local interface in the host netns, so build steps reach it, and buildkit keeps it (not a loopback, so the filter leaves it alone). External buildkit daemons are untouched.

## Verification

- Unit tests (`TestGenerateConfig`) and server boot/startup tests pass; `golangci-lint run ./...` reports 0 issues.
- End to end in the dev env — rebuilt, restarted, ran a real Dockerfile build:
  - Before: `RUN cat /etc/resolv.conf` → `nameserver 8.8.8.8`
  - After: `nameserver 10.8.13.1`, and `nslookup example.com` resolves through it (`exit=0`)

Fixes MIR-1695

### DIFF
--- a/components/buildkit/buildkit.go
+++ b/components/buildkit/buildkit.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -48,6 +49,10 @@ type Config struct {
 
 	// RegistryHost is the hostname for the cluster-local registry (e.g., cluster.local:5000)
 	RegistryHost string
+
+	// DNSNameservers are the nameserver addresses build steps resolve against.
+	// See generateConfig for why we override buildkit's default here.
+	DNSNameservers []string
 }
 
 // Component manages a persistent BuildKit daemon as a containerd container,
@@ -145,7 +150,7 @@ func (c *Component) Start(ctx context.Context, config Config) error {
 	}
 
 	// Generate buildkitd.toml config
-	configContent := c.generateConfig(gcKeepStorage, gcKeepDuration, config.RegistryHost)
+	configContent := c.generateConfig(gcKeepStorage, gcKeepDuration, config.RegistryHost, config.DNSNameservers)
 	configPath := filepath.Join(dataPath, "buildkitd.toml")
 	if err := os.WriteFile(configPath, []byte(configContent), 0644); err != nil {
 		return fmt.Errorf("failed to write buildkit config: %w", err)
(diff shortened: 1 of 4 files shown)


components/buildkit/buildkit.go
```diff
@@ -48,6 +49,15 @@ type Config struct {
...
 	// RegistryHost is the hostname for the cluster-local registry (e.g., cluster.local:5000)
 	RegistryHost string
+
+	// DNSNameservers are the nameserver addresses build steps resolve against.
+	// Miren points these at its own bridge DNS server (reachable from the host
+	// netns the build runs in), which forwards to the host's real upstream
+	// resolver and also answers *.app.miren. Without this, buildkit strips the
+	// host's loopback stub resolver (e.g. systemd-resolved's 127.0.0.53) and
+	// substitutes public DNS (8.8.8.8), which breaks builds on hosts with an
+	// internal or egress-filtered resolver. See generateConfig.
```
phinze: Could we trim this comment back? Looks like the 🤖 was feeling verbose here. Also, builds won’t actually resolve `*.app.miren`: that lookup requires a sandbox source IP, while builds query from the host/router.

--p+🤖

components/buildkit/buildkit.go
```diff
@@ -310,19 +320,40 @@ func (c *Component) Client(ctx context.Context) (*buildkitclient.Client, error)
...
+// host network namespace, but buildkit generates their resolv.conf as if they
+// were sandboxed: it strips any loopback nameserver (the whole 127.0.0.0/8
+// range) and, finding nothing left, substitutes public DNS (8.8.8.8/8.8.4.4).
+// So on a systemd-resolved host whose /etc/resolv.conf is the 127.0.0.53 stub,
+// builds silently resolve against 8.8.8.8 instead of the host's real resolver.
+// That breaks builds on hosts with an internal-only or egress-filtered resolver
+// (MIR-1643). We override it by pointing buildkit at miren's own bridge DNS
+// server, whose address is reachable from the host netns; it forwards to the
+// host's real upstream and also answers *.app.miren. When no nameserver is
+// supplied we emit no [dns] section and leave buildkit's default in place.
```
phinze: Same trim request here. The PR description already carries this context.

--p+🤖

phinze (approved):
The bridge resolver is the right shim for the systemd-resolved case, and the boot wiring looks sound.

I recommend a follow-up to add TCP support to the bridge DNS server and retry truncated upstream responses over TCP. This PR makes that existing UDP-only resolver the builder’s sole DNS path, so larger responses can otherwise fail during TCP fallback.

--p+🤖

components/buildkit/buildkit.go
evanphx: 🤖 Trimmed it down, and dropped the `*.app.miren` claim — you're right, that lookup needs a sandbox source IP and builds query from the host/router.

components/buildkit/buildkit.go
evanphx: 🤖 Trimmed — leaning on the PR description for the full context.

evanphx:
🤖 Good call on the TCP fallback — filed as [MIR-1702](https://linear.app/miren/issue/MIR-1702/add-tcp-support-to-the-bridge-dns-server-and-fall-back-to-it) to add TCP support to the bridge DNS server and retry truncated responses over TCP. Keeping it out of this PR.

PR #1104, opened 2026-08-30
Route embedded BuildKit DNS through interface

evanphx:
## Why this change

BuildKit currently derives each build container’s resolver list from the host. That bypasses the DNS server Miren already runs at the runner interface, so build-time lookups cannot use Miren’s sandbox and service-name resolution.

## Approach and choices

The embedded BuildKit boot now depends on the network boot output and passes its router address to BuildKit. That router address is the same address `SandboxController.Init` configures as the bridge DNS listener on port 53. BuildKit writes it as its sole `[dns]` nameserver, so every build container uses the Miren resolver.

The change is deliberately limited to the embedded daemon. An externally managed BuildKit daemon retains its existing resolver configuration because runtime does not own that process. The DNS address is derived from the active leased subnet instead of being a static address, which keeps it correct when a runner’s network lease changes.

## Review guide

- Confirm that the runner subnet router address is the right handoff for the DNS service.
- Check the additional boot-graph edge: BuildKit must wait until network setup has produced that address.
- Check that keeping external BuildKit untouched is the desired compatibility boundary.

## Validation

- `go test ./components/buildkit -run '^TestGenerateConfig$'`
- `go test ./components/server -run '^TestBuildkitWithoutConfiguredDaemonPublishesEmptyOutput$'`
- `go vet ./components/buildkit ./components/server`
- `make generate-check`
- `go test ./components/buildkit` was attempted, but its container-backed tests cannot access `/run/containerd/containerd.sock` in this sandbox (`permission denied`).

### DIFF
--- a/components/buildkit/buildkit.go
+++ b/components/buildkit/buildkit.go
@@ -48,6 +48,9 @@ type Config struct {
 
 	// RegistryHost is the hostname for the cluster-local registry (e.g., cluster.local:5000)
 	RegistryHost string
+
+	// DNSServer is the Miren interface DNS server used by build containers.
+	DNSServer string
 }
 
 // Component manages a persistent BuildKit daemon as a containerd container,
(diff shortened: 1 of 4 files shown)


phinze:
Whoops, looks like this is a dupe! Closing this one and focusing review on #1103, which is the slightly stronger version.

evanphx:
whooops! Was using this change to test workbench tooling, yeah, dupe.

PR #1109, opened 2026-08-31
Preserve deployment short IDs through lock reservation

phinze:
Deployment lock admission briefly creates a private entity so downgraded runtimes can see that the prospective lock owner is alive. Because that reservation had no kind, it skipped the entity store's automatic short-ID allocation, and publishing it left the deployment without a usable short handle.

Model the reservation as its own private compatibility kind, then carry its allocated short ID into the deployment when the lock is published. The reservation still stays out of deployment history, and store conformance coverage now keeps Ensure behavior aligned between the mock and etcd implementations.

Closes MIR-1703


evanphx (approved):
Approved! Clean fix, and nice that the mock now can't hide this class of bug from us again.

--e+🤖

PR #1111, opened 2026-08-31
Render only the serving deployment as active

phinze:
Legacy deployment rows can retain stale `active` statuses after another deployment takes over. The history RPC canonicalized those rows to successful attempts, then restored the raw status, so `miren app history` could show several active deployments even though `app.active_deployment` named only one.

This makes the app pointer the serving signal: only the successful attempt it names is presented as active; the rest stay succeeded. It also drops the CLI’s `--status` and `--hide-failed` filters, whose mixed attempt/serving semantics and interaction with limits could not give reliable results. Structured consumers can filter `--all --format json`; the RPC filter remains for compatibility.

Closes MIR-1701

### DIFF
--- a/servers/deployment/server.go
+++ b/servers/deployment/server.go
@@ -163,7 +163,11 @@ func (d *DeploymentServer) UpdateDeploymentStatus(ctx context.Context, req *depl
 	if err != nil {
 		return err
 	}
-	deploymentInfo := d.toDeploymentInfo(rec.Deployment)
+	// New clients activate with succeeded; active and rolled_back are legacy aliases for the same transition.
+	serving := status == deploylifecycle.StatusActive ||
+		status == deploylifecycle.StatusSucceeded ||
+		status == deploylifecycle.StatusRolledBack
+	deploymentInfo := d.toDeploymentInfo(rec.Deployment, serving)
 	results.SetDeployment(deploymentInfo)
 
 	d.Log.Info("Updated deployment status",
@@ -211,7 +215,7 @@ func (d *DeploymentServer) UpdateDeploymentPhase(ctx context.Context, req *deplo
 	if err != nil {
 		return err
 	}
-	deploymentInfo := d.toDeploymentInfo(rec.Deployment)
+	deploymentInfo := d.toDeploymentInfo(rec.Deployment, false)
 	results.SetDeployment(deploymentInfo)
 
 	d.Log.Info("Updated deployment phase",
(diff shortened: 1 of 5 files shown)


servers/deployment/server.go
```diff
@@ -300,11 +304,27 @@ func (d *DeploymentServer) ListDeployments(ctx context.Context, req *deployment_
 		versionIDs = append(versionIDs, (&deploylifecycle.Record{Deployment: dwe.deployment}).AppVersion())
 	}
 	versionShortIDs := d.resolveShortIDs(ctx, versionIDs)
+	activeDeployments := make(map[string]entity.Id)
+	resolvedApps := make(map[string]struct{})
+	for _, dwe := range deployments {
+		appName := dwe.deployment.AppName
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch. Renamed the loop-local value to `deployAppName` in the latest push, and the deployment package still passes all 63 tests.

servers/deployment/server.go
```diff
@@ -300,11 +304,27 @@ func (d *DeploymentServer) ListDeployments(ctx context.Context, req *deployment_
...
+	resolvedApps := make(map[string]struct{})
+	for _, dwe := range deployments {
+		deployAppName := dwe.deployment.AppName
+		if _, resolved := resolvedApps[deployAppName]; resolved {
+			continue
+		}
+		resolvedApps[deployAppName] = struct{}{}
+		activeDeployment, err := d.activeDeploymentID(ctx, deployAppName)
+		if err != nil {
+			return err
```
evanphx, line 317: Non-blocking: history now reads the app entity to find the serving pointer, so a non-`NotFound` error from `AppByName` fails the whole listing — a path history never touched before. I lean toward this being right (if this read is erroring, the cluster has bigger problems than the history table), just flagging the new coupling so it's a conscious choice.

servers/deployment/server_test.go
```diff
@@ -154,12 +154,69 @@ func TestToDeploymentInfo(t *testing.T) {
...
+		},
+		{
+			name: "stale canonical succeeded deployment",
+			deployment: &core_v1alpha.Deployment{
+				ID: "deployment/canonical-stale", Outcome: "succeeded", Status: "active",
+			},
+			want: "succeeded",
+		},
+		{
+			name:       "failed deployment remains failed even if pointed to",
```
evanphx, line 203: 👏 easy case to forget to pin.

evanphx (approved):
Approved! Good to have history show what's actually serving instead of a wall of "active." One non-blocking inline note and a 👏 below — neither holds anything up.

--e+🤖

servers/deployment/server.go
phinze: Ya now that the `App` is holding the active deployment pointer it's the data we really need. So failing when it can't be looked up is what we want.

PR #1112, opened 2026-08-31
Make deployment records drive build cancellation

phinze:
Deploy cancellation was accidentally coupled to the CLI's RPC connection. A dropped terminal stopped the build, while `miren deploy cancel` only worked because the initiating CLI kept polling the record and translated that state back into local context cancellation.

This makes the server-owned deployment record the control plane. Builds detach from the transport but remain tied to server shutdown; record cancellation stops action work while saga persistence and compensation finish. Ctrl-C writes that cancellation explicitly, and older servers keep the polling fallback.

Verified with the full serial test suite and a standalone smoke: one build survived its CLI being killed, while SIGINT cancelled the record, interrupted BuildKit, compensated, and released the lock.

Closes MIR-1506

### DIFF
--- a/cli/commands/deploy.go
+++ b/cli/commands/deploy.go
@@ -43,6 +43,43 @@ import (
 
 var deployTracer = otel.Tracer("miren.dev/runtime/cli/deploy")
 
+func isDeploymentCancelled(err error) bool {
+	remote, ok := errors.AsType[cond.ErrRemote](err)
+	return ok && remote.Category == "deployment" && remote.Code == "cancelled"
+}
+
+type deploymentCancellationOutcome int
+
+const (
+	deploymentCancellationUnknown deploymentCancellationOutcome = iota
+	deploymentCancellationConfirmed
+	deploymentCancellationCompleted
+)
+
+// reconcileDeploymentCancellation resolves a failed or ambiguous cancellation
+// request against the authoritative deployment record. The deploy may have
+// finished between the client's interrupt and the server handling its request.
+func reconcileDeploymentCancellation(
+	ctx context.Context,
+	deploymentID string,
+	getter deploymentStatusGetter,
+	cancelErr error,
+) (deploymentCancellationOutcome, error) {
+	status, statusErr := getter.GetStatus(ctx, deploymentID)
+	if statusErr != nil {
+		return deploymentCancellationUnknown, fmt.Errorf("%w; re-reading deployment status: %v", cancelErr, statusErr)
+	}
+
+	switch status {
+	case "cancelled":
+		return deploymentCancellationConfirmed, nil
+	case "succeeded", "active":
+		return deploymentCancellationCompleted, nil
+	default:
+		return deploymentCancellationUnknown, cancelErr
+	}
+}
+
 func Deploy(ctx *Context, opts struct {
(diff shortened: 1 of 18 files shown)


cli/commands/deploy.go
```diff
@@ -460,27 +473,63 @@ func Deploy(ctx *Context, opts struct {
...
+		if ctx.Err() != nil {
+			fresh, connectErr := ctx.rpcClient(cancelCtx, "dev.miren.runtime/deployment")
+			if connectErr != nil {
+				return fmt.Errorf("reconnecting to cancel deployment %s: %w", id, connectErr)
+			}
+			defer fresh.Close()
+			client = deployment_v1alpha.NewDeploymentClient(fresh)
+		}
+
+		result, cancelErr := client.CancelDeployment(cancelCtx, id, "")
```
evanphx, line 552: Non-blocking, worth considering: a narrow race where this reports a failed cancellation for a deploy that actually succeeded.

The record is `in_progress` and cancelable right up to the atomic flip to `succeeded` (and that whole window genuinely rolls back work that already went live, which is great). But between that flip — where `Close` also tears down the watcher — and the client receiving the terminal status arm, the CLI still thinks cancel is on the table. A Ctrl-C in that sliver sends `CancelDeployment`, the server rejects `succeeded → cancelled` as a conflict, and we print "cancellation could not be confirmed" and exit non-zero, for a deploy that shipped.

It fails closed, which is the part that matters — nothing phantom happens to a live deployment. The only rough edge is the message: we can't tell "cancel genuinely failed" from "cancel lost the race to success."

Two cheap ways to close it if you think it's worth it: have `CancelDeployment` surface an "already settled succeeded" signal and report that as success, or on a rejected cancel re-read the record and, if it's `succeeded`, say so and exit 0. Happy to file a follow-up.

servers/build/build.go
```diff
@@ -1384,7 +1390,7 @@ func (b *Builder) BuildFromTar(ctx context.Context, state *build_v1alpha.Builder
 	// -- build.receive_tar span
 	ctx, recvSpan := buildTracer.Start(ctx, "build.receive_tar",
 		trace.WithAttributes(attribute.String("miren.app.name", name)))
-	r := stream.ToReader(ctx, td)
+	r := stream.ToReader(requestCtx, td)
```
evanphx, line 1393: 👏 `requestCtx` staying on the client transport here while the build detaches to `work.action` — the first place you'd look, and it's right.

evanphx (approved):
Approved! This is a really satisfying cleanup.

Non-blocking: one inline note flags a last-sliver race — a Ctrl-C landing right as the deploy succeeds reports a failed cancellation for a deploy that actually shipped. Two cheap fixes in the note; your call whether it's worth a follow-up now. The other inline is just applause.

--e+🤖

cli/commands/deploy.go
phinze: 🤖 Good catch. Fixed with the client-side re-read: when cancellation is rejected, we now fetch the record and distinguish a deployment that completed before cancellation from a cancellation that genuinely failed. A succeeded deploy reports completion and exits cleanly; unknown outcomes still fail closed.

PR #1113, opened 2026-08-31
Generate shared enum types without changing stored entities

phinze:
Our current enum support is built on entity references. For every allowed value, schemagen creates an entity and stores a reference to it. It also creates a different Go type for every field.

That combination produced the disk-provider bug in MIR-940 and led to MIR-942. Two fields can represent the same concept, but converting a value from one field’s Go type to the other leaves the first field’s entity ID intact. The code compiles, but the converted value points to the wrong enum entity.

Not every enum-like field uses references today. Some, including storage filesystem types, are stored as plain strings. Changing those fields to references would change how existing entities are encoded in etcd. Entity values record both their type and their value, so this would be a real data migration rather than a generator-only change.

This PR separates the Go enum we want to use from the way each field is stored. A named enum produces one shared Go type and one set of values. Each field then chooses whether to store those values as a `ref`, `string`, or `keyword`. Existing ref fields keep their current entity IDs, while existing string fields remain strings. All three forms carry their allowed values into the schema and use the same membership validation.

For new enums, `keyword` is the preferred representation. It keeps enum values distinct from free-form strings in the wire format, but does not require a separate entity for every value. Keeping `ref` and `string` as supported representations lets existing fields adopt named enums without first migrating their data.

A few old generated enum strings were also persisted outside the entity wire format, in disk state and soft-delete metadata. Those readers now accept the old values, convert them to the new shared values, and save the upgraded form. Runner status parsing accepts both forms during rollout.

Addresses the enum-membership portion of MIR-1239. Required-field constraints and CEL-backed validation remain follow-up work.

Closes MIR-942


phinze:
Superseded by #1146. We revisited the original enum design and decided to keep membership backed by entity references. The replacement fixes sharing with one member namespace, teaches readers to accept older values, and converges stored entities in the background. That is a different enough design that it deserves a fresh review.

PR #1114, opened 2026-09-01
Apply Bring Your Own Image capabilities across the docs

phinze:
The Bring Your Own Image work gave Miren the ability to preserve an image’s startup contract and resolve its runtime metadata. Much of the documentation still reflected the previous behavior, routing runnable images through Dockerfiles, service commands, and Procfiles that are no longer required.

This sweeps the recipes, language guides, and core documentation to apply those capabilities consistently. OpenHands now runs directly from its upstream image, the Hermes explanation matches its actual metadata, and the retained Headscale and Amp Dockerfiles explain the files or packages they genuinely add.

Source-build tutorials still use Dockerfiles when they compile code or add application files, but those images now carry their own `CMD` instead of repeating it in a Procfile. Direct-image examples inherit the image’s working directory, entrypoint, command, and single exposed TCP port, adding overrides only where the image metadata leaves a gap. Verified with the docs build and local OpenHands and Hermes smoke deploys on current `main`.

Closes MIR-1686

### DIFF
--- a/docs/docs/guides/r.md
+++ b/docs/docs/guides/r.md
@@ -18,7 +18,7 @@ Ask your AI coding agent to "set up this R API on Miren" after installing the
 `0.0.0.0:$PORT`, and deploys — using this page as its reference.
 :::
 
-## Do you need a Dockerfile?
+## Does this source build need a Dockerfile?
 
 Yes. Miren doesn't auto-detect R, so add a `Dockerfile.miren` to your project root.
 Miren builds from it instead of guessing the stack — see
@@ -65,6 +65,8 @@ WORKDIR /app
 COPY . /app
 
 EXPOSE 8080
+ENTRYPOINT ["Rscript"]
+CMD ["/app/entrypoint.R"]
 ```
 
 If you need extra packages, install them in the build:
@@ -76,15 +78,12 @@ If you need extra packages, install them in the build:
 .git
 ```
 
-## Set up the app
+## Deploy
 
-Add a `Procfile`:
+The Dockerfile's `ENTRYPOINT` and `CMD` start the app. Miren uses them as the web
+service's startup default, so you don't need a `Procfile` or service command.
 
-```procfile
-web: Rscript /app/entrypoint.R
-```
-
-Then create `.miren/app.toml` naming your app and deploy from your project root:
+Create `.miren/app.toml` naming your app and deploy from your project root:
 
 ```toml
 name = "r-bench"
@@ -124,7 +123,7 @@ See [App Configuration — Environment Variables](/app-configuration#environment
 
 - **Detection:** none — requires `Dockerfile.miren`
 - **Base image:** `rstudio/plumber:latest` (R + Plumber preinstalled)
-- **Service is required:** define a `Procfile` (`web: Rscript /app/entrypoint.R`) — the image `CMD` is not used
+- **Startup:** inherited from the Dockerfile `ENTRYPOINT` and `CMD`; no `Procfile` or service command needed
 - **Port:** `Sys.getenv("PORT")`; `pr_run(host = "0.0.0.0", port = port)`
 - **Env vars:** `miren env set -e/-s`; read with `Sys.getenv`
 
--- a/docs/docs/recipes/openhands-agent-server.md
+++ b/docs/docs/recipes/openhands-agent-server.md
@@ -35,7 +35,7 @@ the browser) and connect it to this server.
 3. A client (Agent Canvas or the SDK) creates a conversation, passing the LLM model + key and
    the agent's tools. The agent runs **in this container**, reading/writing its workspace on
    the mounted disk.
-4. Conversations, the agent workspace, and bash-event history persist under `/data`.
+4. Conversations, the agent workspace, and bash-event history persist under `/workspace`.
 
 ## Prerequisites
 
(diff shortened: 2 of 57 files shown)


docs/docs/recipes/openhands-agent-server.md
```diff
@@ -126,36 +103,25 @@ required = true
...
-`OH_WORKSPACE_PATH`, and `OH_BASH_EVENTS_DIR` to absolute paths on the mounted disk.
-:::
+The server defaults to port 8000. Setting `OH_SESSION_API_KEYS_0` enables authentication and
+also makes it listen on `0.0.0.0`, so no command-line override is needed.
 
 :::warning[Use a `local` disk, and let it run as the image's user]
-Mount the workspace with `provider = "local"`. Don't add a `USER root` to the Dockerfile — the
-image runs as its own `openhands` user, and a writable disk is made writable for the run user
-automatically; forcing root opts out of that.
+Mount the image's `/workspace` directory with `provider = "local"`. The server's default
```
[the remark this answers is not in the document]
phinze: 🤖 Confirmed against the pinned v1.44.1 source and the smoke deployment. The three defaults are `workspace/conversations`, `workspace/project`, and `workspace/bash_events`; the image sets `WORKDIR /`, so they resolve beneath `/workspace`. The server created `/workspace/conversations` on the mounted disk and reached healthy without path overrides.

docs/docs/guides/r.md
```diff
@@ -65,6 +65,7 @@ WORKDIR /app
 COPY . /app
 
 EXPOSE 8080
+CMD ["Rscript", "/app/entrypoint.R"]
```
evanphx: This one won't survive losing the Procfile. `rstudio/plumber` sets its entrypoint to a whole command:

```
ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); args <- list(host = '0.0.0.0', port = 8000); ...; do.call(pr$run, args)"]
```

A service `command` replaced the process args outright, so that entrypoint never ran. Without it, the `CMD` becomes arguments to it instead — so it plumbs `entrypoint.R` rather than `plumber.R`, and binds a hardcoded 8000. The port lands wrong too: the base image exposes 8000 and this adds 8080, so nothing is inferred and `PORT` falls back to 3000.

`ENTRYPOINT []` above the `CMD` puts it back, same as the Headscale recipe does.

I checked the other guides' runtime base images for the same trap — clojure, julia, jruby, sbcl, frankenphp, deno and temurin all pass the CMD straight through, and caddy has no entrypoint. This is the only one.

docs/docs/services.md
```diff
@@ -230,8 +233,8 @@ Each service can configure:
 |--------|-------------|---------|
 | `command` | Shell command that replaces the image startup command (`/bin/sh -c`) | (none) |
 | `args` | Exec-form arguments that replace image `CMD` while preserving `ENTRYPOINT`; mutually exclusive with `command` | (none) |
-| `image` | Container image to use | App's built image |
-| `port` | Port the service listens on (single-port shorthand) | 3000 (web only) |
+| `image` | Container image to use | App version's primary image |
+| `port` | Port the service listens on (single-port shorthand) | A single TCP port exposed by the primary image; otherwise 3000 (web only) |
```
evanphx: `defaultServicePortFromImage` only touches the service named `web`, and only when it has no image of its own — so a `worker` never inherits, and neither does a Postgres sidecar. The `(web only)` here reads as attaching just to the 3000. Worth widening it to cover both clauses, and same row in `app-toml.md:208`.

evanphx (approved):
Approved! Really nicely done.

Two inline notes, both non-blocking. The R guide one is the only one with teeth. Also: I pulled the OpenHands image config to answer the bot's question above — `WorkingDir` is `/` and it exposes 8000 and 8002, so the recipe is right on both counts.

--e+🤖

docs/docs/guides/r.md
phinze: 🤖 Good catch. The base image sets its whole Plumber launch command as `ENTRYPOINT`, so our `CMD` would be appended to it. We now replace that with `ENTRYPOINT ["Rscript"]` and keep `/app/entrypoint.R` in `CMD`.

docs/docs/services.md
phinze: 🤖 Yep. Reworded both tables so the whole default is scoped to `web` inheriting the primary image: it gets the single exposed TCP port when available, otherwise `3000`.

docs/docs/services.md
```diff
@@ -145,10 +145,13 @@ name = "myapp"
 
 [services.web]
 image = "ghcr.io/example/myapp:latest"
-args = ["serve", "--port", "8080"]
-port = 8080
 ```
 
+With no other service fields, Miren keeps the image's working directory, `ENTRYPOINT`,
+and `CMD`. A single exposed TCP port becomes the web port automatically. Set `port`
+when the image exposes no ports or several, and use `args` or `command` only when its
```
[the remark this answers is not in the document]
phinze: 🤖 This example does inherit the primary image. When `services.web.image` selects a direct-image source, `inheritPrimaryImage` clears the matching service-level reference before `defaultServicePortFromImage` runs, so the single exposed TCP port is applied. A service image only falls back to 3000 when it differs from the selected primary image, which the table’s “inheriting the primary image” qualifier already covers. Keeping this wording as-is.

PR #1115, opened 2026-09-01
Make sandbox DNS forwarding observable and resilient

phinze:
Sandbox DNS failures currently disappear into a silent SERVFAIL, while glibc can spend ten seconds waiting before the app sees an error. This gives sandbox resolvers a bounded retry policy and gives operators actionable, rate-limited Warn records with the query, client, transport, and upstream that failed.

The bridge resolver now serves both UDP and TCP on IPv4 and IPv6. Truncated upstream replies retry over TCP, then get sized for the client's transport. If upstream TCP is unavailable, UDP clients retain the original truncated reply and can retry against the bridge's TCP listener. The upstream walk remains serial, but one 1.5-second budget bounds the whole request rather than each attempt.

Closes MIR-1704
Closes MIR-1702

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
@@ -10,9 +11,11 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/miekg/dns"
+	"golang.org/x/net/netutil"
 	"miren.dev/runtime/api/compute"
 	"miren.dev/runtime/pkg/entity"
 	"miren.dev/runtime/pkg/entity/indexwatch"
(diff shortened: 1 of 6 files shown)


pkg/dns/dns.go
```diff
@@ -330,31 +391,166 @@ func (s *Server) handlePTRQuery(w dns.ResponseWriter, r *dns.Msg, qname string)
...
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), budget)
+	defer cancel()
 
+	var truncatedFallback *dns.Msg
 	for _, upstream := range s.upstreams {
-		upstream = net.JoinHostPort(upstream, "53")
-		response, _, err = s.client.Exchange(r, upstream)
-		if err == nil && response != nil {
+		if ctx.Err() != nil {
```
evanphx, line 407: A blackholing first upstream eats the whole budget, and this `break` means the second never gets tried at all. `TestForwardUsesOneBudgetAcrossUpstreams` pins exactly that (`blocking.calls == 1`). The old client gave each upstream its own 2s, so it did fail over, just slowly — and a blackholed nameserver is the reason people list two in the first place.

`attempts:3` in the new resolv.conf compounds it: each retry restarts the budget on upstream one, so we burn three rounds never reaching upstream two. Softer version of the same thing without a full blackhole — an upstream that takes 1.4s to fail leaves 100ms for everyone else.

Slicing per attempt (`remaining / attemptsLeft`) keeps the hard cap and the failover both. Non-blocking; happy for it to be a follow-up.

Small one while we're in here: `upstreamTimeout` (2s) never binds, since the 1.5s deadline always fires first. `TestNewServerSetsExplicitUpstreamTimeouts` reads like it's load-bearing.

pkg/dns/dns.go
```diff
@@ -937,11 +1135,71 @@ func (s *Server) RemoveSandboxMapping(sandboxID string) {
...
+	if err != nil {
+		tcpListener.Close()
+		return fmt.Errorf("listening for UDP DNS: %w", err)
+	}
+	s.boundAddr = boundAddr
+
+	started := make(chan struct{}, 2)
+	s.udpServer.PacketConn = udpConn
+	s.udpServer.NotifyStartedFunc = func() { started <- struct{}{} }
+	s.tcpServer.Listener = tcpListener
```
evanphx, line 1182: Serving TCP means miekg spawns a goroutine and holds a map entry per accepted connection — no cap, 8s idle timeout — and sandboxes can reach TCP/53 through the bridge rules. Our own workloads rather than the open internet, so this is a guard against a bug more than an attack, which argues for a generous cap rather than none. `golang.org/x/net` is already in go.mod, so `netutil.LimitListener(tcpListener, N)` is a one-liner. Non-blocking.

evanphx (approved):
Approved! Two inline notes, both non-blocking — the shared-budget one is the only one with any teeth.

--e+🤖

pkg/dns/dns.go
phinze: 🤖 Good catch. The test was pinning the wrong failover behavior. We now divide the remaining 1.5-second budget among the upstreams left, so a blackholed primary cannot starve the secondary, and the test requires both to be attempted. We also aligned the client timeout with the actual forwarding cap.

pkg/dns/dns.go
phinze: 🤖 Agreed. We now wrap the TCP listener with a generous 256-connection cap. Normal DNS-over-TCP connections are short-lived, but this keeps a buggy workload from growing miekg’s connection goroutines and map without bound.

PR #1116, opened 2026-09-01
Make component cleanup survive canceled startup

phinze:
The new boot graph made it much easier to cancel etcd while it was adopting an existing container and still waiting for readiness. That cleanup reused the canceled startup context, so containerd never stopped the task, the container could not be deleted, and the component forgot the handles needed to retry on shutdown.

This makes cleanup independent from the startup context while keeping each containerd call bounded. Cleanup failures now stay visible and retain resource state for the graph's later `Stop`; the same contract applies to the other managed components that share this lifecycle. A containerd regression test covers cancellation during etcd adoption.

Part of MIR-1710

### DIFF
--- a/components/appmetrics/appmetrics.go
+++ b/components/appmetrics/appmetrics.go
@@ -4,6 +4,7 @@ package appmetrics
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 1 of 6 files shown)


evanphx (approved):
Approved! Those pesky contexts. 😅

--e+🤖

PR #1117, opened 2026-09-01
Sync exported apps and deployments to Miren Cloud

phinze:
Cloud now has a negotiated entity-sync capability, but runtime still needs a safe way to decide what may leave the cluster and a durable way to fill and resume it. This generates an explicit cloud export contract from entity schemas, marks and backfills opted-in apps, app versions, and deployments, and filters every entity against that contract before sending it.

The exporter waits for the deployment migration sweep before it starts, snapshots the full eligible set, then tails contiguous revisions from Cloud’s acknowledged cursor. Server boot and the shared uplink stay independent, so repair or sync lag never delays local app serving. Cloud can also schedule periodic full snapshots to repair a silently lost watch.

Part of MIR-1651


controllers/deploymentattempts/controller_test.go
```diff
@@ -201,13 +207,87 @@ func TestMigrationFailureDoesNotBlockLaterPhases(t *testing.T) {
 	for range 3 {
 		require.NoError(t, controller.Step(ctx))
 	}
+	select {
+	case <-controller.Ready():
+		t.Fatal("controller reported readiness after a failed sweep")
```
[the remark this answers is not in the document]
phinze: 🤖 No code change needed here: `migrationFailStore` already overrides `PatchEntity` and returns the injected error for this entity. The focused controller test passes through that path, so this finding is already satisfied.

pkg/entity/store.go
```diff
@@ -58,6 +61,21 @@ type Store interface {
...
 
+// SourceEpoch identifies one etcd revision history. Snapshot restore creates a
+// new logical etcd cluster and therefore a new epoch, preventing a cloud cursor
+// from being reused against an unrelated revision sequence.
+func (s *EtcdStore) SourceEpoch(ctx context.Context) (string, error) {
+	endpoints := s.client.Endpoints()
+	if len(endpoints) == 0 {
+		return "", errors.New("etcd has no endpoints")
+	}
+	status, err := s.client.Status(ctx, endpoints[0])
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. `SourceEpoch` now tries each configured etcd endpoint and reports the joined endpoint errors only when none respond successfully.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,569 @@
...
+			if errors.Is(err, rpctypes.ErrCompacted) {
+				err = errCompacted
+			}
+		}
+		if err != nil {
+			cancelWatch()
+			if errors.Is(err, errCompacted) {
+				snapshotRequired = true
+				continue
+			}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. Both snapshot-start and tail compaction restarts now go through the session retry delay before reissuing etcd operations; the compaction recovery test covers the delayed restart.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,554 @@
...
+	if config.ExportSchema != t.contract.Digest() {
+		t.log.Error("cloud selected an unexpected entity export schema",
+			"selected", config.ExportSchema, "local", t.contract.Digest())
+		return
+	}
+	if config.SourceEpoch != t.sourceEpoch {
+		t.log.Error("cloud selected an unexpected entity source epoch",
+			"selected", config.SourceEpoch, "local", t.sourceEpoch)
+		return
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. Invalid or unexpected Cloud selections now log at warning level and still return cleanly.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,554 @@
...
+				return
+			}
+			continue
+		}
+
+		var watcher clientv3.WatchChan
+		didSnapshot := false
+		if snapshotRequired || cursor > currentHead {
+			cursor, watcher, err = s.snapshot(watchCtx, link)
+			didSnapshot = err == nil
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. Snapshot results now replace the acknowledged cursor only after success. A regression test starts from cursor 42, injects a failed snapshot above a regressed source head, and verifies the retry snapshots again from head plus one rather than tailing from revision 1.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		if event.Type == mvccpb.DELETE {
+			source, err := s.exporter.store.GetEntityAtRevision(s.ctx, id, revision-1)
+			if err != nil {
+				if errors.Is(err, rpctypes.ErrCompacted) {
+					return nil, errCompacted
+				}
+				return nil, fmt.Errorf("read deleted entity %s: %w", id, err)
+			}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. A missing prior entity state on an index deletion now logs and skips the stale event while preserving compaction and other error handling. The new regression test covers the no-prior-state case so the tail can continue advancing.

pkg/entitysync/exporter_test.go
```diff
@@ -0,0 +1,768 @@
...
+		entity.Bool(marker, true),
+	)
+	app.SetRevision(4)
+	store.AddEntity(app.Id(), app)
+	s := &stream{exporter: testExporter(store), ctx: t.Context()}
+
+	changes, err := s.changes([]*clientv3.Event{{
+		Type: mvccpb.DELETE,
+		Kv:   &mvccpb.KeyValue{Value: []byte(app.Id()), ModRevision: 5},
+	}}, 5)
```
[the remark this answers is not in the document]
phinze: Fixed in the latest push. Both delete fixtures now match the WithPrevKV watch shape: the deletion revision comes from Kv, Kv.Value is empty, and the entity ID comes from PrevKv.Value. The focused entity-sync tests pass through that production ID-resolution path.

components/coordinate/coordinate.go
```diff
@@ -461,6 +463,21 @@ func (c *Coordinator) NewDeploymentAttemptController() (*deploymentattemptsctrl.
 	return deploymentattemptsctrl.New(c.Log, c.store, c.eac), nil
 }
 
+// BackfillCloudExportMarker makes existing apps and app versions visible
```
evanphx, line 466: 🤖 Non-blocking, but this comment describes a split that isn't what happens. `Step` calls `ensureCloudExportMarker` in every phase except `reconcile`, which covers all three contract kinds, and `prepareEntitySync` waits on `Ready()` before calling this. By the time the backfill runs, the controller has already marked everything.

I checked with a throwaway test running the boot sequence's exact order against three unmarked legacy entities:

```
after controller sweep: app/web           marked=true
after controller sweep: app_version/v1    marked=true
after controller sweep: deployment/legacy marked=true
BackfillMarker after Ready(): scanned=2 marked=0 already=2
```

I don't think it should go — it's the contract-general path, and the only thing that would cover a fourth kind if `exports:` grows. But someone reading this comment will conclude the backfill owns apps and app versions, and go looking in the wrong place when a mark is missing. Worth saying instead that the controller marks everything it sweeps and this is the net for kinds no controller covers.

components/server/boot_deployment_attempt_migration.go
```diff
@@ -4,33 +4,72 @@ package server
...
+	}
+	go b.prepareEntitySync(ctx, coordinator, ready)
+	return deploymentAttemptMigrationBootOutput{entitySyncReady: ready}, nil
+}
+
+func (b *deploymentAttemptMigrationBoot) prepareEntitySync(ctx context.Context, coordinator coordinatorBootOutput, ready chan<- struct{}) {
+	select {
+	case <-ctx.Done():
+		return
+	case <-b.controller.Ready():
```
evanphx, line 55: 🤖 Non-blocking — the fail-closed behavior here is right and I wouldn't change it.

What I'd want in the field: one deployment with a status outside the known set keeps `passFailed` true forever, so this wait never returns, the backfill never runs, and the exporter sits at its one `entity sync waiting for source preparation` line and then goes quiet. The only ongoing signal is a warning from a different component every 10 seconds that never mentions cloud sync. A periodic "still waiting" from the exporter, or naming the gate in the controller's warning, turns that into a two-minute diagnosis instead of a code read.

components/coordinate/coordinate.go
```diff
@@ -1670,56 +1687,59 @@ func (c *Coordinator) Start(ctx context.Context) error {
...
+		uplinkOptions...,
+	)
+	if labs.AppVisibility() {
+		if err := entitysync.NewExporter(
+			c.Log.With("component", "entity-sync"), c.store, core_v1alpha.CloudExportContract,
+			entitysync.WithStartGate(entitySyncReady),
+		).Register(ctx, link); err != nil {
+			// Entity visibility is additive. Local source metadata should not take
+			// Anywhere or cloud RPC off the shared link when it is unavailable.
+			c.Log.Warn("entity sync is unavailable for this uplink session", "error", err)
```
evanphx, line 1723: 🤖 Non-blocking, and agreed on not taking the link down for this. What gives me pause is that there's no second chance: `Register` runs once per process, so if the epoch read fails here, entity sync is off until someone restarts the server, and this Warn is the only trace.

Same call, smaller point: the epoch is sampled once and baked into the offer, so an external-etcd restore under a live coordinator keeps offering the stale one. The `cursor > currentHead` check catches that in practice, so I don't think it's a hole — but the epoch isn't delivering what its comment in `store.go` describes for an in-process restore. Reading it inside `runSession` would fix both.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,558 @@
...
+
+func durationSeconds(seconds int64) time.Duration {
+	const maxDurationSeconds = int64((time.Duration(1<<63 - 1)) / time.Second)
+	if seconds <= 0 || seconds > maxDurationSeconds {
+		return 0
+	}
+	return time.Duration(seconds) * time.Second
+}
+
+func (s *stream) retry(ctx context.Context, message string, err error) bool {
```
evanphx, line 243: 🤖 Non-blocking: `retry` logs before it looks at the context, so stopping the server mid-read gives you `WARN start entity sync session error=context canceled` on the way out. `tail`'s caller already guards this with `if ctx.Err() != nil { return }`; the two call sites above it don't.

controllers/deploymentattempts/controller.go
```diff
@@ -222,6 +235,22 @@ func (c *Controller) migrateDeployment(ctx context.Context, ent *entity.Entity)
...
+func (c *Controller) ensureCloudExportMarker(ctx context.Context, id entity.Id) error {
+	current, err := c.Store.GetEntity(ctx, id)
+	if err != nil {
+		return err
+	}
+	marker := core_v1alpha.CloudExportContract.MarkerID()
+	if attr, ok := current.Get(marker); ok && attr.Value.Kind() == entity.KindBool && attr.Value.Bool() {
+		return nil
+	}
+	_, err = c.Store.PatchEntity(ctx, entity.New(
```
evanphx, line 247: 🤖 Non-blocking. Same operation as `export.ensureMarker`, except that one retries on conflict and this one returns it. Apps get written constantly during a deploy, so losing the `WithFromRevision` race here is likely, and each loss sets `passFailed` and costs another full sweep before the gate can open. Exporting the retrying version and calling it from both places would collapse the pair.

controllers/deploymentattempts/controller.go
```diff
@@ -109,7 +117,11 @@ func (c *Controller) Step(ctx context.Context) error {
 	}
 	var migrateErrs []error
 	for _, ent := range entities {
-		if err := migrate(ctx, ent); err != nil {
+		err := migrate(ctx, ent)
```
evanphx, line 120: 🤖 Predates this change, so purely an observation: `GetEntities` leaves `nil` in the slice for an id whose entity is gone (`store.go:541`), and this loop doesn't guard for it — `migrate` would deref it. `BackfillMarker` guards for exactly this and the exporter's snapshot has a test for the stale-index case, so it looks like a real condition rather than a theoretical one.

api/core/cloud_export_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require.Equal(t, "web", entity.MustGet(filtered, core_v1alpha.DeploymentAppNameId).Value.String())
+	require.Equal(t, "dep-1", entity.MustGet(filtered, entity.DBShortId).Value.String())
+	actor := entity.MustGet(filtered, core_v1alpha.DeploymentDeployedById).Value.Component()
+	require.Equal(t, "user-123", entity.MustGet(actor, core_v1alpha.DeployedBySubjectId).Value.String())
+	_, ok = actor.Get(core_v1alpha.DeployedByUserEmailId)
+	require.False(t, ok)
+
+	encoded, err := json.Marshal(filtered)
+	require.NoError(t, err)
+	for _, excluded := range []string{
```
evanphx, line 60: 🤖 👏 the kind of test that earns its keep.

pkg/entitysync/exporter.go
```diff
@@ -0,0 +1,558 @@
...
+
+func (s *stream) deliver(ack Ack) {
+	s.mu.Lock()
+	waiter := s.waiters[ack.MessageID]
+	s.mu.Unlock()
+	if waiter == nil {
+		return
+	}
+	select {
+	case waiter <- ack:
```
evanphx, line 550: 🤖 👏 subtle and easy to get wrong.

evanphx (approved):
Approved! Really nicely done.

All the inline notes are non-blocking. The backfill one is the only one I'd act on before merge, and even that's a comment fix rather than a code fix.

--e+🤖

PR #1118, opened 2026-09-02
Make `miren cluster add` scriptable

evanphx:
You couldn't script `miren cluster add`. Giving it a cluster name without an address was refused ("both --cluster and --address must be specified, or neither"), so the only ways in were the interactive picker or passing `--address` yourself. Passing the address skips the useful part: trying each address the cluster advertises, asking Miren Cloud to pass commands through when none of them answer, and saving the cluster's certificate to check later connections against. MIR-1592 asked for a way to name a cluster and still get all that.

Now `--cluster prod` on its own runs the same logic the picker does. `--as` sets the local name the picker used to prompt for, and `--organization` separates two clusters that share a name.

A cluster that advertises no address and that Miren Cloud can't reach is greyed out in the picker. Naming it has no row to grey out, and the old code would have fallen back to trying `127.0.0.1:8443`. On a dev box that could save your local cluster's certificate under a remote cluster's name, so it's refused now.

`miren cluster available` is where the name comes from: the clusters Miren Cloud holds for your account, with `--format json`. Checking whether cloud can reach a cluster with no address costs a request each, so that sits behind `--check`.

`cluster add --format json` then reports the outcome. Success says what was written and how the cluster will be reached. Failure gives a code you can branch on and still exits non-zero. A name that doesn't exist and a cluster that's temporarily unreachable get different codes, because they need different responses. JSON mode won't open the picker or ask about overwriting either, even at a terminal. `ctx.Info` writes to stdout, so JSON commands now send progress to stderr, or the document doesn't parse.

Closes MIR-1592

### DIFF
--- a/cli/commands/cluster_add.go
+++ b/cli/commands/cluster_add.go
@@ -332,6 +332,12 @@ func addCluster(ctx *Context, opts addClusterOptions) (*addedCluster, error) {
 	// mode, where cloud is never asked and neither is knowable.
 	var cloudName, organizationName string
 
+	// Whether the caller asked for cloud routing, as opposed to the code
+	// choosing it below. Captured before opts.viaCloud can be set, because the
+	// two cases warrant different treatment: a route chosen here has already
+	// been probed, while one the caller named has not been checked at all.
+	explicitViaCloud := opts.viaCloud
+
 	if address == "" {
 		ctx.Info("Fetching available clusters from identity server...")
 
@@ -416,10 +422,22 @@ func addCluster(ctx *Context, opts addClusterOptions) (*addedCluster, error) {
 		}
 
 		if opts.viaCloud {
-			// No address, no probe, no certificate. Every one of those describes
-			// dialing the cluster, which is the thing this entry exists to avoid
-			// — and probing would just fail slowly before writing the entry that
-			// was going to work.
+			// No address and no certificate: both describe dialing the cluster,
+			// which is the thing this entry exists to avoid.
+			//
+			// The route itself is checked when the caller asked for it by name,
+			// rather than when it was chosen above (which already probed). The
+			// entry is written either way, because --via-cloud says how this
+			// cluster is reached and not that it answers this minute, and a
+			// cluster can reasonably be offline while somebody sets it up. But
+			// an entry that cannot work right now is worth saying out loud,
+			// since otherwise the first sign of it is an unrelated command
+			// failing later.
+			if explicitViaCloud && !canFallBackToCloud(ctx, mainConfig, identityName, identity, selectedCluster) {
+				ctx.Warn("%s did not answer through cloud just now, so commands to it will fail until it reconnects.", selectedCluster.Name)
+				ctx.Warn("Adding it anyway, since --via-cloud records how to reach it rather than that it is up.")
+			}
+
 			announceClusterAdd(ctx, selectedCluster.Name, localName, "routed through Miren Cloud")
 		} else {
 			// Store all available addresses
--- a/cli/commands/cluster_add.go
+++ b/cli/commands/cluster_add.go
@@ -28,7 +28,8 @@ var skipAddresses = map[string]bool{
 	"localhost.localdomain": true,
 }
 
-func ClusterAdd(ctx *Context, opts struct {
+type clusterAddOpts struct {
+	FormatOptions
 	Identity     string `short:"i" long:"identity" description:"Name of the identity to use (optional - will use the only one if single)"`
 	Cluster      string `short:"c" long:"cluster" description:"Name of the cluster to add, looked up in Miren Cloud unless --address is given (optional - will list available)"`
 	Address      string `short:"a" long:"address" description:"Address/hostname of the cluster (optional - will use from selected cluster)"`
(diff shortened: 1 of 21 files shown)


cli/commands/cluster_add_json.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+// The codes a caller can branch on. They exist because the message is written
+// for a person and will be reworded; the code is the part a script is allowed
+// to depend on, so each one names a distinct situation with a distinct fix.
+const (
```
phinze, line 11: Do we want to make this whole taxonomy a public contract here? This is the only command-specific `codedError` / `{ok, cluster|error}` convention in the CLI, and the generated docs neither describe the envelope nor list the codes. The scripting use case clearly needs to distinguish `cluster_not_found` from `cluster_unreachable`, but 16 stable categories feels broader than the behavior a caller can act on (and `cancelled` cannot be emitted in JSON mode). Could we keep the few actionable codes and fold the rest into a generic failure?

--p+🤖

cli/commands/cluster_available.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return err
+	}
+
+	identity, err := lookupIdentity(config, identityName)
+	if err != nil {
+		return err
+	}
+
+	clusters, err := fetchAvailableClusters(ctx, config, identityName, identity)
```
phinze, line 83: This fetching now has four consumers (`cluster add`, `cluster available`, login, and switch), and the adjacent online check has multiple consumers too. Could we seed a small `cloudapi.Client` here rather than adding more HTTP/API machinery to `cli/commands`? It only needs the list and online endpoints now, with base URL, authentication, request decoding, and the fake Cloud fixture moving with it. Identity selection, interaction, rendering, and route decisions can stay in commands, so this does not need to become a broad framework or migrate every endpoint in this PR.

--p+🤖

phinze (approved):
The named add flow and its scripting behavior look solid! I left two non-blocking design comments about keeping the new surface area contained: narrowing the JSON error contract and moving the shared Cloud API calls behind a small client.

--p+🤖

cli/commands/cluster_add_json.go
evanphx: 🤖 Good call, both halves of it. Trimmed to 12 in 83f0e411: `config_load_failed` and `config_write_failed` became `config_error` since a caller can't fix either by asking differently, `no_clusters` folded into `cluster_not_found`, and the two identity-argument codes became `identity_error`. Kept `no_identities` separate because its fix is a person running `miren login` rather than a different flag.

You were right that `cancelled` can't fire in JSON mode; it only comes from the interactive overwrite prompt, which JSON mode never opens. It's a plain error now.

On the documentation gap: the envelope and the full table are the command's extended description, so they render on the generated page. There's a test comparing the table against the constants, since a documented code that never arrives is worse than an undocumented one.

cli/commands/cluster_available.go
evanphx: 🤖 Done in 71d8b510. Your count was right: four callers for the list (`cluster add`, `cluster available`, `login.go:781`, `cluster_switch.go:135`) and two for the online check.

`pkg/cloudapi` now owns the base URL, auth, and decoding, exposing `ListClusters` and `ClusterOnline`. The token is a function rather than a string so refresh stays with the caller. The fake cloud moved too, as `cloudapitest`, so one stand-in serves both the command tests and the client's own. `ClusterResponse` is an alias for `cloudapi.Cluster`, which kept the migration small.

Kept it to the two endpoints as you suggested. Identity selection, rendering, and route decisions stayed in `commands`.

cli/commands/cluster_add.go
```diff
@@ -148,63 +308,96 @@ func addCluster(ctx *Context, opts addClusterOptions) error {
...
+				cloudRoutable = cloudRoutableClusters(ctx, mainConfig, identityName, identity, []ClusterResponse{*selectedCluster})
+			}
+
+			// The picker greys out a cluster with nothing to dial that cloud
+			// cannot reach either. Named, there is no row to grey out, and
+			// falling through would probe localhost as a last resort — which
+			// on a dev machine can pin whatever is listening there under a
+			// remote cluster's name.
+			if !opts.viaCloud && !selectedCluster.hasReachableAddress() && !cloudRoutable[selectedCluster.XID] {
+				return nil, codedErrorf(codeClusterUnreachable, "%s has %s, and cloud cannot reach it either; to connect: %s",
```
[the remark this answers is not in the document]
evanphx: 🤖 You're right that nothing verified the route here. We landed a warning rather than a rejection in 7e5f6d8f.

The reasoning: `--via-cloud` records how a cluster is reached, not that it answers this minute, and refusing to write the entry would block setting up a cluster that's temporarily offline. So the route is now probed when the caller names it explicitly, and a route that doesn't answer gets a warning saying the entry was written anyway. The automatically chosen path already probes before choosing, so it's unchanged.

cli/commands/cluster_add_json.go
```diff
@@ -0,0 +1,174 @@
...
+// the command's page. The codes are a contract callers write against, so they
+// belong somewhere a caller can read without opening the source.
+const clusterAddJSONDoc = "With `--format json`, the command prints one result document and nothing else " +
+	"on stdout; progress and warnings go to stderr. A failure is reported both as a " +
+	"document and as a non-zero exit status.\n\n" +
+	"```json\n" +
+	"{\"ok\": true, \"cluster\": {\"name\": \"prod\", \"cloud_name\": \"prod\", \"xid\": \"cluster-...\",\n" +
+	"                          \"organization\": \"Acme\", \"address\": \"10.0.0.1:8443\",\n" +
+	"                          \"via_cloud\": false, \"identity\": \"cloud\", \"active\": true,\n" +
+	"                          \"config_file\": \"~/.config/miren/clientconfig.d/prod.yaml\"}}\n" +
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch, that document can't exist. `cloud_name` is only set when it differs from the local name, so it can never sit next to an identical `name`. The example now shows the ordinary case without it, plus a sentence saying `cloud_name` appears only when `--as` gave the cluster a different local name, and that `address` is absent for a cloud-routed cluster.

pkg/cloudapi/cloudapi.go
```diff
@@ -0,0 +1,149 @@
...
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := (&http.Client{Timeout: timeout}).Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
evanphx: 🤖 Agreed, capped at 64KB. The body only ever goes into an error message, so reading all of whatever a proxy decides to send would turn a failed request into a failed process.

PR #1145, opened 2026-09-03
Harden entity sync recovery and migration safety

phinze:
Runtime’s entity-sync path is safe to lag behind local serving, but several recovery edges could make that lag silent, sticky, or lossy. A failed source-epoch read lasted for the process lifetime, an unanswered batch stalled a session indefinitely, and catch-up responses could acknowledge the store head before delivering every chunk. Normal index churn could also force another full sweep or panic the migration loop.

This refreshes the epoch offer for each uplink session, bounds acknowledgment waits so the existing retry loop can resume from the last cursor, and advances catch-up cursors through the last delivered event rather than the response header. The preparation gate now logs periodically, marker writes share the conflict-retrying path, stale index rows are skipped, and normal shutdown cancellation stays quiet.

This follows up on the non-blocking review notes from #1117 and folds in the catch-up cursor bug reported by Detail in #1135. Focused packages, the repository compile sweep, and the changed concurrency paths under the race detector pass. The coordinate package also passes in `hack/it`.

Part of MIR-1651
Closes MIR-1739

### DIFF
--- a/controllers/deploymentattempts/controller.go
+++ b/controllers/deploymentattempts/controller.go
@@ -15,6 +15,7 @@ import (
 	"miren.dev/runtime/pkg/cond"
 	"miren.dev/runtime/pkg/deploylifecycle"
 	"miren.dev/runtime/pkg/entity"
+	entityexport "miren.dev/runtime/pkg/entity/export"
 )
 
 const (
@@ -116,10 +117,17 @@ func (c *Controller) Step(ctx context.Context) error {
 		return err
 	}
 	var migrateErrs []error
+	marker := core_v1alpha.CloudExportContract.MarkerID()
 	for _, ent := range entities {
+		if ent == nil {
+			continue
+		}
 		err := migrate(ctx, ent)
 		if err == nil && c.phase != phaseReconcile {
-			err = c.ensureCloudExportMarker(ctx, ent.Id())
+			attr, marked := ent.Get(marker)
+			if !marked || attr.Value.Kind() != entity.KindBool || !attr.Value.Bool() {
+				_, err = entityexport.EnsureMarker(ctx, c.Store, ent.Id(), marker)
+			}
 		}
 		if err != nil {
 			wrapped := fmt.Errorf("migrating %s in phase %s: %w", ent.Id(), c.phase, err)
@@ -232,22 +240,6 @@ func (c *Controller) migrateDeployment(ctx context.Context, ent *entity.Entity) 
 		attrs = append(attrs, entity.Time(core_v1alpha.DeploymentStartedAtId, started))
 	}
 	_, err := c.Store.PatchEntity(ctx, entity.New(attrs), entity.WithFromRevision(ent.GetRevision()))
-	return err
-}
-
-func (c *Controller) ensureCloudExportMarker(ctx context.Context, id entity.Id) error {
-	current, err := c.Store.GetEntity(ctx, id)
-	if err != nil {
-		return err
-	}
-	marker := core_v1alpha.CloudExportContract.MarkerID()
-	if attr, ok := current.Get(marker); ok && attr.Value.Kind() == entity.KindBool && attr.Value.Bool() {
-		return nil
-	}
-	_, err = c.Store.PatchEntity(ctx, entity.New(
-		entity.Ref(entity.DBId, id),
-		entity.Bool(marker, true),
-	), entity.WithFromRevision(current.GetRevision()))
 	return err
 }
 
--- a/pkg/entitysync/exporter.go
+++ b/pkg/entitysync/exporter.go
@@ -25,12 +25,15 @@ const snapshotBatchSize = 100
 const maxLiveBatchesBetweenSnapshotBatches = 8
 const sessionRetryDelay = time.Second
 const defaultMinimumResnapshotInterval = time.Hour
+const defaultAckTimeout = 30 * time.Second
+const defaultPreparationLogInterval = time.Minute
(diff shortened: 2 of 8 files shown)


pkg/entitysync/exporter.go
```diff
@@ -113,23 +115,22 @@ func (t *Exporter) runSession(ctx context.Context, session uplink.Session, link
...
-	resnapshotInterval, nextSnapshot := t.resnapshotSchedule(config)
-	if t.startGate != nil {
-		t.log.Info("entity sync waiting for source preparation")
-		select {
-		case <-ctx.Done():
-			return
-		case <-t.startGate:
-			t.log.Info("entity sync source is ready")
-		}
+	sourceEpoch, err := t.readSourceEpoch(ctx)
```
evanphx, line 122: Non-blocking, and probably a follow-up ticket rather than a change here: this is the only epoch read for the whole session, and the loop below runs indefinitely.

If etcd is restored mid-session, revisions restart low, `cursor > currentHead` trips a full snapshot, and that snapshot goes out under the stale epoch. Cloud rejects it at `SnapshotComplete`, `retry` waits a second, and we upload the whole thing again — until the uplink connection drops, which nothing in this path causes.

Re-reading at the top of the loop and returning when it changes would be close to free, since the loop normally parks inside `tail` for a long time.

pkg/entitysync/exporter.go
```diff
@@ -528,9 +570,11 @@ func (s *stream) sendAndWait(link Link, messageType string, payload any, message
 	if err := link.SendMessageBlocking(s.ctx, messageType, payload); err != nil {
 		return Ack{}, err
 	}
+	waitCtx, cancelWait := context.WithTimeout(s.ctx, s.exporter.ackTimeout)
```
evanphx, line 573: Non-blocking: `SnapshotComplete` shares this budget with change batches, and those have pretty different latency profiles — a change batch is small, but this is where cloud commits an entire snapshot. Missing the 30s there costs a full re-upload.

Do we know what that commit looks like on a large cluster? If it's anywhere near 30s it'd be worth either a separate bound for the snapshot ack or an `Option` so it's tunable — right now `ackTimeout` is only reachable from tests.

controllers/deploymentattempts/controller.go
```diff
@@ -117,9 +118,14 @@ func (c *Controller) Step(ctx context.Context) error {
 	}
 	var migrateErrs []error
 	for _, ent := range entities {
+		if ent == nil {
+			continue
+		}
 		err := migrate(ctx, ent)
 		if err == nil && c.phase != phaseReconcile {
-			err = c.ensureCloudExportMarker(ctx, ent.Id())
+			_, err = entityexport.EnsureMarker(
```
evanphx: Non-blocking: we already have the full entity in `ent`, but `EnsureMarker` opens with its own `GetEntity` before it can check the marker.

`BackfillMarker` guards this — it checks `source.Get(contract.MarkerID())` on the entity in hand and only calls `EnsureMarker` when the marker is actually missing. Same guard here would make the two call sites match and drop ~200 redundant reads every 10s on a cluster that's already converged.

pkg/uplink/client.go
```diff
@@ -241,17 +269,29 @@ func (c *Client) connectCallbacks() []func(ctx context.Context) {
...
-	defer c.mu.Unlock()
-
-	offers := make([]CapabilityOffer, len(c.capabilities))
-	for i, offer := range c.capabilities {
-		offers[i] = offer
-		offers[i].Versions = slices.Clone(offer.Versions)
-		offers[i].Offer = slices.Clone(offer.Offer)
+	capabilities := slices.Clone(c.capabilities)
+	callbacks := slices.Clone(c.onSession)
+	c.mu.Unlock()
```
evanphx, line 276: 👏 clone under the lock, call `provide` outside it

evanphx (approved):
Approved! Really nicely done.

Both new cursor tests go red without the fix, which is the check I most wanted on this one. Three non-blocking notes inline — the mid-session epoch one is the only one I'd bother tracking.

--e+🤖

pkg/entitysync/exporter.go
phinze: 🤖 Good catch. A loop-boundary re-read would stop the repeated upload, but returning would also strand entity sync because the selected epoch is fixed for the current WebSocket session. We need the exporter to invalidate and renegotiate the shared uplink cleanly when the source epoch changes. I’ve tracked that complete fix in [MIR-1764](https://linear.app/miren/issue/MIR-1764/renegotiate-entity-sync-when-the-source-epoch-changes-mid-session).

pkg/uplink/client.go
phinze: 🤖 Thanks! That boundary felt important once providers could do store I/O.

controllers/deploymentattempts/controller.go
phinze: 🤖 Good catch. I added the same bool-true marker guard before `EnsureMarker`, plus a regression asserting that a converged sweep makes no extra point read.

pkg/entitysync/exporter.go
phinze: 🤖 Agreed that snapshot completion has a different cost profile. Cloud only counts and sweeps already-landed rows at completion, but we don’t have large-cluster timing yet to choose a meaningful second bound. I’m keeping the 30s fail-safe here and using the staged rollout to tell us whether snapshot completion needs its own timeout.

PR #1146, opened 2026-09-03
Give shared enums one set of member entities

phinze:
Enum choices in our entity schemas are already references. For every `choices:` value, schemagen creates a member entity and stores a reference to it. That gives enum membership the same referential-integrity behavior as any other entity reference.

The problem is that those members currently belong to a field. Two fields can describe the same concept but use different Go types and different member entities. Converting between those generated values can compile while still writing a reference from the wrong field. That is the bug behind MIR-942.

This adds named enums. Fields that name the same enum share one generated Go type and one set of member entities. Inline `choices:` still works as before, so enums used by only one field need no additional declaration.

Core now shares its port protocol and disk provider enums, while compute shares its port protocol enum. Each declaration chooses one existing member namespace as canonical, so those member entities already exist on upgraded clusters.

The migrated fields use `legacy-choices: true`. That tells schemagen that the field used to own an inline choice set. Generated readers continue to recognize its old field-local member refs, while writers always emit the shared member ref. Writers also recognize the old generated Go strings, which keeps a build saga suspended across the upgrade from silently dropping its disk provider or port protocol.

Existing entities do not need a background rewrite. They remain readable through the aliases, and their refs become canonical the next time the entity is naturally read and written. Unchanged legacy refs remain valid during unrelated patches, including inside repeated components.

Storage is intentionally unchanged. `disk_volume.filesystem` remains an unconstrained string, and this PR adds neither a convergence controller nor a general migration framework.

Compatibility is still one-way during a mixed-version rollout: old binaries do not recognize the new shared member refs. MIR-1520 tracks moving runner entity access behind coordinator RPC, which is the longer-term way to remove that upgrade coupling.

This also establishes the shared-enum groundwork for MIR-1239 and MIR-1447.

Closes MIR-942

## Verification

- `make lint`
- Focused generator, schema, validation, natural-format, entity-server, core, compute, deployment, and sandbox tests
- `make test` ran 7,437 tests with 21 skips. Its five failures were in four integration packages sharing fixed ports or nftables state. The three addon packages passed when rerun serially, and the nftables test passed alone after restarting the test container.


phinze:
@coderabbitai review

api/core/core_v1alpha/schema.gen.go
```diff
@@ -1805,39 +1834,39 @@ const (
 	DisksNameId           = entity.Id("dev.miren.core/disks.name")
 	DisksOwnerId          = entity.Id("dev.miren.core/disks.owner")
 	DisksProviderId       = entity.Id("dev.miren.core/disks.provider")
-	DisksProviderMirenId  = entity.Id("dev.miren.core/provider.miren")
-	DisksProviderLocalId  = entity.Id("dev.miren.core/provider.local")
-	DisksProviderSqliteId = entity.Id("dev.miren.core/provider.sqlite")
+	DisksProviderMirenId  = DiskProviderMirenMemberId
+	DisksProviderLocalId  = DiskProviderLocalMemberId
+	DisksProviderSqliteId = DiskProviderSqliteMemberId
```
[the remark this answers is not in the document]
phinze: Checked this against the generated file on the pre-change parent. The existing IDs were `dev.miren.core/provider.{miren,local,sqlite}` and `dev.miren.core/protocol.{tcp,udp}`, not `disks.provider.*` or `ports.protocol.*`. Those existing IDs are the canonical member IDs in this change, so the decode maps already retain them. I added literal-ID decode/re-encode coverage for all five values to pin that compatibility.

api/storage/schema.yml
```diff
@@ -130,7 +135,9 @@ kinds:
       required: true
 
     filesystem:
-      type: string
+      type: enum
+      enum: disk_filesystem
+      legacy-encoding: string
```
evanphx: This is the change I'd like to pull out, along with `disk_filesystem` above it. I'd rather the enum arrive as a new attribute in a future PR than have this one change type under stored data.

pkg/entity/validation.go
```diff
@@ -180,38 +180,51 @@ func (v *Validator) ValidateAttributes(ctx context.Context, attrs []Attr) error
...
-// appearing unchanged (same id and value) in the original attribute set.
-func exemptUnchangedRefs(original []Attr) refExemptFn {
-	byID := make(map[Id][]Attr, len(original))
-	for _, a := range original {
-		byID[a.ID] = append(byID[a.ID], a)
+// matchOriginalAttrs pairs unchanged attributes as a multiset within their
+// current entity or component. Changed components are then paired with a
+// remaining component at the same scope, so nested exemptions cannot leak
+// between repeated component instances.
+func matchOriginalAttrs(attrs, original []Attr) []*Attr {
```
evanphx, line 190: This one stays even with the sweep gone — it's holding up the shared enums, not the migration. Once the installed `EnumValues` are canonical-only, patching a component that still holds a legacy ref needs this pairing to not fail on the unchanged nested value.

api/core/schema.yml
```diff
@@ -263,7 +272,8 @@ kinds:
                   doc: The name of the disk
                 provider:
                   type: enum
-                  choices: [miren, local, sqlite]
+                  enum: disk_provider
+                  legacy-encoding: ref
```
evanphx: Worth a thought before this lands: the Go value of `ConfigSpecServicesDisksLOCAL` goes from `component.config_spec.services.disks.provider.local` to `local`, and `ConfigSpec` gets JSON'd into saga state between steps (`build_saga.go:1064`) while `coordinate.go:494` recovers build sagas on restart. A deploy suspended across the upgrade resumes holding the old strings, `createConfigVersion` unmarshals them, the encode map has no entry, and the provider drops out — a local disk gets written as a network one. One deploy, heals on redeploy, but it's not in the compat notes.

pkg/model/natural.go
```diff
@@ -138,6 +138,21 @@ func decodeNaturalValue(f *entity.SchemaField, v any) ([]entity.Attr, error) {
...
+	// schema. Older schemas represented ref-backed enums solely as Type "enum"
+	// plus EnumValues; SchemaField.EnumValue continues to handle those as well.
+	if f.Enum != "" || f.Type == "enum" {
+		enum, ok := v.(string)
+		if !ok {
+			return nil, fmt.Errorf("failed to decode field %s: expected string, got %T", f.Name, v)
+		}
+		value, ok := f.EnumValue(enum)
+		if !ok {
+			return nil, fmt.Errorf("failed to decode field %s: enum value %q is not declared, expected one of %v", f.Name, enum, f.EnumMembers)
```
evanphx: `EnumMembers` is only populated for named enums, so every remaining inline `choices:` field reports `expected one of []`. Falling back to the keys of `EnumValues` would cover both. Non-blocking.

evanphx (changes requested):
Requesting changes on scope only — the named enum work is good and I want it to land. The ask is to drop the filesystem conversion and the convergence engine and ship the rest.

I went looking for something that forces a rewrite of the old field-local refs and couldn't find it. They read fine through the alias in the generated `FromId` maps, writes are canonical because the encoder only ever consults `ToId`, an unchanged value is exempted on update, and none of the eight enum fields is indexed, so no query misses an unconverged row. Any read-modify-write converges an entity on its own. Those seven rules can just not run.

`disk_volume.filesystem` is the one rule that isn't ref→ref, and it's the one I'd rather do differently. It's the only attribute here whose type pivots underneath live data, and it was an unconstrained string before — anything outside `{ext4, xfs, btrfs}` decodes to empty now and drops out on the next encode.

Which means storage drops out of this PR entirely: with `disk_volume` back to a string, `disk_filesystem` has one user left, and this PR's own rule says one user is a `choices:` enum.

Inline notes below; the saga one is the only one with teeth.

--e+🤖

api/storage/schema.yml
phinze: 🤖 Done. Storage is unchanged again: `disk_volume.filesystem` remains a string, the added named `disk_filesystem` enum is gone, and no storage-generated or controller files remain in the diff.

api/core/schema.yml
phinze: 🤖 Good catch. Generated `ToId` maps now recognize the old field-specific Go strings and map them to the canonical member ref. I added a test that JSON-round-trips a `ConfigSpec` carrying the old disk-provider and port-protocol values, then verifies both still encode after resume.

pkg/entity/validation.go
phinze: 🤖 Kept. I also retained the location-aware matching so an unchanged nested legacy ref remains patchable without borrowing an exemption from another instance of the same repeated component.

pkg/model/natural.go
phinze: 🤖 Fixed. The error now derives a sorted member list from `EnumValues`, so both inline `choices:` enums and named enums report their actual choices.

phinze:
🤖 Yep, agreed. This is much better scoped now. The new head drops the filesystem conversion, all storage changes, and the convergence engine. What remains is the ref-backed named-enum feature, the genuinely shared core/compute enums, and enough compatibility for old field-local refs and suspended saga values.

servers/entityserver/entityserver.go
```diff
@@ -835,6 +832,28 @@ func (e *EntityServer) MakeAttr(ctx context.Context, req *entityserver_v1alpha.E
 	return nil
 }
 
+func enumRefFromString(values []entity.Value, input string) (entity.Value, bool) {
+	// Ref-backed enum values accept their short member names as input.
+	for _, value := range values {
+		if value.Kind() == entity.KindId && string(value.Id()) == input {
+			return value, true
+		}
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 We’re keeping `TypeEnum` reference-only here. Before this PR, `MakeAttr` already converted enum input to `RefValue`; the new `SchemaBuilder.Enum` now makes that contract explicit with `[]entity.Id` and `EntityElemType: TypeRef`. `EnumValues` itself stays `TypeAny` because the metadata attribute is also used by `schema.Choices` and `db/type`. Accepting a manually constructed string-valued `TypeEnum` would add the multiple physical encodings we intentionally removed from this PR’s scope.

pkg/entity/cmd/schemagen/generator.go
```diff
@@ -677,82 +993,111 @@ func (g *gen) attr(name string, attr *schemaAttr) {
...
+			if legacyType != enumType {
+				g.decodeouter = append(g.decodeouter, j.Type().Id(legacyType).Op("=").Id(enumType))
 			}
+			g.decodeouter = append(g.decodeouter, j.Const().DefsFunc(func(group *j.Group) {
+				for _, value := range enum.Values {
+					legacyConstant := strings.ToUpper(value)
+					if g.isComponent {
+						legacyConstant = g.local + legacyConstant
+					}
+					group.Id(legacyConstant).Id(enumType).Op("=").Id(enumConstant(attr.Enum, value))
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push. A legacy enum constant stays unqualified when it is unique, preserving the existing API. When multiple kind fields would emit the same package-level name, each alias now uses its owning generated field as a prefix. The generator test covers two fields sharing `tcp` and `udp`.

evanphx (approved):
Looks good! Cleans up a bunch of sharp corners in the UX.

PR #1151, opened 2026-09-03
Keep coordinator RPC alive through graceful shutdown

phinze:
The boot graph cancels the coordinator context before dependent components stop. That could send HTTP/3 GOAWAY while the OCI registry was making its final entity RPC for a manifest push, turning an otherwise clean shutdown into a 500.

This gives RPC state an explicit bounded shutdown and invokes it from the coordinator’s ordered stop hook. Server-owned streaming calls are canceled before the drain so long-lived watches from distributed runners cannot consume the deadline, while unary requests remain free to finish. Context-driven cleanup remains in place for standalone RPC state users.

The affected RPC, coordinator, server, registry, and runner suites pass after rebasing onto current main. The RPC package also passes under the race detector.

Closes MIR-1748

### DIFF
--- a/components/coordinate/coordinate.go
+++ b/components/coordinate/coordinate.go
@@ -506,8 +506,10 @@ func (c *Coordinator) ExposeWorkServices() error {
 	return nil
 }
 
-// Stop stops the coordinator and all managed controllers
-func (c *Coordinator) Stop() {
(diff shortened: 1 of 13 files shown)


evanphx (approved):
Approved! Traced the stop ordering end to end and it holds together.

--e+🤖

PR #1152, opened 2026-09-03
Adopt the boot graph for distributed runner startup

phinze:
Distributed runner startup was the remaining hand-built lifecycle after the server adopted the boot graph. Containerd, telemetry, runner construction, and reverse shutdown were coordinated through local variables and defers, leaving failure rollback and ordering implicit.

This moves resource ownership into a three-node graph. Containerd and telemetry publish ready capabilities to the runner, producer failures prevent runner construction, and the same edges put runner shutdown before its dependencies. The containerd lifecycle is now shared with server startup instead of maintained twice.

Command policy, certificate reconciliation, signal handling, and the final background-task wait stay in the CLI. Runner-specific resolver and certificate-file preparation remain ordinary runner setup because they have no independent lifetime. The telemetry token source is armed only after the runner connection exposes its workload issuer.

Focused tests, repository-wide compilation, race tests, and lint pass.

Closes MIR-1690


components/distributedrunner/boot_telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+			TokenSource:   result.tokenSource,
+			Timeout:       b.inputs.timeout,
+		})
+		if err != nil {
+			return telemetryBootOutput{}, fmt.Errorf("building telemetry client: %w", err)
+		}
+		b.client = client
+	}
+
+	if b.inputs.victoriaMetricsAddress != "" {
```
evanphx, line 85: Two comments from the old `runner_start.go` didn't come along, and I think both are worth putting back.

The first explained that telemetry ships to the coordinator's ingest endpoints, and that `VictoriametricsAddress`/`VictorialogsAddress` are never dialed — their presence is only the signal that the cluster records telemetry at all. Without it this reads as a bug: we check one field for emptiness and then build the URL out of a different one.

The second was on the batch writer below — why this path batches, and why the error handler can log at Error without feeding back into the buffer it's reporting on. That second half is the part nobody would reconstruct.

components/distributedrunner/boot_runner.go
```diff
@@ -0,0 +1,213 @@
...
+	deps.Resolver = resolver
+	coordinatorHost, coordinatorPort, splitErr := net.SplitHostPort(b.inputs.coordinator)
+	if splitErr != nil {
+		b.inputs.log.Warn("in-cluster API access disabled: coordinator address has no usable host and port",
+			"coordinator", b.inputs.coordinator, "error", splitErr)
+	} else if coordinatorAddr, err := resolveHost(coordinatorHost); err != nil {
+		b.inputs.log.Warn("could not resolve coordinator address", "host", coordinatorHost, "error", err)
+	} else {
+		hostMapper.SetHost("cluster.local", coordinatorAddr)
+		deps.ApiAddress = net.JoinHostPort(coordinatorAddr.String(), coordinatorPort)
```
evanphx, line 153: The "must be an IP" rationale got lost here and in the trimmed `resolveHost` doc — sandbox DNS answers `app.miren` names and nothing else, so the coordinator's hostname wouldn't resolve inside a container. That's the whole reason `resolveHost` exists, and it's now only implied.

components/distributedrunner/boot_runner.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (b *runnerBoot) stop(context.Context) error {
+	if b.runner == nil {
+		return nil
+	}
+	return b.runner.Close()
+}
+
+func serviceNetworkPrefixes() []netip.Prefix {
```
evanphx, line 188: The `// TODO: These should come from join response` that sat on this list didn't survive the move. It was the only record that these are hardcoded on purpose.

components/distributedrunner/runtime.go
```diff
@@ -0,0 +1,90 @@
...
+	"golang.org/x/sync/errgroup"
+	"miren.dev/runtime/clientconfig"
+	"miren.dev/runtime/pkg/boot"
+	"miren.dev/runtime/pkg/runnerconfig"
+)
+
+const (
+	// ShutdownTimeout is the overall ceiling for stopping the distributed
+	// runner graph. Each component gets a smaller budget so one slow stop does
+	// not consume the time reserved for its producers.
```
evanphx: Non-blocking, but the budgets don't bound anything yet: only the containerd stop passes its context down. `runnerBoot.stop` and `telemetryBoot.stop` both ignore theirs and call `Close()` on things that take no context.

`runnerComponentStopTimeout` also came over verbatim from `components/server`, where the 2 minutes is sized for `stopAllSandboxContainers`. Here stop is just `runner.Close()`, so the number doesn't have a story anymore.

components/containerd/boot.go
```diff
@@ -0,0 +1,146 @@
...
+// daemon. Embedded is nil when the daemon is managed outside this process.
+type BootConfig struct {
+	Log              *slog.Logger
+	DataPath         string
+	SocketPath       string
+	Embedded         *Config
+	ReadinessTimeout time.Duration
+	StopTimeout      time.Duration
+
+	resolveBinary bool
```
evanphx: Small: every other field on `BootConfig` is exported, so a literal `BootConfig{Embedded: &Config{BinaryPath: "containerd"}}` skips the PATH lookup and dies with `containerd binary not found at containerd`. Deriving it from the path shape (`filepath.Base(p) == p`) would make the struct safe to build without the constructors. Both call sites use them today, so nothing's broken.

components/containerd/boot.go
```diff
@@ -0,0 +1,146 @@
...
+		socketPath = b.daemon.SocketPath()
+		b.config.Log.Info("embedded containerd started", "socket", socketPath)
+	}
+
+	client, err := containerdclient.New(socketPath,
+		containerdclient.WithDefaultNamespace(Namespace))
+	if err != nil {
+		return Capability{}, fmt.Errorf("connecting to containerd: %w", err)
+	}
+	b.result = Capability{Client: client, Namespace: Namespace}
```
evanphx, line 114: 👏 moving this above the readiness check is what makes dropping the explicit `client.Close()` safe.

components/distributedrunner/startup_test.go
```diff
@@ -0,0 +1,113 @@
...
+		inputs,
+		boot.ResolvedOutput(containerdBootOutput{}),
+		boot.ResolvedOutput(telemetryBootOutput{tokenSource: tokenSource}),
+	)
+
+	graph := boot.NewGraph()
+	require.NoError(t, graph.Add(runnerBoot.component))
+
+	_, err := tokenSource.Token()
+	require.ErrorIs(t, err, runnertelemetry.ErrIssuerUnavailable)
```
evanphx, line 84: 👏 asserting the before state too.

evanphx (approved):
Approved! Nice refactor.

All the inline notes are non-blocking. The comment ones are where I'd start — a handful of the old `runner_start.go` explanations didn't make the move, and a couple of spots read as surprising without them.

--e+🤖

components/distributedrunner/boot_telemetry.go
phinze: 🤖 Restored both explanations alongside the moved code: why the joined service addresses are capability signals rather than dial targets, and why the runner batches this path without risking recursive error logging.

components/distributedrunner/boot_runner.go
phinze: 🤖 Restored the literal-IP rationale both at API address construction and in `resolveHost`’s doc comment.

components/distributedrunner/boot_runner.go
phinze: 🤖 Restored the TODO at `serviceNetworkPrefixes` so the hard-coded fallback stays visibly temporary.

components/distributedrunner/runtime.go
phinze: 🤖 Good catch. Removed the component-specific budgets for runner and telemetry because their `Close` APIs cannot honor contexts, and rewrote `ShutdownTimeout`’s comment to describe the remaining cooperative deadline honestly. Containerd keeps its 30-second budget because that stop path does accept the context.

components/containerd/boot.go
phinze: 🤖 Fixed. PATH lookup is now derived from whether `BinaryPath` is a bare name, so a direct `BootConfig` literal and the constructor behave consistently.

components/containerd/boot.go
phinze: 🤖 Yep, exactly. Keeping ownership assigned before readiness means graph rollback closes the client on that failure path.

components/distributedrunner/startup_test.go
phinze: 🤖 Agreed, that precondition is what makes the issuer-wiring test useful.

phinze:
Closing this in favor of #1161, which rebuilds the change on top of #1158’s runner capabilities. The three-node graph here still put the whole runner behind one lifecycle boundary. The replacement keeps the shared containerd and telemetry work, but graphs cluster access, storage, the sandbox host, both agents, and node presence separately.

PR #1153, opened 2026-09-03
Add miren top for cluster-wide resource usage

evanphx:
There was no way to ask a Miren cluster what was using the CPU. To diagnose a busy cluster you'd list sandboxes, then dial one runner at a time with `miren sandbox metrics`, which dumped a raw Go struct of lifetime counters for a single sandbox. Nothing aggregated across runners, nothing gave you a rate, and nothing told you whether the load was somebody's app or Miren's own overhead.

This adds `miren top`, which answers the three questions an operator asks in order. `--nodes` says which machine is hot and splits it into sandbox versus platform use, `--apps` totals each app across everything it owns, and the default view lists sandboxes ordered by the busiest. `miren sandbox inspect` covers the last step: what one sandbox runs, how it exited, and whether its pool has given up replacing it.

The nice surprise was that most of the data was already there. Every runner ships its metrics through the coordinator into one store, so a single query sees the whole cluster with no fan-out and no partial answer when a runner is unreachable. It also means the window can be historical, which no live probe could offer, so `--since 1h --aggregate max` finds the spike that caused the page an hour ago.

What was missing was attribution. The metrics could say which sandbox but not which host, and app sandboxes couldn't be told apart from the addon servers and task runs sharing the machine. Two of the five commits are that groundwork: labelling sandbox series with their node, kind and app, and publishing host-level series so there's a capacity figure to divide by. That second one is also what makes the system-versus-sandbox split possible, since containerd, buildkit, the registry and miren itself all run outside every sandbox cgroup and are invisible in the per-sandbox view.

The first commit is a detour worth calling out. The API wants two shapes: a Go caller wants grouped arguments it can't transpose, and a URL can only carry flat scalars. Offering both from one interface meant the URL-shaped method was also a second, worse way for a program to ask the same thing, with thirteen positional arguments including six consecutive strings. So `rest_only: true` is now a thing rpcgen understands. It keeps the HTTP route and withdraws the method from everything else, so reaching for it from Go is a compile error rather than a runtime one.

A couple of decisions that might read as omissions. There's no restart column, because Miren replaces a failed sandbox rather than restarting it, so no single sandbox accumulates a count; repeated failure is a pool-level crash loop, which `inspect` reports. And network, disk IO and PSI aren't here at all. They need new collection rather than new queries, so rather than ship fields that always read zero and can't be told apart from a genuinely idle sandbox, the API only carries what it measures. Adding them later is backward compatible.

Closes MIR-1758

### DIFF
--- a/cli/commands/top.go
+++ b/cli/commands/top.go
@@ -59,7 +59,7 @@ type topOptions struct {
 	Since     string `long:"since" description:"Measure over this window, e.g. 30s, 5m, 1h (default 1m)"`
 	Aggregate string `long:"aggregate" description:"How to collapse the window: avg, max, min, last" default:"avg"`
 
-	Sort  string `long:"sort" description:"Sort by: cpu, memory, app, service, node" default:"cpu"`
+	Sort  string `long:"sort" description:"Sort by: cpu, memory, name, app, service, node" default:"cpu"`
 	Order string `long:"order" description:"Sort direction: desc or asc (default: desc for usage, asc for names)"`
 	Limit int    `long:"limit" description:"Show at most this many rows (0 for all)"`
 
--- a/cli/commands/top.go
+++ b/cli/commands/top.go
@@ -60,7 +60,7 @@ type topOptions struct {
 	Aggregate string `long:"aggregate" description:"How to collapse the window: avg, max, min, last" default:"avg"`
 
 	Sort  string `long:"sort" description:"Sort by: cpu, memory, app, service, node" default:"cpu"`
-	Order string `long:"order" description:"Sort direction: desc or asc" default:"desc"`
+	Order string `long:"order" description:"Sort direction: desc or asc (default: desc for usage, asc for names)"`
 	Limit int    `long:"limit" description:"Show at most this many rows (0 for all)"`
 
 	Watch    bool   `short:"w" long:"watch" description:"Refresh continuously until interrupted"`
--- a/cli/commands/top.go
+++ b/cli/commands/top.go
@@ -1,0 +1,817 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+	"golang.org/x/term"
+
+	"miren.dev/runtime/api/usage/usage_v1alpha"
+	"miren.dev/runtime/pkg/rpc/standard"
+	"miren.dev/runtime/pkg/theme"
+	"miren.dev/runtime/pkg/ui"
+	"miren.dev/runtime/pkg/units"
+)
+
+// ServiceUsage is the cluster-wide resource usage service, served only by the
+// coordinator.
+const ServiceUsage = "dev.miren.runtime/usage"
+
+// mutedStyle renders the trailing summary and warning lines that sit under a
+// table, keeping them subordinate to the rows themselves. It matches the muted
+// label colour the other detail views use rather than borrowing doctor's
+// palette, which is scoped to that command's own output.
+var mutedStyle = lipgloss.NewStyle().Foreground(theme.Muted)
+
+// warnStyle marks a line reporting something the server could not answer.
+var warnStyle = lipgloss.NewStyle().Foreground(theme.Warning)
+
+// minWatchInterval floors how often a watch loop refreshes.
+//
+// Numbers reaching this command are already a few seconds old -- the collectors
+// sample once a second, the writer batches, and the query engine deliberately
+// lags to let late points land. Refreshing faster than that redraws the same
+// figures repeatedly while making a query per redraw, so the floor protects the
+// cluster from a watcher that gains nothing by asking harder.
+const minWatchInterval = 2 * time.Second
+
+type topOptions struct {
+	FormatOptions
+	ConfigCentric
+
+	App     string `long:"app" description:"Only show sandboxes belonging to this app"`
+	Service string `long:"service" description:"Only show sandboxes of this service (e.g. web, worker)"`
+	Runner  string `long:"runner" description:"Only show sandboxes on this runner (name, ID, or short ID)"`
+	Kind    string `long:"kind" description:"Only show sandboxes of this kind (app, addon, run)"`
+	Status  string `long:"status" description:"Only show sandboxes in this status"`
+
+	Nodes  bool `long:"nodes" description:"Show per-node usage instead of per-sandbox"`
+	Apps   bool `long:"apps" description:"Show per-app totals instead of per-sandbox"`
+	System bool `long:"system" description:"Include addon and platform sandboxes"`
+
+	NoAddons bool `long:"no-addons" description:"With --apps, exclude each app's dedicated addons from its total"`
+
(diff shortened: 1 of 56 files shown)


servers/usage/convert.go
```diff
@@ -0,0 +1,126 @@
...
+func (o ordering) descending() bool {
+	switch strings.ToLower(strings.TrimSpace(o.order)) {
+	case "asc":
+		return false
+	case "desc":
+		return true
+	}
+
+	switch strings.ToLower(strings.TrimSpace(o.sort)) {
+	case "name", "app", "service", "node", "runner":
```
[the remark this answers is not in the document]
evanphx: 🤖 This one was fixed in the previous push (bb8ef718 / d99765a7) — `sortSandboxes` now has a `name` case comparing the sandbox short id, falling back to the full entity id.

Rather than add the single missing case, each comparator now declares the direction that reads naturally for it, and `ordering.direction` asks instead of inferring from the key name. The same gap existed for keys the node and app listings do not implement, so a key and its default direction can no longer drift apart. `TestEverySortKeyHasAComparator` walks every documented key and fails if one reaches the fallback comparator.

servers/usage/directory.go
```diff
@@ -0,0 +1,366 @@
...
+
+		var md core_v1alpha.Metadata
+		md.Decode(ent)
+
+		var sch computev1.Schedule
+		sch.Decode(ent)
+
+		// A dead sandbox is not a row. Listing it would show a sandbox using no
+		// resources, which reads as idle rather than as gone.
+		if compute.SandboxDead(sb.Status) {
```
phinze: Normal exits become `STOPPED`, and `SandboxDead` includes both `STOPPED` and `DEAD`. Because `sandboxDetail` resolves through this same directory, a sandbox with a recorded `Exit` is filtered out before `decorateFromEntities` can return that failure history. The same filter means historical sandbox and app views lose a replaced sandbox even while VictoriaMetrics still retains its samples, so `--since 1h --aggregate max` cannot identify a culprit that is no longer live. The overall direction still looks right to me; use your judgment on whether to address this here or narrow the first-cut promise and follow up.

--p+🤖

servers/usage/convert.go
```diff
@@ -0,0 +1,123 @@
...
+// answering a typo'd duration with the default beats a 400.
+func restWindow(since, until, aggregate string) window {
+	end := time.Now()
+	if d, ok := parseRESTDuration(until); ok {
+		end = end.Add(-d)
+	}
+
+	start := end.Add(-defaultWindow)
+	if d, ok := parseRESTDuration(since); ok {
+		start = end.Add(-d)
```
phinze: `until` has already shifted `end` backward here, so subtracting `since` from `end` makes the two durations cumulative. For `?since=2h&until=1h`, this returns a two-hour window from three hours ago to one hour ago; the IDL documents the one-hour span from two hours ago to one hour ago. Calculating both offsets from the same captured `now` would keep the bounds independent. The CLI does not use this REST path, so use your judgment on fixing it here versus following up.

--p+🤖

cli/commands/top.go
```diff
@@ -0,0 +1,817 @@
...
+
+func (m *topModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		switch msg.String() {
+		case "q", "esc", "ctrl+c":
+			m.quitting = true
+			return m, tea.Quit
+		case "r":
+			return m, m.refresh()
```
phinze, line 619: Once a completed refresh has scheduled its next `tea.Tick`, pressing `r` starts an immediate refresh without invalidating that pending tick. The manual result schedules another tick, then the original tick fires and starts a second recurring refresh chain. Each later `r` can multiply the polling rate again, which is especially noticeable in the node view because one refresh issues several metrics queries. This does not change my overall verdict; use your judgment on fixing it here or following up.

--p+🤖

phinze (approved):
Broadly, this is a strong first cut. The collection/query split feels right: runners attach authoritative identity and publish into the shared metrics store, while the coordinator joins metrics to entity state and serves whole-cluster views without fanning out. I smoke-tested the sandbox, app, and node views in a two-node dev cluster with an app scheduled on the remote runner; all three returned live, non-stale data. The targeted package tests and full CI are green.

The inline comments record concrete bugs, but none changes my overall verdict. Use your judgment on what is worth fixing here versus following up.

**Non-blocking UX feedback:**

- The command name led me to expect an interactive view by default.
- The long flags for the display modes feel a little awkward. Maybe those would be better as subcommands, but I am also fine punting this until we figure out the broader CLI shape in MIR-1338.
- I would expect the display modes to be toggleable from the interactive view.

All minor stuff!

--p+🤖

servers/usage/directory.go
evanphx: 🤖 Good catch, and the inspect half is the worse of the two — `decorateFromEntities` exists to report `sb.Exit`, so the filter meant inspect could never reach the case its exit reporting was written for. Fixed: `filter` gained `includeDead`, and `sandboxDetail` sets it. Listings keep the old behaviour, since a dead row there still reads as idle rather than gone. Also dropped the now-stale "or it is not running" from the not-found error.

The historical half I'm narrowing rather than fixing. Including a replaced sandbox means reconciling the window against each sandbox's death time, and dead entities get collected eventually — so the honest fix is sourcing historical rows from the metric series instead of the entity store, which is a design change rather than a patch. The IDL already promises "one row per live sandbox"; I've extended the comment at the filter to name the tradeoff so the next reader doesn't have to rediscover it.

servers/usage/convert.go
evanphx: 🤖 Right, and the doc comment three lines above promises exactly the behaviour the code didn't have. Fixed: both offsets now come off one captured `now`, so `?since=2h&until=1h` is the hour it names. Added the same inverted-span fallback `resolveWindow` uses for the RPC surface, so `?since=1h&until=2h` answers with the default length rather than a backwards window. Covered in a new `convert_test.go` — neither window function had tests.

cli/commands/top.go
evanphx: 🤖 Confirmed, and it compounds exactly as you describe. Fixed with a generation counter: `refresh()` bumps it, both message types carry the generation they were issued under, and anything from a superseded chain is dropped. That also covers two `r` presses in quick succession — cancelling just the pending tick would still leave two results in flight, each scheduling its own. Tests in a new `cli/commands/top_test.go` walk the interleaving.

evanphx:
🤖 Heads up: force-pushed after a rebase onto main, which had landed the "owned runtime capabilities" startup refactor. Three conflicts needed real resolution rather than a textual merge:

- The usage service's `ExposeValue` lived in the block that refactor deleted, so it's now its own capability — `components/coordinate/resource_usage.go` plus `components/server/boot_resource_usage.go`, modeled on the secret store.
- `RunnerDeps` ended up with two `MetricsWriter` fields (main added one independently); dropped ours and kept the explanatory comment on main's.
- Docs links moved to the `./command/x.md` style; converted the generated rows.

Verified against a live dev cluster, not just the suite: `m top --nodes` reports real host metrics, which exercises both the node-usage publisher and the relocated service.

PR #1156, opened 2026-09-04
Make disk backup and restore work from anywhere

evanphx:
`miren disk backup` and `miren disk restore` were the last disk commands that made you SSH into the server. Everything else (create, list, attach, delete) already worked from your laptop. These two opened the disk image directly and stopped early with "data path /var/lib/miren not found" when it wasn't there.

That restriction quietly broke both runbooks RFD-75 promised. Disaster recovery means the original host is gone, so there is nothing to SSH into and the whole job is rebuilding on a new one. Right-sizing means restoring onto a differently-sized VM that doesn't exist yet. A restore that has to run on the target server can't help with either.

So this implements RFD-108. Backup and restore become thin orchestration over the cluster's existing authenticated RPC channel. Two things genuinely have to happen on the server, moving the image bytes and talking to miren.cloud with the cluster's key, and both stay there. There is deliberately one execution path whether you run it on the box or on your laptop, which is why the old refusal is gone rather than made conditional: with no local-versus-remote detection, there is no detection to guess wrong.

`disk undelete` and `disk list-deleted` had the same problem (worse, since undelete also needed root) and got the same treatment.

RFD-108 also settles that a multi-gigabyte transfer has to checkpoint and resume, so that's here too. It matters most over the Miren Anywhere relay, where a dropped cluster uplink ends every in-flight session and the cluster can take a minute to come back, making the failure structural rather than rare. The client picks the transfer id, since one the server handed back would be lost along with the call that failed. Each direction checkpoints where bytes actually land, and the retry loop refuses to retry a refusal, because "the disk is in use" means the same thing on all six attempts.

Two pre-existing bugs turned up along the way, both silent.

Restoring into a mounted disk never worked. `LoopAttach` hands the kernel an open file descriptor, so a loop device holds the image's inode rather than its path, and renaming a restored image over an attached one leaves the mounted filesystem reading the old one. That restore has always reported success and changed nothing. It now refuses, with no `--force` escape, since overloading `--force` would only let you ask for a silent no-op.

Related and worse, `FindAllLoopBackings` stripped the kernel's " (deleted)" marker off a loop's backing file, so a device holding an unlinked inode looked like a live attachment at that path. `ensureVolumeMount` would adopt it, find the old filesystem, decide no formatting was needed, and mount the data the operator had just replaced. A restore is exactly how you get there. I confirmed the kernel semantics against a real loop device, and they're worth knowing because they aren't what the name suggests: a rename is followed with no marker, and the marker appears only once the inode has no name left.

Smaller third one: the warning about backing up a live disk keyed off `disk.status == ATTACHED`, which nothing in the runtime ever sets. It had never fired once, and `was_attached` on every restore point was always false. All three now come from the same question asked of the kernel.

The old image-touching implementations aren't deleted. They move under `miren debug disk` as break-glass for a host whose RPC listener is down.

Not covered, deliberately: quiescing a mounted volume so an in-place restore can work rather than being refused. That needs an exported quiesce API on the volume controller and a way to suspend its reconcile loop, neither of which exists, and the refusal has to sit underneath it as the guard for when quiesce fails anyway. Tracked separately.

Closes MIR-1772


blackbox/disk_backup_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func waitDiskProvisioned(t *testing.T, m *harness.Miren, name string) {
+	t.Helper()
+	harness.Poll(t, "disk provisioned: "+name, 60*time.Second, 2*time.Second,
+		func() (bool, string) {
+			r := m.Run("debug", "disk", "list")
+			if !r.Success() {
+				return false, "debug disk list failed"
+			}
+			if r.OutputContains(name) && r.OutputContains("provisioned") {
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch on the diagnosis, we fixed it. We didn't take the suggested diff though: `debug disk list` prints `Name:` and `Status:` on separate lines of an entry, so matching both on one line would never match and the poll would just hang to timeout. Added a `diskIsProvisioned` helper that scans the named disk's entry and stops at the next one.

servers/disk/transfer.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	if offset != have {
+		return nil, refuse(
+			"this transfer is at %d bytes but the client offered to continue from %d — ask transferOffset where to resume, or use a new transfer id to start over",
+			have, offset,
+		)
+	}
+
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0600)
```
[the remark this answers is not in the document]
evanphx: 🤖 Agreed, fixed. We took the lock wider than the Stat+Open pair: it's held across the whole staging-and-stream and receive-and-install, since the window that actually matters is the one where another call could truncate a staged snapshot mid-copy. It's reference counted so the map doesn't grow one entry per backup.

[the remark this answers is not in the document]
evanphx: 🤖 Fixed, and you and miren-code-agent both landed on this one. The lock spans the complete upload or staging lifecycle as you suggested, not just the file opens.

cli/commands/debug_disk_undelete.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	eac := entityserver_v1alpha.NewEntityAccessClient(client)
+	ec := entityserver.NewClient(ctx.Log, eac)
+	resolver := diskresolve.New(eac, ec)
+
+	// Check if a disk with this name already exists
+	if _, err := resolver.FindDisk(context.Background(), meta.DiskName); err == nil {
+		return fmt.Errorf("a disk named %q already exists — rename or delete it before restoring", meta.DiskName)
+	}
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, and we pushed it one layer further. `snapshot.PrepareRestore` had the same shape, where any `FindDisk` error meant "create the disk", so a brief entity-store outage during a restore would mint a duplicate. Absence is now its own error type and only absence means create.

pkg/diskresolve/resolver.go
```diff
@@ -88,26 +90,18 @@ func (r *entityDiskResolver) FindVolume(ctx context.Context, diskID string) (*sn
 // controller ignores it while restore writes the image. The returned
 // RestoreTarget includes a Finalize callback that creates the disk_volume
 // entity and transitions the disk to PROVISIONED.
-func (r *entityDiskResolver) CreateDiskAndVolume(ctx context.Context, name string, sizeBytes int64, filesystem string, dataPath string) (*snapshot.RestoreTarget, error) {
-	sizeGb := sizeBytes / (1 << 30)
+func (r *Resolver) CreateDiskAndVolume(ctx context.Context, name string, sizeBytes int64, filesystem string, dataPath string) (*snapshot.RestoreTarget, error) {
+	// Round up. Truncating would hand back a disk that claims less capacity
+	// than the image it is about to be given, so a 1.5 GiB image would land on
+	// a disk reporting 1 GiB.
+	sizeGb := (sizeBytes + (1 << 30) - 1) / (1 << 30)
```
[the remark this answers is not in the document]
evanphx: 🤖 Right, and we introduced it: the round-up from the last round added `gib-1`, which wraps near the top of the range. Switched to quotient-plus-remainder so it can't, and the size is validated on the way in now, since it comes out of a snapshot header the caller handed us. Negative is rejected, and so is anything past the point where the controller's own conversion back to bytes would overflow.

servers/disk/undelete.go
```diff
@@ -0,0 +1,307 @@
...
+	meta := entry.Metadata
+
+	// A live disk already owning this name would end up with two entities
+	// answering to it, and every lookup by name is ambiguous from then on.
+	//
+	// Only an actual absence clears the way. A lookup that merely failed says
+	// nothing about whether the name is taken, and treating it as free is how
+	// a moment of entity-store trouble becomes the duplicate this check exists
+	// to prevent.
+	switch _, err := s.disks.FindDisk(ctx, name); {
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed for the case that can actually happen here: recoveries now serialize on the disk name. We went process-local rather than a conditional unique-name create, and left the reasoning in the code — disk volumes are pinned to the coordinator, so one process serves every recovery on a cluster. Your point holds the moment disks schedule anywhere else, which is what the comment says to do about it.

servers/disk/transfer.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			s.log.Warn("could not sweep abandoned transfers", "error", err)
+		}
+		return
+	}
+
+	cutoff := time.Now().Add(-transferTTL)
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".part") {
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, the sweep takes both halves of a staged backup now.

servers/disk/restore.go
```diff
@@ -0,0 +1,384 @@
...
+	if name == "" {
+		return refuse("disk name is required")
+	}
+
+	// An uploaded snapshot lives in a file keyed by transfer id, and it stays
+	// there until the image is installed. Hold the id for the whole handler, so
+	// an overlapping call cannot append into the same file or pull it out from
+	// under the install. A restore point needs none of this: it comes from the
+	// cloud and touches no transfer file.
+	if id := args.TransferId(); id != "" && args.RestorePoint() == "" {
```
phinze, line 50: Per-transfer staging and locking look good. I think `Restore` wants the same disk-name lock as undelete too. We tried two clients restoring the same snapshot to the same new name; one completed, while the other caught the disk halfway through creation, got `no disk volume found`, and fell into the transfer retry loop. Different transfer IDs keep the staging files separate, but they do not protect the disk entity or the shared `<image>.restore.tmp`. Could we hold `s.names` across the full restore handler?

--p+🤖

servers/disk/restore.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+	}
+
+	prog.Message("Restoring %s (%d bytes)", name, meta.SizeBytes)
+
+	if err := s.writeImage(target.ImagePath, src, total, meta, prog); err != nil {
+		return err
+	}
+
+	if target.Finalize != nil {
```
phinze, line 109: The restored payload came back byte-for-byte, which is great. The lifecycle around a fresh restore still has a race, though: each one in our dev smoke test left two `DV_READY` volumes for the same disk, both mounted. It looks like `Finalize` patches the disk to `PROVISIONED` before creating the restored `disk_volume`, giving `DiskController` a window to see no volume and self-heal by formatting a second blank one. The next `miren disk backup` then fails with `multiple disk volumes found`. Could fresh restore use the `PROVISIONING` / `DV_PENDING` handshake that undelete uses, so only one path owns this transition?

--p+🤖

docs/docs/addons.md
```diff
@@ -411,38 +415,42 @@ miren disk backup -n pg-pg-myapp-sCZDabc123-data
...
 
-<CliCommand context="server">
+<CliCommand context="client">
 ```miren
 miren disk restore -s myapp-db.miren.zst -n new-disk-name
 ```
 </CliCommand>
 
+:::warning[Stop the app before restoring]
+Restore refuses to write over a disk that is still attached. A disk in use is held open by the kernel, so a restore into it would leave the running database on the old data while reporting success — the command stops rather than let that happen. Scale the app to zero, or restore into a new disk name and switch over.
```
phinze: The live-image refusal and warning are good guardrails; they stopped us from getting a fake-success restore. One docs wrinkle: we could not make the “stop the app” path reach a restorable universal image. Disk-backed services use fixed concurrency, app config rejects `num_instances = 0`, and a manual pool scale to zero was reconciled back to one. Even during the brief zero-instance window, the universal volume stayed mounted and `Restore` refused it. Restoring a local snapshot under a new name is a good escape hatch, but `--from-cloud` needs the target disk’s existing cloud-volume ID and cannot use it. Could we call out that in-place universal restore is not supported today and document the cloud recovery path that does work? I am not asking this PR to solve quiescing, just to set the expectation.

--p+🤖

cli/commands/disk_restore.go
```diff
@@ -2,144 +2,206 @@ package commands
...
-	if !target.Created {
-		if _, err := os.Stat(target.ImagePath); err == nil {
-			if !opts.Force {
-				return fmt.Errorf("disk image already exists at %s — use --force to overwrite", target.ImagePath)
-			}
-			ctx.Warn("Overwriting existing disk image at %s", target.ImagePath)
-		} else if !os.IsNotExist(err) {
-			return fmt.Errorf("checking existing image at %s: %w", target.ImagePath, err)
+		res, rerr := dc.Restore(ctx, name, "",
+			stream.ServeReader(ctx, snapFile, stream.WithBulkBatching()),
```
phinze: The client-owned transfer ID plus server-side offset checkpoint is a nice design. There is one file-lifetime snag here: `ServeReader` closes an underlying `io.Closer` at EOF, so if the server consumes the upload and then returns an error, the next attempt seeks the already closed `snapFile`. We hit this after the server accepted all 8.1 MB and returned `no disk volume found`; attempts 2 through 6 all ended with `seek ... file already closed`. Could we leave the file owned by the outer defer, perhaps by passing a non-closing reader wrapper here, and add a retry test where the failure happens after the full upload?

--p+🤖

phinze (changes requested):
Really like the direction here. Moving backup and restore behind RPC makes this feel like a normal product feature instead of server surgery, and putting the direct data-path commands under `debug disk` gives the supported path a clean boundary. Client-owned transfer IDs and server-side staging also feel like the right shape for resumable transfers.

The basic flow was nice in dev. From `/tmp`, we backed up an 8 MiB payload, restored it under a new name, attached it to a fresh app, and got the same SHA-256 back. The progress was easy to follow, and the live-disk warning was both clear and honest.

I’m requesting changes on three restore correctness issues: fresh restores create a second blank mounted volume and break the next backup, post-upload retries reuse a snapshot `ServeReader` has closed, and overlapping restores to the same name are not serialized. The universal/cloud comment is about documenting what works today, not asking this PR to solve application quiescing. CodeRabbit’s remaining node-lookup cleanup finding also looks valid, but it is not driving my verdict.

--p+🤖

servers/disk/restore.go
evanphx: 🤖 Fixed, and thanks for pinning the mechanism exactly. `Finalize` now creates the `disk_volume` first and hands the disk over as `PROVISIONING`, so it is the same handshake undelete uses and one path owns the transition. `DiskController` ignores a `RESTORING` disk, which means there is no longer any window where the disk is visible without its volume.

`Cleanup` had to move with it. It used to remove the image before touching the disk, on the reasoning that no volume could exist yet. A partial `Finalize` can now leave one the controller is already mounting, and pulling an image out from under a loop device leaves it holding an unlinked inode instead of releasing it, so it marks the disk `DELETING` first and lets the teardown path unmount and detach.

Reproduced your smoke test on a dev cluster: one volume, and the follow-up `miren disk backup` succeeds. The server log shows `found existing disk_volume` then `disk provisioned via disk_volume entity`, with no format, so the self-heal never fires.

cli/commands/disk_restore.go
evanphx: 🤖 Good catch, and thanks for the trace. `serveReader.Recv` closes the reader at clean EOF when it is an `io.Closer`, so the first complete upload closed the snapshot and every retry then failed on the seek rather than on anything real. The file is wrapped in something exposing only `Read` now, so the stream has no `Close` to find and ownership stays with the outer defer.

Added the test you asked for, in two halves: that the wrapper survives a read to EOF and can still be re-seeked and re-sent, and that an unwrapped `*os.File` is the shape that would have been closed.

servers/disk/restore.go
evanphx: 🤖 Agreed, done. `s.names` is held across the whole restore handler now, alongside the transfer lock, since the two cover different things: the name covers the disk entity and the shared `<image>.restore.tmp`, the transfer id covers the uploaded snapshot.

Ran your scenario on a dev cluster. Two clients restoring the same snapshot to the same new name now give one completed restore and one clean "disk is in use" refusal, with no `no disk volume found` and no closed-file retries. One disk and one volume at the end.

docs/docs/addons.md
evanphx: 🤖 You are right that the advice was wrong rather than just incomplete: the docs told operators to scale the app to zero, which as you found is not a thing that can be done. Rewritten to say in-place restore of a live disk is not supported, with the reasons you hit (fixed concurrency, `num_instances = 0` rejected, a manual scale to zero reconciled back, and the volume mounted regardless), and to lead with restoring into a new disk and moving the app across, including the app.toml change.

Also called out that `--from-cloud` resolves the restore point through the target disk's own cloud volume, so it needs that disk to already exist and cannot target a new name. Which means the move-across path needs a local snapshot file, so the docs now say to keep one.

phinze (changes requested):
Great follow-up. The four original issues look resolved, and the new smoke test passed: fresh restore left one volume, the follow-up backup succeeded, and two same-name restores produced one success and one clean refusal.

One blocker remains before approval. `Cleanup` marks the disk `DELETING` and immediately removes its image, but controller teardown is asynchronous. After a partial `Finalize`, that can unlink an image while its loop device is still attached. CodeRabbit’s existing comment has the right request: wait for teardown to finish, or leave image removal to the controller.

Once that is fixed, I think we’re there.

--p+🤖

servers/disk/restore.go
evanphx: 🤖 You're right, and my note above oversold what the reorder bought. Marking the disk `DELETING` only *starts* the teardown; the controller unmounts, detaches and soft-deletes the volume some time later, so removing the image on the next line still raced it. That's the same stranded-inode case I said the reorder avoided.

Rather than wait for teardown, which would mean polling the entity store from an error path that is already handling a failure, `Finalize` now records that it committed the volume and cleanup removes the image only when it did not. That's the common case anyway, since most cleanups run before `Finalize`, and it is exactly the case where nothing ever opened the image. Once a volume exists the image is the controller's, which is the option you offered.

Two tests pin both halves: an image survives cleanup after a committed `Finalize`, and is reclaimed when no volume was created.

evanphx:
🤖 Addressing CodeRabbit's two remaining findings.

**`FindNodeId` failure stranded a `RESTORING` disk.** Fixed. `CreateDiskAndVolume` commits the disk before the node lookup, and a lookup failure returned an error with no `RestoreTarget`, so the caller had no `Cleanup` to call. The disk sat in `RESTORING` and the next restore of that name found it, skipped creating one, and failed looking for the volume it never got. It now transitions the disk to `DELETING` before returning, with a test that fails the node lookup and asserts the disk is on its way out.

**The restore admonition in `addons.md`.** Split per the one-concept rule: the warning carries only the refusal and why it happens, and the release limitations and the recommended new-disk path moved to prose alongside it.

pkg/diskresolve/resolver.go
```diff
@@ -0,0 +1,390 @@
...
+			// soft-deletes the volume directory some time later. Removing the
+			// image here would race it, and losing that race unlinks a file
+			// whose loop device is still attached, which leaves the kernel
+			// holding an inode nobody can reach instead of releasing it.
+			//
+			// Waiting for the teardown instead would mean polling the entity
+			// store from an error path that is already handling a failure.
+			// Letting the controller finish the job it already does is both
+			// simpler and the thing that cannot race.
+			if volumeCommitted {
```
phinze, line 184: 🤖 Tiny leak window, not blocking: if the PROVISIONING patch fails and the volume controller coalesces the create and the `ABSENT` flip into one reconcile, `deleteVolume` finds no local state, marks the volume `DELETED`, and this image is left under `volumes/<volId>/` with nothing to GC it. Leaked space beats a stranded inode, so this is the right side of the trade; just flagging it in case a `RemoveVolumeDir` in that "volume not found in state" branch is cheap later.

phinze (approved):
This does what we asked. `Finalize` now records that it committed the volume and `Cleanup` leaves the image to the controller's teardown in that case, which I traced through `handleDeletion` → `deleteVolume` → `softDeleteVolume` to confirm the image really does get moved into `deleted-volumes`. The `FindNodeId` unwind closes the stuck-`RESTORING` gap too. All three new tests pass locally alongside the existing 21 in `pkg/diskresolve`.

One non-blocking note inline about a narrow orphaned-image window in the handoff. Not worth holding this up over.

--p+🤖

PR #1157, opened 2026-09-04
Ship the lbd kernel module with miren

evanphx:
Accelerator mode has never worked on a real install. It needs the lbd kernel module and `lbdctl`, and miren shipped neither, so every disk fell back to loop devices unless someone had built lbd by hand. lbd ships a `dkms.conf`, but DKMS needs a C toolchain and DKMS itself installed on every node.

`miren disk accelerator install <node>` now handles it end to end. The module's C source is embedded in the miren binary, so there is nothing to fetch. Your cluster builds a small toolchain image from an embedded Dockerfile using the BuildKit and registry it already runs for your apps, the named node pulls that image from `cluster.local`, compiles the module against its own kernel, and loads it. Nothing is downloaded from us and there is no published artifact for us to version.

The work splits across two machines deliberately. The toolchain image is built once on the coordinator, where BuildKit and the registry already live. The module is built on the target node, because it has to be compiled against the kernel running there and loaded into it. That is also why the CLI names a node: it is a property of the machine, not the cluster.

Kernel headers come from the host when it has them, bind-mounted read-only at their real paths so the build tree's own absolute symlinks still resolve. When it has none, the builder installs them for itself, which works on Debian and Ubuntu because that is what the image is built from. A module only loads on the kernel it was built for, so the server notices at startup when a kernel upgrade has stranded one and rebuilds — only on nodes that opted in, and under a ten-minute cap, because accelerator mode is a speed-up and a wedged build must never hold up a node coming back from a reboot.

Things worth knowing:

- **`c8873ab2` changes behaviour on existing nodes.** Three places picked the disk mode by looking for `lbdctl` on `PATH`. That was a fair proxy while nothing installed `lbdctl`; it breaks now that we install it ourselves, because a node would find the binary, choose accelerator mode, and then fail at attach with no module loaded. All three now check that the module is loaded and its control device is present.
- **Artifact GC needed to learn about system artifacts.** It archives everything no `AppVersion` references, and the toolchain image belongs to no app, so it would have been collected within the hour and its blobs deleted underneath the nodes still pulling it. An artifact is named after the tag it was pushed under, so the tag carries a reserved prefix and GC retains what matches. A genuinely orphaned artifact is still collected.
- **Fedora and RHEL need `kernel-devel` installed by hand.** The builder can only fetch headers for itself on the Debian family, since that is what the image is built from. Other distributions get an error naming the package. Shipping a Dockerfile rather than an image makes a second builder cheap when we want one.
- **Secure Boot must be off**, because a self-built module is unsigned. Clang-built kernels are refused outright.
- This also carries a one-line fix to `hack/dev-distributed`, which could not bootstrap from a git worktree at all. It is unrelated to the feature but it is what made the verification below possible.

Verified on a coordinator and a separate runner: the toolchain image was built and pushed to the cluster registry, the runner pulled it over `cluster.local` with a coordinator-minted token, and compiled `lbd.ko` for its own kernel. Earlier, on a throwaway VM, the full install-and-load path was exercised against a real kernel including a filesystem round-trip on `/dev/lbd0` and a kernel-upgrade rebuild. The one seam not covered by either is `depmod` and `modprobe` running on a distributed runner, because the peer image has no `kmod`.

Closes MIR-840


pkg/lbdmod/lock.go
```diff
@@ -0,0 +1,67 @@
+package lbdmod
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
```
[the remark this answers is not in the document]
evanphx: 🤖 Checked this: `pkg/lbdmod` still builds clean for darwin (`GOOS=darwin go build ./pkg/lbdmod/...`), because `x/sys/unix` provides `Flock` and the `LOCK_*` constants there too. Adding `//go:build linux` to `lock.go` would mean adding a stub for `acquireBuildLock` to keep `build.go` compiling elsewhere, which is more machinery than the boundary is worth right now. The package is already Linux-only in practice by way of its callers — `cli/commands/disk_accelerator.go` is `//go:build linux` with a stub sibling. Happy to revisit if a non-Linux consumer ever appears.

pkg/lbdmod/ctrbuild/ctrbuild.go
```diff
@@ -0,0 +1,268 @@
...
+// -- `ctr images import` plus --image is the only way to run a builder on a
+// node with no reachable registry. It is safe here because the builder
+// reference is pinned to a tag that never moves.
+func (b *Builder) resolveImage(ctx context.Context, ref string) (containerd.Image, error) {
+	if img, err := b.cc.GetImage(ctx, ref); err == nil {
+		b.log.Info("using the lbd builder image already on this node", "image", ref)
+		return img, nil
+	}
+
+	b.log.Info("pulling the lbd builder image", "image", ref)
```
phinze, line 174: Could we take the self-bootstrapping shape one step further and have Miren build the builder image too?

The Dockerfile is really just Ubuntu, a handful of build packages, and the build script. It doesn't contain the lbd source. Downloading that baked image isn't very different from having BuildKit install those packages once and cache the layers on the coordinator. Embedding the Dockerfile alongside the source would remove the public registry dependency, the release job, and an artifact we'd otherwise have to keep publishing and versioning.

The rough shape I have in mind is: the coordinator uses its existing BuildKit to build a content-addressed builder image from the embedded Dockerfile and stores it in the cluster-local registry. Distributed runners already map `cluster.local` back to the coordinator and pull app images from there, so they can pull and cache the builder in their own containerd the same way. Each runner then launches a small node-local system task with its own `/lib/modules`, `/usr/src`, source, and output directories mounted in. The toolchain image is shared, but the actual module compilation still happens next to the kernel it targets.

I don't think the public `miren app run` path is quite the right abstraction because this must run on a specific node with privileged host mounts. But we can reuse the BuildKit, registry, and authenticated image-resolution machinery underneath it, then keep the custom task adapter focused on those host mounts, logging, and cleanup. That keeps the special code at the actual host boundary instead of also making us own another published artifact and a parallel image-distribution path.

Does that seem like a workable direction?

--p+🤖

phinze (changes requested):
This is coming along nicely! The source embedding and local install/rebuild flow look good to me.

I'm requesting changes provisionally because I think we can take the self-bootstrapping shape one step further and avoid maintaining a separately published builder image. I wrote out the rough shape in one comment. Mostly I want to see what you think before we commit to that artifact. Overall, though, this is looking good.

--p+🤖

pkg/lbdmod/ctrbuild/ctrbuild.go
evanphx: Oh I like it! That also let's us support other distros more easily in the future by shipping Dockerfiles rather than maintaining images. On it!

components/runner/nodeadmin.go
```diff
@@ -0,0 +1,59 @@
...
+		Log: s.log,
+		Builder: ctrbuild.New(s.deps.CC, s.log, ctrbuild.WithClusterRegistry(&ctrbuild.ClusterRegistry{
+			Resolver: s.deps.Resolver,
+			Issuer:   s.deps.WorkloadIssuer,
+		})),
+		Options: lbdmod.HostOptions(s.deps.DataPath),
+		Image:   req.Args().Image(),
+	}
+
+	status, err := installer.Install(ctx, req.Args().Force())
```
phinze, line 60: This endpoint turns its caller-supplied image into host code: it runs `/usr/local/bin/build-lbd` from the image, then installs the resulting `lbd.ko` and `lbdctl` as root and loads the module. The runner RPC state is constructed without an authenticator or authorizer, so it falls back to `NoOpAuthenticator`; its TLS listener accepts certless clients, and this handler never verifies that the peer is the coordinator. Since the registry resolver also falls back to public registries for non-`cluster.local` hosts, anyone who can reach the runner API can ask it to pull an arbitrary image and load arbitrary kernel code.

Could we close both sides of that boundary? I think the handler should require the verified coordinator certificate (the coordinator dials with its `miren-api` certificate), and reject any image outside the cluster-local `miren-system/lbd-builder` repository. The image restriction gives us defense in depth and keeps this RPC input aligned with its one intended use.

--p+🤖

cli/commands/disk_accelerator_other.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// DiskAcceleratorInstall is not supported on non-Linux platforms
+func DiskAcceleratorInstall(ctx *Context, opts struct {
+	ConfigCentric
+
+	Force bool   `short:"f" long:"force" description:"Rebuild even when the module is already current"`
+	Node  string `position:"0" usage:"Runner to install on (name, ID, or short ID)" required:"true"`
+}) error {
+	return fmt.Errorf("disk accelerator install is only available on Linux")
```
phinze: Now that install only makes an RPC and does not inspect or modify the local host, this stub prevents an operator from installing on a remote runner from the macOS client. Could we move `DiskAcceleratorInstall` into an untagged file and leave only status and uninstall in the platform-specific siblings?

--p+🤖

phinze (changes requested):
This is now the self-bootstrapping architecture I was hoping for. Embedding and content-hashing the Dockerfile, building it through the coordinator’s existing BuildKit path, caching it in the cluster registry, and leaving only the kernel-specific compile/load on the target runner all fit together cleanly.

I did find one important authorization hole at the new node-admin boundary, plus a smaller cross-platform CLI regression. I think those need one more pass before approval; details inline. Apart from those, the implementation and coverage are looking good.

--p+🤖

components/runner/nodeadmin.go
evanphx: 🤖 You're right, and thank you for catching it. I confirmed the whole chain: `components/runner/runner.go:359` builds the runner's RPC state with `rpc.WithSkipVerify` and no authenticator, so every caller arrives anonymous, and the handler happily pulled and ran whatever image it was handed before installing the result as root.

Closed both sides as you suggested. `requireCoordinator` now rejects anonymous callers, non-certificate methods, and certificate holders whose subject isn't the coordinator — a registered runner holds a valid cluster cert, so "presented a cert" was never enough on its own, and there's a test for that case specifically. `lbdmod.IsBuilderImage` pins the reference to the cluster-local `miren-system/lbd-builder` repository. The tag is deliberately not checked, since a coordinator on a newer miren carries a different content hash and asking a node to build with it is legitimate.

I also replaced the bare `"miren-api"` string with `rpc.CoordinatorCertSubject`, shared between the coordinator issuing the cert and the runner checking it, so the two cannot drift.

cli/commands/disk_accelerator_other.go
evanphx: 🤖 Good catch, fixed. `DiskAcceleratorInstall` moved to an untagged `cli/commands/disk_accelerator_install.go`, and the stub is gone from the non-Linux sibling. `status` and `uninstall` stay platform-split since they read `/proc` and load modules. Verified with a darwin build of `./cmd/miren`.

servers/runner/disk_accelerator.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		s.Log.Error("failed to build the lbd toolchain image", "error", err)
+		results.SetError(fmt.Sprintf("building the lbd toolchain image: %v", err))
+		return nil
+	}
+
+	s.Log.Info("installing the lbd kernel module on a runner",
+		"node", node.ID, "address", node.ApiAddress, "image", image)
+
+	cl, err := s.RPC.Connect(node.ApiAddress, string(rpc.ServiceNodeAdmin))
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed, added `defer cl.Close()`.

docs/docs/command/disk-accelerator.md
```diff
@@ -0,0 +1,69 @@
...
+that: your cluster builds the toolchain image with the BuildKit and registry it
+already runs, the named node pulls it from there, and the module is compiled and
+loaded on that node. Nothing is downloaded from us, and there is no published
+image to keep up to date.
+
+## Getting started
+
+```bash
+miren disk accelerator status          # can this host run it?
+miren disk accelerator install runner1  # build and load it there
```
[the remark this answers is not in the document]
evanphx: 🤖 Fixed at the source in `cli/commands/disk_accelerator_doc.go` and regenerated — that page is generated, so editing it directly would not have stuck.

pkg/lbdmod/ctrbuild/ctrbuild.go
evanphx: 🤖 This is done, in `5576f83a` and `391c058d`. The Dockerfile is embedded and content-hashed, the coordinator builds it through BuildKit into the cluster registry, and runners pull it over `cluster.local` with a coordinator-minted token, leaving only the kernel-specific compile and load on the node.

Verified on a coordinator and a separate runner: the image was built and pushed, runner1 pulled it and compiled `lbd.ko` for its own kernel. Artifact GC needed teaching about system artifacts along the way, or the toolchain image would have been collected within the hour with its blobs deleted underneath the nodes still pulling it.

docs/docs/command/disk-accelerator.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Requirements
+
+- The kernel headers for your running kernel. On Debian and Ubuntu the builder
+  fetches them itself if the host has none. Everywhere else you install them
+  first, and `status` names the package -- `kernel-devel-$(uname -r)`
+  on Fedora and RHEL.
+- Secure Boot disabled. A self-built module is unsigned, and firmware with Secure
+  Boot enforcing will refuse to load it.
+- A kernel built with GCC. Clang-built kernels are not supported.
```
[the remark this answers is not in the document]
evanphx: 🤖 Skipping this one — I think it reads the house rule backwards. `docs/CLAUDE.md` says not to stack admonitions: "More than two or three admonitions in a row and the 'this is important' signal collapses into wallpaper," and "when a section genuinely has many gotchas, one scannable summary (a roadblock checklist, a table) beats a column of boxes." Three titled boxes in a row is exactly the shape it warns against.

The one-concept-per-admonition rule applies to admonitions; this section has none. It is a plain bullet list here and a table on the narrative page, which is what that guidance points at. The docs lint enforces these rules and passes.

components/runner/nodeadmin.go
```diff
@@ -0,0 +1,103 @@
...
+// every caller arrives anonymous unless it brought one. That is tolerable for
+// the services it already exposes; it is not tolerable here, where the handler
+// pulls an image, runs it, and loads the result into the kernel as root.
+// Anyone who could reach the port would otherwise have root on the node.
+//
+// The coordinator dials with its API certificate, so its subject is what
+// separates it from everything else that can route to this port.
+func requireCoordinator(ctx context.Context) error {
+	identity := rpc.IdentityFromContext(ctx)
+	if identity == nil || identity.Method == rpc.AuthMethodAnonymous {
```
phinze, line 93: This check is the right shape, but nothing on the runner ever produces an identity it can accept, so it rejects the coordinator too and `miren disk accelerator install` fails on every call right now.

The runner's `rpc.State` is built at `components/runner/runner.go:448` (standalone) or through `clientconfig.Config.State` (distributed), and neither passes `WithAuthenticator`. `rpc.NewState` fills that gap with `NoOpAuthenticator` (`pkg/rpc/state.go:491`), which stamps every caller `AuthMethodAnonymous`, so the coordinator arrives anonymous and trips the first branch here. I confirmed it over the wire with a scratch test: a listener built the way the runner builds it, dialed by a state holding a CA-issued `miren-api` cert, gets `presented none` in both shapes. The unit tests miss it because they hand-build the `Identity` in the context.

The fix needs two things on the runner listener: `LocalOnlyAuthenticator` so a verified cert becomes an identity, and a cluster CA so the server actually requests and verifies the client cert. The distributed runner already has the CA from its client config; the standalone one has neither today, and with `WithSkipVerify` alone the listener never asks for a client cert at all (`ClientAuth` stays at its zero value, `pkg/rpc/state.go:612`). One thing to watch while doing it: that listener also serves `dev.miren.runtime/exec` and `sandbox.metrics`, and an authenticator that returns nil for certless callers turns into a 401 on their non-public methods, so it is worth checking who reaches those without a cert.

A test that drives the real listener the way `TestServerVerifiesClientCertChain` does in `pkg/rpc/mtls_test.go` would have caught this and would make the fix self-proving.

--p+🤖

phinze (changes requested):
This is one commit away. Everything from the last round landed the way we talked about, and the darwin move, the shared cert subject constant, and the cancellable build lock all look right. The image pin and the coordinator check are exactly the boundary I wanted.

The catch is that the coordinator check closes the door on the coordinator as well: the runner's listener never turns a client cert into an identity, so `install` currently fails on every call in both modes. Details and a reproduction inline. Once the runner listener can see the coordinator's cert, I'm ready to approve.

--p+🤖

PR #1158, opened 2026-09-04
Model coordinator and runner startup as boot capabilities

phinze:
`Coordinator` and `Runner` remain useful names for node roles. In our code, though, the corresponding types had become catch-all bags for “everything this process starts.” That hid the real readiness boundaries, so a lot of the more intelligent boot graph edges could not be drawn without opening those bags and laying their parts directly onto the graph.

This introduces a good chunk of new names. I tried to keep each one descriptive and to bundle together things that are either categorically similar or genuinely need to boot together. I know they are not perfect, but each boundary now represents a capability another part of the system can actually use.

The coordinator side now breaks down like this:

| Name | Purpose |
| --- | --- |
| `cluster-foundation` | Gets the cluster’s identity, RPC server, and entity data online. |
| `app-version-migration` | Finishes updating old app-version records before anything else changes them. |
| `secret-store` | Loads the cluster keyring and secret backends, then keeps the keys rotating. |
| `runner-endpoints` | Opens the few coordinator APIs a runner needs to join and recover. |
| `application-management` | Makes apps, addons, and runs available to operators and other components. |
| `workload-control` | Turns that stored intent into workloads that are placed, running, and reachable. |
| `entity-maintenance` | Repairs indexes and cleans up old records in the background. |
| `cloud-control` | Publishes workload identity and keeps the cloud informed about the cluster. |
| `admin-api` | Opens the admin API once the things it manages are actually available. |

The runner side becomes:

| Name | Purpose |
| --- | --- |
| `cluster-access` | Gets the runner connected to the coordinator and the cluster data it needs. |
| `node-storage` | Finds and restores the node’s disks before anything starts reconciling them. |
| `sandbox-host` | Finds surviving sandboxes, restores local networking, and opens exec. |
| `storage-agent` | Starts watching for disk changes and bringing local storage into line. |
| `sandbox-agent` | Starts watching sandbox and service state once placement can safely resume. |
| `node-presence` | Announces the node as ready only after its host and agents really are. |

Once the graph is up, `ControlPlane` and `Runner` are role-level views reconstituted from these pieces. They do not hide another startup sequence. `Ingress` intentionally remains outside `ControlPlane`; it is part of the data plane.

What this buys us:

- Consumers can start as soon as the capability they actually need is ready. Slow migrations and maintenance only gate their real dependents.
- Typed outputs carry usable values, while `DependsOn` expresses readiness-only ordering without inventing fake data.
- Restart safety is visible in the graph. Existing sandboxes are adopted before placement resumes, and node presence and ingress wait for runner recovery.
- Shutdown follows the same ownership boundaries in reverse, with foundation RPC kept alive while dependents drain.
- MIR-904 now has a clearer seam around ingress and traffic-preserving handoff, without requiring this PR to split the process.

Repository-wide compilation, focused tests, race tests, and vet pass.

Closes MIR-1689


components/coordinate/control_plane.go
```diff
@@ -0,0 +1,108 @@
...
+		return err
+	}
+	if err := c.workloads.Start(ctx); err != nil {
+		return err
+	}
+	if err := c.maintenance.Start(ctx); err != nil {
+		return err
+	}
+	return c.cloud.Start(ctx)
+}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in 598a2ba1. `ControlPlane.Start` now rolls back a partial start through `Stop` with a fresh bounded context, while preserving the original startup error.

components/runner/runner.go
```diff
@@ -333,98 +426,135 @@ func (r *Runner) ContainerdContainerForSandbox(ctx context.Context, id entity.Id
...
-		client, err = rs.Client("entities")
-		if err != nil {
-			return err
+		if err == nil {
+			client, err = rs.Client("entities")
 		}
 	}
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in 598a2ba1. The cleanup defer is installed immediately after RPC state creation and before `Connect`/`Client`, so every later startup failure closes the state and clears the retained pointer.

components/coordinate/cloud_control.go
```diff
@@ -0,0 +1,579 @@
...
+
+	// Collect resource usage metrics
+	resourceUsage := c.collectResourceUsage()
+
+	// Build status report
+	status := &cloudauth.StatusReport{
+		ClusterID:     c.CloudAuth.ClusterID,
+		State:         "active",
+		Version:       versionInfo.Version,
+		NodeCount:     1, // Static value for now
```
evanphx, line 330: Not critical but I'd love us to fix this number!

components/coordinate/coordinate.go
```diff
@@ -375,1714 +283,3 @@ func loadX509Cert(path string) (*x509.Certificate, error) {
 	}
 	return x509.ParseCertificate(block.Bytes)
 }
-
```
evanphx, line 378: Image

components/coordinate/workload_control.go
```diff
@@ -0,0 +1,312 @@
...
+
+	cm := controller.NewControllerManager()
+	// Add addon controller (reconciles addon associations for provisioning/deprovisioning)
+	addonController := addonctrl.NewController(c.Log, ec, eac, addonRegistry, addonFw.Storage)
+	if err := addonController.Init(ctx); err != nil {
+		c.Log.Error("failed to initialize addon controller", "error", err)
+		return err
+	}
+
+	addonReconciler := controller.NewReconcileController(
```
evanphx, line 108: I've always wondered if there was a way to cleanup this serial, interdependent, woven sequence. No need now, just remarking.

pkg/rpc/state.go
```diff
@@ -719,12 +724,14 @@ func (s *State) startListener(ctx context.Context, so *stateOptions) error {
...
-	// network servers to drain. Unary request contexts remain transport-owned,
-	// so in-flight calls can still finish during the graceful shutdown window.
+	// Stop server-owned background work, streaming RPCs, and outgoing calls
+	// before asking the network servers to drain. Unary request contexts from
+	// other peers remain transport-owned, so in-flight calls can still finish
+	// during the graceful shutdown window.
 	if s.cancel != nil {
 		s.cancel()
 	}
+	s.closeOutboundConnections()
```
evanphx, line 738: Curious what exposed this (it's fine to have just curious)

evanphx (approved):
Let's do it!

pkg/rpc/state.go
phinze: Yeah this could use better justification and a local test. It was necessary to fix a integration shutdown race exposed by the new boot/shutdown timing.

components/coordinate/cloud_control.go
phinze: 🤖 Yep, it is hard-coded to one even for multi-runner clusters. Filed [MIR-1778](https://linear.app/miren/issue/MIR-1778/stop-hard-coding-one-node-in-cluster-status-reports) to count the actual node entities and pin down the semantics.

components/coordinate/coordinate.go
phinze: 🤖 Extremely satisfying deletion. 🔥

components/coordinate/workload_control.go
phinze: 🤖 Agreed. Filed [MIR-1779](https://linear.app/miren/issue/MIR-1779/untangle-workloadcontrols-controller-wiring-sequence) to make those construction and wiring dependencies explicit without turning every controller into an outer boot node.

pkg/rpc/state.go
phinze: 🤖 Added the missing local regression test and expanded the ownership comment. The test leaves a state-owned WebTransport session between its upgrade and first control stream, which reproduces the HTTP/3 shutdown timeout if outbound connection closure is removed.

PR #1159, opened 2026-09-04
Expose entity sync diagnostics through miren debug cloud-sync

phinze:
When entity sync stalls, runtime currently leaves operators piecing together uplink logs and guessing which phase is stuck. This adds `miren debug cloud-sync`, backed by the local debug API, so operators can see connection and capability state, source preparation, epoch and schema, the committed cursor and watch revision, snapshot progress, and the last acknowledgment or error.

The diagnostics follow uplink reconnects and exporter transitions without changing the sync protocol. Snapshot-tail acknowledgments stay provisional until the full snapshot commits, so the command does not overstate what cloud has durably accepted. The focused runtime suite, race tests, package compilation, vet, and lint pass.

Part of MIR-1711

### DIFF
--- a/components/coordinate/control_plane.go
+++ b/components/coordinate/control_plane.go
@@ -3,6 +3,9 @@ package coordinate
 import (
 	"context"
 	"time"
+
+	"miren.dev/runtime/api/core/core_v1alpha"
+	"miren.dev/runtime/pkg/entitysync"
 )
 
 type ControlPlaneParts struct {
@@ -22,6 +25,15 @@ func NewControlPlane(foundation *Foundation, supplied ...ControlPlaneParts) *Con
 	if len(supplied) > 0 {
 		parts = supplied[0]
 	}
+	diagnostics := entitysync.NewDiagnostics(core_v1alpha.CloudExportContract.Digest())
+	if parts.Cloud != nil {
+		diagnostics = parts.Cloud.EntitySyncDiagnostics()
+	} else if parts.Applications != nil && parts.Applications.entitySyncDiagnostics != nil {
+		diagnostics = parts.Applications.entitySyncDiagnostics
+	}
+	if parts.Applications != nil && parts.Applications.entitySyncDiagnostics == nil {
+		parts.Applications.entitySyncDiagnostics = diagnostics
+	}
 	if parts.Secrets == nil {
 		parts.Secrets = NewSecretStore(foundation)
 	}
@@ -29,7 +41,7 @@ func NewControlPlane(foundation *Foundation, supplied ...ControlPlaneParts) *Con
 		parts.RunnerEndpoints = NewRunnerEndpoints(foundation)
 	}
 	if parts.Applications == nil {
-		parts.Applications = NewApplicationManagement(foundation, parts.Secrets)
+		parts.Applications = NewApplicationManagement(foundation, parts.Secrets, diagnostics)
 	}
 	if parts.Workloads == nil {
 		parts.Workloads = NewWorkloadControl(foundation, parts.Applications)
@@ -38,7 +50,7 @@ func NewControlPlane(foundation *Foundation, supplied ...ControlPlaneParts) *Con
 		parts.Maintenance = NewEntityMaintenance(foundation)
 	}
 	if parts.Cloud == nil {
-		parts.Cloud = NewCloudControl(foundation, parts.Applications)
+		parts.Cloud = NewCloudControl(foundation, parts.Applications, diagnostics)
 	}
 	return &ControlPlane{
 		Foundation:      foundation,
--- a/pkg/entitysync/diagnostics.go
+++ b/pkg/entitysync/diagnostics.go
@@ -1,0 +1,241 @@
+package entitysync
+
+import (
+	"maps"
+	"sync"
+	"time"
+
+	"miren.dev/runtime/pkg/uplink"
+)
+
+// Event records the most recent acknowledgement or failure.
+type Event struct {
+	At        time.Time
+	Stage     string
+	Message   string
+	MessageID string
+	Cursor    int64
+}
+
+// SnapshotProgress describes the snapshot currently being transmitted.
+type SnapshotProgress struct {
+	ID           string
+	HeadRevision int64
+	NextRevision int64
+	PagesSent    int64
+	EntitiesSent int64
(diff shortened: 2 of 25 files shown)


pkg/entitysync/exporter.go
```diff
@@ -583,6 +654,9 @@ func (s *stream) sendAndWait(link Link, messageType string, payload any, message
 	case <-waitCtx.Done():
 		return Ack{}, fmt.Errorf("await %s ack: %w", messageType, waitCtx.Err())
 	case ack := <-waiter:
+		if s.exporter.diagnostics != nil {
+			s.exporter.diagnostics.acknowledge(messageType, ack)
+		}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed. Rejected acknowledgments now return before diagnostics record success, and the regression test verifies `LastAcknowledgment` remains empty.

pkg/entitysync/exporter.go
```diff
@@ -90,9 +96,14 @@ func (t *Exporter) Register(_ context.Context, link Link) error {
 	link.OfferCapabilityFunc(uplink.CapabilityEntitySync, []uint{Version1}, func(ctx context.Context) (json.RawMessage, bool) {
 		sourceEpoch, err := t.readSourceEpoch(ctx)
 		if err != nil {
+			t.fail("read-source-epoch", err)
+			t.setMode("waiting", "source-epoch-unavailable")
```
evanphx: 🤖 This reason gets overwritten before anyone can read it. The offer callback runs inside `establishSession`, so it lands *before* the client reports `connected`; `ObserveUplink` then sees a session with the capability unselected and unconditionally sets `WaitReason` to `capability-not-selected`. The `fail()` on the line above survives as a `LastError` event, so the information isn't gone — but the summary line an operator reads first will say the vague thing.

Non-blocking. Either have `ObserveUplink` leave an existing reason alone when the capability is missing, or drop this `setMode` since the error event already carries it. (The identical `setMode` at line 150 is fine — that one runs after the connect.)

pkg/entitysync/exporter.go
```diff
@@ -497,6 +562,12 @@ func (s *stream) sendWatchResponse(link Link, response clientv3.WatchResponse, c
 	if ack.Cursor != to {
 		return cursor, fmt.Errorf("change ack cursor %d does not match batch end %d", ack.Cursor, to)
 	}
+	if s.exporter.diagnostics != nil {
+		s.exporter.diagnostics.setNextWatchRevision(to + 1)
+		if committed {
```
evanphx, line 566: 🤖 👏

Non-blocking but worth doing: nothing tests this. `testExporter` never attaches diagnostics, so every diagnostics call in the exporter is a no-op across the whole suite except the one rejection subtest that wires one up by hand.

`TestLiveChangePreemptsArchiveSnapshot` is the exact scenario — it already drives a `TypeChangeBatch` through `drainLiveChanges` mid-snapshot. Attach a `Diagnostics` and assert `CloudCursor` is still 0 in the `TypeSnapshotComplete` branch, and you've got a test that goes red if this flag ever flips.

pkg/entitysync/diagnostics.go
```diff
@@ -0,0 +1,241 @@
...
+	return status
+}
+
+func (d *Diagnostics) ObserveUplink(status uplink.Status) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.status.UplinkState = status.State
+	d.status.RetryAt = status.RetryAt
+	if status.LastError != "" {
+		d.status.LastError = &Event{At: time.Now().UTC(), Stage: "uplink", Message: status.LastError}
```
evanphx, line 95: 🤖 Every routine reconnect lands here — the uplink's `retrying` status carries `read: EOF` — and nothing clears `LastError` when it connects again. A cluster that bounced once last Tuesday still shows an error line today. The timestamp keeps it from being actively wrong, but it's the shape that teaches people to skip the field.

Non-blocking. Clearing it on `connected`, or stamping a last-success next to it, would make a populated Last Error mean "something is wrong right now."

components/coordinate/control_plane.go
```diff
@@ -22,14 +25,23 @@ func NewControlPlane(foundation *Foundation, supplied ...ControlPlaneParts) *Con
 	if len(supplied) > 0 {
 		parts = supplied[0]
 	}
+	diagnostics := entitysync.NewDiagnostics(core_v1alpha.CloudExportContract.Digest())
```
evanphx, line 28: 🤖 In production every part is supplied, so this allocates a `Diagnostics` and throws it away on every boot, and neither branch below ever fires. It's here because `NewApplicationManagement` and `NewCloudControl` take diagnostics as a variadic optional — which buys not touching seven test call sites, at the cost of constructors that will happily accept three of them.

Non-blocking, but a field on `ControlPlaneParts` plus a required parameter would say the same thing with none of the branches.

evanphx (approved):
Approved! Really nicely done.

Four inline notes, all non-blocking — the untested `committed` flag is the only one I'd bother with before merging.

`--e+🤖`

pkg/entitysync/exporter.go
phinze: 🤖 Confirmed. We removed the overwritten `setMode`; the specific source failure remains in `LastError`, while `capability-not-selected` describes the resulting negotiated state.

pkg/entitysync/exporter.go
phinze: 🤖 Good catch. We attached diagnostics in `TestLiveChangePreemptsArchiveSnapshot` and assert that `CloudCursor` is still zero when `TypeSnapshotComplete` is sent. The test now fails if catch-up advances the durable cursor early.

pkg/entitysync/diagnostics.go
phinze: 🤖 We’re keeping `LastError` historical for now. Its timestamp makes that useful, while clearing it on reconnect would also need rules for exporter and preparation failures.

components/coordinate/control_plane.go
phinze: 🤖 Yep, this is awkward. We’re keeping the compatibility plumbing in this PR to avoid widening the constructor churn; production still converges on the shared diagnostics instance.

PR #1161, opened 2026-09-04
Model distributed runner startup as boot capabilities

phinze:
#1158 exposes the runner’s startup phases as capabilities, but the standalone command would still hide them behind `Runner.Start`. That would make distributed startup look graph-managed while its lifecycle and readiness boundaries remained implicit.

This carries the capability model through distributed startup. Cluster access and containerd begin independently; storage and telemetry wait for cluster access; the sandbox host waits for their ready outputs; and node presence is published only after both agents are active. Telemetry now receives the coordinator-backed workload issuer before its writers start.

The shared storage-agent, sandbox-agent, and node-presence adapters live alongside the runner capabilities. Containerd’s boot adapter is shared by server and runner composition too. The CLI keeps command policy, signal handling, and the final background-task wait, while the graph owns startup rollback and reverse teardown.

Focused component and CLI tests pass, including containerd’s integration test, and repository lint is clean.

Closes MIR-1690


cli/commands/runner_start.go
```diff
@@ -166,203 +87,33 @@ func RunnerStart(ctx *Context, opts struct {
...
-		return fmt.Errorf("failed to start runner: %w", err)
+		return fmt.Errorf("failed to start distributed runner dependency graph: %w", err)
 	}
-
-	// Start acquired the coordinator-backed issuer, so telemetry can now mint.
-	if issuer := r.WorkloadIssuer(); issuer != nil {
-		tokenSource.SetIssuer(issuer)
-	} else if telemetryClient != nil {
-		ctx.Log.Error("no workload identity issuer available; telemetry cannot be shipped")
+	stop := func() {
```
evanphx: Why is this a function that is just called one place rather than being down below at the call site?

evanphx (approved):
Great cleanup!

cli/commands/runner_start.go
phinze: 🤖 Good call. We inlined the shutdown context and `runtime.Stop` at the only call site.

PR #1162, opened 2026-09-04
Make stale pool drains converge after node removal

phinze:
Disk-aware deploys could wedge when a historical, diskless sandbox remained STOPPED on a node that no longer existed. Each resync rewrote the already-decommissioned pool, waited for the same timeout, and blocked the app from reconciling.

Make launcher drain updates idempotent and treat STOPPED diskless sandboxes as unable to conflict with a disk-backed replacement, while preserving the DEAD barrier for sandboxes that mounted volumes. Nodehealth now completes STOPPED to DEAD after the existing five-minute grace period and periodically collects terminal sandboxes whose scheduled node is gone, so old tombstones eventually disappear instead of pinning pools forever.

Closes MIR-1780


evanphx (approved):
Those poor orphans!

PR #1163, opened 2026-09-04
Report runtime-owned app health at cloud-requested cadences

phinze:
Entity sync tells cloud which apps exist, but nothing currently fills their health samples. This adds runtime-owned health reporting over the uplink, using the same classification as `miren app list` and keeping health separate from durable app identity.

Cloud chooses a slow background cadence and temporarily accelerates sampling while someone is viewing a panel. Viewer demand expires on its own, and both intervals travel in the protocol so cloud can change them without a runtime release.

Verified with a real local runtime/cloud stack and Chromium: health and counts matched the CLI, sampling slowed after the panel closed, and reopening produced a fresh reading. Cloud counterpart: mirendev/cloud#237.

Part of MIR-1777

### DIFF
--- a/pkg/apphealthsync/reporter.go
+++ b/pkg/apphealthsync/reporter.go
@@ -1,0 +1,192 @@
+// Package apphealthsync samples runtime-owned health for cloud's ephemeral cache.
+package apphealthsync
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+
+	"miren.dev/runtime/pkg/apphealth"
+	"miren.dev/runtime/pkg/uplink"
+)
+
+const (
+	defaultBackground = 5 * time.Minute
+	minInterval       = time.Second
+	maxInterval       = time.Hour
+	maxLease          = time.Minute
+	maxBatch          = 100
+)
+
+// Source shares the classifier used by the app list.
+type Source interface {
+	ListAppHealth(context.Context) ([]apphealth.State, error)
+}
+
+type Link interface {
+	OfferCapability(uplink.CapabilityOffer)
+	OnSession(func(context.Context, uplink.Session))
+	Handle(string, uplink.MessageHandler)
+	SendMessageBlocking(context.Context, string, any) error
+}
+
+type samplingSession struct {
+	id     string
+	ctx    context.Context
+	demand chan Demand
+}
+
+type Reporter struct {
+	log     *slog.Logger
+	source  Source
+	mu      sync.Mutex
+	session *samplingSession
+}
+
+func NewReporter(log *slog.Logger, source Source) *Reporter {
+	return &Reporter{log: log, source: source}
+}
+
+func (r *Reporter) Register(_ context.Context, link Link) error {
+	link.OfferCapability(uplink.CapabilityOffer{Name: uplink.CapabilityAppHealth, Versions: []uint{Version1}})
+	link.Handle(TypeDemand, r.handleDemand)
+	link.OnSession(func(ctx context.Context, session uplink.Session) {
+		selection, ok := session.Capability(uplink.CapabilityAppHealth)
+		if !ok {
+			return
+		}
+		background := defaultBackground
+		var config Config
+		if len(selection.Config) > 0 {
+			if err := json.Unmarshal(selection.Config, &config); err != nil {
+				r.log.Warn("invalid app health configuration", "error", err)
+				return
+			}
+			if config.BackgroundSeconds > 0 {
+				seconds := min(config.BackgroundSeconds, int(maxInterval/time.Second))
+				background = max(minInterval, time.Duration(seconds)*time.Second)
+			}
+		}
+		active := &samplingSession{id: session.ID, ctx: ctx, demand: make(chan Demand, 1)}
+		r.mu.Lock()
+		r.session = active
+		r.mu.Unlock()
+		go r.run(active, link, background)
+	})
+	return nil
+}
+
+func (r *Reporter) handleDemand(_ context.Context, raw json.RawMessage) error {
+	var demand Demand
+	if err := json.Unmarshal(raw, &demand); err != nil {
+		return err
+	}
+	if demand.IntervalSeconds < 1 || demand.IntervalSeconds > int(maxInterval/time.Second) || demand.LeaseSeconds < 1 || demand.LeaseSeconds > int(maxLease/time.Second) {
+		return fmt.Errorf("invalid app health sampling demand")
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	active := r.session
+	if active == nil || active.ctx.Err() != nil || active.id != demand.SessionID {
+		return nil
+	}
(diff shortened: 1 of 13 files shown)


phinze:
@coderabbitai review

pkg/apphealthsync/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case demand := <-active.demand:
+			now := time.Now()
+			wasActive := now.Before(until)
+			interval = min(background, time.Duration(demand.IntervalSeconds)*time.Second)
+			until = now.Add(time.Duration(demand.LeaseSeconds) * time.Second)
+			if !wasActive {
```
evanphx, line 124: 🤖 Non-blocking, but nothing pins this branch. I changed `wasActive := now.Before(until)` to a constant `true` — collapsing fresh arrival and renewal into one case — and the whole suite still passed. The existing tests only reach here with `last` at the zero time, where both branches happen to agree.

This one catches it:

```go
func TestReturningViewerGetsFreshSample(t *testing.T) {
	synctest.Test(t, func(t *testing.T) {
		_, _, link, cancel := startSampler(t)
		defer cancel()
		demand(t, link, "s1", 5, 6) // sample now and at +5s, lease ends at +6s
		time.Sleep(7 * time.Second)
		synctest.Wait()
		before := link.sendCount()
		demand(t, link, "s1", 5, 20)
		require.Equal(t, before+1, link.sendCount(), "a returning viewer gets a fresh reading")
	})
}
```

Red under the mutation, green clean. Say the word and I'll push it.

pkg/apphealthsync/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+	active := r.session
+	if active == nil || active.ctx.Err() != nil || active.id != demand.SessionID {
+		return nil
+	}
+	// Keep the latest request without blocking the shared uplink read loop.
+	select {
+	case <-active.demand:
+	default:
+	}
+	active.demand <- demand
```
evanphx, line 101: 🤖 👏 easy to get wrong

servers/app/app.go
```diff
@@ -187,17 +195,14 @@ func (r *AppInfo) ListApps(ctx context.Context) ([]*app_v1alpha.AppInfo, error)
 				entry.activeVersion = &appVer
 				entry.activeVersionEntity = verEnt
 				if resolvedCfg, err := coreutil.ResolveRuntimeConfig(ctx, r.EC.EAC(), &appVer); err == nil {
```
evanphx: 🤖 `ResolveRuntimeConfig` does the config lookup plus a per-app list of addon associations and a variable merge, but health only reads `spec.Services` (names, concurrency mode) and `spec.Tasks`. The addon step only touches variables and `Services[i].Disks`, so `ResolveConfig` gives identical health for one fewer entity list per app.

Didn't matter when this only ran on a CLI call; now it's on a timer that can reach once a second. Non-blocking, and it moves `ListApps` too, so worth a look before switching.

pkg/apphealthsync/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		background := defaultBackground
+		var config Config
+		if len(selection.Config) > 0 {
+			if err := json.Unmarshal(selection.Config, &config); err != nil {
+				r.log.Warn("invalid app health configuration", "error", err)
+				return
+			}
+			if config.BackgroundSeconds > 0 {
+				background = time.Duration(min(config.BackgroundSeconds, 3600)) * time.Second
```
evanphx: 🤖 `background_seconds` has a ceiling but no floor, and unlike a demand it never expires — cloud saying `1` here pins a cluster at a full health derivation every second for the life of the session. Cloud is ours so this isn't a trust question, just a lot of blast radius on one config value. `minInterval` is already sitting there if we want a floor.

Small thing while you're in here: that `3600` is `maxInterval` in disguise, and the demand check a few lines down uses the constant.

pkg/apphealthsync/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		timer.Reset(time.Until(wake))
+	}
+}
+
+func (r *Reporter) sample(ctx context.Context, link Link) {
+	observed := time.Now().UTC()
+	apps, err := r.source.ListAppHealth(ctx)
+	if err != nil {
+		r.log.Debug("failed to sample app health", "error", err)
```
evanphx: 🤖 Debug is right for the send below — that fails on any normal disconnect and shouldn't be loud. But a failing `ListAppHealth` is a genuinely degraded runtime, and nothing above Debug will ever say so, which makes a broken health feed look identical to a healthy one. Worth splitting the two, even if the fast-mode rate argues against a plain Warn.

evanphx (approved):
Approved! Nicely done.

Four inline notes, all non-blocking. The missing test on the returning-viewer path is the only one I'd act on before merge — I have it written and verified if you want it pushed.

--e+🤖

pkg/apphealthsync/reporter_test.go
```diff
@@ -0,0 +1,315 @@
...
+		require.Equal(t, before+1, link.sendCount(), "a returning viewer gets a fresh reading")
+	})
+}
+
+func TestDerivationLogsOnlyFailureAndRecoveryTransitions(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		reporter, source, link, cancel := startSampler(t)
+		defer cancel()
+		var logs bytes.Buffer
+		reporter.log = slog.New(slog.NewTextHandler(&logs, nil))
```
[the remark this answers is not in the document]
phinze: 🤖 This test runs inside a synctest bubble. startSampler calls synctest.Wait before returning, leaving the sampler durably blocked. The logger assignment then precedes the demand-channel send, which orders it before run resumes sampling and reads the logger. Buffer reads also follow synctest.Wait. The test passes under the race detector; no additional lock is needed here.

pkg/apphealthsync/reporter.go
phinze: 🤖 Added your test and verified it passes clean and fails with wasActive forced to true. Thanks for writing it.

servers/app/app.go
phinze: 🤖 Switched the shared health path to ResolveConfig. Health only reads service concurrency and task presence, so addon variables and disks do not affect it. The app and reporter tests pass under race.

pkg/apphealthsync/reporter.go
phinze: 🤖 Replaced the literal ceiling with maxInterval and made the minInterval clamp explicit. The effective floor remains one second: background_seconds is a positive integer here and minInterval is already one second.

pkg/apphealthsync/reporter.go
phinze: 🤖 Derivation failures now warn once per failure streak and log recovery once. Send failures remain at Debug, and session cancellation does not emit a warning. Added coverage for repeated failures, recovery, and send failures.

PR #1164, opened 2026-09-04
Coalesce controller work and reconcile current state

phinze:
ReconcileController currently queues full event snapshots in a bounded channel. A busy controller can therefore replay stale state, drop work, or wait for another event after a transient failure.

This replaces event replay with a keyed dirty queue: one entry per entity, a fresh read immediately before reconciliation, live work ahead of repair scans, and bounded jittered retries. Index watches resume from their revision cursor and take full current snapshots after compaction without retaining full index membership in every controller. Concurrency stays local to each controller instead of sharing a process-wide gate, and per-controller metrics make queue pressure and write volume visible.

The watcher shutdown audit also found a transport race: canceling a WebTransport dial before its per-call dialer initialized could dereference a nil internal cancel function or leak a connection returned after cleanup began. The RPC path now waits for that initialization boundary and closes late connections deterministically.

The caller audit moves parent bridges to ID-only signals and tightens the few places where automatic retry changes policy, including certificate pacing, default-route repair, and service deletion without a tombstone.

Closes MIR-1391


phinze:
@coderabbitai review

phinze:
@coderabbitai review

phinze:
@coderabbitai review

pkg/controller/controller.go
```diff
@@ -289,17 +193,68 @@ func (c *ReconcileController) WriteTracker() WriteTracker {
...
+		}
+		for id := range c.known {
+			if _, ok := next[id]; !ok {
+				c.enqueueSignal(workSignal{id: id, priority: workRepair, present: false})
+			}
+		}
+		c.known = next
+
+	case indexwatch.EventAdded, indexwatch.EventUpdated:
+		c.known[event.Id] = struct{}{}
```
evanphx: Do we need to have a map that tracks effectively EVERY live entity? We track this so we can detect deletions during a sync, but this is quite expensive.

pkg/controller/workqueue.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+type workQueueStats struct {
+	depth     int
+	oldestAge time.Duration
+}
+
+// dirtyQueue is a keyed, two-lane work queue. An entity has one entry across
+// queued, in-flight, and retrying states. Urgent watch/manual work always runs
+// before anti-entropy and retry work.
```
evanphx, line 72: I really like this concept!

evanphx (approved):
Great cleanup! Mostly worried about the in-memory tracking of entities ids. Perhaps not a big deal, but it's an unbounded map so it's worth keeping an eye on.

pkg/controller/controller.go
phinze: 🤖 Yep, good catch. The map was bounded by current index cardinality rather than historical churn, but that still meant every controller retained every ID in its index, often duplicating broad indexes. We removed it. `EventSync` now queues only current entities; live delete events still use the delete path. This gives up synthesizing delete callbacks after compaction, but the map could not cover process restarts anyway, so controllers requiring durable cleanup still need level-based sweeps.

pkg/controller/workqueue.go
phinze: 🤖 Thanks! The keyed dirty state ended up being the useful center of the design.

phinze:
@coderabbitai review

phinze:
@coderabbitai review

PR #1166, opened 2026-09-07
Stop the ReplaceEntity index leak and drain the backlog

phinze:
A saga that had finished kept showing up in the list of sagas that hadn't. Every restart, recovery would pick up tens of thousands of long-dead teardown executions and try to resume them, and retention would refuse to delete anything whose parent looked similarly unfinished.

The entity store's index had been lying. `ReplaceEntity` is meant to drop index entries for values the replacement no longer carries, but it compared the new attributes against themselves, so it never dropped any: every entity that changed an indexed attribute kept the old value indexed forever. Sagas got hit hardest because they're saved on every status change, so each one ended up listed under all five statuses it had ever been in.

The first rev fixes that, and teaches the two etcd-backed saga storages to check the status they actually got back. That second part isn't just belt and braces. Listing an index is two separate reads, ids first and then the entities, so what the index said can be out of date by the time the entities arrive. The entity is the real answer and the index is only a pointer to it. `MemoryStorage` has no index and filters the executions directly, which is exactly why it was the one backend that never had this bug.

The second rev cleans up what's already sitting in etcd. The background sweep used to ask only whether an entry pointed at an entity that still exists. Now it asks whether that entity still actually carries the value the entry is indexing. Orphans fall out of the new question for free, since an entity that's gone carries nothing at all. Answering it means reading the entity's attributes instead of just checking it exists, and doing that a page at a time is what let us drop the full entity-ID snapshot the sweep used to build on every pass. That's MIR-1400, which this closes too.

Closes MIR-1735
Closes MIR-1400

### DIFF
--- a/controllers/indexgc/gc.go
+++ b/controllers/indexgc/gc.go
@@ -1,11 +1,13 @@
 // Package indexgc runs a bounded, best-effort background sweep that removes
-// stale entity-store index (collection) entries whose backing entity is gone.
+// entity-store index (collection) entries the backing entity no longer
+// justifies: the entity is gone, or it still exists but no longer carries the
+// value the entry indexes.
 //
-// The source of these orphans was fixed upstream (atomic deletes, lease-bound
-// session index entries), so this exists to drain any legacy backlog on upgrade
-// and to keep a cluster self-healing if a future leak ever reappears, without
-// anyone having to run `miren debug reindex` by hand. It deliberately does its
-// work off the read path: foreground reads stay pure and never issue deletes.
+// It drains the backlog those leaks left and keeps a cluster self-healing
+// without anyone running `miren debug reindex` by hand. Mismatched entries in
+// particular cannot drain any other way, since no write path removes an indexed
+// value the entity has already stopped carrying. It deliberately works off the
+// read path: foreground reads stay pure and never issue deletes.
 package indexgc
 
 import (
@@ -16,35 +18,57 @@ import (
 	"miren.dev/runtime/pkg/entity"
 )
 
-const (
-	// initialDelay keeps the first sweep clear of the boot storm and any
-	// startup-time additive reindex before it begins deleting.
-	initialDelay = 1 * time.Minute
-)
+// defaultInitialDelay keeps the first sweep clear of the boot storm and any
+// startup-time additive reindex before it begins deleting.
+const defaultInitialDelay = 1 * time.Minute
 
 // GCConfig tunes the background sweep. Defaults are intentionally gentle: this
 // is convergence insurance, not active firefighting, so it drains slowly and
 // stays out of the way.
 type GCConfig struct {
-	// CheckInterval is how often to run a sweep.
+	// CheckInterval is how often to run a sweep. A sweep resolves the entities
+	// behind every entry it scans whether or not it deletes anything, so a
+	// drained store pays a full read of itself on every tick. That cost, not
+	// the delete rate, is what sets this.
 	CheckInterval time.Duration
-	// MaxDeletesPerSweep caps deletions per sweep so a large legacy backlog
-	// drains over several sweeps rather than one thundering pass. Zero means
-	// unbounded.
+	// MaxDeletesPerSweep caps deletions per sweep so a large backlog drains over
+	// several sweeps rather than one thundering pass. Zero means unbounded.
+	//
+	// Deletes are not the sweep's dominant cost, since every sweep resolves the
+	// entities behind the entries it scans regardless, so throttling them hard
+	// mostly stretches the drain.
 	MaxDeletesPerSweep int
 	// BatchPause is slept periodically during deletion to rate-limit write
 	// pressure. Zero disables pacing.
 	BatchPause time.Duration
-	// SweepTimeout bounds a single sweep. A truncated sweep is fine: the next
-	// one resumes the work, since the pass is idempotent.
+	// SweepTimeout bounds a single sweep. A truncated sweep is mostly fine: the
+	// pass is idempotent and the next one starts over, so finished work sticks.
+	//
+	// The caveat is that it restarts from the head of the keyspace, so a store
+	// too large for one sweep to scan would never reach its tail. A truncated
+	// sweep logs at Warn, which is the signal this needs a resume cursor.
 	SweepTimeout time.Duration
+	// InitialDelay is how long to wait before the first sweep. Zero means
+	// defaultInitialDelay; it is configurable so a test can drive the schedule.
+	InitialDelay time.Duration
 }
 
-// DefaultGCConfig returns the default (gentle) configuration.
+// initialDelay resolves the configured first-sweep delay.
+func (c GCConfig) initialDelay() time.Duration {
+	if c.InitialDelay > 0 {
+		return c.InitialDelay
+	}
+	return defaultInitialDelay
+}
+
+// DefaultGCConfig returns the default (gentle) configuration. At six hours and
+// 5000 deletes a sweep, a backlog in the low hundreds of thousands drains in
+// about a week, which is the right pace for entries that are bloat rather than
+// a correctness problem.
 func DefaultGCConfig() GCConfig {
 	return GCConfig{
-		CheckInterval:      1 * time.Hour,
-		MaxDeletesPerSweep: 500,
+		CheckInterval:      6 * time.Hour,
+		MaxDeletesPerSweep: 5000,
 		BatchPause:         1 * time.Second,
 		SweepTimeout:       10 * time.Minute,
 	}
@@ -60,6 +84,9 @@ type GCController struct {
 	Config GCConfig
 
 	cancel context.CancelFunc
+	// stopped closes when the sweep loop has returned, so Stop can be observed
+	// rather than merely requested.
+	stopped chan struct{}
 }
 
 // Start begins the periodic sweep.
@@ -69,14 +96,22 @@ func (c *GCController) Start(ctx context.Context) {
 		"max_deletes_per_sweep", c.Config.MaxDeletesPerSweep)
 
 	ctx, c.cancel = context.WithCancel(ctx)
-	go c.run(ctx)
+	c.stopped = make(chan struct{})
+	go func() {
+		defer close(c.stopped)
+		c.run(ctx)
+	}()
 }
 
-// Stop gracefully stops the controller.
+// Stop cancels the sweep loop and waits for it to return, so a stopped
+// controller is known to have stopped issuing deletes. Safe on a controller that
+// was never started, and safe to call more than once.
 func (c *GCController) Stop() {
-	if c.cancel != nil {
-		c.cancel()
+	if c.cancel == nil {
+		return
 	}
+	c.cancel()
+	<-c.stopped
 }
 
 func (c *GCController) run(ctx context.Context) {
@@ -85,7 +120,7 @@ func (c *GCController) run(ctx context.Context) {
 	// keeps the first interval a true CheckInterval and avoids a back-to-back
 	// double-fire if initialDelay is ever tuned longer than CheckInterval.
 	select {
-	case <-time.After(initialDelay):
+	case <-time.After(c.Config.initialDelay()):
 		c.sweep(ctx)
 	case <-ctx.Done():
 		c.Log.Info("stale index GC controller stopped")
@@ -134,6 +169,8 @@ func (c *GCController) sweep(ctx context.Context) {
 		c.Log.Info("stale index GC sweep complete",
 			"scanned", stats.CollectionEntriesScanned,
 			"found", stats.StaleEntriesFound,
+			"orphaned", stats.OrphanedEntriesFound,
+			"mismatched", stats.MismatchedEntriesFound,
 			"removed", stats.StaleEntriesRemoved,
 			"cas_conflicts", stats.CASConflicts,
 			"by_collection", stats.RemovedByCollection)
--- a/pkg/entity/cleanup.go
+++ b/pkg/entity/cleanup.go
@@ -10,7 +10,6 @@ import (
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
-	"miren.dev/runtime/pkg/cond"
 )
 
 // cleanupDeleteBatchSize is how many stale entries we delete between rate-limit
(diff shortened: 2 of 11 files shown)


pkg/entity/cleanup.go
```diff
@@ -41,133 +46,295 @@ type CleanupStats struct {
...
+		return nil, nil, fmt.Errorf("cleanup: failed to resolve entities: %w", err)
+	}
+
+	justified = make(map[Id]map[string]bool, len(ids))
+	unverifiable = make(map[Id]bool)
+	for i, id := range ids {
+		ent := entities[i]
+		if ent == nil {
+			continue
 		}
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, this was a real regression and it is fixed in the latest push.

The old sweep survived it because a nil from the batched read only made an entry a candidate; it then did a confirming linearizable `GetEntity` and treated anything other than a clean not-found as "could not verify, leave it alone". Dropping that confirming read is what made nil load-bearing, and `decoder.Unmarshal` failures were sharing it with genuinely absent keys.

`getEntities` now returns the ids whose key was present but whose payload would not decode, and `resolveJustified` folds those into `unverifiable`, so their entries are skipped rather than judged. The decode failure also logs now instead of being swallowed. `TestCleanup_LeavesUndecodableEntityAlone` corrupts a live entity's payload in place and asserts the sweep finds nothing stale and leaves the index entry alone; it fails without the fix.

pkg/entity/cleanup.go
```diff
@@ -41,133 +46,295 @@ type CleanupStats struct {
...
+// is stale, queues the delete and enforces the pass's delete budget. It returns
+// errStopScan once the budget is spent.
+func (p *cleanupPass) judge(ctx context.Context, e collectionEntry, justified map[Id]map[string]bool) error {
+	collection := collectionFromKey(e.key, p.collectionPrefix)
+	values, present := justified[e.entityID]
+
+	switch {
+	case present && values[collection]:
+		return nil
+	case present && e.sessionScoped:
```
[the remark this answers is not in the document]
phinze: 🤖 Fair, the comment overclaimed. The behaviour is what you describe and is intended: a session-scoped entry whose entity is gone is an orphan and gets deleted like any other, which is also what the sweep did before this change. Only the live-entity case is deferred to the lease.

Reworded in the latest push to say so explicitly rather than leaving the ordering to carry the meaning.

controllers/indexgc/gc.go
```diff
@@ -69,14 +90,22 @@ func (c *GCController) Start(ctx context.Context) {
...
-	go c.run(ctx)
+	c.stopped = make(chan struct{})
+	go func() {
+		defer close(c.stopped)
+		c.run(ctx)
+	}()
 }
 
-// Stop gracefully stops the controller.
+// Stop cancels the sweep loop and waits for it to return, so a stopped
```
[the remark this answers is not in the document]
phinze: 🤖 Leaving this as is. A second `Stop` calls `cancel` again, which is documented as safe, and reads an already-closed channel, which returns immediately; an unstarted controller returns early on the nil `cancel`. So it holds today for reasons the language guarantees rather than by accident.

Nilling the fields out would introduce the failure mode it is meant to prevent, since a `Stop` racing another `Stop` could then read a nil `stopped` and block forever. If the controller ever gains restart, that is the change that has to revisit this, and it will have to think about the channel regardless.

phinze:
🤖 Answering biscuit's transaction-math question from the review body, since there is no thread on it.

The arithmetic is right: `etcdMaxTxnOps` is 128 and `GetEntities` issues two ops per entity, so `maxEntitiesPerBatch` is 64 and a 200-entry resolve page always fans out into four transactions. That is intentional. The page size is chosen against how much entity payload we are willing to hold at once, not against the transaction limit, and for a background sweep on an hourly timer four round trips per page is not a cost worth tuning. Aligning it to a multiple of 64 would tidy the arithmetic without changing anything that matters.

pkg/entity/store.go
```diff
@@ -539,13 +558,19 @@ func (s *EtcdStore) GetEntities(ctx context.Context, ids []Id) ([]*Entity, error
...
 				continue
 			}
 
 			var entity Entity
 			err = decoder.Unmarshal(primaryResp.Kvs[0].Value, &entity)
 			if err != nil {
+				// The key is there, so this entity exists; we just cannot read
+				// it. Always worth a line, whatever warnMissing says.
+				s.log.Warn("failed to decode entity from etcd", "id", batchIds[i], "error", err)
+				undecodable[batchIds[i]] = true
```
evanphx, line 573: 🤖 Non-blocking, and latent rather than live: the session-attribute block right below this still swallows a decode failure with a bare `continue`. Those attributes then just aren't on the entity, and the sweep reads "attribute absent" as "entry unjustified" — the same load-bearing-nil shape this line just fixed for the primary payload, with the same consequence of deleting a live entity's index entries.

I checked whether it can fire today and it can't: `compute/node/status` and `actor/actor/node` are the only `session: true` attributes and neither is indexed. It goes live the day someone adds `index: true` to a session attribute, which is a long way from where anyone would think to look.

controllers/indexgc/gc.go
```diff
@@ -16,35 +18,51 @@ import (
...
+	}
+	return defaultInitialDelay
 }
 
 // DefaultGCConfig returns the default (gentle) configuration.
 func DefaultGCConfig() GCConfig {
 	return GCConfig{
 		CheckInterval:      1 * time.Hour,
-		MaxDeletesPerSweep: 500,
+		MaxDeletesPerSweep: 5000,
```
evanphx, line 71: 🤖 Non-blocking, more of a "worth a moment" thing. Once a store is drained the delete budget never trips, so every hour the sweep pulls every entity's full payload out of etcd — and saga entities carry the action-output blobs the `terminalFetchBatch` comment warns about. The old shape only paid that for suspected orphans.

Might be worth backing the interval off after a sweep that finds nothing, so a healthy cluster isn't re-reading itself hourly forever. Ties into the `SweepTimeout` caveat too: the clusters most likely to run out of sweep are the ones carrying the backlog this is here to drain.

pkg/entity/cleanup.go
```diff
@@ -41,133 +46,299 @@ type CleanupStats struct {
...
+// The pass is idempotent and bounded, so it is safe to run repeatedly as a
+// background sweep.
 func (s *EtcdStore) CleanupStaleCollectionEntries(ctx context.Context, log *slog.Logger, opts CleanupOptions) (*CleanupStats, error) {
-	stats := &CleanupStats{RemovedByCollection: map[string]int64{}}
+	pass := &cleanupPass{
+		store:            s,
+		log:              log,
+		opts:             opts,
+		stats:            &CleanupStats{RemovedByCollection: map[string]int64{}},
+		collectionPrefix: s.prefix + "/collections/",
```
evanphx, line 86: 🤖 Tiny one: `ListAllEntityIDs` has no callers left anywhere now, not even the `Store` interface. Worth deleting so nobody reaches for the O(everything) thing this change just got rid of.

pkg/entity/cleanup.go
```diff
@@ -41,133 +46,299 @@ type CleanupStats struct {
...
-		}
-		return stats, fmt.Errorf("cleanup: scan failed: %w", scanErr)
+// justifiedCollections returns the collection segments the entity's current
+// indexed attributes entitle it to, hashed the way the write path hashes them.
+//
+// Strict on purpose: it fails if any attribute's schema cannot be read, and the
+// caller then leaves that entity's entries alone. Reindex's tolerant variant
+// would drop the unreadable attribute, which here reads as "not justified" and
+// deletes a live entry.
+func (s *EtcdStore) justifiedCollections(ctx context.Context, ent *Entity) (map[string]bool, error) {
```
evanphx, line 315: 🤖 ✨ subtle and very easy to get backwards.

pkg/entity/cleanup_test.go
```diff
@@ -57,6 +60,168 @@ func TestCleanup_RemovesStaleKeepsLive(t *testing.T) {
...
+
+	ids, err = store.ListIndex(ctx, indexed)
+	require.NoError(t, err)
+	assert.Contains(t, ids, live.Id(), "the index entry must survive")
+}
+
+// TestCleanup_LeavesAConsistentStoreAlone is the safety net for the mismatch
+// verdict: a bug in the justified set shows up as deletions on a store with
+// nothing wrong with it, system and schema entities included.
+func TestCleanup_LeavesAConsistentStoreAlone(t *testing.T) {
```
evanphx, line 228: 🤖 👏 this is the one that makes the rest safe to ship.

evanphx (approved):
🤖 Approved! Really nicely done.

Three inline notes, all non-blocking; the `getEntities` one has the most teeth, though it's latent rather than live today. Happy to file tickets for the sweep-cost note and for the resume cursor the `SweepTimeout` comment already calls for.

pkg/entity/store.go
phinze: 🤖 Fixed in the latest push. You are right that it is the same shape, and I had walked past it.

The session block now breaks out on a decode failure, logs, and marks the entity undecodable so `resolveJustified` routes it to `unverifiable` alongside a bad primary payload. A partially decoded entity never reaches a caller, which matters more than the specific consequence here: anything judging an entity by what it carries would draw the wrong conclusion from a partial decode, not just the sweep.

It is testable today even though it cannot fire today, so `TestCleanup_LeavesEntityWithUnreadableSessionAttrAlone` declares a `Session: true, Index: true` attribute, corrupts the session blob under a live entity, and asserts the index entry survives. Verified it fails without the fix. That pins the behaviour before someone indexes a session attribute rather than after.

pkg/entity/cleanup.go
phinze: 🤖 Gone. And it turned out to be holding up more than itself: with `ListAllEntityIDs` deleted, `(*EtcdStore).scanPaged` was left with no callers either, since that was the last thing using the materializing variant of the scan. Removed too. The package-level `scanPaged` stays, `shortid_migrate.go` still uses it.

controllers/indexgc/gc.go
phinze: 🤖 Agreed, and this is a genuinely new cost rather than one MIR-1400 already signed up for. That issue reasoned about an existence check, which could have stayed a cheap key-only probe. The mismatch verdict needs the attributes, so the full payload read is now unavoidable per entry per sweep, and I should have flagged that the trade it analysed no longer applies.

Not fixing it here, since backing the interval off after an empty sweep is a scheduling change and it interacts with the resume-cursor question the `SweepTimeout` comment already raises. The clusters that would run out of sweep are the ones carrying the backlog, exactly as you say, so the two want deciding together rather than one landing on its own.

Taking you up on the ticket offer for both, unless you would rather I write them; I have the context loaded either way. My rough shape: hold the interval while a sweep is finding things, back it off on consecutive empty sweeps up to some ceiling, and reset on the first non-empty. That gives a draining cluster the hourly cadence and a healthy one something much quieter.

phinze: 🤖 Filed both, and took a partial mitigation here after talking it through with Paul.

MIR-1790 owns the steady-state read cost, MIR-1791 the resume cursor, related to each other because any interval backoff has to treat a deadline-truncated sweep differently from a clean one. Both carry your framing that the clusters most likely to run out of sweep are the ones carrying the backlog.

In this PR: `CheckInterval` default goes 1 hour to 6. The hourly cadence was picked when a sweep was a keys-only snapshot plus a few point reads, and that stopped being true here, so the constant was simply wrong now rather than needing new logic. Cuts the steady-state cost six-fold; the drain slows from roughly a day to roughly a week, which is fine given these entries are bloat rather than a correctness problem. The adaptive version belongs in MIR-1790 with measurements behind it, since nobody has sized the real read volume yet.

One thing I added to MIR-1790 that is worth knowing when someone does size it: entries are ordered by collection key, so an entity's several index entries land in different collections and therefore different resolve pages. The per-page dedup in `distinctEntityIDs` saves very little across a full sweep, and reads scale with entries rather than entities.

pkg/entity/store.go
```diff
@@ -539,25 +558,39 @@ func (s *EtcdStore) GetEntities(ctx context.Context, ids []Id) ([]*Entity, error
...
+				}
 				continue
 			}
 
 			var entity Entity
 			err = decoder.Unmarshal(primaryResp.Kvs[0].Value, &entity)
 			if err != nil {
+				// The key is there, so this entity exists; we just cannot read
+				// it. Always worth a line, whatever warnMissing says.
+				s.log.Warn("failed to decode entity from etcd", "id", batchIds[i], "error", err)
```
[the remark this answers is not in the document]
phinze: 🤖 Taken, both now log at Error. A record the store wrote and cannot read back is the platform failing at something it owns, which is the Error case in our log-level guidance; cleanup handling it gracefully does not make the underlying corruption less our problem.

pkg/entity/store.go
```diff
@@ -1074,8 +1112,10 @@ func (s *EtcdStore) ReplaceEntity(
 		return nil, err
 	}
 
-	// Keep track of original indexed attributes for removal
-	originalIndexedAttrs, err := s.collectIndexedAttributes(ctx, repl.attrs)
+	// Must come from the stored entity, not the replacement: buildCollectionOps
+	// deletes the original indexed values the replacement no longer carries, and
+	// diffing repl against itself emits no deletes at all.
+	originalIndexedAttrs, err := s.collectIndexedAttributes(ctx, originalEntity.attrs)
```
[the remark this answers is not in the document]
phinze: 🤖 Confirmed and filed as MIR-1793 rather than fixed here, for two reasons.

The arithmetic is right, and the link to this PR is tighter than the finding states. Before this change `originalIndexedAttrs` came from the replacement rather than the stored entity, so `buildCollectionOps` diffed the new attributes against themselves and emitted no deletes at all — that was the bug. A replace therefore cost N+1 ops and only broke at 128 changed values. Correcting it means deletes are emitted, so it now costs 2N+1 and breaks at 64. The ceiling is pre-existing but this halves the headroom, and that belongs in the issue.

It is not fixable the way the suggestion implies, though. Chunking the transaction is exactly what we cannot do: writing the entity and its collection entries atomically is the invariant the stale-index sweep depends on, since an entry whose entity is absent is only provably orphaned because creates are atomic. Splitting the write would undermine the change this PR is making. The real options are raising `--max-txn-ops`, validating indexed-value count on write, or making index maintenance non-atomic with the entity write, and picking between those is a design decision rather than a fix to land alongside a bug fix.

Severity as filed: it needs a single entity carrying 64+ indexed values, nothing is known to do that today, and it fails loudly with `too many operations in txn request` rather than writing partially. Boundary test noted on the issue.

PR #1167, opened 2026-09-07
Page saga recovery and retention so a backlog cannot exhaust startup memory

phinze:
Upgrading a 4 GB host stalled in sandbox saga recovery and never brought application ingress up. Recovery asks "which sagas do I need to resume?", which ANDs a status predicate with an ownership one, and the entity store only answers equality on a single indexed attribute at a time. So it drove off status, got the whole cluster's incomplete set, and loaded every payload to find the handful it owned. On that host the answer was three executions and the set was around 95,000, each carrying JSON blobs of every action's output.

The first rev is really about the entity store. Listing an index is two reads, the ids and then the entities those ids name, and every caller in the tree paired them by hand and unpinned. `ListIndexEntitiesPage` makes that one operation, pinned to a single revision per page, and the three walks that already open-coded it collapse onto it and get shorter. `list_page` is the same operation over RPC, since `EntityAccess.list` materializes the whole index server-side too, which is how one restarting runner took the coordinator down with it. `list` keeps its hundred-plus callers for now; migrating them is MIR-1302, and the new method's doc says which way the migration runs.

The second rev pages recovery and retention on top of that, and picks up a second unbounded read on the way: retention was listing every in-flight execution in the cluster just to check whether a finished child's parent was still running.

Worth being explicit about what this doesn't do. It makes the walk bounded, not cheap. Recovery still visits the whole incomplete set, 200 payloads at a time rather than all at once. Filtering by ownership before loading payloads needs either an index-membership primitive or an attribute that answers ownership directly, and on a system that isn't pathological the incomplete set is a handful, so that's its own design rather than a fifth rev here.

#1166 has landed, which fixes the stale index entries that made this set look considerably bigger than it was. The two are the same fact seen from either end: the index naming things the store will not return. That one makes it stop being true; this one has to stay correct while it still is, which is why the paging carries its own defense against a page that resolves to nothing rather than waiting on the repair.

Part of MIR-1785

https://claude.ai/code/session_015bxhbDRnsP7uM37K2eQyNg

### DIFF
--- a/pkg/saga/eac_storage.go
+++ b/pkg/saga/eac_storage.go
@@ -69,42 +69,53 @@ func (s *EACStorage) Get(ctx context.Context, id string) (*Execution, error) {
 	return entityToExecution(sagaEntity)
 }
 
-// ListTerminal summarizes every execution in a terminal state via EAC.
-func (s *EACStorage) ListTerminal(ctx context.Context) ([]TerminalExecution, error) {
+// ListTerminalPage summarizes one bounded page of finished executions via EAC.
+func (s *EACStorage) ListTerminalPage(ctx context.Context, q TerminalQuery) (*TerminalPage, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	stage, inner, err := decodeStageCursor(q.Cursor, len(terminalStatuses))
+	if err != nil {
+		return nil, err
+	}
+
+	status := terminalStatuses[stage]
+
+	resp, err := s.eac.ListPage(ctx, entity.Ref(saga_v1alpha.SagaStatusId, status), inner, int64(clampLimit(q.Limit)))
+	if err != nil {
+		return nil, fmt.Errorf("listing sagas with status %s: %w", status, err)
+	}
+
+	next := nextStageCursor(stage, resp.Cursor(), len(terminalStatuses))
+
+	// An empty page is the shortest possible short page, not the end of the
+	// walk. The server drops ids it could not resolve into entities, so a page
+	// of nothing but stale index entries comes back with no values and a live
+	// cursor into the rest of this index. Treating that as an exhausted index
+	// would skip everything after it.
+	if len(resp.Values()) == 0 {
+		return &TerminalPage{Cursor: next}, nil
+	}
+
 	var result []TerminalExecution
 	var staleNonTerminal int
-	seen := make(map[string]struct{})
 
-	for _, status := range []entity.Id{
-		saga_v1alpha.SagaStatusCompletedId,
-		saga_v1alpha.SagaStatusFailedId,
-	} {
-		resp, err := s.eac.List(ctx, entity.Ref(saga_v1alpha.SagaStatusId, status))
-		if err != nil {
-			return nil, fmt.Errorf("listing sagas with status %s: %w", status, err)
-		}
-		for _, v := range resp.Values() {
-			id := v.Id()
-			if _, dup := seen[id]; dup {
-				continue
-			}
-			seen[id] = struct{}{}
-
-			summary, verdict := terminalSummary(v.Entity())
-			switch verdict {
-			case summaryOK:
-				result = append(result, summary)
-			case summaryNotTerminal:
-				staleNonTerminal++
-			case summaryNoTimestamp:
-				s.log.Warn("terminal saga has no usable timestamp, skipping", "id", id)
-			}
+	for _, v := range resp.Values() {
+		summary, verdict := terminalSummary(v.Entity())
+		switch verdict {
+		case summaryOK:
+			result = append(result, summary)
+		case summaryNotTerminal:
+			staleNonTerminal++
+		case summaryNoTimestamp:
+			s.log.Warn("terminal saga has no usable timestamp, skipping", "id", v.Id())
 		}
 	}
 
 	logStaleTerminal(s.log, staleNonTerminal)
 
-	return result, nil
+	return &TerminalPage{Executions: result, Cursor: next}, nil
 }
 
 // Delete removes a saga execution entity via EAC.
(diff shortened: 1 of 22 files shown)


servers/entityserver/entityserver.go
```diff
@@ -692,6 +692,35 @@ func (e *EntityServer) List(ctx context.Context, req *entityserver_v1alpha.Entit
 	return nil
 }
 
+// ListPage reads a bounded page of entities from an index.
+func (e *EntityServer) ListPage(ctx context.Context, req *entityserver_v1alpha.EntityAccessListPage) error {
+	args := req.Args()
+
+	page, err := e.entityPage(ctx, args.Index(), args.HasIndex(), args.Cursor(), args.Limit())
```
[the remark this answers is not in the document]
phinze: Good catch, and a real regression rather than a theoretical one. `Index()` is `return *v.data.Index`, and Go evaluates arguments before the call, so the `hasIndex` parameter I was threading into `entityPage` could never have fired. Both handlers would have panicked on a request without an index, where the pre-refactor code returned the error correctly.

Fixed by checking `HasIndex()` at both call sites before touching `Index()`, and dropping the parameter entirely so the trap cannot be reintroduced. `entityPage` now carries a note saying why the guard has to live at the call site.

pkg/saga/eac_storage.go
```diff
@@ -118,63 +133,68 @@ func (s *EACStorage) Delete(ctx context.Context, id string) error {
...
-			s.log.Warn("failed to convert saga entity, skipping", "id", eacEnt.Id(), "error", err)
-			continue
+			return nil, fmt.Errorf("listing sagas with status %s: %w", status, err)
 		}
-		if isTerminal(exec.Status) {
-			staleTerminal++
+
+		next := nextStageCursor(stage, resp.Cursor(), len(incompleteStatuses))
+
+		if len(resp.Values()) == 0 {
```
evanphx: This treats "the server sent zero entities" as "this index is exhausted" and falls through to the next stage, throwing `next` away. Over the wire those aren't the same thing: the server's `resolve` (`servers/entityserver/entityserver.go:1069`) drops any id it couldn't turn into an entity, so a page made entirely of stale index entries comes back with no values and a live cursor into the rest of the index.

Reproduced against the conformance harness — two stale pending entries sorting ahead of three real pending sagas, read two at a time:

```
--- FAIL: TestPaging/EACStorage
    expected: {"zzz-real-00", "zzz-real-01", "zzz-real-02"}
    actual  : {}
```

`EntityStorage` passes the same case, since there an empty page really does imply an empty cursor. Runners are on `EACStorage`, and a page of nothing but stale entries is the shape #1166 exists to repair.

Both branches collapse into one:

```go
if len(resp.Values()) == 0 {
    return &IncompletePage{Cursor: next}, nil
}
```

An empty page with a cursor is just the shortest possible short page, which every doc in this PR already tells callers to handle. Worth a conformance case too — the suite runs all three backends and would have caught this.

pkg/saga/eac_storage.go
```diff
@@ -69,42 +69,57 @@ func (s *EACStorage) Get(ctx context.Context, id string) (*Execution, error) {
...
 			return nil, fmt.Errorf("listing sagas with status %s: %w", status, err)
 		}
-		for _, v := range resp.Values() {
-			id := v.Id()
-			if _, dup := seen[id]; dup {
-				continue
+
+		next := nextStageCursor(stage, resp.Cursor(), len(terminalStatuses))
+
+		if len(resp.Values()) == 0 {
```
evanphx: Same as the incomplete walk below: a page the server filtered to nothing ends this status index early and skips the rest of it, so retention stops seeing executions it should collect.

pkg/saga/retention.go
```diff
@@ -108,52 +123,95 @@ func RunRetention(ctx context.Context, storage Storage, cfg RetentionConfig, log
...
+//
+// Spending the budget is not the same as being cut short. A sweep that deletes
+// its last permitted execution and has nothing left to look at did the whole
+// job, and reporting that as capped would have an operator chasing a backlog
+// that is not there.
+//
+// Deciding takes a look ahead, because the cursor alone cannot answer it. The
+// walk covers several status indexes in sequence, so a cursor can point at the
+// head of a next index that turns out to be empty. Backends never return an
+// empty page together with a cursor, so one page of lookahead settles it, and
```
evanphx: Non-blocking, but this isn't quite true of `EntityStorage` — a terminal page whose entities all get dropped (deleted between the two reads, non-terminal status, no usable timestamp) returns empty with a live cursor. Costs only a `Capped: false` on a sweep that really did stop early, but the lookahead is resting on the claim.

pkg/saga/paging.go
```diff
@@ -0,0 +1,106 @@
...
+	default:
+		return limit
+	}
+}
+
+// incompleteStatuses are the status indexes a recovery walk covers, in the
+// order it covers them. Pending first because an execution that crashed between
+// its initial save and its first transition has done the least and is the
+// cheapest to resume.
+var incompleteStatuses = []entity.Id{
```
evanphx, line 54: Non-blocking: worth writing down the other reason this order matters. Sagas only move forward through pending → running → undoing, so walking the indexes in that same order is what keeps an unpinned walk from dropping an execution that transitions partway through. The stated reason is cost, and someone reordering on cost grounds later wouldn't know they'd broken anything.

servers/entityserver/entityserver.go
```diff
@@ -1024,11 +1125,14 @@ func decodeCursor(cursor string) (string, error) {
 //
 // Session membership is not an attribute index and the store cannot page it, so
 // that case lists and slices. Everything else pages in the store.
-func (e *EntityServer) pageIds(ctx context.Context, index entity.Attr, cursor string, limit int64) ([]entity.Id, string, int64, error) {
+// pageIds reads one page of ids from an index and reports the store revision it
```
evanphx: Two doc comments stacked here — the old one is still sitting above. Non-blocking, but while you're in it: `pageIds` has one caller left (the session branch of `entityPage`) and it discards the revision with `_`, so the tail below is unreachable and the new return value has no reader. Might be cleaner to inline the session case and drop the function.

pkg/saga/executor.go
```diff
@@ -721,6 +791,9 @@ func (e *Executor) Recover(ctx context.Context) error {
 				"saga", exec.DefinitionName, "execution", exec.ID)
 			continue
 		}
+
+		attempted[exec.ID] = struct{}{}
```
evanphx, line 795: 👏 placed exactly right.

evanphx (changes requested):
One blocker: over the wire, a page that comes back all-stale ends the walk of that status index early, so recovery can skip sagas it should have resumed. Reproduced it — details on `eac_storage.go`. The other three inline notes are non-blocking, and everything else reads really well.

--e+🤖

pkg/saga/eac_storage.go
phinze: 🤖 Confirmed and fixed, took your collapse as written.

Wrote the conformance case first and watched it fail with exactly your output before touching the code: `expected {zzz-real-00, -01, -02}` / `actual {}`. It's `TestStorageConformance_PagingPastStaleEntries`, running on both index-backed backends through a new `makeIndexed` factory that hands back the `MockStore` so a test can seed stale entries the way production grows them.

Two knock-ons. I applied the same shape to `EntityStorage` even though it cannot hit the bug, since its nils hold their place and an all-stale page has non-zero length. Two backends behind one contract should not read differently, and callers now rely on "an empty page can carry a cursor" generally. And once the `continue` was gone, staticcheck pointed out the stage loop always returns on its first iteration, so all four paged methods are now straight-line: decode the cursor, read one page, return.

One of my tests had to give. I had asserted an empty store ends the walk *immediately*, which is no longer true, since a backend cannot tell an exhausted index from a page that resolved to nothing without asking. It now asserts the walk terminates and yields nothing, and says why. Costs an empty store one round trip per status index.

pkg/saga/eac_storage.go
phinze: 🤖 Same fix here. The conformance case exercises the incomplete walk directly; retention gets the identical shape and the same comment.

pkg/saga/retention.go
phinze: 🤖 Right, and the blocker fix makes it worse rather than better: an empty page with a live cursor goes from an edge case to something the backends now produce deliberately, so a one-page lookahead would call a truncated sweep complete fairly often.

It now reads forward until it either finds an execution or the walk genuinely ends, instead of concluding from a single empty page. Still only runs when the delete budget lands exactly on a page boundary.

pkg/saga/paging.go
phinze: 🤖 Good catch, I only had the cheap half of the reason. Written down now: executions only move forward through pending → running → undoing, and the walk is not revision-pinned, so walking the same direction means a saga transitioning mid-walk is seen twice at worst. Reverse it and one sitting in running when undoing is read, and in undoing by the time running is read, appears in neither. The comment says the ordering is the requirement and the cheapness is the coincidence.

servers/entityserver/entityserver.go
phinze: 🤖 Both gone. The stacked comment was mine from the consolidation, and the indexed tail went unreachable the moment `entityPage` started going through `ListIndexEntitiesPage`, taking the revision return with it.

Replaced with `pageSessionIds`, which only does the session case and has no dead branch. Kept it on the server rather than pushing it into the store, since reaching `ListSessionEntities` means base58-decoding a session id out of the attribute value.

evanphx (approved):

PR #1180, opened 2026-09-08
Count service traffic without a named counter

evanphx:
Fixes MIR-1795.

Every service and nodeport chain body opened with `counter name "services"`. That is the nftables objref expression, and it needs `CONFIG_NFT_OBJREF`. The affected host runs 6.1.158+ built without it, so nft rejected that rule and rolled back the batch carrying it.

## What this broke

`Init` references no counters, so it succeeded: the table, the three verdict maps and the static chains all installed. Every batch after that carries a chain body, so every batch was rejected in full — no endpoint chains, no service chain bodies, no DNAT.

The result is a service IP that resolves to nothing. `addServiceChain` and its verdict-map element are in the same rejected batch, so on a host in this state the chain may not exist at all; where an earlier build had already created it, the map routed traffic into an empty chain. Either way the connection hangs rather than being refused. Nothing ever repaired it, because every reconcile and every GC pass failed identically.

## Why it was misread as a missing counter

nft reports the failure like this:

```
Error: Could not process rule: No such file or directory
add rule inet miren service_9yATg4... counter name "services"
                                                   ^^^^^^^^^^
```

ENOENT pointing at an object name reads as "that counter does not exist". It actually means "this kernel cannot look one up". The two are indistinguishable from the error text, and #1165 took the first reading — it declared the counter in the same transaction that referenced it, guarding an absence that was never real. On the affected host `nft list counters table inet miren` showed both counters present the whole time.

The distinction is directly visible there:

- `nft add counter inet ktest c1` succeeds. Named counters can be created.
- `nft add rule inet ktest ch1 counter name "c1"` returns ENOENT — against a counter that demonstrably exists.
- `zcat /proc/config.gz | grep NFT_OBJREF` gives `# CONFIG_NFT_OBJREF is not set`.

Everything else the controller depends on is present on that kernel: `NFT_NUMGEN`, `NFT_NAT`, `NFT_MASQ`, `NF_NAT`, `NF_TABLES_INET`. `OBJREF` is the only gap, and it is one rule.

## The fix

Use an anonymous `counter`. It needs nothing beyond the nf_tables core — there is no `CONFIG_NFT_COUNTER` to be missing, since the expression is built into nf_tables in 6.1 — and it counts per service chain instead of summing every service into one figure, which is more useful when the question is which service is taking traffic. Nothing in the tree read the aggregate.

Per service does not mean cumulative. `Periodic` flushes and rebuilds every chain body on its five minute tick, and an anonymous counter goes away with its rule, so what you read is a rolling window rather than a running total. The named counter lived in the table and survived those flushes. That trade seems right for what these counters are actually used for — during an incident the question is whether traffic is reaching this chain now, which a large cumulative number does not answer — but it is a real difference and worth stating rather than implying. Making the rebuild conditional would need something to diff a live body against, and knftables' `ListRules` exposes `Comment` and `Handle` but not `Rule`; that is follow-up work, not this PR.

`Init` stops creating the two named counters. Tables that already have them keep two inert objects; deleting them would fail on any table that doesn't have them.

Kernels that do have `OBJREF` are unaffected. An anonymous counter works everywhere the named one did.

This removes the counter-declaration half of #1165 and leaves its cache-ordering fix in place. That one stands on its own: a rejected batch should not leave the chain cache asserting rules the kernel never took, whatever the reason for the rejection.

## Verification

By hand on the affected host, in a scratch table mirroring the real chain shapes:

- the old body shape reproduces the production ENOENT and rolls back
- the new body shape applies, giving `counter`, the mark-for-masq jump and the numgen vmap
- a full datapath check — service IP, verdict map, `nat-output` hook, DNAT to a real listener — returns HTTP 200, with the anonymous counter incrementing

`controllers/service` under iso: 30 tests. `make lint`: clean.

The suite cannot reproduce the missing kernel option, because the kernel under iso has `OBJREF`. What it pins instead is the property that makes the option irrelevant. `TestChainBodyDependsOnNoNamedObjects` applies a body in a table where no named counter exists and asserts `Init` creates none; `TestChainBodyUsesAnonymousCounter` fails if `counter name` ever returns.

The two test files are renamed, because their old names described a cause that turned out not to exist.

## Follow-up

`Init` should refuse to start, loudly, when a required nftables expression is missing, rather than writing batches that can never apply. Every expression it emits is load-bearing, and any of them could be absent on the next unusual kernel. A missing kernel config line cost days of investigation purely because the failure was silent and looked like state corruption.

### DIFF
--- a/controllers/service/gc.go
+++ b/controllers/service/gc.go
@@ -357,7 +357,7 @@ func (s *ServiceController) reconcileLiveChains(tx *knftables.Transaction, targe
 	}
 	for _, sp := range target.serviceSpecs {
 		s.addServiceChain(tx, sp.ip, int(sp.port), sp.proto)
-		s.writeChainBody(tx, s.serviceChain(sp.ip, sp.port, sp.proto), "services", sp.endpoints)
+		s.writeChainBody(tx, s.serviceChain(sp.ip, sp.port, sp.proto), sp.endpoints)
 	}
 	for _, sp := range target.nodePortSpecs {
 		chain := s.nodeportChain(sp.nport, sp.proto)
@@ -367,7 +367,7 @@ func (s *ServiceController) reconcileLiveChains(tx *knftables.Transaction, targe
 			Key:   []string{sp.proto, strconv.Itoa(sp.nport)},
 			Value: []string{"goto " + chain},
 		})
-		s.writeChainBody(tx, chain, "nodeports", sp.endpoints)
+		s.writeChainBody(tx, chain, sp.endpoints)
 	}
 }
 
--- a/controllers/service/service.go
+++ b/controllers/service/service.go
@@ -293,9 +293,9 @@ func (s *ServiceController) addServiceChain(tx *knftables.Transaction, ip netip.
 }
 
 // setEndpoints fills in the body of a service-IP or nodeport chain. Skips the
-// flush+rebuild when the endpoint set hasn't changed from the cache to avoid
-// resetting the named counter on each event-driven reconcile. For the
-// unconditional-rebuild path used by Periodic, see writeChainBody.
+// flush+rebuild when the endpoint set hasn't changed from the cache, which also
+// spares the chain's counter from being zeroed on each event-driven reconcile.
+// For the unconditional-rebuild path used by Periodic, see writeChainBody.
 //
 // Anything it writes is recorded in pending rather than in the cache. A cache
 // entry has to mean "nft accepted this body", so the caller commits pending
@@ -345,7 +345,9 @@ func (s *ServiceController) writeChainBody(tx *knftables.Transaction, chain stri
 	// cannot look one up". That rejection rolled back the whole batch, so on
 	// such a kernel no service chain body ever installed and every service IP
 	// was a black hole. A per-chain counter needs no kernel option beyond the
-	// nf_tables core and reports per service rather than one global total.
+	// nf_tables core and reports per service rather than one global total. That
+	// is a rolling window, not a running total -- Periodic rebuilds every body
+	// on its five minute tick, and an anonymous counter goes away with its rule.
 	tx.Add(&knftables.Rule{Chain: chain, Rule: "counter"})
 
 	if len(endpoints) == 0 {
--- a/controllers/service/service.go
+++ b/controllers/service/service.go
@@ -204,12 +204,6 @@ func (s *ServiceController) Init(ctx context.Context) error {
 		Type: "inet_proto . inet_service : verdict",
 	})
 
-	// Named counters that the per-service and per-nodeport chains reference
-	// for traffic visibility. Created here so chain bodies can `counter name`
-	// them without races.
-	tx.Add(&knftables.Counter{Name: "services"})
-	tx.Add(&knftables.Counter{Name: "nodeports"})
-
 	// Static chains. We Add to ensure existence and Flush to wipe stale rules
 	// from prior controller versions, then re-add the canonical rule set.
 	// The duplicate `jump services` rules the bug report flagged came from old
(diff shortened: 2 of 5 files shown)


controllers/service/service.go
```diff
@@ -306,7 +300,7 @@ func (s *ServiceController) addServiceChain(tx *knftables.Transaction, ip netip.
 // Anything it writes is recorded in pending rather than in the cache. A cache
 // entry has to mean "nft accepted this body", so the caller commits pending
 // with commitChainCache once the batch applies -- see Create.
-func (s *ServiceController) setEndpoints(tx *knftables.Transaction, pending map[string][]string, chain, counterName string, endpoints []string) {
+func (s *ServiceController) setEndpoints(tx *knftables.Transaction, pending map[string][]string, chain string, endpoints []string) {
```
phinze, line 303: 🤖 Tiny leftover: the comment above still says the cache is here "to avoid resetting the named counter on each event-driven reconcile." No named counter anymore, and the anonymous one goes away with its rule on every flush. What it describes still happens, since skipping the rebuild does keep the count between events. It just points at something that isn't there now.

controllers/service/gc.go
```diff
@@ -357,7 +357,7 @@ func (s *ServiceController) reconcileLiveChains(tx *knftables.Transaction, targe
 	}
 	for _, sp := range target.serviceSpecs {
 		s.addServiceChain(tx, sp.ip, int(sp.port), sp.proto)
-		s.writeChainBody(tx, s.serviceChain(sp.ip, sp.port, sp.proto), "services", sp.endpoints)
+		s.writeChainBody(tx, s.serviceChain(sp.ip, sp.port, sp.proto), sp.endpoints)
```
phinze, line 360: 🤖 Just noting what this line costs now, since it surprised me. `writeChainBody` starts with `tx.Flush`, and an anonymous counter dies with its rule, so every five minute `Periodic` pass zeroes the count on every service and nodeport chain. The old named counter lived in the table and sailed through flushes. This one doesn't. Per-service counters and an unconditional rebuild can't really coexist, and the PR quietly picks the counters.

No change needed here though. The unconditional rebuild earns its keep, since it unpins the orphan leaves before the delete pass in the same transaction, which is the whole MIR-1129 fix. And you couldn't easily make it conditional anyway: knftables' `ListRules` gives you `Comment` and `Handle` but not `Rule`, so there's nothing to diff a live body against. You'd need a rule-count check, or a comment carrying the endpoint-set hash. Feels like a follow-up ticket rather than this PR.

phinze (approved):
🤖 Nice detective work. A counter that `list counters` shows sitting right there, failing lookup with ENOENT, next to `# CONFIG_NFT_OBJREF is not set`, is about as clean as a diagnosis gets.

Couple of things I checked, and they hold up. Nothing in the tree ever read the `services` or `nodeports` totals, so nobody loses anything when they go. And the named counter was the only objref expression in these chains anyway. `addEndpointChain`'s `counter dnat` and the `counter jump` rules right below it were already anonymous, so this actually makes the body consistent with itself.

One small thing about the description rather than the code. The per-service framing reads like you get durable accounting, but `Periodic` flushes and rebuilds every chain body every five minutes, and an anonymous counter goes away with its rule. So it's really a rolling window. I think that's the better trade anyway, since during an incident you want to know whether traffic is hitting this chain right now, and a big cumulative number won't tell you that. Just worth saying out loud so nobody plans around the stronger version. More in the inline.

CI is red on `TestControlPlaneParse`, which is the HTTP/3 drain flake and not you. #1177 has the fix.

controllers/service/service.go
evanphx: 🤖 Good catch, fixed in c824b3c7. You are right that the behavior it describes still holds — skipping the rebuild does keep the count between events — so the sentence stays and just points at the counter the chain actually has now.

controllers/service/gc.go
evanphx: 🤖 Agreed, leaving the rebuild alone. It has to stay unconditional — it unpins the orphan leaves before the delete pass in the same transaction, which is the whole MIR-1129 fix. And you are right that making it conditional needs something to diff a live body against, which `ListRules` will not give us: `Comment` and `Handle` but not `Rule`. A rule-count check or an endpoint-set hash carried in a rule comment would do it, as its own ticket rather than here.

The part worth fixing was that neither the code nor the description said any of this out loud. c824b3c7 reworks the comment on `writeChainBody` to call it a rolling window rather than a running total, and the PR description now says the same and notes what the named counter used to give us instead.

PR #1181, opened 2026-09-08
Catch up changelog ahead of release

phinze:
The Unreleased section was empty and 72 PRs had merged since the v0.14.0 tag (`9e0b7f4`, Aug 27), so this fills it in: 24 entries across two breaking changes, one security fix, four features, eight improvements, and nine bug fixes.

The biggest single item is the Bring Your Own Image work, folded into one entry under Improvements rather than given its own marquee. `services.web.image` is a complete build source now, with the entrypoint, working directory, and single exposed TCP port inherited from the image, so five PRs collapse into one thing a reader can act on.

Worth a second look before this ships: both breaking changes. `post_import` in `app.toml` now fails validation rather than being quietly ignored, and `miren app history` loses `--status` and `--hide-failed`. Neither is likely to bite many people, but they're the entries where being wrong costs the most.

What's deliberately not here: nine intra-release regressions, each introduced and fixed inside this window so no tagged release ever shipped them; five PRs behind the disabled `appvisibility` labs flag; twelve dead-code chores; two CI sharding changes; and #1093 and #1134, neither of which has an observable surface. Docs-only entries were cut as below the fold.

Docs lint passes.

https://claude.ai/code/session_017jPGiLjm9pGQYMq3rPzjHg


docs/docs/changelog.md
```diff
@@ -11,6 +11,40 @@ All notable changes to Miren Runtime will be documented in this file.
...
+- **Scriptable `miren cluster add`** - Scripting `miren cluster add` meant passing `--address` yourself, which skipped the useful part: trying each address the cluster advertises, asking Miren Cloud to relay when none answer, and saving the cluster's certificate to check later connections against. `--cluster prod` on its own now does all of that, `miren cluster available` lists the names, and `--format json` reports the outcome with a code you can branch on. ([#1118](https://github.com/mirendev/runtime/pull/1118))
+- **Managed application metrics** - A service can opt into managed metrics without exposing its endpoint publicly. The coordinator finds each running sandbox, scrapes it privately, and sends samples to your Prometheus Remote Write destination with Miren's own app, version, service, sandbox, runner, and cluster labels attached. Scrape and delivery failures show up in `miren logs system vmagent`. ([#1087](https://github.com/mirendev/runtime/pull/1087))
+
```
evanphx, line 26: Let's get the new `top` command in here.

evanphx (approved):

docs/docs/changelog.md
phinze: Ah yeah this was generated before that merged, we'll do a small catchup in the release process for the ones that landed since

PR #1185, opened 2026-09-09
Clear out saga executions that get stuck forever

phinze:
Some saga executions get stuck where nothing will ever finish them and nothing will ever delete them. A retry finds its earlier attempt by name, so one with a generated name has nobody left who can name it again. And retention only deletes executions that have finished, because one still in progress is exactly what crash recovery goes looking for. The record stays in the store forever, and every recovery pass pays to read it. One pre-upgrade snapshot has about 19,964 of them for a single addon association. Nothing new is piling up, since convergent IDs shipped in v0.14.0, so this is about clearing out what older versions left behind and it can afford to be slow and careful.

Rather than delete these, the sweep marks them failed. That hands them back to rules we already have, and it works whichever kind of record it hits: one named after its entity gets cleared and retried on the next reconcile, and one with a generated name gets deleted by ordinary retention a week later. Marking also leaves a week where someone can still look at what got cleared and why.

To decide what is stuck, we ask how long an execution has gone without changing. Every step writes a timestamp, so anything actually running stays well clear of a seven-day window. The catch is that v0.11.1's saga schema had no timestamp fields at all, so the very records we are clearing read back with nothing. That is what the first rev is for: a summary read of the in-progress set that keeps the entity store's own timestamps, which get dropped when you decode into an `Execution`.

No new config. Setting `saga.retention_period` to 0 already means leave saga records alone while I investigate, and it now stops both sweeps instead of just the deleting half.

Closes MIR-1788

https://claude.ai/code/session_01SZsBA225fbbLveUiDGmn1W

### DIFF
--- a/pkg/saga/stalled.go
+++ b/pkg/saga/stalled.go
@@ -131,8 +131,9 @@ func RunStalledSweep(ctx context.Context, storage StalledStorage, cfg StalledCon
 			// parent re-finds its children by ID rather than re-running them,
 			// so failing one under a live parent fails the parent for a reason
 			// that was never true. A stranded parent is non-terminal and so
-			// reads as live, which means a stranded tree converges one level
-			// per sweep rather than all at once.
+			// reads as live, which means a stranded tree converges at least one
+			// level per sweep. Often more: a parent forced earlier in the same
+			// walk is already terminal by the time its child is read.
 			if summary.ParentID != "" && parents.isLive(ctx, summary.ParentID, log) {
 				result.Skipped++
 				continue
@@ -151,7 +152,12 @@ func RunStalledSweep(ctx context.Context, storage StalledStorage, cfg StalledCon
 				continue
 			}
 
-			log.Info("forced stalled saga execution to failed",
+			// Debug rather than Info: draining a backlog puts a full sweep's
+			// budget of these into the tier an operator reads, every tick, for
+			// hours. Retention logs per record only when a delete fails, and
+			// the sweep's summary carries the count worth noticing. The id is
+			// one -v away.
+			log.Debug("forced stalled saga execution to failed",
 				"id", summary.ID, "status", summary.Status,
 				"last_changed", summary.LastChanged)
 			result.Forced++
--- a/pkg/saga/stalled.go
+++ b/pkg/saga/stalled.go
@@ -1,0 +1,253 @@
+package saga
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"time"
+)
+
+// StalledError is the error recorded on an execution the sweep forces to
+// failed. It is deliberately a sentence rather than a code: it is read by
+// whoever runs `m debug saga show` on a forced record, and "this saga was never
+// going to finish" is the thing they need to know.
+const StalledError = "forced to failed by the stalled-saga sweep: this execution " +
+	"had not changed state for longer than the stall window, so nothing was " +
+	"driving it and nothing would have collected it"
+
+// StalledConfig tunes a stalled-saga sweep.
+type StalledConfig struct {
+	// StaleAfter is how long an execution may sit without changing state before
+	// the sweep declares it stranded. Zero disables the sweep, which is the
+	// escape hatch if a cluster needs its in-flight sagas left exactly as they
+	// are for an investigation.
+	//
+	// This is a measurement, not a guess about provenance. Save writes a
+	// timestamp on every transition and every action completion, so a saga that
+	// is being driven, or that is retrying its undos on a loop, keeps a recent
+	// one no matter how old the run is. What sits untouched for days is what
+	// nothing is driving.
+	StaleAfter time.Duration
+
+	// MaxForces caps transitions in one sweep so a backlog drains over several
+	// passes rather than one thundering herd of writes. Zero means unbounded.
+	MaxForces int
+}
+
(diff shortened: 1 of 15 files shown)


pkg/saga/stalled.go
```diff
@@ -0,0 +1,204 @@
...
+					"id", summary.ID, "status", summary.Status,
+					"last_changed", summary.LastChanged, "error", err)
+				result.Failed++
+				continue
+			case !forced:
+				result.Recovered++
+				continue
+			}
+
+			log.Info("forced stalled saga execution to failed",
```
evanphx: 🤖 Non-blocking. Draining the 20k backlog at 1000 a tick puts 1000 Info lines in the log every half hour for about ten hours. Retention only logs the summary line plus per-record failures. There's a case for keeping the ID here, since once the backlog is gone any line means something is still stranding sagas, so fine to leave as is. Debug is the other reasonable choice.

pkg/saga/storage.go
```diff
@@ -243,6 +298,68 @@ func (s *EntityStorage) ListTerminalPage(ctx context.Context, q TerminalQuery) (
...
+	exec.Error = reason
+	exec.UpdatedAt = time.Now()
+
+	forced, err := executionToEntity(exec)
+	if err != nil {
+		return false, err
+	}
+
+	if _, err := s.store.ReplaceEntity(ctx, forced, entity.WithFromRevision(ent.GetRevision())); err != nil {
+		if errors.Is(err, cond.ErrConflict{}) {
```
evanphx, line 354: 🤖 Non-blocking, and not this PR's to fix. The refusal we swallow here is logged at Error by the store: `entity revision mismatch` in checkRevisionConflict, and the txn-failed branch logs `error=<nil>`. So the one outcome this function calls not-an-error shows up as one in the entity store's log. Rare in practice. We'll file a ticket to get those lines down to Warn or Debug.

pkg/saga/storage.go
```diff
@@ -254,55 +371,92 @@ func (s *EntityStorage) Delete(ctx context.Context, id string) error {
...
+//
+// It prefers the saga's own updated_at and falls back to the entity store's
+// system timestamps, which is what makes both sweeps safe across an upgrade:
+// v0.11.1's schema had neither field, so its executions read back with a zero
+// saga timestamp, and treating that as infinitely old would collect a saga
+// finished seconds ago or force one that is actively running.
+//
+// The false return is not a "just use now". An execution we cannot date is one
+// we must not act on, and both callers skip it and say so.
+func lastChanged(ent *entity.Entity, s *saga_v1alpha.Saga) (time.Time, bool) {
```
evanphx, line 400: 🤖 👏 easy to have missed.

pkg/saga/stalled_test.go
```diff
@@ -0,0 +1,586 @@
...
+			ctx := context.Background()
+			storage := backend.make(t)
+
+			saveAged(t, storage, "stranded-parent", StatusPending, 30*24*time.Hour)
+			saveStalledChild(t, storage, "stranded-child", "stranded-parent", 30*24*time.Hour)
+
+			result, err := RunStalledSweep(ctx, storage, weekStalled(), testutils.TestLogger(t))
+			require.NoError(t, err)
+			assert.Equal(t, 1, result.Forced)
+			assert.Equal(t, 1, result.Skipped)
```
evanphx, line 214: 🤖 Non-blocking. This holds because `stranded-child` sorts before `stranded-parent` in the walk. If the parent came first it would be forced, the child's liveness read would see it terminal, and both would go in one sweep. Right outcome, failed assertion, so a rename would break this with no bug behind it. Worth a line here saying so. Same thing makes the doc comment in stalled.go more accurately "at least one level per sweep".

evanphx (approved):
Approved. Three non-blocking notes inline, none of them need a change before merge. The store-side Error log is a follow-up ticket against the entity store, not this PR.

--e+🤖

pkg/saga/stalled.go
phinze: 🤖 Took Debug. The signal that survives the backlog is the count, and that is already in the sweep summary at Info, so this now matches retention, which logs per record only when a delete fails. The ids are one `-v` away if someone needs them.

pkg/saga/storage.go
phinze: 🤖 Agreed, leaving it out of this PR. Filed MIR-1801 for the entity-store side, covering all three spots: the two `entity revision mismatch` lines (store.go:798 and :943) and the `ReplaceEntity` branch at :1252 that logs `error=<nil>` and calls a CAS conflict a failure.

pkg/saga/storage.go
phinze: 🤖 Thanks, though it was more theft than insight. Retention already had the fallback for the same upgrade reason, so this was mostly noticing that the stalled sweep needed it more, since v0.11.1 records are exactly the population it is draining.

pkg/saga/stalled_test.go
phinze: 🤖 Both done. The test now says the counts hold because `stranded-child` sorts ahead of `stranded-parent`, that reversing them is a better outcome and a failed assertion, and that the property to keep asserting is that the tree is fully forced by the end. `stalled.go` now says at least one level per sweep, since a parent forced earlier in the same walk is already terminal when its child is read.

PR #1191, opened 2026-09-10
Make the saga path the only one for builds and sandbox creation

phinze:
Sagas went default-on in v0.14.0 with `--labs -sagas` as the way back, and v0.15.0 shipped a full release that way without anyone reaching for it. The pre-saga code has not run in production since August, so we have been carrying two implementations of build and sandbox creation and exercising one. Keeping them honest cost real ceremony: `sandbox_frozen_test.go` hashed three files and broke the build when you touched them, so every behavior change had to be consciously mirrored into the saga path or consciously declared inherited.

Three revs. The first deletes `buildFromDir` and the two `Builder` entry points that called it, leaving `SagaBuilder` as the only implementation and `Builder` as the primitives it wraps. The second folds `SagaSandboxController` into `SandboxController`, which had only ever existed to wrap it and replace one method. The third drops the flag.

The middle one got an audit rather than a rip. That frozen test's accumulated comments are a ledger of twelve divergences, each either inherited through `sandboxOps` or mirrored by hand, and I checked all twelve before deleting the guard. A normalized diff of the two `Create` bodies came down to three redundant Error logs. The one real difference is that the old `createSandbox` mutated the struct handed to it where the saga writes the entity, which moved five `TestSandbox` assertions onto the durable record; those subtests now exercise the saga path under containerd for the first time.

Two things this surfaced are filed rather than fixed. MIR-1807 covers `sandboxOps` being a seam that separates nothing, since twelve of the fourteen methods it adapts are exported with no caller outside the package. MIR-1808 covers the saga executor emitting no spans, which is why deploy traces have had no middle since v0.14.0.

Closes MIR-1460

https://claude.ai/code/session_01FwTXwxDf8C6UQmEMP8oyFg


controllers/sandbox/sandbox.go
```diff
@@ -1835,20 +1619,6 @@ func (c *SandboxController) deleteEndpoints(ctx context.Context, id entity.Id, s
 }
 
 // deallocateNetwork releases the network resources allocated for a sandbox
```
evanphx: Non-blocking: this is `deallocateNetwork`'s old doc comment, now sitting on top of `AllocateNetwork`.

evanphx (approved):
Approved! Good to be down to one path.

One inline note and one non-blocking nit that has no diff line to hang on: the comment above `runBuildkitBuild` in `servers/build/build_saga_buildkit.go` (line 151) still says it's shared with the pre-saga `buildFromDir` path.

--e+🤖

controllers/sandbox/sandbox.go
phinze: 🤖 Fixed. Rather than just dropping the orphaned line I gave `AllocateNetwork` its own doc, since it's exported and deleting alone would have left it bare.

phinze:
🤖 Both notes fixed. The buildkit one made me re-run the sweep for stale references, which turned up a third you didn't hit: `sandboxCgroups` described its key convention as "the way `createSandbox` keys them," and `createSandbox` is gone too. It now points at the create-sandbox saga's boot-containers action, which is where that convention actually lives (`create_saga.go:332`).

On `runBuildkitBuild`, I dropped the "both paths share this" rationale and replaced it with why `status` can be noop, since that's the part still worth knowing: `SenderFor` hands back a noop sender when no stream is registered, which is exactly what a recovered build looks like.

Each fix went into the rev that created the staleness rather than riding on top, so no rev in the stack carries a comment that contradicts its own diff. That force-pushed over what you approved — the only difference is those three comments.

PR #1192, opened 2026-09-10
Retire the distributedrunners labs flag

phinze:
Distributed runners have been on by default since v0.13.0, and `--labs -distributedrunners` was there so an operator could back out if the new topology caused trouble. Two releases went by without anyone reaching for it, so the escape hatch has done its job. The two gates it fed, `miren runner` command registration and etcd mTLS setup at boot, are unconditional now. A server still passing the flag logs `unknown labs feature flag` and starts normally, the same landing as `-sagas` below.

The labs tests come along for a different reason. They were always about how `Init` parses flags, but they named whichever real features happened to exist, so every graduation churned them. They use synthetic names now, which also recovers a case that was about to go vacuous: with `appvisibility` the only real feature left, nothing distinguishes "the exclusion disabled the one I named" from "the exclusion disabled everything."

One docs consequence worth a look. `appvisibility` defaults to off, so labs.md has no honest "turn off a feature that's on by default" example anymore. I reworded that section to describe the case rather than fake a snippet for it.

Stacked on #1191, which retires the `sagas` flag.

Closes MIR-1461

### DIFF
--- a/docs/docs/changelog.md
+++ b/docs/docs/changelog.md
@@ -13,6 +13,7 @@ All notable changes to Miren Runtime will be documented in this file.
 
 **Breaking Changes**
 - **The `sagas` escape hatch is gone** - Builds and sandbox startup have run on the saga engine by default since v0.14.0, and `--labs -sagas` existed to put a cluster back on the previous code path if that went badly. It shipped a full release without anyone needing it, so both the flag and the path it selected are now removed. `-sagas` is no longer a known feature, so a server still passing it logs `unknown labs feature flag` at boot and starts normally; drop it from your `--labs` or `MIREN_LABS` setting to quiet the warning. Nothing else changes, because the saga path is already what you have been running. ([#1191](https://github.com/mirendev/runtime/pull/1191))
+- **The `distributedrunners` escape hatch is gone** - Distributed runners have been on by default since v0.13.0, with `--labs -distributedrunners` there to hide the `miren runner` commands and skip etcd mTLS setup if the new topology caused trouble. Two releases went by without anyone reaching for it, so the flag and its gating are removed: the runner commands are now always registered, and an embedded etcd always comes up with mTLS. As with `-sagas`, a server still passing `-distributedrunners` logs `unknown labs feature flag` at boot and starts normally; drop it from your `--labs` or `MIREN_LABS` setting to quiet the warning. ([#1192](https://github.com/mirendev/runtime/pull/1192))
 
 ---
 
--- a/pkg/labs/labs_test.go
+++ b/pkg/labs/labs_test.go
@@ -3,6 +3,7 @@ package labs
 import (
 	"bytes"
 	"log/slog"
+	"sort"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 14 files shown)


docs/docs/changelog.md
```diff
@@ -13,6 +13,7 @@ All notable changes to Miren Runtime will be documented in this file.
...
+- **The `distributedrunners` escape hatch is gone** - Distributed runners have been on by default since v0.13.0, with `--labs -distributedrunners` there to hide the `miren runner` commands and skip etcd mTLS setup if the new topology caused trouble. Two releases went by without anyone reaching for it, so the flag and its gating are removed: the runner commands are now always registered, and an embedded etcd always comes up with mTLS. As with `-sagas`, a server still passing `-distributedrunners` logs `unknown labs feature flag` at boot and starts normally; drop it from your `--labs` or `MIREN_LABS` setting to quiet the warning.
```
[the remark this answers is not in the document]
phinze: Declining this one. `docs/lint.ts:120` only matches a bold admonition keyword followed by a colon (`**NOTE:**`, `**Warning**:`); `**Breaking Changes**` is neither, and `make lint` passes on this file as written. It's also the section heading used at all seven release sections in this changelog, so converting just these two would break the file's own convention, and docs/CLAUDE.md specifically warns against stacking adjacent admonitions.

phinze:
Heads up for human reviewers: both bots reviewed `main`...`HEAD` rather than this PR's base, so most of their findings are about #1191's revs (`controllers/sandbox/*`, `servers/build/build.go`). This PR is 11 files, +280/-267. The saga_controller and sandbox.go findings belong on #1191.

On biscuit's etcd mTLS caveat: `distributedrunners` has defaulted to `true` since v0.13.0, so an installation "running without the flag" has been starting embedded etcd with mTLS for two releases already. The only clusters that change behavior here are ones explicitly passing `-distributedrunners`, which is the population this issue exists to retire.

docs/docs/changelog.md
evanphx: Non-blocking: the sagas entry above ends with a link to its PR, this one doesn't. Worth a `([#1192](...))` for symmetry.

pkg/labs/labs_test.go
```diff
@@ -12,43 +13,93 @@ func quiet() *slog.Logger {
...
+		t.Error("alpha should be disabled after '-alpha'")
 	}
 }
 
-func TestDistributedRunnersEnabledByDefault(t *testing.T) {
-	Reset()
+// A feature that ships on by default keeps its flag for a release purely so an
+// operator can put the old behavior back. That escape hatch has to work from a
+// cold start, with nothing else named.
+func TestNegativePrefixAloneDisablesDefaultOnFeature(t *testing.T) {
```
evanphx, line 75: 👏 Checked that this goes red if the `-` prefix stops working (made it a no-op in `Init`, three tests failed including this one).

evanphx (approved):
Approved! Nicely done.

Two leftovers, both non-blocking. The comment at `.github/workflows/test.yml:484` still says the distributed suite runs with distributedrunners enabled. And the `--labs` help on `server container install` (`cli/commands/server_install_container.go:39`) still says `(e.g. sagas)`, which flows into the generated docs. That one is really #1191's, fine to fix on either.

--e+🤖

docs/docs/changelog.md
phinze: 🤖 Added, thanks. The entry closes with `([#1192](https://github.com/mirendev/runtime/pull/1192))` now, same as the sagas one.

pkg/labs/labs_test.go
phinze: 🤖 Thanks for actually mutating `Init` to check that. A test pinning an escape hatch is exactly the kind that can quietly assert nothing, so it is good to have it confirmed from the outside.

phinze:
🤖 Both body leftovers fixed here rather than splitting them across the stack.

`.github/workflows/test.yml:484` no longer says the distributed suite runs "with distributedrunners enabled" — with the flag gone it just exercises the multi-node topology.

`server_install_container.go:39` now reads `(e.g. appvisibility)`. You were right that it reaches the generated docs; `docs/docs/command/server-container-install.md` moved with it. Took it here since this PR is what leaves `appvisibility` as the only feature left to name.

docs/docs/changelog.md
```diff
@@ -13,6 +13,7 @@ All notable changes to Miren Runtime will be documented in this file.
...
+- **The `distributedrunners` escape hatch is gone** - Distributed runners have been on by default since v0.13.0, with `--labs -distributedrunners` there to hide the `miren runner` commands and skip etcd mTLS setup if the new topology caused trouble. Two releases went by without anyone reaching for it, so the flag and its gating are removed: the runner commands are now always registered, and an embedded etcd always comes up with mTLS. As with `-sagas`, a server still passing `-distributedrunners` logs `unknown labs feature flag` at boot and starts normally; drop it from your `--labs` or `MIREN_LABS` setting to quiet the warning. ([#1192](https://github.com/mirendev/runtime/pull/1192))
```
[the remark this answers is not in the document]
phinze: 🤖 Declining again, same as https://github.com/mirendev/runtime/pull/1192#discussion_r3981948161 (which you resolved earlier this run).

`docs/lint.ts:120` only matches a bold *admonition keyword* followed by a colon — `**NOTE:**`, `**Warning**:`. `**Breaking Changes**` is neither in that keyword set nor colon-suffixed, so the rule does not fire, and `make lint` passes on this file as written (`✓ docs lint passed`). It is also the section heading used at all seven release sections in this changelog, so converting only these two would break the file's own convention, and docs/CLAUDE.md warns specifically against stacking adjacent admonitions.

The only change since your last pass was appending the PR link to the entry.

PR #1194, opened 2026-09-10
Make `miren deploy` output scriptable: clean non-TTY text, JSON, and a JSONL event stream

evanphx:
When `miren deploy` is run from CI, a script, or an agent, its output was built for a person watching a terminal. The upload meter rewrote its line with clear-line escape codes, which land in a captured log as `[K` noise. Every BuildKit step was replayed verbatim. Success was implicit, and the version ID only existed inside a sentence, so anything programmatic had to scrape prose. The `--summary-json` file covered the GitHub Action but was awkward for `miren deploy | jq`. This is MIR-660.

The first commit fixes the plain path. When stdout isn't a terminal the upload progress is ordinary newline-terminated lines, BuildKit's plain renderer is used even if stderr happens to be a terminal (so `deploy | tee log` stays clean), the build is condensed to one `Build & push image (0.5s) - 14 steps, 9 cached` line, and every deploy ends with `✓ Deploy successful` plus a `Version: <full id>` line that round-trips into `--version`. `--format json` moves the commentary to stderr and writes one document to stdout, on failure too, and `--quiet` drops the progress streams while keeping the summaries.

The second commit adds `--format jsonl`. It's the fully structured mode: one JSON object per line as the deploy progresses (upload, build steps and their logs, deployment phases, health, warnings, crash logs), ending in a `result` line that carries the same fields as the JSON document. Nothing unstructured is printed on either stream in this mode, so the CLI's own logger is routed into the stream as `log` events and a failure returns a bare exit code rather than an `Error:` line.

Dogfooding that immediately found a real bug, which is the third commit. Running `miren deploy --format jsonl > out.jsonl` from inside the project fails with `archive/tar: write too long`, because the output file is part of the upload and grows while it's being packed. Any file being appended to during a deploy (a log, say) could hit this. The archiver now writes exactly the size the header promised, cutting a file that grew and zero-padding one that shrank, so the upload survives and the next deploy picks up the newer content.

The last commit adds `miren debug deploy-events`, which turns a saved JSONL stream (or a live pipe) back into the readable output the deploy would have printed. `--timestamps` prefixes each line with the elapsed time since the first event, which turns out to be the quickest way to see where a slow deploy actually spent its time.

Closes MIR-660


cli/commands/health_wait.go
```diff
@@ -274,7 +315,10 @@ func waitForActivation(ctx *Context, getter appInfoGetter, tailer logTailer, app
 
 	text, ok := healthOutcomeText(versionDisplay, outcome, snap)
 	ctx.Printf("%s\n", healthSummaryLine(ok, text, elapsed))
-	return reportHealthResult(ctx, tailer, appName, snap, versionDisplay, ok)
+	if obs != nil {
+		obs.healthVerdict(versionID, text, ok, elapsed)
```
phinze: 🤖 This hands the observer a bool when it's standing on top of much more. `outcome` is a six-value `terminalOutcome` and `snap` carries `.health` (an `apphealth` string) plus ready/desired/crashCount/cooldownSeconds, and none of it reaches the event.

The collapse shows up in the output. `outcomeScaledToZero` returns `ok=true`, so an app with no instance running emits `{"event":"health","status":"healthy"}` while the `message` right next to it reads "scaled to zero, no instance running right now." `outcomeTaskOnly` lands on `healthy` too, though `apphealth` already names that state `Ready`. On the other side, a crash-loop and a timeout both become `failed` and the difference survives only in the English.

So a consumer asking "did this actually come up serving traffic" has to parse `message` to find out, which is the thing `--format jsonl` exists to make unnecessary.

Passing `snap.health` through as its own field would close it, and `pkg/apphealth` is already the shared contract for that value. Its package doc says it lives there "so that both the app RPC surface and the cloud health reporter can name the same value without either depending on the other." A deploy event stream reads like the third consumer that sentence anticipated.

cli/commands/deploy.go
```diff
@@ -1593,9 +1773,51 @@ func buildStepsSummary(count int) string {
...
+// is reported as "cancelled" rather than "failed" so a script can tell the two
+// apart. It also pins URLs to a stable array (never JSON null) so consumers
+// see a fixed schema.
+func (s *deploySummary) finalize(err error) {
+	if s.URLs == nil {
+		s.URLs = []string{}
+	}
+	switch {
+	case err == nil:
+		s.Status = "success"
```
phinze: 🤖 `success` diverges from `deploylifecycle.StatusSucceeded`, which spells it `succeeded`. The two cases below it, `cancelled` and `failed`, match their `deploylifecycle` counterparts exactly, so this is the one value out of three that doesn't line up.

That's an awkward thing to freeze into a published format, because both spellings are reachable from the same CLI for the same deployment. `miren deploy --format json` gives you `{"status":"success","deploy_id":"deployment-N"}`, and `miren app history --format json` reports `"status":"succeeded"` for that same id, since it passes `dep.Status()` through untouched. Anything branching on `status == "succeeded"` across both surfaces quietly reads every successful deploy as unsuccessful.

Reusing `deploylifecycle.Status` here, or mapping to it explicitly, would settle it. It would also give `deployment.phase` a typed `deploylifecycle.Phase` instead of a pass-through string that agrees with the server today only because both sides happen to spell it the same way.

phinze (changes requested):
The behavior change here is great and I want it. What I'd like to sort out first is where the vocabulary lives, so we're not growing a second taxonomy next to the ones we already have, and so `cli/commands` keeps trending toward being just CLI plumbing.

`--format jsonl` is a new public contract, documented in `deploy.md` as one, and most of the values it carries already have owners. `pkg/deploylifecycle` owns deploy status and phase and guards them with an `init()` drift check. `pkg/apphealth` owns health classification, and its package doc says it sits in its own package specifically so the RPC surface and the cloud health reporter can name the same value without either depending on the other. The stream defines its own values for all three instead of reusing them. Two inline comments have the details: a `success` versus `succeeded` split that's already reachable two ways from the same CLI, and a health field lossy enough that a consumer has to fall back to parsing English.

The event names themselves (`start`, `upload_complete`, `build_step`, `result`) are genuinely the stream's own vocabulary and read well. It's the status values inside the events I'd want to reconcile, mostly because a JSONL format is frozen the moment someone scripts against it. That's the only reason this is request-changes rather than a note on an approval.

Moving `IsJSONL` off the shared `FormatOptions` in the latest push is the same instinct, and it marks the seam for the rest. The rendering, the plain progress meter, and `debug deploy-events` are CLI work and belong where they are. The event schema and its status vocabulary look like they want to sit next to `deploylifecycle` and `apphealth`, where Miren Cloud can import them too.

--p+🤖

phinze (approved):
Nice pivots - looking good!

cli/commands/health_wait.go
evanphx: 🤖 Agreed, the bool threw away exactly what a consumer wants. The observer now receives the poll's outcome and the snapshot behind it, and the `health` event carries a typed `outcome` (`waiting`, `healthy`, `scaled_to_zero`, `task_only`, `crashed`, `timeout`), an `ok` flag, the `apphealth` value as `health`, and `ready`/`desired`/`crash_count`/`cooldown_seconds`. A scaled-to-zero app now reads `{"outcome":"scaled_to_zero","ok":true,"health":"idle"}`, and crash-loop and timeout are distinct without touching `message`. Fixed in 9760a792.

cli/commands/deploy.go
evanphx: 🤖 Good catch, and the `app history` cross-reference made it concrete. `Result.Status` is now a `deploylifecycle.Status` (`succeeded`, `failed`, `cancelled`), `deployment.phase` is a `deploylifecycle.Phase`, and the whole event schema moved into a new `pkg/deployevents` beside `deploylifecycle` and `apphealth` so Cloud can decode the stream with the same types. The rendering, the progress meter, the BuildKit conversion, and `debug deploy-events` stayed in `cli/commands`. Fixed in 9760a792.

PR #1195, opened 2026-09-10
Pin container install image to the build version instead of :latest

evanphx:
`miren server container install` has been defaulting `--image` to `oci.miren.cloud/miren:latest`. That meant a v0.14.0 CLI would happily install whatever `latest` pointed at that day, and running the same command a month later could land you on a different server. The systemd installer never had this problem: it defaults `--branch` from the version baked into the binary, so a release CLI installs its own release.

This brings the container path in line. The image tag now follows the build. A tagged release pins its own tag, a `main` or detached-HEAD build uses `main` (the release workflow publishes both), and any other dev build stops with an error telling you to pick with the new `--version` flag (or `--image` for a fully custom reference). Nothing quietly falls back to `latest` anymore, though `--version latest` still works if you actually want it.

One small extra: detached checkouts report their branch as `HEAD`, which nothing publishes under. `version.Branch()` now maps that to `main`, which also fixes the systemd, runner, prepare, and download-release paths that would otherwise try to fetch a release bundle for a channel named HEAD.

Closes MIR-1090

### DIFF
--- a/cli/commands/commands.go
+++ b/cli/commands/commands.go
@@ -1132,6 +1132,10 @@ miren deploy --format jsonl | jq -c 'select(.event == "build_step")'
 			Name: "Force a specific runtime",
 			Body: "miren server container install --runtime podman",
 		}),
+		WithExample(mflags.Example{
+			Name: "Install a specific release",
+			Body: "miren server container install --version v0.15.0",
+		}),
 	))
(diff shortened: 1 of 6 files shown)


evanphx:
@coderabbitai review

phinze (approved):
Read through resolveContainerImage and the BranchOf refactor, checked the tag grammar against what release.yml actually publishes (`:main` and `:v*`), ran the new tests locally. LGTM.

--p+🤖

PR #1196, opened 2026-09-10
Surface image pull failures in miren logs

evanphx:
When a node can't pull a sandbox's image, the boot fails and the sandbox gets marked DEAD, but the reason only ever reached that node's journal. The case that prompted this (MIR-1541) is a distributed runner that can't open port 5000 on the coordinator: the deploy reports "crash-looped" with no logs at all, because no container ever ran, and `miren logs` is just as empty. You'd have to ssh to the runner and read journalctl to find out why.

We already had the right mechanism for this. `EmitSandboxEvent` writes a `[miren] ...` line onto the sandbox's log stream, keyed by the app entity, and the volume and port-binding failures use it. Image pulls just never did. This change factors the duplicated GetImage/Pull/GetImage block out of `BuildSpec` and `buildSubContainerSpec` into one `ensureImage` helper that emits `[miren] failed to pull image <ref>: <err>` on failure before returning the same error as before. The line now shows up in `miren logs <app>`, `miren logs sandbox <id>`, and the "Recent logs" tail a failed deploy prints. When the ref is on the cluster registry and the error is a network one, the line also says to check that the node can reach the coordinator on port 5000, since the raw containerd error names an IP but not what should be listening there.

Verified in the dev environment by removing an app's image from containerd, blocking port 5000, and restarting the app: the full line landed in `miren logs` within a few seconds, one per boot attempt.

Fixes MIR-1541

### DIFF
--- a/controllers/sandbox/image_pull.go
+++ b/controllers/sandbox/image_pull.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	containerd "github.com/containerd/containerd/v2/client"
+	"github.com/containerd/errdefs"
 
 	compute "miren.dev/runtime/api/compute/compute_v1alpha"
 	"miren.dev/runtime/components/ocireg"
@@ -25,6 +26,12 @@ func (c *SandboxController) ensureImage(ctx context.Context, sb *compute.Sandbox
 	img, err := c.CC.GetImage(ctx, ref)
 	if err == nil {
 		return c.logImageReady(ctx, img)
+	}
+	// Only a missing image justifies a pull. Anything else (a containerd RPC
+	// failure, a cancelled context) would just fail again and be misreported
+	// as a pull failure.
+	if !errdefs.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get image %s: %w", ref, err)
 	}
 
 	_, err = c.CC.Pull(ctx, ref, containerd.WithPullUnpack, containerd.WithResolver(c.resolver()))
--- a/controllers/sandbox/image_pull.go
+++ b/controllers/sandbox/image_pull.go
@@ -29,7 +29,13 @@ func (c *SandboxController) ensureImage(ctx context.Context, sb *compute.Sandbox
 
 	_, err = c.CC.Pull(ctx, ref, containerd.WithPullUnpack, containerd.WithResolver(c.resolver()))
 	if err != nil {
-		c.EmitSandboxEvent(sb, shortID, describePullFailure(ref, err))
+		// A cancelled context means we were superseded or are shutting down,
+		// not that the pull failed. Skip the event so the app's logs don't
+		// report a failure (and, via the url.Error wrapper, a network hint)
+		// for something that never went wrong. Same guard as the port wait.
+		if ctx.Err() == nil {
+			c.EmitSandboxEvent(sb, shortID, describePullFailure(ref, err))
+		}
 		return nil, fmt.Errorf("failed to pull image %s: %w", ref, err)
 	}
 
@@ -56,8 +62,8 @@ func (c *SandboxController) logImageReady(ctx context.Context, img containerd.Im
 // error names an IP and port but not what is listening there, and the
 // distributed-runner case that motivated MIR-1541 is exactly a runner that
 // cannot open port 5000 on the coordinator.
-const registryUnreachableHint = "check that this node can reach the cluster image registry (" +
-	ocireg.Host + ", served by the coordinator on port 5000)"
+const registryUnreachableHint = "check that this node can reach the cluster image registry at " +
+	ocireg.Host + " on the coordinator"
 
 // describePullFailure builds the user-facing line for a failed image pull.
 // The containerd error is kept verbatim because its tail carries the root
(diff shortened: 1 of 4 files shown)


controllers/sandbox/image_pull.go
```diff
@@ -0,0 +1,75 @@
...
+func (c *SandboxController) ensureImage(ctx context.Context, sb *compute.Sandbox, shortID, ref string) (containerd.Image, error) {
+	img, err := c.CC.GetImage(ctx, ref)
+	if err == nil {
+		return c.logImageReady(ctx, img)
+	}
+
+	_, err = c.CC.Pull(ctx, ref, containerd.WithPullUnpack, containerd.WithResolver(c.resolver()))
+	if err != nil {
+		c.EmitSandboxEvent(sb, shortID, describePullFailure(ref, err))
+		return nil, fmt.Errorf("failed to pull image %s: %w", ref, err)
```
phinze, line 46: 🤖 This emits on any Pull error, including a cancelled context, so a superseded deploy or a shutdown that interrupts an in-flight pull writes a failure line into the user's app logs for something that did not actually fail.

I ran it against real containerd. With an already-cancelled context the line is `[miren] failed to pull image <ref>: context canceled`. Cancelled mid-dial it comes through as the `*url.Error`-wrapped `context deadline exceeded`.

On a cluster-registry ref it goes a step further, because `*url.Error` satisfies `net.Error` regardless of what it wraps, so `describePullFailure` appends the reachability hint too:

```
failed to pull image cluster.local:5000/app:v1: ...: context canceled; check that
this node can reach the cluster image registry (cluster.local:5000, served by the
coordinator on port 5000)
```

That points the operator at the network when the real cause was cancellation.

The port-wait loop in `createSandbox` already guards exactly this case (sandbox.go:1579):

```go
if ctx.Err() != nil {
	// We're shutting down, not looking at a port mismatch — skip
	// diagnosis so we don't emit misleading "listening elsewhere" events.
	return err
}
```

An early return in the same shape here, before the emit, would leave the wrapped error return unchanged.

controllers/sandbox/image_pull.go
```diff
@@ -0,0 +1,75 @@
...
+// The containerd error is kept verbatim because its tail carries the root
+// cause (a dial timeout, a 404, an auth failure); the hint is only added when
+// the failure is a network error against the cluster registry.
+func describePullFailure(ref string, err error) string {
+	msg := fmt.Sprintf("failed to pull image %s: %v", ref, err)
+
+	var netErr net.Error
+	if strings.HasPrefix(ref, ocireg.Host+"/") && errors.As(err, &netErr) {
+		msg += "; " + registryUnreachableHint
+	}
```
phinze, line 85: 🤖 The hint matches one spelling of the cluster registry, but both resolvers accept two. `SandboxController.resolver` (sandbox.go:1985) and your `directImageRegistryHosts` (servers/build/direct_image.go:149) both route `cluster.local` and `cluster.local:5000` to the local registry, while this only prefix-matches `ocireg.Host + "/"`. So a ref written as `cluster.local/foo:v1` gets pulled from the cluster registry and then misses the reachability hint on exactly the failure the hint exists for.

In practice every producer writes the `:5000` spelling (build.go:1172, build_saga_buildkit.go:284), so this only bites a hand-written bare-host `image` in a miren.hcl going through the first-class image path. Worth knowing `image_gc.go:266` makes the same single-spelling assumption, so it is a pre-existing convention rather than something you introduced here.

Entirely your call whether to widen scope and teach both spots the two spellings, or file it as a follow-up. Not a reason to hold this PR.

Tiny thing while you are in here: the rendered hint reads `(cluster.local:5000, served by the coordinator on port 5000)`, so port 5000 lands twice in one parenthetical and four times in the full line once the containerd error is included.

phinze (approved):
Nice improvement! The mechanism is the right one. `EmitSandboxEvent` was already how volume and port-binding failures reach a sandbox's log stream, so image pulls were a conspicuous hole. Collapsing the two duplicated GetImage/Pull/GetImage blocks into one `ensureImage` is a good bonus on top.

This'll need to be rebased over the frozen sandbox test going away, since we dropped the dual saga path today. #1192 removed `sandbox_frozen_test.go` outright, so the hash bump and its audit note can go with it.

Two inline notes, neither of which needs to hold this up: one on context cancellation that I think is worth the couple of lines, and one on the `cluster.local` spelling that is yours to fix here or defer.

--p+🤖

controllers/sandbox/image_pull.go
evanphx: 🤖 Good catch, and thanks for running it against real containerd. Fixed in 1367506c: we now return the wrapped error without emitting when `ctx.Err()` is set, same shape as the port-wait guard. The containerd-backed test got a second pass with an already-cancelled context that asserts no entry is written.

controllers/sandbox/image_pull.go
evanphx: 🤖 Went ahead and widened it in 1367506c: `isClusterRegistryRef` now accepts both `cluster.local:5000/` and `cluster.local/`, with a table case for the bare spelling. Also took the wording nit, the hint now reads `check that this node can reach the cluster image registry at cluster.local:5000 on the coordinator`, so the port is named once. We left `image_gc.go`'s single-spelling assumption alone since it predates this change and nothing in the build pipeline writes the bare form.

PR #1197, opened 2026-09-10
Keep commas in env values and accept bare KEY=VALUE args in env set

evanphx:
Fixes MIR-1168.

`miren env set -e KEY=a,b,c` was setting three variables (`KEY=a`, `b`, and `c`) because the flag library splits every repeatable string flag on commas. Once split, the pieces cannot be told apart from three separate `-e` flags, so the fix lives in mflags: a `split:"false"` struct tag that keeps each flag occurrence whole (mirendev/mflags#16, merged).

This PR:

- Bumps mflags to the merge commit of mirendev/mflags#16.
- Applies `split:"false"` to the `-e` and `-s` flags of `env set` and `deploy`, and to `--org` on `auth provider add github`, whose documented `name:team1,team2` form was being split the same way.
- Lets `env set` accept bare `KEY=VALUE` arguments with no flag, treated the same as `-e`. Docs and an example cover this.

Tests go through the real flag set for all three commands. The command docs were regenerated and show no drift beyond the `env set` changes.


phinze (approved):
Approved upstream so you can repin to the merged commit

PR #1203, opened 2026-09-11
Stop streaming RPC leaks on v0.15

phinze:
Periodic controller watch resyncs in v0.15.0 left streaming connections alive, accumulating goroutines and memory even without app traffic. Close each call's dedicated connection when its session ends or its dial fails, and backport #1202 to release closed connections from diagnostics.

A minimal patch to the v0.9 WebTransport fork also releases closed sessions from its registry. The transport protocol and QUIC version stay at their v0.15.0 versions. The fork is pinned to an exact commit, and the v0.15.1 changelog is included.

Repeated-call and watch-resync probes stay flat in goroutines, with retained heap bounded in the repeated-call probe. RPC/watch tests, focused race checks, mixed-version watch/exec tests with packet loss, and lint pass.

Part of MIR-1820

### DIFF
--- a/docs/docs/changelog.md
+++ b/docs/docs/changelog.md
@@ -13,6 +13,14 @@ All notable changes to Miren Runtime will be documented in this file.
 
 ---
 
+## v0.15.1
+*2026-09-11*
+
+**Bug Fixes**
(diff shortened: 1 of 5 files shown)


phinze:
🤖 Biscuit's dependency checks are covered by [mirendev/webtransport-go#2](https://github.com/mirendev/webtransport-go/pull/2). The fork is public, based on upstream v0.9.0, and pinned here to commit `180830f79577`. Its lifecycle regression fails without the registry cleanup and passes with it, including 10 race-detector repetitions. The dependency PR's CI is green. I also fetched the pinned version from the public Go proxy into a fresh module cache with checksum-database verification enabled; both checksums match the exact commit we tested locally.

evanphx (approved):

PR #1209, opened 2026-09-11
Make the CLI nicer on an unknown command

evanphx:
Typing a command name wrong gave an unhelpful message:

```
$ miren app update
ERROR: error parsing flags: unexpected arguments: [update]
```

Nothing about flags was wrong, the Go slice syntax leaked into user-facing
output, and it never said what the user probably meant — even though the
dispatcher knew every valid command name.

Now:

```
$ miren app lst
ERROR: unknown command "lst" for "miren app"

Did you mean?
  list

Run 'miren app --help' to see available commands.
```

```
$ miren deploy --aap x
ERROR: unknown flag: --aap

Did you mean?
  --app
```

## Where the work is

The fix lives in **mflags** — the only place with both the mistyped word and
the command registry. See mirendev/mflags#17, now merged — this branch is
pinned to the merge commit on mflags main, so there is no cross-repo ordering
left to worry about.

This repo carries the pin bump plus tests that exercise the messages against
the real command tree, since the wording is the product here and mflags can
only test a synthetic dispatcher.

## What it fixes

Three paths produced bad behavior, all now covered:

| Input | Before | Now |
|---|---|---|
| `miren depoy` | `unknown command: depoy` | names the word, suggests `deploy` |
| `miren app lst` | `error parsing flags: unexpected arguments: [lst]` | names the word, suggests `list` |
| `miren runner update` | printed help, **exit 0** | errors with exit 1 |

The third was the quiet failure: section commands tolerate unknown flags, so
they skip the extra-argument check and a typo looked like success.

Two smaller repairs: `unknown command: depoy myapp` used to blame every word
typed and now blames only the one at fault, and `unexpected arguments: [foo]`
no longer prints a Go slice.

## Suggestion quality

Against the real command tree, every realistic typo I tried lands on exactly
one correct suggestion (`lst`→`list`, `wohami`→`whoami`, `upgrde`→`upgrade`,
`verisons`→`versions`, `rollbak`→`rollback`). Nonsense gets no guess.

`miren app update` is itself an example of staying quiet — nothing under `app`
is close to `update`, so it names the word and points at `--help` without
guessing. That is deliberate: a confident wrong suggestion is worse than none.

## Behavior change

`miren runner update` and its shape go from exit 0 to exit 1.

## Testing

`go test ./cli/...` passes and `golangci-lint run ./cli/...` is clean.

`cli/commands/unknown_command_test.go` covers each error shape against the
full dispatcher, plus the inputs the new check must leave alone: help
keywords, pass-through arguments, and value-taking global flags before a
section name.

Also checked by hand against a built binary that valid invocations are
untouched — `app run echo hi`, `logs -a foo`, `-C prod app list`, `auth help`,
`app --help`, `app list --format json`.

Note: `make lint` fails locally on 4 staticcheck (SA4023) findings in
`pkg/cloudrpc` and `pkg/rpc`. Those are a local toolchain artifact, not
something this PR introduces: the files are identical to main, and the checks
only fire on golangci-lint 2.13.x, while CI pins 2.12.2. CI lint is green.

MIR-1823


phinze (approved):
Much nicer UX. Related Q do we wanna do the “shortest uniq abbrev” thing jj does?

evanphx:
@phinze in what way? I'm familiar with it for revs but not what it does for commands.

phinze:
@evanphx ah yeah I meant “shortest unique prefix works” so like `jj b a` works as `jj bookmark advance`

kinda convenient, not sure if they apply it genetically or by hand

evanphx:
@phinze I'm open, let's revisit in a separate PR.

PR #1210, opened 2026-09-12
Bound the control process with a systemd memory limit

evanphx:
A goroutine leak grew the miren control process until the host ran out of memory and stopped responding to anything, including SSH. Nothing capped it, so the only outcome on the table was the whole machine going down. This puts a memory limit on the unit, so the kernel stops and restarts the server instead. Miren's availability is the price, which is the right trade.

The thing that makes this safe to turn on by default is where the cgroups actually sit, and it's worth spelling out because the tree is not what you'd guess. `miren.service` holds the coordinator, containerd, and one shim per container. It does not hold the containers. Sandboxes get created at `/miren/sandbox-<id>` and everything else containerd runs at `/<namespace>/<id>`, both absolute paths that runc resolves against the cgroup root, so `/miren` ends up a top-level cgroup sibling to `system.slice` holding every app, addon database, etcd, buildkit and the two Victoria services. Capping the service cannot touch any of them. (A comment in `runtime_memory_linux.go` asserted the opposite. It was wrong, and it's fixed here. Verified with `systemd-cgls` on staging.)

That also changes what a sensible number looks like. Since apps are outside the cap, sizing it near the whole machine would be close to useless: memory outside the service can exhaust the host well before the coordinator's own cgroup fills up. So the cap is sized to what a control plane actually needs, a quarter of host RAM floored at 2 GB and capped at 16 GB. Generous against a healthy control plane that idles around 600 MB, and tight enough that it would have caught the leak that started this.

The limit ships as a drop-in rather than going into the unit template, and that part matters more than it looks. The base unit gets written once at install and then left alone forever unless someone passes `--force`, so editing the template would reach nobody who already has miren installed, which is precisely the population at risk. A drop-in can be rewritten unconditionally on install and on upgrade, so existing hosts pick this up on their next `miren server upgrade` with nothing to do by hand. It also leaves `systemctl edit` free as the override, and regenerating on upgrade means a resized VM gets a limit matched to its new size.

The other half of the ticket is telling someone it happened. This is harder than it sounds, because systemd resets a unit's `Result` to `success` the moment the replacement instance activates. A server that inspects its own unit during startup always sees a healthy unit, no matter how violently its predecessor died. The one window where the reason is readable is an `ExecStopPost` hook, so that's where we grab it and write it to disk for the next boot to find.

The report then goes out twice, deliberately. Once before startup begins, which lands in the journal and is the only output a server stuck in a memory crashloop will ever produce. Once after observability comes up, which is what puts the line in `miren logs system` where people actually look. Only the second one retires the record, so a crashlooping server keeps saying something and still gets the durable line on the first boot that works. One duplicated journal line on a healthy boot is a much smaller problem than a silent crashloop.

Two drive-bys came along. The systemd test harness was writing the unit from its own hand-copied heredoc, which had drifted and been missing `ExecReload` for months, so the harness was faithfully verifying a unit nobody shipped. It now calls `miren server install` and asserts on what systemd actually resolved. And `configs/miren.service.example` was a fourth copy of the unit, drifted worse than the others and read by no code at all, so it's gone.

Closes MIR-1822


evanphx:
Thanks, both real. Fixed in fed0620a.

**`quoteExecArg` escaping** — correct on both halves. `systemd.service(5)` says "To pass a literal dollar sign, use \`$$\`", and `systemd.unit(5)` says "Use \`%%\` in place of \`%\`". A backslash escapes neither. Since both are interpreted whether or not the value is quoted, the doubling now happens ahead of the quoting decision rather than inside it, and `unquoteExecArg` undoes it when reading a path back out of an existing drop-in. Round-trip test covers `$`, `%n`, a trailing `%`, and the combination with a space.

**`set-property --runtime` cleanup path** — also correct, and worse than it looks: since `/run` outranks `/etc` in drop-in precedence, the squeeze surviving its own cleanup means the service restarts still capped at 32 MB. Now clears both `system.control` and `system`.

**The RAM-detection regression** was already fixed one commit earlier, in f3f657bb. `installServiceLimits` does still call `servicelimits.Write` unconditionally, so reading the caller alone gives the impression the write goes through, but `Write` itself returns early and keeps the existing file when `!l.Known()` and a drop-in is already present. It reports whether it wrote so the caller can say which happened. Covered by `TestWriteKeepsAnExistingLimitWhenRAMIsUnknown`, with `TestWriteStillInstallsWhenRAMIsUnknownAndNothingExists` for the first-install case that still needs the hook.

hack/systemd/entrypoint.sh
```diff
@@ -57,44 +161,25 @@ setup_miren() {
...
-    systemctl daemon-reload
-    systemctl enable miren.service
-
-    # Don't start automatically - let user control when to start
+    # Install the unit through the real code path rather than a copy of it.
+    # This file used to carry its own heredoc of the unit, which silently went
+    # stale: it was missing ExecReload for months, so the harness tested a unit
+    # nobody shipped.
+    log "Installing systemd service via 'miren server install'..."
+    if ! miren server install --force --no-start --without-cloud --skip-system-check \
```
phinze, line 169: 🤖 The local-binary path ends up testing main's server, not the local one. `/miren-local/miren` is copied into the release dir above, but `server install` then calls `ensureReleaseBundlePresent` (`server_install.go:243`), which keys on `release/containerd` existing. It does not on a fresh container, so it downloads the `$RELEASE` bundle and `promoteReleaseDir` swaps the whole directory in, `miren` included (`hack/package-release.sh:18` confirms the bundle carries the binary).

So the local build runs the install and writes the drop-in, and then the unit's `ExecStart` and the `ExecStopPost` hook both run the downloaded binary. Until this PR is on main, `test-oom-restart.sh` from a local build fails because main has no `internal record-exit`, which is unrelated to the code under test. The README's "a change that hasn't been pushed anywhere can still be tested" only holds for the install side today.

Simplest fix is probably to re-copy `/miren-local/miren` over `$MIREN_RELEASE_DIR/miren` right after this install call, so the bundle supplies containerd et al and the local build supplies `miren`.

cli/commands/runner_install.go
```diff
@@ -139,12 +139,15 @@ WantedBy=multi-user.target
...
 		ctx.Completed("Service file created at %s", runnerServicePath)
+	}
 
-		ctx.Info("Reloading systemd daemon...")
-		cmd := exec.Command("systemctl", "daemon-reload")
-		if output, err := cmd.CombinedOutput(); err != nil {
-			return fmt.Errorf("failed to reload systemd: %w\nOutput: %s", err, output)
-		}
+	// Outside the guard above on purpose — see installServiceLimits.
+	installServiceLimits(ctx, runnerServiceName, opts.DataPath)
```
phinze, line 152: 🤖 Non-blocking: the runner gets the limit and the hook, so `/var/lib/miren/runner/last-exit.json` is written on an abnormal exit, but nothing on the runner side reads it. The only `exitrecord.Read` callers are `cli/commands/server.go` and `components/server/boot_exit_report.go`, and `Record.LogTo` hardcodes "previous miren server run". Happy to see the cap land on the runner now; just flagging that the reporting half is server-only so it does not get mistaken for done. A short comment here or a follow-up ticket would cover it.

pkg/servicelimits/limits.go
```diff
@@ -0,0 +1,372 @@
...
+	// quarter of RAM would be a cap the coordinator could never realistically
+	// reach and so would not bound a leak in time.
+	memoryCapBytes = 16 * gib
+
+	// memoryHeadroomPercent bounds the limit against physical RAM, so a tiny
+	// machine that falls under memoryFloorBytes still leaves the rest of the
+	// system something to run in.
+	memoryHeadroomPercent = 90
+
+	// memoryHighPercent sets MemoryHigh relative to MemoryMax. Crossing
```
phinze: 🤖 Non-blocking design note. This comment and the troubleshooting doc ("often avoids the hard stop at MemoryMax entirely") lean on the GC responding to `MemoryHigh`, but nothing sets `GOMEMLIMIT` or calls `debug.SetMemoryLimit` in the coordinator (the only hit in the tree is etcd tuning). The kernel can reclaim page cache from containerd pulls, but it cannot shrink a Go heap, so between 85% and 100% the process mostly gets throttled rather than recovering. It still converges on the kill, and this is the standard systemd shape, so fine to ship as is. A follow-up where the server reads its own cgroup's `memory.high` at boot and calls `SetMemoryLimit` would make the grace period real.

docs/docs/changelog.md
```diff
@@ -11,6 +11,11 @@ All notable changes to Miren Runtime will be documented in this file.
 ## Unreleased
 *main*
 
+**Improvements**
+- **The server can no longer take the machine down with it** - A goroutine leak grew the control process until the host ran out of memory and stopped responding to anything, including SSH. The systemd unit now carries a memory limit, so the kernel stops and restarts the server instead of letting it exhaust the machine. Apps and addon databases run in their own cgroups outside the server's, so the limit doesn't apply to them and they keep running through a restart. The limit is a quarter of the machine's memory, floored at 2 GB and capped at 16 GB, and lands in a drop-in that is refreshed on every upgrade — so existing installs pick it up on their next `miren server upgrade` with nothing to do by hand. Raise it with `sudo systemctl edit miren` if your cluster genuinely needs more; an override there wins and survives upgrades.
```
phinze: 🤖 Nit: missing the `([#1210](https://github.com/mirendev/runtime/pull/1210))` link the neighboring entries carry.

phinze (approved):
🤖 Traced the cgroup claim that makes this safe (`/miren/sandbox-*` and `/<ns>/<id>` are absolute, so apps, addons, etcd, buildkit and the Victorias sit beside `system.slice`, not under the unit) and the install/upgrade/rollback paths that refresh the drop-in. The design holds up and the bot findings from earlier heads are all addressed. One real issue in the harness drive-by: the "test a local binary" path gets its binary replaced by the downloaded bundle during `server install`, so it doesn't test what it says it does. Rest is non-blocking.

hack/systemd/entrypoint.sh
evanphx: 🤖 Good catch, and verified: `promoteReleaseDir` renames the whole staging directory into place and `hack/package-release.sh:16` puts `bin/miren` in the bundle, so the copy made before `server install` is gone by the time the unit starts. The harness was installing with the local build and testing the downloaded one.

Took your suggested fix in 7d4593b2: the local binary is restored over the release directory right after the install call, so the bundle supplies containerd, runc and the shims while the local build supplies `miren`. The README claimed more than the code did, so it now spells out that the bundle is still downloaded and only `miren` comes from your build, including why the copy happens twice.

pkg/servicelimits/limits.go
evanphx: 🤖 Right on both counts, and worth being precise about what the band does buy. The kernel can reclaim page cache under `MemoryHigh` (containerd pulls, mmap'd files), but it cannot shrink a Go heap, so a leak in Go memory gets throttled between 85% and 100% rather than recovering, and still converges on the kill.

The comment and the troubleshooting doc both claimed the GC gets a chance to recover, which overstated it. Reworded both in 7d4593b2 to say the band buys a slowdown and a signal rather than a recovery, and to point at reading the cgroup's `memory.high` at boot and calling `debug.SetMemoryLimit` as the thing that would make the grace period real. Tracked as a follow-up rather than folded in here.

cli/commands/runner_install.go
evanphx: 🤖 Confirmed, and worth stating plainly: the runner is capped but silent. It writes `<data-path>/last-exit.json` on an abnormal exit and nothing reads it, since both `exitrecord.Read` callers are server-side and `LogTo` says "previous miren server run".

Added a comment at the call site in 7d4593b2 so it reads as half-done on purpose rather than finished, and filed a follow-up to do the reporting half.

Went with a comment over making `LogTo` unit-aware now because that message text is quoted verbatim in `troubleshooting.md`, the changelog entry, and `test-oom-restart.sh`'s grep — churning all of it for a path nothing reads yet seemed worse than a clear note. Happy to do it properly in this PR if you would rather it not ship asymmetric.

docs/docs/changelog.md
evanphx: 🤖 Fixed in 7d4593b2.

cli/commands/runner_install.go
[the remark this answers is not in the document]
evanphx: 🤖 Deliberately keeping this non-fatal. If the drop-in write fails, failing the whole install leaves the operator with no miren at all, which is a worse outcome than a working server without a memory cap.

We did take the point that it was too quiet, though — the warning now says explicitly that the service is unprotected and how to set a limit by hand, instead of a one-line "couldn't set resource limits" that reads as cosmetic.

The unknown-RAM half of your comment is fixed properly: `Write` now keeps an existing drop-in rather than replacing it with one that has no memory directives, and reports whether it wrote so the caller can say which happened.

hack/systemd/test-oom-restart.sh
```diff
@@ -0,0 +1,139 @@
...
+start_time=$(date '+%Y-%m-%d %H:%M:%S')
+
+# The early report runs before the dependency graph starts, so it does not
+# depend on the server booting successfully. That makes it the one part of
+# reporting this script can assert on unconditionally, and a regression here is
+# a real failure rather than a quirk of running inside a container.
+warned=0
+for _ in $(seq 1 60); do
+    if journalctl -u miren --no-pager --since "$start_time" 2>/dev/null \
+        | grep -qi "killed for exceeding its memory limit"; then
```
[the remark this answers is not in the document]
evanphx: 🤖 Good catch on the mechanism, and reproduced it: piping a 5 MB stream with the match on line 1 into `grep -q` under `set -uo pipefail` returns 141 every time. A working feature would have read as broken.

The suggested fix does not hold, though. GNU grep applies the same early-exit optimisation when its output is `/dev/null`, so `grep -i ... >/dev/null` returns 141 in the same repro:

```
--- grep -q (old form) ---
141 141 141
--- grep >/dev/null (suggested) ---
141 141 141
```

Redirecting to a real file does consume the stream, but leaving a file behind to dodge a signal seemed like a poor trade.

Fixed in dccead30 by reading the journal into a variable and matching with a bash glob, which removes the pipe entirely so there is nothing to signal. Side benefit: the assertion is now case-sensitive against the exact message the server logs, slightly stricter than the `grep -i` it replaces.

The two other journalctl pipelines in the script are diagnostic output guarded with `|| true`, not assertions, so they are left as they are.

docs/docs/changelog.md
```diff
@@ -11,6 +11,15 @@ All notable changes to Miren Runtime will be documented in this file.
...
 *main*
 
+**Improvements**
+- **The server can no longer take the machine down with it** - A goroutine leak grew the control process until the host ran out of memory and stopped responding to anything, including SSH. The systemd unit now carries a memory limit, so the kernel stops and restarts the server instead of letting it exhaust the machine. Apps and addon databases run in their own cgroups outside the server's, so the limit doesn't apply to them and they keep running through a restart. The limit is a quarter of the machine's memory with a floor of 2 GB, and lands in a drop-in that is refreshed on every upgrade — so existing installs pick it up on their next `miren server upgrade` with nothing to do by hand. Raise it with `sudo systemctl edit miren` if your cluster genuinely needs more; an override there wins and survives upgrades. ([#1210](https://github.com/mirendev/runtime/pull/1210))
+
+  **If you already set a memory limit by hand**, move it to `sudo systemctl edit miren`. A limit written directly into `/etc/systemd/system/miren.service` is overridden by the managed drop-in, silently and possibly downward, because systemd applies drop-ins after the unit file.
```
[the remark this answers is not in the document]
evanphx: 🤖 Agreed on the rule, but moved it rather than converting it in place.

The changelog has no admonitions anywhere in its history, and it has a direct precedent for the bold form for this exact kind of note — the v0.3.1 entry uses `**For existing installations:**` for another systemd unit-file action. A single admonition in the middle of a chronological release list would read oddly and would put every future entry writer on the spot.

So in 13c92c5f the warning lives on the troubleshooting page, right after the `systemctl edit` instructions it qualifies, alongside the admonitions that page already uses. That is also where someone who hits the problem will actually look. The changelog keeps a one-line note and links to it.

PR #1215, opened 2026-09-12
Serve app usage from the metrics store, with time series and a per-sandbox breakdown

evanphx:
Asking "what has this app been using?" over anything longer than the current deployment didn't really work. Only `getSandbox` could return a time series at all; the app views collapsed the window into a single number, and they built their rows from the sandboxes that happened to be alive at that moment. A question about last week, for an app redeployed on Tuesday, could only ever cover since Tuesday.

The odd part is that the data was never missing. A dead sandbox's *entity* gets collected about an hour after it dies, but its samples sit in VictoriaMetrics for a month, and every one of them already carries `miren_app`, `miren_kind`, the service, the version and the node. The only thing in the way was that the queries were keyed on sandbox ids, which is precisely the thing that disappears.

The first commit moves app rows and totals onto `sum by (miren_app, miren_kind)` and adds a `getApp` RPC method that mirrors `getSandbox` and carries a series. The entity store still has a job: it supplies each app's id, and it keeps a row for an app that exists but has gone quiet, so a broken telemetry pipeline still looks broken instead of looking like an idle cluster. Two behaviour changes come with that, both deliberate. An app with nothing left running now appears for as long as the window reaches its samples (marked `historical`, and `(gone)` in the table), and the three count fields now mean "sandboxes that reported" rather than "sandboxes that were scheduled".

The second commit answers the obvious follow-up, which is which sandboxes actually made up that number. Over a week that's mostly sandboxes nobody can look up any more, so each one is reconstructed from its own metric labels, plus the span it was reporting for. That span is what makes a week's worth readable as a sequence of deployments rather than a pile of ids.

Worth calling out, because I got it wrong first: measuring contributors the way everything else here measures things reported zero for every sandbox over a week. Every collapse in this package evaluates a fixed twenty points across the window, so at seven days the step lands at over eight hours, and a sandbox that lived ten minutes falls between two evaluation points. It still appears in the breakdown, with zeroes beside it, which reads as "used nothing" rather than "not measured". Applied straight to the raw metric instead, `avg_over_time` and `increase` scan the samples themselves, so they're exact at any window length and describe the ten minutes the sandbox was actually up. That's also where `cpu_seconds` comes from. Cores is a rate, so it can't be compared between sandboxes that ran for different durations, but seconds add up and they sum to the app's consumption.

Since the dev cluster only ever holds minutes of history, I checked the week path by backfilling seven days of samples across three deployment generations. The series came back with 60 points spanning the week, and the per-sandbox cores and memory matched what I'd synthesized, with `cpu_seconds` summing exactly to the expected total.

On the cardinality question from the issue: it turns out to be a non-issue. There are two series per sandbox, total. The `instance` label that would have multiplied them is never actually written, because `Setup()` is never called on the sandbox collectors, so it stays empty and an empty label value doesn't exist as far as the store is concerned.

Two small things fell out along the way. App versions outlive the sandboxes that ran them, so the directory now holds on to the version map it was already loading and a departed sandbox can still name a version you can type. And `formatDuration` never rolls up past minutes (fine for a download ETA, less good for a week, which it rendered as `10080m0s`), so these views got a formatter whose unit follows the number.

```
miren top --apps --app shop --since 168h --series --contributors
```

Closes MIR-1829

### DIFF
--- a/servers/usage/apps.go
+++ b/servers/usage/apps.go
@@ -1,8 +1,10 @@
 package usage
 
 import (
+	"cmp"
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -29,18 +31,20 @@ type appListing struct {
 // different kind. Doing it here means "what is app X using" has one answer that
 // does not depend on the caller reconstructing Miren's model.
 func (s *Server) listApps(ctx context.Context, f filter, w window, ord ordering) (*appListing, error) {
-	// Every kind is loaded regardless of the addon setting: addons are needed
-	// either to be counted or to be reported as the zero they were excluded at.
-	// The app filter is applied per row rather than in the directory, since an
-	// addon's app comes from a different label than a service's.
+	// The directory is still loaded, but for a narrower job than it used to do.
+	// It no longer supplies the figures -- those come from the store, grouped
+	// by app -- only each app's entity id, and a row for an app that exists but
+	// has gone quiet. Every kind is loaded regardless of the addon setting:
+	// addons are needed either to be counted or to be reported as the zero they
+	// were excluded at.
 	dir, err := s.loadDirectory(ctx, filter{node: f.node, includeSystem: true})
 	if err != nil {
 		return nil, err
 	}
 
-	cpu, memory, warnings := s.appSamples(ctx, f.node, w, dir)
+	m, warnings := s.appSamples(ctx, f.node, w, dir)
 
-	rows, cluster := buildAppRows(dir, cpu, memory, f.app, f.includeAddons)
+	rows, cluster := buildAppRows(dir, m, f.app, f.includeAddons)
 
 	out := &appListing{cluster: cluster, warnings: warnings, total: int32(len(rows))}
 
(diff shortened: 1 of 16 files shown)


servers/usage/apps.go
```diff
@@ -257,20 +512,104 @@ func (s *Server) appSamples(
...
+				store(k, r.value)
+			}
+		}
 	}
 
-	if vals, err := s.instantByLabel(ctx, memoryBytesQuery(labelSandbox, selector, dur, w.aggregate), labelSandbox, w.end); err != nil {
-		warnings = append(warnings, "memory usage unavailable: "+err.Error())
-	} else {
-		memory = vals
+	collect("cpu usage", cpuCoresQuery(group, selector, dur, w.aggregate),
```
phinze, line 531: 🤖 Moving the sum into the query is the right call, and for windows of 3h or more it gives the right answer too. It turns out that is VictoriaMetrics covering for us rather than `rate` doing it. VM's `rate(m[d])` is `dv/dt` over the series' *own* samples, not over `d`. At ≥3h an internal rewrite turns `sum(rate(m[d]))` into `sum(increase(m[d])) / d`, which is window-weighted; below that there is no rewrite. Once two sandboxes' series are summed inside the query, a dead one keeps contributing its full while-alive rate for as long as the window still contains its samples. Per sandbox this never mattered, because a series cannot double itself.

I backfilled a week for a 1-core app across three sequential generations into the same VM image the dev env uses (v1.106.1) and ran the strings `cpuCoresQuery` produces here. The 7-day row came back 0.857 (6/7, correct). `--since 1h` thirty minutes after a redeploy came back **2.0**. The `--series` range path in `appSeries` never gets the rewrite, so the 7-day series has a 2.0 point at every deployment boundary, and `--aggregate max` evaluates the rate inside a subquery, so every deploy reads as a spike.

Spelling it as what VM's rewrite does, `sum by (...) (increase(cpu[d])) / <d in seconds>`, gave ≈1.0 at 1h, 2h and 7d, a flat series through both boundaries, and identical results to `rate` in steady state. It would also make the contributors' `cpu_seconds` and the row agree by construction, since they would be the same `increase`.

servers/usage/apps.go
```diff
@@ -257,20 +512,104 @@ func (s *Server) appSamples(
...
+		func(k appKey, v float64) { m.cpu[k] = v })
+
+	collect("memory usage", memoryBytesQuery(group, selector, dur, w.aggregate),
```
phinze, line 534: 🤖 Same shape as the CPU one above, different mechanism. Inside `avg_over_time(sum by (...)(memory)[7d:8h])` the bare selector's lookbehind defaults to the subquery step, so at a 7-day window a sandbox that died keeps its last sample "current" for up to 8h after. Four hours after one generation died, `sum(memory)` returned 200 MiB for a 100 MiB app, and the 7-day average landed at 117 MiB, which is exactly 19/17: two doubled evaluation points out of seventeen. An app redeployed ten times a week would have a dead sandbox inside roughly half its points.

Pinning the inner lookbehind, `sum by (...) (last_over_time(memory_usage_bytes{...}[15s]))` inside the subquery, brought it back to 100 MiB. That bounds the bleed to about a scrape interval at any window length, and `--aggregate max` stops seeing a spurious 2× at every deploy for the same reason.

servers/usage/apps.go
```diff
@@ -230,23 +476,32 @@ func buildAppRows(
...
+// number of sandboxes.
 func (s *Server) appSamples(
 	ctx context.Context,
 	node string,
 	w window,
 	dir *directory,
-) (cpu, memory map[string]float64, warnings []string) {
-	cpu = map[string]float64{}
-	memory = map[string]float64{}
+) (appMetrics, []string) {
```
phinze, line 506: 🤖 Non-blocking. With the row set now coming from the store, the selector fallback just below changes meaning. `loadDirectory` returns an empty directory for an unknown node, but `appSamples` re-resolves it, gets nil, and leaves the selector empty, so the three queries run cluster-wide and every app in the cluster comes back as a `(gone)` row. Before, an empty directory meant zero rows. `--apps --runner <typo>` listing the whole cluster as historical is a confusing answer to get. Returning early when a node was named and nothing matched would keep the "unknown node is an empty listing" behaviour the directory comment promises.

servers/usage/contributors.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	collect := func(what, query string, store func(*contributorTally, float64)) {
+		rows, err := s.instantRows(ctx, query, w.end)
+		if err != nil {
+			warnings = append(warnings, what+" unavailable: "+err.Error())
+			return
+		}
+		for _, r := range rows {
+			if id, ok := identityOf(r.labels); ok {
```
phinze: 🤖 Non-blocking. `--no-addons` takes addons out of the row and the series but not out of the breakdown: nothing in here looks at `includeAddons`, so the contributors table still lists the database sandboxes, and their `cpu_seconds` no longer sum to the row's figure, which is the property the `cpu_seconds` doc promises. Skipping `id.kind == addon` here when addons are excluded would keep the two in step.

phinze (approved):
🤖 This is the right design and a genuinely good write-up. Keying on the labels instead of the sandbox id is what makes a week-long question answerable at all; the `stale`/`historical` split gives an operator two honest answers instead of one ambiguous one; the contributors path sidesteps the 20-point subquery trap on purpose and explains why; and the tests pin the things that would otherwise regress silently (live-before-historical ordering, per-timestamp summing, seen-times never inventing a row). The comments throughout are the kind you can debug from.

Approving, with one thing I'd ask you to fix before merging rather than after. Moving the sum across sandboxes into the query exposes two VictoriaMetrics behaviours that were harmless per sandbox: `rate` is `dv/dt` over a series' own samples, and a bare selector inside a subquery inherits the subquery step as its lookbehind. Both double count at a deployment boundary, which is exactly the situation this PR sets out to describe. I ran the package's own query strings against a backfilled week in the dev VM image and got 2.0 cores for a 1-core app on `--since 1h` after a redeploy, and 117 MiB for a 100 MiB app over 7 days. The fixes are small and spelled out inline. Two further notes on the unknown-node fallback and `--no-addons` with `--contributors`, take or leave.

servers/usage/apps.go
evanphx: 🤖 Confirmed and fixed. Reproduced your exact result: backfilled a steady 1-core app that redeployed 30 minutes ago, and `sum(rate())` answered 2.000 at both 1h and 2h, 1.000 at 7d. The 3h rewrite boundary explains why my own verification passed, since I only ever tested a week-long window.

Switched to `sum by (...) (increase(cpu[d])) / <d>` as you suggested. Every shape `cpuCoresQuery` emits now returns ~1.000 across 1h/2h/24h/7d on both avg and max, and the `--series` range path is flat through the deployment boundary (min 0.903, max 1.000) where it previously spiked to 2.0.

Worth noting this also fixes `nodes.go:280`, which groups by node and so sums across sandboxes the same way. That one predates this branch.

servers/usage/apps.go
evanphx: 🤖 Confirmed and fixed. It took getting the timing right to see it: with the death 30 minutes back the 7d grid happened to miss it, but a 1-day window with `--aggregate max` gave exactly your 200 MiB for a 100 MiB app.

Pinned the inner lookbehind as you suggested. I went with 60s rather than 15s: the writer flushes every 5s (`defaultMetricFlushPeriod`), so 60s is twelve flushes of slack, which keeps a slow-batching runner from punching holes in the long-window series while still bounding the overlap to about one scrape. Now reads exactly 100 MiB at every window and aggregate.

servers/usage/apps.go
evanphx: 🤖 Good catch, fixed. The root cause was that the node got resolved twice, once in `loadDirectory` and again in `appSamples`, and only the first one treated "no match" as meaningful. Collapsed it into a single `nodeSelector` that returns an `ok` bool, so an unmatched node cannot silently read as unconstrained. A mistyped `--runner` now gives "No apps are reporting usage." again.

servers/usage/contributors.go
evanphx: 🤖 Fixed, thanks. Skipping addon-kind contributors when addons are excluded. The two now agree by construction: on the check I ran, the contributors' `cpu_seconds` summed to 86220 and the row's cores implied exactly 86220 over the same window.

evanphx:
🤖 Pushed `f0f135a0` addressing all four. Two were real wrong answers and both are worth the summary: `sum(rate())` reported 2.0 cores for a 1-core app at short windows, and memory read 200 MiB for a 100 MiB app on a 1-day max. Both reproduced first, then re-verified across 1h/2h/24h/7d on avg and max.

The CPU one is a good lesson on my part: I verified this feature against a week-long window, which is exactly the range where VictoriaMetrics' rewrite hides the bug. Appended rather than force-pushed so your review state stays intact.

PR #1217, opened 2026-09-14
Snapshot etcd before an upgrade and restore it on rollback

phinze:
Rollback in the lifecycle executor (#1216) puts the previous binary back but not the data it ran against. A new build that migrates etcd on boot and then fails readiness leaves the old build serving migrated data, which is the one outcome a rollback is supposed to rule out. This is the etcd-only primitive; the full RFD-75 bundle stays with MIR-969.

Upgrades get a `backing_up` phase that snapshots etcd (after the download rather than before, so the snapshot is as fresh as possible and a failed download costs nothing) and records the reference on the operation as `backup_ref`. The executor never learns what the data is: it calls a `DataBackup` and, on rollback, writes a restore request onto the operation and restarts. A new `data-restore` component early in the boot graph, between containerd and etcd, performs the restore before etcd opens the data directory and answers in a result file beside the record. The restore itself runs `etcdutl snapshot restore` as a one-shot task from the same etcd image the server uses, so the data directory is built by the version that reads it, rather than pulling the etcd server module into the binary and betting on cross-version formats.

One premise from the design didn't survive contact: embedded etcd only outlives the server on an unclean death. A graceful `systemctl restart` stops it through the boot graph. The boot-graph placement still holds, and is in fact the only spot that handles both states, since the executor cannot tell which it will get.

A restore that fails keeps the server down rather than starting on the data the rollback was meant to replace. The request stays pending across the executor's timeout so the next boot retries, the operation record carries the restore error, and `miren server operations abandon <id>` is the operator's exit, which also covers an executor that died mid-operation. Snapshots are kept three deep under the lifecycle directory, and `operations show` surfaces the backup and the restore outcome.

The systemd e2e's forced-rollback case now proves the data comes back: the broken fixture boots etcd on the data directory, writes, and fails, and the harness checks the write is gone afterwards. Two harness leaks got fixed on the way (an asset server that outlived its EXIT trap, and a port collision that silently served stale fixtures).

Closes MIR-1843


pkg/serverlifecycle/executor.go
```diff
@@ -253,19 +294,115 @@ func (e *Executor) rollback(ctx context.Context, op *Operation) error {
 			return e.fail(ctx, op, fmt.Errorf("%s; rollback failed: %w", op.Error, err))
 		}
 	}
+	if op.BackupRef != "" && op.DataRestore == nil {
```
evanphx: :robot: The old binary becomes launchable when `installer.Rollback` returns, before this restore request is persisted. Since `miren.service` has `Restart=always`, a failed upgraded service can be started automatically in that window and open migrated data with the old binary. Can rollback stop/inhibit autonomous starts around this sequence, or establish the durable restore barrier before replacing the executable?

pkg/serverlifecycle/store.go
```diff
@@ -126,12 +126,103 @@ func (s *Store) Active() (*Operation, error) {
...
+// PendingRestore returns the operation whose rollback asked the server to
+// restore data and has not been answered, or nil. The server calls this
+// early in boot, before the data it would restore is in use.
+//
+// The request outlives its operation on purpose. An executor that waited on
+// a server refusing to boot marks the operation failed and exits, and if
+// that ended the request the very next boot would start on the data the
+// rollback was meant to replace. Only a restore or Abandon settles it.
+func (s *Store) PendingRestore() (*Operation, error) {
+	ops, err := s.List()
```
evanphx: :robot: `List` deliberately skips operation records it cannot decode, so `PendingRestore` can return nil when the corrupted record is precisely the one carrying the restore request. That contradicts the boot path’s fail-closed contract. Could this path enumerate/decode operation files strictly and propagate any read or decode error?

pkg/serverlifecycle/store.go
```diff
@@ -154,7 +245,7 @@ func (s *Store) write(op *Operation) error {
 		os.Remove(tmpName)
 		return err
 	}
-	if err := os.Rename(tmpName, s.path(op.ID)); err != nil {
+	if err := os.Rename(tmpName, path); err != nil {
```
evanphx, line 265: :robot: This file is acting as a durable restore barrier, but syncing the temporary file does not make the subsequent rename durable across power loss. Please fsync the containing directory after the rename; the same applies to restore-result writes through this helper.

evanphx (changes requested):
The snapshot and pre-etcd restore architecture looks right, but there are gaps in the durable restore barrier and fail-closed behavior. In particular, an automatic systemd restart can expose the rolled-back binary before the request is recorded, and a malformed operation can currently be mistaken for no pending restore.

— e + :robot:

pkg/serverlifecycle/executor.go
phinze: 🤖 Good catch, and the window is real. The request is now persisted before `installer.Rollback`, so a `Restart=always` launch of the old binary finds it. That opened the mirror-image hole: the build being rolled back *from* can still be crash looping in the same window, and if it honored the request it would restore, migrate the data again, fail, and leave a settled request behind for the old build. So the request also names the build it is for (`for_version`/`for_commit`, from the operation's previous-build fields), and the boot component leaves a request meant for another build alone. We considered inhibiting systemd (stop, swap, start) instead, but the version gate is needed anyway for the resume-after-executor-death case, and with it the ordering fix closes the window without touching the unit.

pkg/serverlifecycle/store.go
phinze: 🤖 Agreed. `PendingRestore` now reads through a strict variant that returns the first read or decode error, and the boot component fails closed on it. `List` keeps skipping so `operations list` still shows the rest; test pins both.

pkg/serverlifecycle/store.go
phinze: 🤖 Done: `writeFileAtomic` fsyncs the containing directory after the rename, which covers operation records and restore results alike.

evanphx (approved):
The requested changes are addressed. The restore barrier now precedes binary replacement and is scoped to the intended build, restore discovery fails closed on malformed records, and atomic record writes sync the containing directory. The focused tests cover the important ordering and recovery cases.

— e + :robot:

PR #1222, opened 2026-09-14
Render server operations list/show with pkg/ui

phinze:
`miren server operations list` was the last command still rendering through `ctx.DisplayTable`, a pre-`pkg/ui` lipgloss table with its own styles and column sizing, so it looked subtly different from its 22 neighbors and would miss any theme or width fix made in `pkg/ui`. `operations show` hand-padded its labels with `Printf`. Both now go through `pkg/ui` with the same fields in the same order, phase colored by outcome, and `--json` byte-for-byte unchanged. `DisplayTable` and the never-called `DisplayTableTemplate` go with them.

Moving `show` onto `NamedValueList` surfaced that the list painted every string value green, so a failed phase looked exactly like the id above it. The first rev flips that default for everyone: labels bold, strings plain, typed values keep their colors, and a new `ui.NewStyledValue` colors one value on purpose. That changes the look of the dozen existing `NamedValueList` callers (`auth provider show`, `auth ci list`, `route protect`, ...) from green-everything to bold-label/plain-value, which is worth a glance if you have opinions about those.

Checked against a live cluster and a fake `--dir` ledger covering failed, rolled-back and in-flight records.

A review pass turned up that protecting most of the list's columns could defeat the squeeze entirely on a narrow terminal, so the middle rev bounds `AutoSizeColumns` when the protected columns already fill the width, fixes the floor bump that left tables a few columns wider than the terminal, and the list now protects only ID, PHASE and STARTED.

Closes MIR-1850


cli/commands/server_operations.go
```diff
@@ -199,14 +202,22 @@ func ServerOperationsList(ctx *Context, opts struct {
...
-			op.ID, string(op.Action), string(op.Phase), operationVersions(op),
+		rows = append(rows, ui.Row{
+			op.ID, string(op.Action), phaseStyle(op.Phase).Render(string(op.Phase)), operationVersions(op),
 			op.CreatedAt.Local().Format("2006-01-02 15:04:05"), op.Error,
 		})
 	}
-	ctx.DisplayTable([]string{"ID", "ACTION", "PHASE", "VERSIONS", "STARTED", "ERROR"}, rows)
+	// A long error must not squeeze the identifying columns; it is the one
+	// column that can lose its tail, since 'show' has the full text.
+	columns := ui.AutoSizeColumns(headers, rows, ui.Columns().NoTruncate(0, 1, 2, 3, 4))
```
evanphx: :robot: Protecting columns 0–4 can make their combined minimum width exceed a normal terminal before ERROR receives any space. In that case `AutoSizeColumns` skips scaling and the table overflows instead of truncating ERROR as intended. Please allow at least VERSIONS to truncate, or make the sizing logic collapse scalable columns when protected columns consume the available width, and add a narrow-terminal regression test.

evanphx (changes requested):
Moving these commands to `pkg/ui` preserves the JSON paths and improves consistency. However, protecting the first five list columns defeats the stated narrow-terminal behavior and can cause the full table to overflow. Please correct the sizing policy and cover the narrow-width case.

— e + :robot:

evanphx (approved):
Approving so this can merge once the narrow-terminal sizing issue is addressed.

— e + :robot:

cli/commands/server_operations.go
phinze: 🤖 Good catch, and it turned into a bit more than the two options. The list now protects only ID (what gets pasted into `show`), PHASE (a styled cell renders whole, so honoring its width keeps the row aligned) and STARTED (a fixed-width timestamp); VERSIONS and ERROR truncate.

On the `pkg/ui` side, `AutoSizeColumns` had two problems on a narrow terminal. It skipped the squeeze entirely when the protected columns already filled the width, which is the unbounded overflow you spotted. And even when it did squeeze, it bumped short columns up to the 10-char floor after handing out the proportional shares, so a table with one long column came out a few columns too wide, which renders as a blank line under every row. The squeeze now pins a column that would fall below its floor (or its natural width) and re-shares the remaining budget among the rest, and when the protected set alone exceeds the terminal everything else sits on its floor. That's the middle rev (`ui: share the squeeze so floors cannot push a table past the terminal`), with the math pulled into `sizeColumns(…, termWidth)` so it's testable without a TTY. `TestSizeColumnsNarrowTerminal` covers this table's shape at 120, 100 and 40 columns; the 40-column case came out 325 wide on the old code.

PR #1224, opened 2026-09-14
Reclaim images left behind by deleted apps

phinze:
Deleting an app took its artifact entities with it but left the containerd images behind, and the image GC kept any image it couldn't find an artifact for. So those images were never collected by any sweep. On a 40GB server this showed up as the GC firing on disk pressure every hour and finding nothing to delete, while fifteen orphans from a renamed app and a removed one sat there until a manual `ctr images rm`.

Two revs. The first makes the image GC treat a miren-managed image with no artifact as collectable after a 24h grace period, which matches what the blob GC already does and clears orphans already sitting on live servers. The second stops app delete from removing artifacts at all and leaves them to the artifact GC, which archives whatever no version references. That also fixes a latent bug: artifacts are deduped by manifest digest, so two apps can share one, and deleting the first app used to break the second's next pull.

Closes MIR-1851


controllers/sandbox/watchdogtest/image_test.go
```diff
@@ -321,6 +321,70 @@ func TestImageWatchdog(t *testing.T) {
...
+		watchdog := &sandbox.ImageWatchdog{
+			Log:       slog.Default(),
+			CC:        cc,
+			EAC:       eac,
+			Namespace: ii.Namespace,
+			DataPath:  "/tmp",
+			Config:    cfg,
+		}
+
+		result, err := watchdog.RunGC(ctx)
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, fixed in the latest push. containerd stamps CreatedAt itself so we can't back-date the image; the test now uses a 100ms grace and sleeps past it, so the age comparison is actually exercised.

controllers/sandbox/image_gc.go
```diff
@@ -230,22 +240,24 @@ func (w *ImageWatchdog) RunGC(ctx context.Context) (*ImageGCResult, error) {
...
-		if !found {
-			// No artifact entity found - keep the image (safe default)
-			result.RetainedImages++
-			continue
-		}
-
-		// Only delete if artifact is explicitly archived
-		if status != core_v1alpha.ARCHIVED {
+		switch {
+		case !found:
```
evanphx, line 271: :robot: A missing Artifact is not sufficient proof that this image is unreferenced. Existing installations may have a surviving AppVersion whose shared Artifact was deleted with another app. After upgrading, this branch can delete that version’s image whenever no sandbox is running. Please also retain images referenced by surviving AppVersions; deletion should require both Artifact and version ownership checks to report the image unreferenced.

servers/app/delete.go
```diff
@@ -75,6 +82,9 @@ func DeleteAppTransitive(ctx context.Context, client *entityserver.Client, log *
 		}
 
 		attrId := entity.Id(schema.Id())
+		if attrId == core_v1alpha.ArtifactAppId {
```
evanphx, line 85: :robot: Retaining Artifacts introduces a resurrection path that the registry does not currently handle. `putManifest` can deduplicate against an `ARCHIVED` Artifact without reactivating it, after which image and blob GC may delete content referenced by the new AppVersion. Please make digest reuse atomically reactivate the Artifact, or create a new active Artifact for the new lifecycle.

evanphx (changes requested):
Retaining deduplicated Artifacts rather than deleting them with their original app is the right ownership direction. Two destructive lifecycle gaps remain: historical missing Artifacts can still be referenced by surviving AppVersions, and archived Artifacts can be reused without reactivation. Both can allow GC to remove content needed by a durable version.

— e + :robot:

evanphx (approved):
Approving so this can merge once the AppVersion reference and archived-Artifact reuse issues are addressed.

— e + :robot:

controllers/sandbox/image_gc.go
phinze: 🤖 Agreed, and the upgrade case is exactly where the old shape bites. Added a rev that collects the images and artifact IDs every AppVersion references (by `artifact` and by `image_url`) and retains those before the artifact status is consulted at all, so deletion needs both the artifact and the versions to disown the image. Checking versions regardless of status also covers a push reusing an artifact in the same instant the artifact GC archives it. Test builds the dangling case the way it actually happens: artifact created, version pointed at it, artifact deleted underneath.

servers/app/delete.go
phinze: 🤖 Good catch. Version retention could already produce this, but leaving artifacts behind on delete makes it the normal path for delete-then-redeploy, so it belongs in this PR. `putManifest` now reactivates an archived match before answering 201, with a revision-checked patch that re-reads on conflict since the competitor is the artifact GC archiving the same artifact. If it can't land, the push fails rather than minting a version whose image is about to vanish. I went with reactivation over a fresh artifact so the existing image name and blobs stay valid for the new version.

controllers/sandbox/image_gc.go
```diff
@@ -25,6 +25,9 @@ type ImageGCConfig struct {
 	PressureCheckInterval time.Duration
 	// DiskPressureThreshold is the disk usage percentage that triggers immediate GC (default: 80%)
 	DiskPressureThreshold float64
+	// OrphanGracePeriod is how old a miren-managed image must be before it is
+	// reclaimed when no Artifact entity exists for it (default: 24h)
+	OrphanGracePeriod time.Duration
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push: a zero or negative `OrphanGracePeriod` now normalizes to the 24h default inside `RunGC`, so a partial config can't switch the grace off.

controllers/sandbox/image_gc.go
```diff
@@ -230,22 +247,35 @@ func (w *ImageWatchdog) RunGC(ctx context.Context) (*ImageGCResult, error) {
...
+		// moment. Neither is a reason to pull the image out from under a
+		// deploy or rollback.
+		if versionRefs.images[imgName] || versionRefs.artifacts[artifactID] {
 			result.RetainedImages++
 			continue
 		}
 
-		// Only delete if artifact is explicitly archived
-		if status != core_v1alpha.ARCHIVED {
+		status, found := artifactStatuses[artifactID]
```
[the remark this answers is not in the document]
phinze: 🤖 Skipping this one deliberately. The containerd image is a cache of what the registry serves, so if a sweep working from a stale snapshot deletes it under a version created mid-sweep, the next sandbox start re-pulls it. The durable state is what has to stay consistent, and that's covered: reactivation in `putManifest` is revision-checked, and blob GC takes its own fresh snapshot after image GC. A revalidate-then-delete would still race without a transaction spanning etcd and containerd, and the same window already existed on the archived path before this PR. Worst case here is one extra pull.

PR #1227, opened 2026-09-15
Make MockStore.WatchIndex replay from a revision like etcd

phinze:
The appmetrics target discovery test failed on an unrelated PR with its middle `Eventually` timing out at 1s. It looked like a timing flake, but no timeout would have saved it. `indexwatch.Watcher` lists at revision R and then opens `WatchIndex` from R+1, trusting the store to replay whatever landed in between. `MockStore` recorded that `fromRev` and ignored it, so when the test's write beat the watcher's registration, the event was gone for good. A 50ms delay in the mock's registration path reproduces the CI failure exactly.

The mock already documented this gap and told tests to call `WaitForIndexWatcher` first. This PR closes it instead: `MockStore` now keeps an etcd-style store-wide revision and an ordered log of index events, and `WatchIndex` replays from `fromRev` and registers the live channel under one lock. That also means events carry real `CreateRevision`/`ModRevision`, so the entity server's create-vs-modify classification is finally honest on the mock (it used to report every put as a create).

The visible change for other tests is the revision scheme: entities are stamped with the global revision rather than counting from 1 per entity, matching etcd. Nothing in the suite depended on the old numbers. Two new store conformance cases pin the replay and classification contracts against both backends, and the build saga cancel test drops its now-redundant `WaitForIndexWatcher` guard.

Closes MIR-1863


pkg/entity/mock.go
```diff
@@ -82,12 +112,18 @@ func (m *MockStore) GetEntityAtRevision(ctx context.Context, id Id, rev int64) (
...
-// AddEntity is a thread-safe helper to directly add an entity to the mock store
+// AddEntity is a thread-safe helper to directly add an entity to the mock store.
+// It is fixture setup rather than a write: no revision is assigned and no watch
+// event is produced. A fixture that carries its own revision moves the store
+// head up to it so a watch resumed from that head starts past the fixture.
 func (m *MockStore) AddEntity(id Id, entity *Entity) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	entity.Fixup()
 	m.Entities[id] = entity
```
evanphx, line 123: :robot: `AddEntity` makes these index entries visible in the initial snapshot but does not seed `indexEntryCreated`. When a fixture-backed entity is later updated, `commitLocked` treats every existing index entry as newly created (`CreateRevision == ModRevision`), so the entity server emits Create rather than Update—the classification this PR intends to fix. Please consider seeding the current index entries here, or deriving existing membership from `prevEntity`, and adding a fixture-then-update regression test. This is non-blocking since it is limited to mock fixture behavior.

evanphx (approved):
The revision replay mechanism and lock ordering correctly close the list/watch gap, and the conformance coverage exercises the production contract against both backends. `AddEntity` leaves one mock-only classification edge case, noted inline, but that is non-blocking for this test-infrastructure fix. Approve.

— e + :robot:

pkg/entity/mock.go
phinze: 🤖 Good catch, that was a real hole. Fixed in `526f3ad9`: `AddEntity` now records the fixture's index entries at its own revision (and `RemoveEntity` drops them), so the first write against a fixture reads as a modify. Went with seeding rather than deriving from `prevEntity` so the classification has one source of truth. `TestMockStoreFixtureUpdateReadsAsModify` pins the fixture-then-update path; I checked it fails without the seed.

PR #1228, opened 2026-09-15
Recover from leaked containerd task records on restart, and stop force-killing the dev server so eagerly

phinze:
`TestPOP` flaked in the pop job with the replacement server dying on `task miren-buildkit: already exists` right after `hack/dev-server stop` had escalated to `kill -9`. Reproducing it locally turned up something worse than a timing race: kill miren mid-`task.Delete` and containerd keeps a task record that it reports as NotFound on lookup but AlreadyExists on create, forever. The server's existing "evict the old task, then create" logic can't see it, so every subsequent start fails the same way until something recreates the container.

The server side adds `base.ReplaceTask` and routes buildkit, etcd, victoriametrics and victorialogs through it. When lookup and create disagree like that, it reaps the record by container ID via the raw task Delete RPC (the only path that reaches it) and creates again. This is the part that also helps a production host after a hard crash, since that is the same state a crashed miren leaves behind. Killing the dev server at various points mid-shutdown reproduced the leak for both buildkit and etcd, and every restart recovered.

The harness side stops manufacturing the problem: 60s of graceful shutdown before SIGKILL instead of 5s, the log tail printed when it does escalate, and the previous server's log rotated to `.prev` rather than truncated so a bad restart keeps its evidence. CI dumps that file on failure.

One thing deliberately left alone: a SIGKILLed miren orphans its embedded containerd, which keeps the socket and the bolt lock, so every later miren's "new" containerd blocks on that lock while miren dials the old one. That is why the leak never self-heals here. Fixing it means deciding what containerd's lifecycle should be across a miren restart, which belongs with the per-component graceful restart work rather than a flake fix.

Verified with the unit tests for the helper, the containerd-backed component tests under iso, and both restart phases of `TestPOP` going through the new harness cleanly. `TestPOP` itself fails on my machine at the POP forwarding step on trunk too, so the pop job here is the real check for that.

Closes MIR-1856


components/base/replace_task.go
```diff
@@ -0,0 +1,101 @@
...
+		log.Warn("containerd holds a leaked "+componentName+" task record (lookup says not found, create says already exists); reaping it", "attempt", attempt, "error", err)
+		if err := reap(ctx, container.ID()); err != nil {
+			return nil, fmt.Errorf("reap leaked %s task: %w", componentName, err)
+		}
+	}
+}
+
+// ReapLeakedTask asks containerd to delete a task by container ID without first
+// loading it. The loaded-task path (container.Task followed by task.Delete) is
+// closed to a leaked record because the lookup itself fails, but the delete RPC
```
evanphx, line 101: :robot: After `NewTask` returns `AlreadyExists`, consider rechecking `container.Task()` before issuing the raw delete. Another server or process could create a legitimate task between the preceding lookup and create; the current path would reap that fresh task based on the stale `NotFound`. Startup appears singleton today, so this is defensive hardening rather than blocking.

hack/dev-server
```diff
@@ -18,6 +18,43 @@ EXTRA_FLAGS="${DEV_SERVER_FLAGS:-}"
...
+}
+
+# force_kill PID: SIGKILL a server that ignored its shutdown window, and show
+# what it was doing so the escalation leaves evidence behind.
+force_kill() {
+  local pid=$1
+  echo "Warning: server did not exit within ${STOP_TIMEOUT}s, forcing termination (containerd tasks may be orphaned)"
+  echo "Last log lines before forced kill:"
+  tail -n 20 "$LOGFILE" 2>/dev/null | sed 's/^/  | /'
+  kill -9 "$pid" 2>/dev/null || true
```
evanphx, line 65: :robot: Could we verify process identity immediately before SIGKILL, ideally using `/proc/$pid/stat` start time captured at launch? There is a TOCTOU window between `wait_for_exit`’s final probe and this kill where PID reuse could target an unrelated process.

hack/dev-server
```diff
@@ -18,6 +18,43 @@ EXTRA_FLAGS="${DEV_SERVER_FLAGS:-}"
...
+  sleep 0.5
```
evanphx: :robot: A fixed 500ms sleep does not prove the process exited. If SIGKILL leaves it in uninterruptible sleep, `stop` removes the PID file and `restart` launches another server anyway. Consider waiting for disappearance with a bounded timeout and failing instead of reporting “Server stopped” while the PID remains.

evanphx (approved):
The leaked-task recovery is bounded and well tested, and the shutdown change materially reduces eager force-killing. The inline notes are defensive hardening under the current singleton development model and are non-blocking. Approve.

— e + :robot:

components/base/replace_task.go
phinze: 🤖 Agreed, done in 0052cfb. On `AlreadyExists` we now look the task up once more and, if one has appeared, send it through the graceful eviction path instead of the raw delete; the reap only fires when the recheck still says NotFound. Unit test covers the appeared-between-lookup-and-create case and asserts the reaper is never called.

hack/dev-server
phinze: 🤖 Took this in a lighter form: before signalling, `stop` and `restart` check `/proc/PID/cmdline` is actually a `miren server`, and refuse (keeping the PID file) if it isn't. That covers the realistic case, a stale PID file whose number got reused, without tracking start times across invocations. If a reused PID landed on *another* miren server we'd still be fooled, but at that point the PID file is wrong in a way no identity check on this side fixes.

hack/dev-server
phinze: 🤖 Yep. `force_kill` now waits up to 10s for the PID to disappear and returns failure if it's still there; `stop`/`restart` then exit 1 and keep the PID file instead of reporting stopped and starting another server on top.

PR #1229, opened 2026-09-16
Stop polling cluster status once cloud negotiates its replacements

phinze:
The runtime POSTs a status report to cloud every five minutes. mirendev/cloud#266 gave everything in it a home on the negotiated uplink session; this is the runtime half, so a supported runtime no longer needs the poll and cloud can retire the endpoint once the compatibility window closes.

Two new capabilities carry what the session did not already say. `cluster-network` sends how the cluster can be reached (advertised addresses, CA fingerprint, netcheck's verdict, containerized) on session start, whenever the facts change, and at least hourly. `cluster-resources` sends whole-host CPU, memory, and storage at the cadence cloud asks for. Both are built from the same code the poll uses, so the two sources cannot disagree while both are live. The welcome now carries the workload identity anchor the poll's response used to, and the runtime records it from there.

The compatibility switch is negotiation itself: when a session selects both capabilities the periodic poll goes quiet for as long as that session holds, and a cloud that declines either one gets the poll exactly as before. No version checks on either side. The startup report and the poll code stay until the window closes; the report just stops carrying fields cloud never read.

Node export through entity sync is deliberately not here. Cloud declines entity sync outright on a contract digest it does not know, so nodes land together with cloud's projection in the next PR.

Part of MIR-1861


evanphx (approved):
The capability boundary, reconnect ownership, and fallback behavior look sound. Negotiated suppression preserves legacy polling for older Cloud versions, and the focused tests and CI checks pass. Approve.

— e + :robot:

PR #1232, opened 2026-09-16
MIR-1870 Reattach component task logs on adoption

evanphx:
## Summary

- reattach slog consumers when adopting existing etcd, VictoriaMetrics, and VictoriaLogs tasks
- add a reusable `slogout.AttachLogger` adapter for existing containerd FIFOs
- bound the etcd maintenance status call and warn when its health check times out
- cover stdout and stderr draining beyond pipe capacity with a regression test

## Verification

- `hack/it ./pkg/slogout`
- `hack/it ./components/etcd`
- `hack/it ./components/victoriametrics`
- `hack/it ./components/victorialogs`
- `golangci-lint run ./...`
- `make test-serial` ran 7,849 tests; its only failures were 19 pre-existing `pkg/perf` tracepoint tests because tracefs is unavailable in the orb

Linear: MIR-1870


phinze (approved):
🤖 Traced the adopt paths end to end: the attach sits above both etcd branches, its options match the create path, `cio.NewAttach` runs the same `copyIO` loop as the creator against the FIFO paths from the task GET, and the other `Task(ctx, nil)` sites in `components/` are all stop/delete paths (buildkit and appmetrics always evict and recreate), so etcd, VictoriaMetrics, and VictoriaLogs are the complete set. The maintenance timeout's real value is that the loop keeps ticking and warning instead of going silent the way it did at 09:03:58; startup was already bounded by `waitForHealthy`.

Biscuit's `orig-level` finding is real but predates this PR (the create path has clamped since `etcd.go:194` existed), so it's filed as MIR-1871 rather than taken here.

PR #1233, opened 2026-09-16
MIR-1869 Reject valueless stream frames before they reach any callback

evanphx:
## Summary

Prod crashed eight times since the Sep 11 deploy on a nil `*EntityOp` inside the index watcher's `WatchIndex` callback (MIR-1869). The nil is not specific to the watcher. `NetworkCall.Args` drops its decode error, so when a call stream ends between the call header and its args frame, every stream callback receives the zero value of `T`. Every callback in the tree dereferences a pointer on its first line, so both activator watches, computex, the entity client, and the CLI log tails carried the same latent panic.

The byte-stream writer adapter had the quieter version of the same bug: a torn frame became a successful zero-byte write.

This PR checks for the missing value once, in the three `SendStream` adapters that hand frames to consumers (`stream.Callback`, `stream.StreamRecv`, and `stream.ServeWriter`), and reports it as `cond.Closed`. A frame with no value never comes from a sender, since `SendStreamClient.Send` always populates it and the byte writer only sends non-empty chunks, so the only reading is that the stream tore. The inline router already treats `ErrClosed` as a quiet stream end and the watcher already treats a nil error from `WatchIndex` as a clean end to resume from, so no call site changes.

An earlier revision of this branch guarded the watcher's callback directly; that guard is superseded and dropped so the fix lives in one place.

## Verification

- `go test ./pkg/rpc/stream ./pkg/entity/indexwatch`
- `golangci-lint run ./pkg/rpc/stream ./pkg/entity/indexwatch`
- `TestWatcher_TornFrameDoesNotPanic` reproduces the incident through the real dispatch path: a wrapped entity server hands the callback a valueless frame on its first `WatchIndex`. Without the guard it panics on the same `OperationType` line as the production stack (`watcher.go:379`); with it the watcher logs a clean end and resumes.
- The iso `make test` sweep has not been run locally; CI covers it.

Linear: https://linear.app/miren/issue/MIR-1869/nil-entityop-delivered-to-indexwatch-watchindex-callback-panics-the


phinze (approved):
🤖 Took the pen on this one, as discussed. The branch is now a single commit that moves the guard from the watcher callback into the three `SendStream` adapters (`Callback`, `StreamRecv`, `ServeWriter`), since `NetworkCall.Args` dropping its decode error means every stream callback in the tree had the same latent nil, and the byte writer was recording a torn frame as a successful zero-byte write. The watcher is untouched from `main`; `TestWatcher_TornFrameDoesNotPanic` reproduces the incident through the real dispatch path and dies on `watcher.go:379` without the guard. CodeRabbit's `serveWriter` note is folded in. Merging on Evan's behalf while he's out.

PR #1234, opened 2026-09-16
MIR-881: Add declarative deploy targets

evanphx:
## Summary
- add `.miren/deploy.toml` targets with default and named target selection
- resolve targets by stable Miren Cloud cluster ID across user-specific local aliases
- add `miren deploy target` commands to add, list, remove, and choose the default target without hand-editing IDs
- document the workflow and generate command reference pages

## Verification
- `go test -p 1 ./appconfig ./cli/commands`
- `make lint`
- `cd docs && bun run lint && bun run build`
- containerized package tests: 1,101 passed

Linear: https://linear.app/miren/issue/MIR-881/declarative-deploy-targets


cli/commands/deploy_target.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+
+	"miren.dev/runtime/appconfig"
+	"miren.dev/runtime/clientconfig"
+	"miren.dev/runtime/pkg/ui"
+)
+
+type deployTargetAddOpts struct {
+	AppCentric
+	Name        string `position:"0" usage:"Name for the deployment target" required:"true"`
```
phinze, line 14: I think I'd expect bare `miren deploy target add` to prompt for the name and then drop into the cluster picker, so the whole command follows the "omit an arg, get prompted for it" convention rather than half of it. That is already how `env set`, `secret set`, and the login identity-name flow behave via `ui.PromptForInput`, and the picker fallback here already handles the non-TTY case for the cluster arg, so the name could mirror it: drop `required:"true"`, prompt when empty, error with "name is required in non-interactive mode" when there is no terminal.

--p+🤖

cli/commands/deploy.go
```diff
@@ -105,6 +107,69 @@ type deployOpts struct {
...
+	}
+	if o.Cluster != "" {
+		return nil
+	}
+
+	target, err := dc.Target(o.Target)
+	if err != nil {
+		return err
+	}
+	o.Cluster, err = o.clusterNameForTarget(target)
```
phinze, line 137: Worth being deliberate about a side effect here: once the target lands in `o.Cluster`, `AppCentric.LoadCluster` takes the same branch it does for `-C` and calls `saveClusterState`, so every deploy re-pins the app's cluster in `app-state.toml` for every other command. After a bare `miren deploy` (default = staging), `miren logs` follows to staging; after `miren deploy --target prod` it follows to prod; the next bare deploy flips it back. That may be exactly the behavior you want ("wherever I last deployed"), but it is a committed file steering unrelated commands, and neither the PR body nor `deployment.md` mentions it. If it is intended, a sentence in the docs would save someone a confused afternoon; if not, `Validate` could stash the resolved name somewhere other than `o.Cluster` so the pin only moves on an explicit `-C`.

--p+🤖

cli/commands/deploy.go
```diff
@@ -105,6 +107,69 @@ type deployOpts struct {
...
+	}
+	return nil
+}
+
+func (o *deployOpts) clusterNameForTarget(target *appconfig.DeployTarget) (string, error) {
+	if target.ClusterID == "" {
+		return target.Cluster, nil
+	}
+
+	cfg, err := o.LoadConfig()
```
phinze, line 149: Small first-run regression: a teammate who clones a repo with a `cluster_id` target and has never run `miren login` now dies here with `error validating options: no config file found`, because `LoadConfig` returns `ErrNoConfig` and `Validate` runs before the deploy body. Before this change the same person got as far as the body's "no client configuration available; run `miren login`..." message. Treating `ErrNoConfig` here the same way as the XID-miss case below (not configured; run `miren login` / `miren cluster add`) would keep the friendlier error for the exact people most likely to hit it.

--p+🤖

cli/commands/deploy.go
```diff
@@ -105,6 +107,69 @@ type deployOpts struct {
...
+	}
+	if dc == nil {
+		if o.Target != "" {
+			return fmt.Errorf("--target requires %s", appconfig.DeployConfigPath)
+		}
+		return nil
+	}
+
+	if o.Target != "" && o.Cluster != "" {
+		return fmt.Errorf("--target and --cluster cannot be combined")
```
phinze, line 127: `o.Cluster` is also fed by `MIREN_CLUSTER`, and a CI job that exports that and passes `--target prod` is probably the most likely way to trip this. The message then names a flag nobody typed. Something like "--target cannot be combined with --cluster or MIREN_CLUSTER" would point at the right thing.

--p+🤖

cli/commands/deploy.go
```diff
@@ -86,6 +86,8 @@ func reconcileDeploymentCancellation(
 type deployOpts struct {
 	AppCentric
 
+	Target string `long:"target" description:"Deployment target from .miren/deploy.toml"`
```
phinze, line 89: Bigger-picture question, anchored here because the confirmation code itself is not in the diff: does the `Deploy app X to cluster Y? [Y/n]` gate still make sense once a project has targets? With a `deploy.toml` present I think I'd expect bare `miren deploy` on a TTY to show me a picker over the targets (default first, so Enter still means "the default") instead of a yes/no about a cluster name I did not type. Same keystroke to accept, but it lets you switch targets without knowing the `--target` flag, and it makes the committed file feel like the source of truth rather than a thing that pre-fills a prompt. `--force`, `--target`, and non-TTY would all skip it as they do today. Totally fine as a follow-up if you agree; not asking for it in this PR.

--p+🤖

phinze (approved):
This came out nicer and simpler than I was bracing for. It's entirely client sugar, the server never sees anything new, and the XID-based resolution is exactly the piece that makes a committed file survive different people naming the same cluster differently. Approving.

Inline are a handful of things to consider, none blocking: finishing the "omit it, get prompted" pattern for the target name, a side effect where every deploy re-pins the app's cluster for other commands, a first-run error that got less friendly, an error message that should mention `MIREN_CLUSTER`, and a bigger-picture question about whether the y/n gate should become a target picker now that targets exist. Happy to talk through any of them.

--p+🤖

cli/commands/deploy_target.go
evanphx: :robot: Implemented the prompt-first flow. A bare `miren deploy target add` now prompts for the target name and then opens the cluster picker; non-interactive use returns `name is required in non-interactive mode`.

cli/commands/deploy_targets_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"miren.dev/runtime/appconfig"
+	"miren.dev/runtime/clientconfig"
+)
+
+func writeDeployToml(t *testing.T, dir, content string) {
+	t.Helper()
+	require.NoError(t, os.WriteFile(filepath.Join(dir, ".miren", "deploy.toml"), []byte(content), 0644))
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed — `writeDeployToml` now creates `.miren` before writing, so the helper no longer depends on `writeAppToml` running first.

cli/commands/deploy.go
evanphx: :robot: Agreed that a target picker would improve discovery. I’ve kept it as a follow-up as suggested so this PR retains the existing deploy confirmation behavior.

cli/commands/deploy.go
evanphx: :robot: Target resolution now uses a deploy-only cluster override and bypasses `AppCentric`’s app-state persistence. I also documented that selecting a target does not alter the app’s pinned cluster for other commands.

cli/commands/deploy.go
evanphx: :robot: Added a friendlier first-run error directing users through `miren login` and `miren cluster add`, with regression coverage for a target with a cloud ID and no client config.

cli/commands/deploy.go
evanphx: :robot: Updated the conflict error to name both `--cluster` and `MIREN_CLUSTER`.

cli/commands/deploy.go
```diff
@@ -105,6 +108,88 @@ type deployOpts struct {
...
+	return nil
+}
+
+func (o *deployOpts) LoadCluster() (*clientconfig.ClusterConfig, string, error) {
+	if o.targetCluster == "" {
+		return o.AppCentric.LoadCluster()
+	}
+	config := o.ConfigCentric
+	config.Cluster = o.targetCluster
+	return config.LoadCluster()
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed. A deploy target without `cluster_id` must now match a locally configured cluster name; otherwise validation fails with guidance to run `miren cluster add` instead of falling through to address probing. Added a regression test for the missing-name case.

docs/docs/command/server-identity-anchor.md
```diff
@@ -11,7 +11,7 @@ Move where this cluster's workload identity is anchored
 ## Usage
 
 ```bash
-miren server identity-anchor <anchor> [flags]
+miren server identity-anchor [anchor] [flags]
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed by marking `Anchor` required and regenerating the command reference, which now shows `<anchor>`.

docs/docs/command/server-operations-abandon.md
```diff
@@ -11,7 +11,7 @@ Give up on an unfinished operation, including a data restore that keeps the serv
 ## Usage
 
 ```bash
-miren server operations abandon <id> [flags]
+miren server operations abandon [id] [flags]
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed by marking the server operation ID required and regenerating the command reference.

docs/docs/command/server-operations-show.md
```diff
@@ -11,7 +11,7 @@ Show one restart or upgrade operation
 ## Usage
 
 ```bash
-miren server operations show <id> [flags]
+miren server operations show [id] [flags]
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed by marking the server operation ID required and regenerating the command reference.

hack/gen-command-docs
```diff
@@ -89,7 +89,13 @@ build_usage_line() {
...
       local arg_name
       arg_name=$(echo "$pos_args" | jq -r ".[$i].name" | tr '[:upper:]' '[:lower:]')
-      line="$line <$arg_name>"
+      local required
+      required=$(echo "$pos_args" | jq -r ".[$i].required // false")
+      if [ "$required" = "true" ]; then
+        line="$line <$arg_name>"
+      else
+        line="$line [$arg_name]"
+      fi
```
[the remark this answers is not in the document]
evanphx: :robot: Fixed by marking `AuthCIRemove.ID` required and regenerating its command reference. The route arguments remain optional.

PR #1235, opened 2026-09-16
Backport component adoption and index watch fixes to v0.15

evanphx:
Backports #1232 and #1233 to `release/0.15`. The #1232 cherry-pick is unchanged. The #1233 cherry-pick follows that PR's rewrite: the nil guard now lives in the stream adapters rather than the index watcher, so every stream callback is covered.

## Summary

- reattach stdout and stderr logging when adopting existing etcd, VictoriaMetrics, and VictoriaLogs tasks after an unclean exit
- bound the etcd maintenance loop's health check so a wedged etcd keeps producing warnings instead of silencing the loop
- reject valueless stream frames in the `SendStream` adapters before they reach any callback or writer, ending the stream cleanly instead of panicking (or recording a zero-byte write)
- document both fixes in the v0.15 release notes

## Verification

- `go build ./...` on the release branch
- `go test ./pkg/rpc/stream ./pkg/entity/indexwatch ./pkg/slogout`
- `golangci-lint run ./pkg/rpc/stream ./pkg/entity/indexwatch`
- `jj interdiff` between the #1233 cherry-pick and its mainline commit is empty
- The iso `make test` sweep has not been run locally; CI covers it.


phinze (commented):
🤖 Rebuilt around the #1233 rewrite: the #1232 cherry-pick is unchanged, the #1233 cherry-pick is an empty `jj interdiff` against its mainline commit, and the changelog rev is redone with the watch entry describing the adapter-level fix and the etcd entry saying what the timeout actually does (the loop keeps warning instead of going silent; startup was already bounded). The `orig-level` caveat is pre-existing and tracked as MIR-1871; not taken in a byte-identical backport. Merging on Evan's behalf while he's out.

PR #1237, opened 2026-09-16
Install the whole base bundle on server upgrade, and report what the server is running

phinze:
`miren server install` puts containerd, runc, the shim, nerdctl and ctr into `/var/lib/miren/release` beside miren, and an upgrade downloads that same base tarball. But the extractor kept only the miren entry and threw the rest away with its temp dir, so a host upgraded any number of times was still on the containerd and runc it was first installed with. The `.iso/Dockerfile` bump that moves the bundle forward only ever reached fresh installs.

The first rev makes `pkg/release` stage and install every file in the archive, miren last, with the previous bundle kept in `release/.previous/` for rollback. The server keeps running through an upgrade and its shims exec runc from that directory, so each swap is a hard-link backup plus a single rename, and the path never stops resolving. There's no new lifecycle phase: a graceful restart already stops the embedded containerd through the boot graph, and containerd 2.x re-adopts the running shims on boot, so the existing `restarting` phase brings the new containerd up under the existing sandboxes. Runner upgrades share the same code and get the same fix.

The second rev closes the loop on "did it take": the containerd boot reports the daemon and runc versions it's driving, that lands in the health endpoint's `server.components`, and the executor records it on the operation so `operations show` says what the restarted server is actually on. The systemd e2e upgrades to a bundle whose runc reports a distinctive version and checks that report after the upgrade and after a rollback, with a `ctr`-started container required to survive all of it on the same pid.

One thing for the release notes: the upgrade *to* this release is still done by the old executor, so bundle updates start applying from the upgrade after it.

Closes MIR-1842


pkg/release/installer.go
```diff
@@ -124,60 +119,243 @@ func (i *binaryInstaller) Install(ctx context.Context, downloaded *DownloadedArt
...
+	mirenBackup := i.opts.InstallPath + i.opts.BackupSuffix
+	if !i.exists(mirenBackup) && bundled == nil {
+		return fmt.Errorf("no backup found at %s", mirenBackup)
+	}
 
-	// Check if backup exists
-	if _, err := os.Stat(backupPath); err != nil {
-		return fmt.Errorf("no backup found at %s: %w", backupPath, err)
+	targetDir := filepath.Dir(i.opts.InstallPath)
+	for _, rel := range bundled {
```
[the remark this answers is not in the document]
phinze: 🤖 Took the comment route: `Rollback` now says outright that miren-last is the only ordering that matters and the bundle files go back in whatever order they are listed. For what it is worth, `filepath.WalkDir` does promise lexical order by contract, so `bundleBackups` was already deterministic; the doc comment on it now says so too.

pkg/release/installer.go
```diff
@@ -124,60 +119,243 @@ func (i *binaryInstaller) Install(ctx context.Context, downloaded *DownloadedArt
...
 func (i *binaryInstaller) Rollback(ctx context.Context) error {
-	backupPath := i.opts.InstallPath + i.opts.BackupSuffix
+	bundled, err := i.bundleBackups()
+	if err != nil {
+		return fmt.Errorf("failed to list bundle backups: %w", err)
+	}
+	mirenBackup := i.opts.InstallPath + i.opts.BackupSuffix
+	if !i.exists(mirenBackup) && bundled == nil {
+		return fmt.Errorf("no backup found at %s", mirenBackup)
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 Fixed in the latest push: `HasBackup`, the `Rollback` guard, and the resume check all key on regular files under `.previous/` now (`len(bundled) > 0`), and `undo()` removes the tree once nothing is left in it, nested directories included. `TestBinaryInstaller_EmptyBackupTreeIsNoBackup` and `TestBinaryInstaller_UndoRemovesNestedBackupDirs` pin both.

pkg/release/installer.go
```diff
@@ -47,74 +47,76 @@ func NewInstaller(opts InstallOptions) Installer {
...
-	// Atomic rename from downloaded location to install path
-	if err := os.Rename(downloaded.Path, i.opts.InstallPath); err != nil {
-		// If rename fails (e.g., cross-device), fall back to copy
-		if err := i.copyFile(downloaded.Path, i.opts.InstallPath); err != nil {
-			// Restore backup if we created one
-			if backedUp {
-				i.Rollback(ctx)
-			}
-			return fmt.Errorf("failed to install binary: %w", err)
+	for _, rel := range downloaded.Bundled {
```
evanphx, line 99: :robot: Non-blocking: installation replaces files present in the new archive but does not remove files that disappeared from it; rollback likewise restores only replaced files. Future component renames or removals can therefore leave a mixed bundle. Consider a bundle manifest or explicit obsolete-file policy.

pkg/release/downloader.go
```diff
@@ -94,17 +94,18 @@ func (d *assetDownloader) Download(ctx context.Context, artifact Artifact, opts
 		}
 	}
 
-	// Extract the binary
-	binaryPath, err := d.extractTarGz(archivePath, opts.TargetDir)
+	staged, err := d.extractTarGz(archivePath, opts.TargetDir)
```
evanphx, line 97: :robot: Non-blocking: the checksum and artifact are downloaded from the same origin and the checksum is not rooted in separately trusted metadata. HTTPS covers ordinary transport corruption, but signed release metadata would provide stronger provenance against origin compromise or misconfiguration.

pkg/release/downloader.go
```diff
@@ -271,91 +272,104 @@ func (d *assetDownloader) verifyChecksum(filePath, expectedChecksum string) erro
...
+				return nil, err
 			}
-
 			outFile, err := os.Create(targetPath)
 			if err != nil {
-				return "", err
+				return nil, err
 			}
-
 			if _, err := io.Copy(outFile, tarReader); err != nil {
```
evanphx, line 346: :robot: Non-blocking: extraction rejects unsafe paths and links but has no limits on expanded bytes, entry count, or required free space. Consider archive limits and destination-space preflight so an oversized bundle cannot exhaust the temporary or destination filesystem.

evanphx (approved):
The bundle installation and rollback design has strong recovery semantics, atomic per-file replacement, integrity verification, and meaningful end-to-end coverage. The inline notes are future-proofing and hardening rather than merge blockers. Approve.

— e + :robot:

pkg/release/installer.go
phinze: 🤖 Agreed, and of the three it is the one that is a real gap in this design rather than a pre-existing one. The tarball is the only source of truth for "what is in the bundle" and we do not keep the previous one's listing, so an install cannot tell a file that was renamed away from an operator's file. Filed MIR-1889 to give the bundle an explicit file list the next install can diff against. Leaving as-is here since a stale extra binary is inert.

pkg/release/downloader.go
phinze: 🤖 Fair. That path is unchanged in this PR (the checksum has always come from the same origin as the artifact), but the bundle now carries the runtime binaries, so the blast radius of an origin compromise is bigger than it was. Filed MIR-1890 for signed release metadata; it belongs with the release pipeline rather than the installer.

pkg/release/downloader.go
phinze: 🤖 Agreed. Folded into MIR-1890 alongside the signing work since both are about trusting the artifact less: a total-bytes and entry-count cap on extraction is cheap, a free-space preflight against the destination a bit more.

PR #1239, opened 2026-09-16
Export nodes to cloud as the cluster's runner inventory

phinze:
Companion to mirendev/cloud#270. **Merge order matters: cloud#270 has to be merged and deployed first.** Cloud declines entity sync outright on a contract digest it doesn't know, so a runtime carrying this change against today's cloud would lose app and deployment sync along with the node export. This stays a draft until that's live.

`node_count` was the last thing the console read from the legacy status poll, and this runtime only ever sent 1. Instead, the runtime now exports its node entities through entity sync, so cloud can keep a real runner inventory: name, runner id, version, status, cordon state, address, labels, registration time. That's also what a cloud-driven coordinator-plus-runner upgrade (MIR-1875) will need to sequence and watch a fleet.

Nodes live in `dev.miren.compute` while the cloud contract is declared in `dev.miren.core`, so the first rev lets a contract span domains: a schema names the owner of its export, and the owner's `go:generate` merges it in with `-export-merge`. An owner with nothing to merge produces the exact contract it did before, and a test fails the build if a contributor is declared but not merged.

Nodes are also the first exported kind written through a session, which surfaced two things the exporter got wrong for session-written entities. The store leases a second marker-index key to the writer's session, so a runner losing its coordinator session fired a DELETE the exporter read as the node being removed; it now re-reads the entity at that revision and exports it as a change when it's still there. And `GetEntityAtRevision` read the primary key alone, dropping every session-scoped attribute, so cloud would never have seen a node's status; it now goes through the same pinned read as the batch form.

The poll's hardcoded `node_count` goes with this.

Part of MIR-1861


evanphx (approved):
Node identity, deletion behavior, export allowlisting, snapshot/change handling, and mirror lifecycle look correct. The compatibility guard and tests support the intended runner-inventory behavior without exposing secret-bearing fields. Approve.

— e + :robot:

PR #1241, opened 2026-09-16
Carry the full commit and build date in the uplink hello

phinze:
The uplink hello identifies the running build as `runtime_version`, which for a main build is `main:abc1234`. That short sha is all cloud has to compare a cluster against the main channel, so the Updates tile matches on seven characters and the build-date tiebreak that `miren upgrade --check` applies locally has nothing to work with.

This adds `runtime_commit` and `runtime_build_date` to the hello, straight from `version.GetInfo()`. Both are omitted when the binary wasn't stamped by ldflags, the same shape `runtime_instance_id` took, so there's no handshake rev and an older cloud just doesn't see them. `version.Commit` defaults to the literal string `"unknown"`, which is fine for `miren version` output but not for a field whose absence means something, so a small `KnownCommit` helper normalizes it on the way out.

Companion to mirendev/cloud#272, which reads the fields. Either can merge first.

Part of MIR-1883


pkg/uplink/session.go
```diff
@@ -59,9 +59,17 @@ type SessionHello struct {
...
-	RuntimeInstanceID string            `json:"runtime_instance_id,omitempty"`
-	ClientTime        time.Time         `json:"client_time"`
-	Capabilities      []CapabilityOffer `json:"capabilities"`
+	RuntimeInstanceID string `json:"runtime_instance_id,omitempty"`
+	// RuntimeCommit and RuntimeBuildDate identify the build exactly.
+	// RuntimeVersion carries only a short sha, which is ambiguous once git
+	// lengthens abbreviations under a collision and carries no order, so
+	// cloud needs the full commit to match a build against its channel and
+	// the build date to say which of two main builds is newer. Both are
+	// absent from a binary built without hack/build.sh's ldflags.
```
evanphx, line 68: :robot: Non-blocking: `make release` / `hack/build-release.sh` and `make dist` still stamp only Version, so locally packaged binaries omit the new commit and build date. Consider aligning those build paths in a follow-up.

evanphx (approved):
This is a focused, backward-compatible protocol extension with appropriate omission semantics and proportional tests. Production CI and normal development builds supply the intended metadata; the remaining local release/dist build-path inconsistency is non-blocking. Approve.

— e + :robot:

pkg/uplink/session.go
phinze: 🤖 Good catch, and small enough to do here rather than in a follow-up. `a37ed8d` stamps Commit and BuildDate in both `hack/build-release.sh` and `hack/build-dist.sh`, matching what `hack/build.sh` and `build-ci.sh` already do, so every build path now produces the same identity.

hack/build-release.sh
```diff
@@ -58,7 +68,7 @@ fi
 
 if ! test -f $dir/miren-linux-amd64.tar.gz || ! test -f $dir/miren-linux-amd64.zip; then
   echo "Linux / amd64"
-  GOOS=linux GOARCH=amd64 go build -ldflags "-X miren.dev/runtime/version.Version=$version" -o $dir/miren ./cmd/miren
+  GOOS=linux GOARCH=amd64 go build -ldflags "$ldflags" -o $dir/miren ./cmd/miren
```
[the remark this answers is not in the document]
phinze: 🤖 The unquoted `$dir` predates this change (this PR only touches the ldflags), and `VERSION` is something we set ourselves when running `make release`, so leaving it alone here.

PR #1243, opened 2026-09-17
Honor ctx in inline RPC calls and stop cloudrpc Recv dropping a frame on close

phinze:
Two small read-path fixes from the Detail batch, same family as #1219 and the streaming-watch hang: places where a blocked read outlives what the caller asked for, or a close wins a race it shouldn't.

`inlineClient.Call` was the one ctx-taking read in `pkg/rpc` without a `CancelRead` watcher, so an inline callback whose peer stalled ignored the caller's deadline entirely. First rev gives it the same watcher `handleCallStream` and `roundTrip` already use, stopped before the stream goes back in the pool, and maps a cancelled read to the context error.

`cloudrpc.session.Recv` could return EOF with a frame still sitting in `inbound` when the frame and the close landed in the same instant, and since `readPump` treats EOF as final that frame was gone. Second rev takes one more non-blocking look at the buffer from the close branches, the read-side twin of the `SendContext` guard from `cb7ca262`. Both come with tests that fail on the old code.

Closes MIR-1743
Closes MIR-1728


pkg/rpc/inline.go
```diff
@@ -157,9 +157,27 @@ func (c *inlineClient) Call(ctx context.Context, method string, args any, ret an
...
+				conn.stream.CancelRead(cancelReadCode)
+			case <-stop:
+			}
+		}()
+	}
+
 	// Return stream to pool when done (unless there's an error)
 	shouldReturn := true
 	defer func() {
+		close(stop)
```
evanphx, line 200: :robot: Closing `stop` does not guarantee the cancellation watcher has exited before `returnStream` publishes the stream back to the pool. If context cancellation races cleanup, both `ctx.Done()` and `stop` can be ready; cleanup can pool the stream, another call borrows it, then the old watcher selects `ctx.Done()` and calls `CancelRead` on the new call. Please synchronize watcher completion before pooling, e.g. with a done channel closed by the watcher and awaited after `close(stop)`, and add a test that forces cancellation and cleanup to race.

evanphx (changes requested):
The cloudrpc close-frame fix is sound, but the inline RPC cleanup still has a pooled-stream race. Closing `stop` does not establish that the watcher exited before the stream is returned, so an old watcher can cancel a subsequent borrower. Please synchronize watcher completion before pooling and add a regression test that forces cancellation and cleanup to race.

— e + :robot:

pkg/rpc/inline.go
phinze: 🤖 Good catch, fixed in `3c1a0a1b`. The watcher now closes a `done` channel on exit and cleanup waits on it before touching the pool. Waiting turned out to be necessary but not sufficient: a `msgStream` that took a `CancelRead` stays cancelled, so the watcher also reports whether it fired and a cancelled stream is closed instead of pooled.

`TestInlineClientCancelRacingCleanupDoesNotPoisonPool` cancels right after a successful `Call`, which leaves `stop` and `ctx.Done` both ready for the parked watcher, then asserts the next borrower's call succeeds. On the previous rev it fails at iteration 3.

pkg/rpc/inline.go
```diff
@@ -157,9 +157,42 @@ func (c *inlineClient) Call(ctx context.Context, method string, args any, ret an
...
+	stop := make(chan struct{})
+	done := make(chan struct{})
+	fired := false
+	if ctx.Done() != nil {
+		go func() {
+			defer close(done)
+			select {
+			case <-ctx.Done():
+				fired = true
+				conn.stream.CancelRead(cancelReadCode)
```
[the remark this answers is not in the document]
phinze: 🤖 Right, on msgStream the write side fails with errStreamCancel too. Both Encode branches now go through the same helper as the Decode ones (renamed to `callErr`).

pkg/rpc/inline_test.go
```diff
@@ -0,0 +1,167 @@
...
+	serveInlineOK(t, peer)
+
+	ic := &inlineClient{
+		log:     slog.Default(),
+		oid:     OID("cap"),
+		session: client,
+	}
+
+	var out struct{}
+	for i := range 2000 {
```
evanphx: :robot: Non-blocking: this test makes the previous race likely to surface but does not deterministically force cancellation and cleanup into the critical interleaving. Consider adding a barrier or instrumented stream so the old implementation reliably fails.

evanphx (approved):
Approved. Cleanup now waits for cancellation-watcher completion before returning a stream to the pool, and a stream where cancellation fired is discarded. This establishes the required happens-before relationship and resolves the pooled-stream corruption race. The nearby cancellation error mapping is also corrected, and current checks pass.

— e + :robot:

pkg/rpc/inline_test.go
phinze: 🤖 Agreed, done in `e8fd25c0`. A hooked stream cancels the ctx as it delivers the last response byte and holds that `Read` until `CancelRead` has actually landed, so when `Decode` returns the watcher has provably fired and cleanup is holding a poisoned stream. Against the pre-fix `inline.go` it fails 20 of 20 runs at the "returned to the pool" assertion.

PR #1244, opened 2026-09-17
doctor reports whether the CLI or server is behind the latest release

phinze:
`miren doctor` could tell you whether your CLI and the server disagreed, but not whether either was behind what's actually published. Against a v0.9 server it said "server is behind CLI" and stopped there. The Version row now states the three facts that matter, `CLI v0.15.0, server v0.9.0, latest v0.15.2`, and warns with the upgrade command for whichever is stale.

Latest comes from the same channel and the same comparison `miren upgrade` uses, so a warning here means that command would act. Skew between CLI and server on its own stops being a warning, since an upgrade does nothing for a prerelease ahead of a current server; it's the fallback evidence only when the asset service can't be reached, and the line says so. The row reads the same from a laptop and from an SSH shell on the server host, because each part means one thing: the binary you invoked, the selected cluster, the latest channel.

Two cleanups come first. Doctor's evidence was entirely cluster-shaped and only gathered after a cluster loaded; it's now gathered per scope, so the CLI is judged even with nothing configured, and auth runs in the gather so every check is a pure function over the evidence. The `doctor config|server|auth` subcommands are removed rather than aliased, since the single sweep has been the product since the rebuild and they only filtered it. The upgrade advice also moves from `sudo miren server upgrade` to `sudo miren upgrade`; on hosts older than v0.15 that refuses and names the server form, a one-step detour we're accepting for the sake of one spelling.

Closes MIR-1832


cli/commands/doctor_version.go
```diff
@@ -1,53 +1,130 @@
...
+	case skew == skewUnordered:
+		parts = append(parts, "server "+server.Version+" (different build)")
+	default:
+		parts = append(parts, "server "+server.Version)
+	}
+
+	if latestKnown {
+		parts = append(parts, "latest "+latest.Version)
+	} else {
+		parts = append(parts, "latest unknown ("+fetchFailure(env.latestErr)+")")
```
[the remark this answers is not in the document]
phinze: 🤖 Checked: `GetVersionMetadata` never returns `(nil, nil)`. A non-200 is an error, and the success path returns whatever `ParseMetadata` gives back, which is a value or an error. The `"no metadata"` branch is unreachable defensive text, so we are leaving it as is.

cli/commands/doctor_auth.go
```diff
@@ -1,11 +1,125 @@
...
+		}
+
+		result.Claims, _ = auth.ParseUnverifiedClaims(token)
+		result.Method = string(identity.Type)
+
+		result.UserInfo, _ = fetchCloudUserInfo(ctx, authServer, token)
+
+	case clientconfig.IdentityCertificate:
+		result.Method = "certificate"
+	}
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, and pre-existing: rev 1 only moved this code. Fixed in `e0a71df` as its own rev: `checkAuthentication` now reads `Method == "none"`, which `tryAuthenticate` already leaves set on every failure path, so a certificate identity passes. `authTried` is gone. Table test added covering token, certificate, and the two failure shapes.

cli/commands/doctor_version.go
```diff
@@ -1,53 +1,130 @@
...
+	}
+
+	server, serverKnown := serverVersionInfo(env)
+	tooOld := env.configured() && env.connErr == nil &&
+		errors.Is(env.serverVersionErr, errServerVersionUnsupported)
+
+	// Without latest to judge against, CLI-versus-server skew is all there is.
+	skew := skewNone
+	if !latestKnown && serverKnown {
+		skew = compareVersions(cli.Version, server.Version)
```
evanphx, line 47: :robot: Non-blocking: when fetching latest fails, fallback skew logic can give false upgrade advice. A stable CLI against a prerelease server can be classified as CLI-behind even though `miren upgrade` targets stable latest and may do nothing. Consider suppressing actionable advice for prerelease/channel mismatches when latest is unknown.

cli/commands/doctor.go
```diff
@@ -297,7 +154,6 @@ func statusText(r checkResult) string {
 func printDoctorJSON(checks []check, results []checkResult) error {
 	type checkJSON struct {
 		Name    string   `json:"name"`
-		Group   string   `json:"group"`
```
evanphx, line 300: :robot: Non-blocking compatibility note: removing `group` changes the established `doctor --json` schema, alongside removal of `doctor config|server|auth`. Consider retaining `group` in JSON and/or providing a deprecation window for subcommands.

evanphx (approved):
Approved with two non-blocking notes. The implementation correctly judges both CLI and server against the latest stable release when metadata is available, keeps an offline fallback, and has focused coverage for version and authentication behavior. The remaining concerns are bounded prerelease/offline guidance and compatibility for existing doctor consumers; neither needs to block this testing-machinery improvement.

— e + :robot:

cli/commands/doctor_version.go
phinze: 🤖 Good catch, fixed in the follow-up rev. When latest is unknown and either side is a prerelease, the skew is now a fact in the summary line with no upgrade advice, since `miren upgrade` targets stable and would do nothing. Two table cases cover both directions.

cli/commands/doctor.go
phinze: 🤖 We went back and forth on the subcommands and landed on deleting them outright: the single sweep has been the product since the rebuild, and they only filtered it. `group` only ever mirrored those subcommands, so with them gone it would be a constant with no meaning; better to drop it than keep a field that lies. Anyone scripting `doctor --json` keys on `name` and `status`, which are unchanged.

cli/commands/doctor_version.go
```diff
@@ -1,53 +1,136 @@
...
 		}
 	}
 
-	server := env.serverVersion
-	summary := fmt.Sprintf("CLI %s, server %s", cli.Version, server.Version)
-	if !server.Ready {
-		summary += " (server still starting)"
+	// Skew across a prerelease is a fact with no fix: `miren upgrade` targets
+	// stable, so on either side it would do nothing about the gap.
+	if (skew == skewServerBehind || skew == skewCLIBehind) && (isPrerelease(cli) || isPrerelease(server)) {
```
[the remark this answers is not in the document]
phinze: 🤖 Agreed, took the suggested condition: advice is suppressed only when the newer side is the prerelease. Added a case for an rc CLI behind the stable server, which keeps its `miren upgrade`.

PR #1245, opened 2026-09-17
Fix cookie Secure flag leaking between concurrent ingress auth requests

phinze:
Every ingress auth middleware (OIDC, connector, password) was writing its request's scheme into the one shared `SessionManager` and then reading it back when the handler emitted cookies. With an http and an https request inflight at the same time, whichever wrote last decided the other's `Secure` flag. For the OIDC callback the gap between write and read includes the round trip to the IdP, so this wasn't a narrow window. It also showed up as a plain data race.

The handler cache was already keyed by host and scheme, so the flag really belongs to the handler, not the request. `SetSecure` is gone; `WithSecure` returns a copy that shares the encryption key, and each cached handler gets its own per-scheme copy at construction. The shared instance is never mutated after startup. The cookies were already encrypted and HttpOnly, so the exposure was replay of the opaque cookie value over plaintext, not token contents.

The `Server` field is renamed from `oidcSessionManager` to `sessionManager` since it backs all three auth backends now. New tests cover `WithSecure` and drive the password middleware with concurrent http and https logins under `-race`; I confirmed the middleware test fails with the shared-mutation shape restored.

Closes MIR-1734


evanphx (approved):
Fix is sound: cookie security state is now immutable per scheme-specific cached auth handler, eliminating the data race and cross-request leakage while preserving cookie semantics and key compatibility. Tests and current checks pass. Non-blocking follow-up: restrict or sanitize forwarded-protocol headers at the trust boundary to prevent client-controlled secure downgrades.

— e + :robot:

PR #1249, opened 2026-09-18
Pull iso services and blackbox fixture base images through oci.miren.cloud

phinze:
Today's main release run ([first attempt](https://github.com/mirendev/runtime/actions/runs/35376074815/attempts/1)) lost two blackbox shards at "Start dev environment" when Docker Hub's token endpoint timed out mid-pull of `postgres:16-alpine`. No test ran, and rerunning the failed jobs got the release through. But it's a hole worth closing: postgres and valkey were the only two iso services still pulled straight from Docker Hub, and every blackbox and distributed shard pulls them fresh, so one Hub hiccup can take out several jobs at once.

First rev points both at the `oci.miren.cloud` pull-through proxy, same tags, the way etcd and the Victoria images already are. Second rev does the same for the four testdata Dockerfiles that still `FROM alpine:3.20`, since BuildKit has no docker.io mirror configured and rebuilds them on every shard.

That leaves one Hub pull in the blackbox path: `testdata/image-only` uses the bare `nginx:alpine` on purpose so the test can assert on the normalized `docker.io/library/` form, and the proxy doesn't alias nginx anyway. If Hub keeps biting, adding nginx to the proxy is the better fix than bending that test.


evanphx (approved):
This all hangs together. The proxy is serving the same upstream platform digests for Alpine, PostgreSQL, and Valkey, including amd64 and arm64, and the change stays put in development services and blackbox fixtures. The tags are still mutable and there is no fallback if the Miren proxy has a bad day, but those are the same tradeoffs the existing setup already makes. Nice tidy way to get Docker Hub auth failures out of this path.

— e + :robot:

PR #1250, opened 2026-09-18
Emit process start time and build info from the coordinator

phinze:
The central store can see the coordinator's heap, RSS and goroutine count, but not when the process started or which build it is. So a restart shows up as a gap and a memory reset, an upgrade looks the same, and telling the two apart means going to the commit log and guessing from the auto-upgrade timer's jitter. That was the biggest source of noise while calibrating the MIR-1825 alert rules.

This adds a `ProcessInfo` collector next to `RuntimeMemory` that pushes two standard series with the same `entity="miren/control"` label: `process_start_time_seconds`, and `miren_build_info{version, commit, channel} 1` in the usual info-metric shape. They ride the existing operational fanout, so they land in the embedded VictoriaMetrics and get the cluster identity stamped on the way out. Since both values are constant for the life of the process, they push once a minute rather than every ten seconds, with an extra push at start and again when the off-cluster shipping sink attaches, so a process that dies young still leaves a record where the restart and fleet-skew rules can see it.

Restart markers, upgrade annotations and the "which build is everyone on" query are now a Grafana problem rather than a runtime one. Runner processes are not covered yet; MIR-1827 picks that up.

Closes MIR-1864


metrics/process_info_test.go
```diff
@@ -0,0 +1,106 @@
...
+	assert.NotEmpty(t, pi.Version)
+	assert.NotEmpty(t, pi.Commit)
+}
+
+func TestProcessInfo_EmitNilWriterIsNoop(t *testing.T) {
+	require.NoError(t, NewProcessInfo(testLogger(), nil).Emit(context.Background()))
+	var pi *ProcessInfo
+	require.NoError(t, pi.Emit(context.Background()))
+}
+
```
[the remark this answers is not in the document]
phinze: 🤖 Good catch, added in 6c1ab13: `TestProcessInfo_MonitorEmitsBeforeFirstTick` runs Monitor against a channel-backed PointWriter and fails if no batch lands within a second, so moving Emit back inside the select would trip it.

components/server/boot_app_metrics.go
```diff
@@ -118,6 +118,13 @@ func (b *appMetricsBoot) start(
 	b.shipping = &metrics.Labeled{Sink: b.shipWriter, Labels: identityLabels}
 	b.operational = observability.operationalMetrics
 	b.operational.Attach(b.shipping)
+	// The identity sample the process pushed at boot only reached the
+	// embedded store, since this sink did not exist yet. Push it again so a
+	// process that dies before its next tick still records its start time
+	// and build where the restart and skew rules can see it.
+	if err := observability.processInfo.Emit(ctx); err != nil {
```
evanphx: :robot: A little non-blocking test suggestion: this late-sink re-emission is important startup-order wiring, and the collector tests do not quite prove that `Attach` happens before `Emit` or that the shipping sink receives this batch. A small fanout or boot-level regression test would help keep this fence mended.

evanphx (approved):
Looks good. The names and units follow Prometheus conventions, the labels stay nicely bounded, and the second push after the shipping sink attaches closes the startup-order gap. The collector and concurrency pieces are in good shape. I left one small test suggestion, but there is nothing here worth holding the train for.

— e + :robot:

components/server/boot_app_metrics.go
phinze: 🤖 Good call, added in 08dd23e9. `start()` brings up vmagent so the block wasn't reachable in a unit test; pulled the attach-then-emit into `attachShipping` and `TestAttachShippingReemitsProcessIdentity` hands it a fanout with recording sinks. It checks the shipping sink gets both identity series with the cluster label stamped, and I confirmed it fails when the two calls are swapped.

